PR #11, opened 2017-03-24
Make dependencies git submodules

jmct:
The idea here is to have a better handle of what versions of dependencies are required for the build to succeed. 

Submodules aren’t nice, but sometimes they are necessary.

Because this requires a change in people’s workflow I think it’s worth some discussion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,4 @@
 /build/
-/dependencies/
 .cabal-sandbox
 .stack-work
 **/cabal.config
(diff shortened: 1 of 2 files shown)


joehendrix:
I'm undecided whether it makes sense to have a specific set of submodules on this repo.  So far I've been using submodules that are specific to the project I am working on rather than crucible itself.  This allows me to pull in all the repos I need and specify versions that work with all the constraints.

atomb:
From the SAWScript perspective, we'll probably move toward submodules directly in the `saw-script` repository, which would be redundant with any in the `crucible` repository. Having them probably wouldn't cause trouble, but may be unnecessary.

robdockins:
I think it doesn't make sense to maintain dependencies as submodules directly in this repository.  Instead, downstream projects should be maintaining consistent sets of dependencies via whatever method makes sense for them.

Instead, we should be managing the potential for build failures via better version bound discipline.

PR #24, opened 2017-08-09
Fix SimpleBuilder.boundVars when the expression is just a bound var

jmgrosen:
Arguably, we should also insert `Set.empty` into the hash table for all other elements, but I think it's a reasonable default to assume that if an ID isn't in the table, then it's empty.


robdockins:
Is this a correctness fix, or just an efficiency change?

jmgrosen:
Correctness fix. If I pass `varExpr sym v`, i.e., an expression consisting of solely a variable, to `boundVars`, I get back an empty hash table. In my mind, this is not expected behavior.

PR #26, opened 2017-08-10
Update to match new llvm-pretty types

TomMD:
Update crucible to match the new llvm-pretty types.


TomMD:
This patch was also part of the tree merged in #27, so I'm closing.

PR #27, opened 2017-08-10
Fix/invoke

TomMD:
Update crucible llvm translation to translate Invoke as Call + Jump.  This translation ignores the exception path and thus will never result in execution reaching a landingpad.

@joehendrix I think you're the gate keeper here and I don't know about any merge procedures you'd like followed so I'll either wait till I hear back or let you press the button.


robdockins:
I can't get this to compile.  Are the new `decComdat` fields not part of llvm-pretty `master` yet?

TomMD:
@robdockins  See pull request #28.

WRT compilation, I can compile it with SAW but have all sorts of missing packages when I try to build with the scripts provided under crucible.  Not sure how most people dev cycle goes but that would be helpful to know.

PR #30, opened 2017-08-16
Fix/metadata

TomMD:
Sometimes metadata does not include an explicit symbol and instead uses ValUndef.  It seems to occur in some situations when using `llvm-link`.  At the expense of less-useful error messages, we can simply ignore these sorts of calls.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
@@ -331,6 +331,8 @@ instrResultType instr =
     L.Conv _ _ ty -> ty
     L.Call _ (L.PtrTo (L.FunTy ty _ _)) _ _ -> ty
     L.Call _ ty _ _ -> error $ unwords ["unexpected function type in call:", show ty]
+    L.Invoke (L.FunTy ty _ _) _ _ _ _ -> ty
+    L.Invoke ty _ _ _ _ -> error $ unwords ["unexpected function type in invoke:", show ty]
     L.Alloca ty _ _ -> L.PtrTo ty
     L.Load x _ _ -> case L.typedType x of
                    L.PtrTo ty -> ty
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
```diff
@@ -331,6 +331,8 @@ instrResultType instr =
     L.Conv _ _ ty -> ty
     L.Call _ (L.PtrTo (L.FunTy ty _ _)) _ _ -> ty
     L.Call _ ty _ _ -> error $ unwords ["unexpected function type in call:", show ty]
+    L.Invoke (L.FunTy ty _ _) _ _ _ _ -> ty
+    L.Invoke ty _ _ _ _ -> error $ unwords ["unexpected function type in invoke:", show ty]
```
robdockins: Why is `invoke` treated differently than `call` here?  Looking at the LLVM manual, I don't see a reason they should be...

TomMD: Well `Invoke` is not `Call` and its type is packed differently iiuc.  Let's look at a simple example:

```
int h()
{
    throw 2;
}

int g()
{
    int x;
    try {
        x = h();
    } catch(int err) {
        x = err;
    }
    return x;
}

int test()
{
    int a = g();
    return a;
}
```

And the bitcode for those call and invokes is parsed into an AST of:

```
  (Invoke
    (FunTy (PrimType (Integer 32)) [] False)
    (ValSymbol (Symbol "_Z1hv"))
    []
    (Anon 4)
    (Anon 5))
```

vs

```
(Call
    False
   (PtrTo (FunTy (PrimType (Integer 32)) [] False))
   (ValSymbol (Symbol "_Z1gv"))
   [])
```

So invoke gives us a type of `FunTy (PrimType (Integer 32)) [] False` while call gives `PtrTo (FunTy (PrimType (Integer 32)) [] False)`.

robdockins:
Sorry, @TomMD, I lost the thread here.  Did these fixes get merged another way, or do I still need to merge them?

TomMD:
AFAIK these were never merged in any way.  Merging would be a useful but not complete step towards working with real C++ code.

PR #34, opened 2018-01-05
Qualify Context import to resolve update ambiguity

dmjio:
When attempting to build `cryptol-gen`
```
src/Lang/Crucible/Utils/SymMultiDimArray.hs:43:5: error:
    Ambiguous occurrence ‘update’
    It could refer to either ‘Ctx.update’,
                             imported from ‘Data.Parameterized.Context’ at src/Lang/Crucible/Utils/SymMultiDimArray.hs:70:1-50
                             (and originally defined in ‘Data.Parameterized.Context.Unsafe’)
                          or ‘Lang.Crucible.Utils.SymMultiDimArray.update’,
                             defined at src/Lang/Crucible/Utils/SymMultiDimArray.hs:567:1
```

### DIFF
--- a/crucible/src/Lang/Crucible/Utils/SymMultiDimArray.hs
+++ b/crucible/src/Lang/Crucible/Utils/SymMultiDimArray.hs
@@ -67,7 +67,7 @@ import           Control.Monad.State.Strict
 import           Data.Foldable
 import           Data.Maybe
 import           Data.Parameterized.Classes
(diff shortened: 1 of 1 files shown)


robdockins:
Does this problem go away if you pull in the latest `parameterized-utils`?  I think Joe renamed the `update` function so this conflict doesn't occur any more.

dmjio:
Yea, had an old version, this isn't needed.

PR #46, opened 2018-07-06
Preserve the context when collecting proof obligations.

yav:


crucible/src/Lang/Crucible/Backend/AssumptionStack.hs
```diff
@@ -110,6 +115,23 @@ data ProofGoal pred assumeMsg assertMsg =
   , proofGoal        :: LabeledPred pred assertMsg
   }
 
+type CruxGoalGCollector pred assumeMsg assertMsg =
```
robdockins: No really a fan of this name

robdockins:
I think we need to look a bit more at `cloneAssumptionStack`.  Probably, leaving an `emptyGoalCollector` there is the wrong thing to do; in the cloned state, we need to recreate the assumption context so that asserted goals have the appropriate hypothesis in scope.

PR #52, opened 2018-07-13
Simulator refactor

robdockins:
This branch contains a fairly large-scale refactoring of the core Crucible symbolic simulator.  The goals of this refactoring are:

* fix bugs involving exception handling in the simulator;
* simplify and document the main simulator data structures and operations; and
* pave the way for other uses and enhancements (debugging support, profilers, etc).

In the process of this refactoring two major changes were made to the core operations.  First, the simulator has been modified into a "small-step" style, whereas previously the simulator would compute all-at-once to the final result.  The new small-step style should make  it easier to incorporate different kinds of orchestration loops, and has already enabled bugfixes for the exception handling code.  The other major change is that several data items that used to contain Haskell closures (return handlers and paused frames) have instead been defunctionalized, replacing the continuations with explicit data constructors indicating what actions need to be taken.  I believe this makes the code significantly easier to understand.

In addition there is more complete documentation, and the `ExectionTree` and `CallFns` modules have been reorganized into new modules with better organized APIs and layering: `ExecutionTree`, `Operations` and `EvalStmt`.  There is also a new reexport module `Lang.Crucible.Simulator` that reexports the most relevant symbols for downstream use.

I've taken wide latitude to change names, alter style, and move code around in this refactoring, so it will be a pretty noisy diff.  However, the external, downstream effects should be fairly limited.


robdockins:
I view this branch to be largely complete, but I plan to hold off on merging until I've done some additional testing and benchmarking to gain confidence that I haven't introduced correctness or performance bugs.

PR #53, opened 2018-07-14
Abstract floating point unit

andreistefanescu:


what4/src/What4/Expr/GroundEval.hs
```diff
@@ -123,6 +123,7 @@ defaultValueForType tp =
     BaseStringRepr  -> mempty
     BaseArrayRepr _ b -> ArrayConcrete (defaultValueForType b) Map.empty
     BaseStructRepr ctx -> fmapFC (GVW . defaultValueForType) ctx
+    BaseFloatRepr _ -> undefined
```
robdockins: Need to fill this in somehow

what4/src/What4/Expr/GroundEval.hs
```diff
@@ -323,6 +324,38 @@ evalGroundApp f0 a0 = do
...
+    FloatIsNeg{} -> undefined
+    FloatIte{} -> undefined
+    FloatCast{} -> undefined
+    BVToFloat{} -> undefined
+    SBVToFloat{} -> undefined
+    RealToFloat{} -> undefined
+    FloatToBV{} -> undefined
+    FloatToSBV{} -> undefined
+    FloatToReal{} -> undefined
+
```
robdockins: Need to fill these in too

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
```diff
@@ -1208,7 +1208,8 @@ printfOps sym valist =
   , printfGetFloat = \i _len ->
      case valist V.!? (i-1) of
        Just (AnyValue (FloatRepr _fi) x) ->
-         return $ asRational x
+         -- TODO: handle interpreted floats
```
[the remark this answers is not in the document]
andreistefanescu: As far as I understand the code around, this only works for concrete values. For floats, I don't keep any concrete values, I just send the literals to the solver, so I don't think there is much I can do.

PR #55, opened 2018-07-24
CFG rewriting

lukemaurer:
New module for doing rewrites on registerized CFGs (easier than rewriting under SSA invariants).


crucible/src/Lang/Crucible/Utils/RegRewrite.hs
```diff
@@ -0,0 +1,213 @@
+-----------------------------------------------------------------------
+-- |
+-- Module           : Lang.Crucible.Utils.RegRewrite
+-- Description      : Operations for manipulating registerized CFGs
+-- Copyright        : (c) Galois, Inc 2014-2018
+-- License          : BSD3
+-- Maintainer       : Joe Hendrix <jhendrix@galois.com>
```
robdockins: Minor issue, but let's not blame Joe for code he didn't write :-)

robdockins:
What's the status here?  Should I start looking at this?

lukemaurer:
Sure. I've been writing against it, and there are a few conveniences we might want to add but the core functionality seems solid.

PR #56, opened 2018-08-07
Integer update

robdockins:
Updates to what4 and crucible regarding the integer theory and coercions to/from integers.


robdockins:
Some other updates in this branch also, including a refactoring of the `AssumptionStack` data structures.

Issue #73, opened 2018-08-21
Implement more LLVM intrinsic functions


langston-barrett:
Currently, Crucible implements a subset of available [LLVM intrinsic functions](https://llvm.org/docs/LangRef.html#intrinsic-functions). Which ones? Find the latest list:
```bash
curl https://raw.githubusercontent.com/GaloisInc/crucible/master/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs | grep "let nm = " | perl -p -i -e 's/let nm = "(.+)" in/\1/' | sort
```
They are implemented [here](https://raw.githubusercontent.com/GaloisInc/crucible/master/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs). 

Here's a list of the ones suggested in the `FIXME` comment:
 - [x] llvm.ctlz
 - [x] llvm.cttz
 - [x] [llvm.bitreverse](https://llvm.org/docs/LangRef.html#llvm-bitreverse-intrinsics)
 - [x] [llvm.bswap](https://llvm.org/docs/LangRef.html#llvm-bswap-intrinsics)
 - [x] [llvm.ctpop](https://llvm.org/docs/LangRef.html#llvm-ctpop-intrinsic)
 - [x] llvm.sadd.with.overflow
 - [x] llvm.uadd.with.overflow
 - [x] llvm.ssub.with.overflow,
 - [x] llvm.usub.with.overflow
 - [x] llvm.smul.with.overflow
 - [x] llvm.umul.with.overflow,

siddharthist:
Here's a second list which may be helpful in verifying C++ code: 
 - [ ] [llvm.lifetime.start](https://llvm.org/docs/LangRef.html#llvm-lifetime-start-intrinsic)
 - [ ] [llvm.lifetime.end](https://llvm.org/docs/LangRef.html#llvm-lifetime-start-intrinsic)

(I'll probably edit this list as I find more).

siddharthist:
To support verification of C++ code, we'll need to handle the [exception handling intrinsics](https://llvm.org/docs/LangRef.html#exception-handling-intrinsics). Since exception handling control flow is difficult, we should likely advise users to ensure they use enough preconditions to ensure exceptional behavior doesn't arise, and implement an override for `throw` that just calls `abort`. 

NB: for C++ code that doesn't use _any_ exceptions, you can compile the libc++ with the following configuration:
```bash
      cmake ../llvm/projects/libcxx \
[...]

PR #78, opened 2018-08-30
implement the llvm.bswap.* intrinsics

siddharthist:
Progress on #73


what4/src/What4/Interface.hs
```diff
@@ -1971,6 +1982,64 @@ class (IsExpr (SymExpr sym), HashableF (SymExpr sym)) => IsExprBuilder sym where
...
+-- apply this newtype.
+newtype SymBV' sym w = MkSymBV' { getSymBV :: SymBV sym w }
+
+-- | Join a @Vector@ of smaller bitvectors
+bvJoinVector :: forall sym n w. (1 <= w, IsExprBuilder sym)
+             => sym
+             -> NatRepr w
+             -> Vector.Vector n (SymBV sym w)
+             -> IO (SymBV sym (n * w))
+bvJoinVector sym w = coerce $ Vector.joinWithM @IO @(SymBV' sym) @n bvConcat' w
```
robdockins: I think you can just use `getSymBV` here

what4/src/What4/Interface.hs
```diff
@@ -1971,6 +1982,64 @@ class (IsExpr (SymExpr sym), HashableF (SymExpr sym)) => IsExprBuilder sym where
...
+
+-- | Split a bitvector to a @Vector@ of smaller bitvectors
+bvSplitVector :: forall sym n w. (IsExprBuilder sym, 1 <= w, 1 <= n)
+              => sym
+              -> NatRepr n
+              -> NatRepr w
+              -> SymBV sym (n * w)
+              -> IO (Vector.Vector n (SymBV sym w))
+bvSplitVector sym n w x =
+  fmap (getSymBV @sym) <$>
```
robdockins: Here is where `coerce` would actually get you something

what4/src/What4/Interface.hs
siddharthist: That didn't work out.

what4/src/What4/Interface.hs
siddharthist: Yep, changed!

PR #97, opened 2018-09-27
Change handling of global variables

siddharthist:
**Note**: This is highly backwards-incompatible. SAW scripts that used to run will no longer with this change! Should we make a release note, or something similar? I've bumped the version so that library clients can select which behavior they need. 

Instead of providing a CFG that initializes all global variables, the result of translating a module now contains a map from symbols (the name of the global) to their initializers (represented as a constant LLVM value).

Clients to this API (e.g. saw-script) can now initialize globals at will by turning them into LLVM expressions.

Enables progress on https://github.com/GaloisInc/saw-script/issues/203.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
@@ -66,7 +66,6 @@
 --  * Alternate calling conventions
 ------------------------------------------------------------------------
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE EmptyDataDecls #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE ImplicitParams #-}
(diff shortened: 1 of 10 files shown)


siddharthist:
This isn't quite ready to merge, I want to find a better home for the `constToLLVMVal` function. Please make suggestions!

crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
```diff
@@ -152,298 +159,36 @@ import           Lang.Crucible.Types
...
+type GlobalInitializerMap = Map L.Symbol (Either String LLVMConst)
+
 -- | The result of translating an LLVM module into Crucible CFGs.
 data ModuleTranslation arch
    = ModuleTranslation
-      { cfgMap :: ModuleCFGMap arch
-      , initMemoryCFG :: C.SomeCFG (LLVM arch) EmptyCtx UnitType
+      { cfgMap        :: ModuleCFGMap arch
       , _transContext :: LLVMContext arch
+      , globalMap     :: GlobalInitializerMap
```
siddharthist: This should be renamed to `globalInitMap`, it conflicts with `Lang.Crucible.Simulator.SymGlobalState.globalMap`.

crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
```diff
@@ -152,298 +159,36 @@ import           Lang.Crucible.Types
...
+-- * The declaration is ill-typed
+-- * The global isn't linked ("extern global")
+--
+-- These failures are as granular as possible (attached to the values)
+-- so that simulation still succeeds if the module has a bad global that the
+-- verified function never touches.
+--
+-- To actually initialize globals, saw-script translates them into
+-- instances of @MemModel.LLVMVal@.
+type GlobalInitializerMap = Map L.Symbol (Either String LLVMConst)
```
siddharthist: This map should probably contain more information than it does (optimally, anything someone down the road would want to know about a given global). The least "lossy" way of doing this is probably
```haskell
type GlobalInitializerMap = Map L.Global (Either String LLVMConst)
```

PR #100, opened 2018-10-01
Remove constraint that floats are interpreted as reals.

andreistefanescu:


crucible-jvm/src/Lang/Crucible/JVM/Overrides.hs
```diff
@@ -513,20 +512,26 @@ showInt jty e = case W4.asSignedBV e of
...
               Nothing -> show $ W4.printSymExpr e
 
-showFloat :: (W4.IsExpr e) => e BaseRealType -> String
-showFloat e = case W4.asRational e of
+showFloat :: W4.IsExpr e => e tp -> String
+showFloat e = case floatAsRational e of
   Just rat -> show (fromRational rat :: Double)
   Nothing -> show $ W4.printSymExpr e
 
+floatAsRational :: W4.IsExpr e => e tp -> Maybe Rational
```
robdockins: Not really a fan of this approach.  We should instead make it so that we can retrieve concrete values from all the floating point interpretations by making it an operation in the interpreted floating point API.

sweirich:
I'm swamped with POPL PC meeting this week. However, I'm happy to see the simplified class constraints.

crucible-jvm/src/Lang/Crucible/JVM/Overrides.hs
andreistefanescu: I thought about doing that, but it would not be very useful. Because the IEEE floating-point operations are parametric in the number of bits in the exponent and the significant, and in the rounding mode, I haven't implemented constant folding. So it would only be possible to retrieve a constant if it was created as a constant, which is not the case in the JVM tests.

robdockins: OK.  Let's open a separate ticket for the constant-folding aspects then, because it is desirable and important that computations on concrete values yield concrete values.

But, I suppose it shouldn't hold up this pull.

PR #104, opened 2018-10-05
Add features for "active acknowledgment" to solver interactions.

robdockins:
This allows us to more closely synchronize with solvers if we induce
them to emit success acknowledgement messages after successful commands.

In addition, add support for entering "early unsat" states.  Yices, for example,
will sometimes detect early (i.e., before a "check-sat" command) that the context
is unsatisfiable when new facts are asserted.  In these cases, it immediately
prints "unsat" and enters unsat mode.  We need to detect this because it makes
certain other commands illegal (e.g. "push").  When we are in an early unsat
state, we simply short-cut all check commands and highjack the usual push/pop/reset
logic so that we only start interacting with the solver again once an appropriate
"pop" or "reset" puts us back into a potentially-satisfiable state.


robdockins:
Still some work to do here.

I'm trying to decide if it makes sense to set up the SMTLib2 flavor of solvers (Z3, CVC4, etc) to also use active acknowledgement.  This can be done by setting the `:print-success` option to true.

PR #106, opened 2018-10-09
Improve error message for undefined LLVM intrinsics

siddharthist:
Makes it clear that it is a system, rather than a user error, and provides a way for us to track which would be most helpful to implement.


siddharthist:
Perhaps predictably, here's what it looks like in practice:
```
Loading file "foo.saw"
Encountered unresolved LLVM intrinsic 'llvm.lifetime.start.p0i8'
Please report this on the following issue:
https://github.com/GaloisInc/crucible/issues/73
```

PR #107, opened 2018-10-09
[WIP] Skip overloaded llvm.lifetime.* intrinsics at translation time

siddharthist:
These intrinsics add valuable metadata, but it is not used at runtime and so can be safely skipped. They were only skipped in their non-overloaded forms, this PR generalizes that behavior to their overloaded use.

It is worth noting that they are not advertised as overloaded in the reference manual: https://llvm.org/docs/LangRef.html#llvm-lifetime-start-intrinsic.

cc: #73


siddharthist:
This change doesn't have the effect I intended, namely I'm still getting the following error:
```
Could not resolve function: llvm.lifetime.start.p0i8
```

robdockins:
@siddharthist, what's the status of this code?

siddharthist:
I have been dealing with another issue, though I expect I'll return to this soon. This isn't the right solution, though. I'll close until I have a better one.

PR #114, opened 2018-11-30
[WIP] Polymorphic CFGs

robdockins:
For discussion/review. Not ready to merge in it's current state.


crucible/src/Lang/Crucible/CFG/Core.hs
```diff
@@ -518,6 +534,193 @@ instance ExtendContext (TermStmt blocks ret) where
       extC  = extendContext  diff
 
 
+------------------------------------------------------------------------------------
+-- Instantiation / type substitution
+
```
robdockins: I suspect that a lot of this instantiation stuff can be reorganized to use the `traverseFC` methods, which would reduce significantly the amount of traversal boilerplate.

crucible/src/Lang/Crucible/CFG/Expr.hs
```diff
@@ -1360,3 +1399,207 @@ foldApp f0 r0 a = execState (traverseApp (go f0) a) r0
 mapApp :: TraversableFC (ExprExtension ext)
        => (forall u . f u -> g u) -> App ext f tp -> App ext g tp
 mapApp f a = runIdentity (traverseApp (pure . f) a)
+
+-------------------------------------------------------------------------------------
+-- Type Instantiation
+
+-- App :: Type -> (CrucibleType -> Type) -> CrucibleType -> Type
+type instance Instantiate subst App = App  
+instance (IsSyntaxExtension ext) => InstantiateFC (App ext) where
```
robdockins: Here, in particular, I think a traversal would eliminate most of this boilerplate.

crucible/src/Lang/Crucible/CFG/Reg.hs
```diff
@@ -676,12 +683,38 @@ data CFG ext s (init :: Ctx CrucibleType) (ret :: CrucibleType)
          , cfgNextLabel :: !Int
          -- ^ A number greater than any label ID appearing in the CFG.
          }
+   -- An instantiated CFG.
+   -- TODO: maybe we can fold this instantiation into the FnHandle type
+   | forall subst init' ret'.
+      (Instantiate subst init' ~ init, Instantiate subst ret' ~ ret) =>
+     ICFG { cfgIHandle :: !(FnHandle init' ret')
```
robdockins: I wonder... is there ever a good reason to have an instantiated version of a registerised CFG?

robdockins:
@sweirich, I haven't fully understood all the details of what's going on here, but I feel like this approach is probably worth continuing to pursue.  Some of the instantiation scaffolding and such could perhaps be generalized somewhat and pushed down into `parameterized-utils`, but generally I think the basic design is carving out a useful point in the design space.

sweirich:
Thanks. I will look at your comments more fully tomorrow, when I’m back on Galois work. 

I’ll also try to identify and point out any potential pitfalls for clients. (Did you look at my most recent commit?) 

Stephanie

crucible/src/Lang/Crucible/CFG/Core.hs
sweirich: That would be nice, but traverseFC does not have the right signature for this function. There's boilerplate here, and it would be good to automate it, but it will probably need its own TemplateHaskell to do so.

crucible/src/Lang/Crucible/CFG/Expr.hs
sweirich: I heartily agree. But note: there are a few special cases for reasoning about closed subcomponents.

crucible/src/Lang/Crucible/CFG/Reg.hs
sweirich: You're right. This isn't necessary.

robdockins:
I've lost track of the status of this branch a little bit.  Is this ready for review?

PR #125, opened 2018-12-18
Add support for What4 array memory stores.

andreistefanescu:


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -857,6 +928,21 @@ setMem sym w ptr val sz m =
...
+  :: (IsSymInterface sym, 1 <= w)
+  => sym
+  -> NatRepr w
+  -> LLVMPtr sym w {- ^ Pointer -}
+  -> SymArray sym (SingleCtx (BaseBVType w)) (BaseBVType 8) {- ^ Array value -}
+  -> SymBV sym w {- ^ Array size -}
+  -> Mem sym
+  -> IO (Pred sym, Mem sym)
+writeArrayMem sym w ptr arr sz m =
+  do p <- isAllocated sym w ptr sz m
```
robdockins: This probably needs to call `isAllocatedMut` instead

PR #131, opened 2019-01-03
strlen override

siddharthist:
Motivating CPP example: This works with SAW built with this commit, and doesn't without it. 
```cpp
  #include <string>

  int test_me() {
    std::string x = "foo";
    return x.length();
  }

  int main() {
    return test_me();
  }
```
with saw-script:
```javascript
  m <- llvm_load_module "./test-strlen.bc";

  let test_spec = do {
      crucible_execute_func [];
      crucible_return (crucible_term {{ 3 : [32] }});
  };

  crucible_llvm_verify m "_Z7test_mev" [] true test_spec z3;
```

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
@@ -177,6 +177,7 @@ do_register_overrides = do
    , register_llvm_override llvmCallocOverride
    , register_llvm_override llvmFreeOverride
    , register_llvm_override llvmReallocOverride
+   , register_llvm_override llvmStrlenOverride
    , register_llvm_override llvmPrintfOverride
    , register_llvm_override llvmPutsOverride
    , register_llvm_override llvmPutCharOverride
@@ -1450,6 +1451,25 @@ llvmPutsOverride =
   (KnownBV @32)
   (\memOps sym args -> Ctx.uncurryAssignment (callPuts sym memOps) args)
 
+llvmStrlenOverride
+  :: (IsSymInterface sym, HasPtrWidth wptr, wptr ~ ArchWidth arch)
+  => LLVMOverride p sym arch (EmptyCtx ::> LLVMPointerType wptr) (BVType 64)
+llvmStrlenOverride =
+  let nm = "strlen" in
+  LLVMOverride
+  ( L.Declare
+    { L.decRetType = L.PrimType $ L.Integer 64
+    , L.decName    = L.Symbol nm
+    , L.decArgs    = [ L.PtrTo $ L.PrimType $ L.Integer 8
+                     ]
+    , L.decVarArgs = False
+    , L.decAttrs   = []
+    , L.decComdat  = mempty
+    }
+  )
+  (Empty :> PtrRepr)
+  (KnownBV @64)
+  (\memOps sym args -> Ctx.uncurryAssignment (callStrlen sym memOps) args)
 
 llvmPrintfOverride
   :: (IsSymInterface sym, HasPtrWidth wptr, wptr ~ ArchWidth arch)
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
```diff
@@ -1450,6 +1451,25 @@ llvmPutsOverride =
   (KnownBV @32)
   (\memOps sym args -> Ctx.uncurryAssignment (callPuts sym memOps) args)
 
+llvmStrlenOverride
+  :: (IsSymInterface sym, HasPtrWidth wptr, wptr ~ ArchWidth arch)
+  => LLVMOverride p sym arch (EmptyCtx ::> LLVMPointerType wptr) (BVType 64)
```
robdockins: This should really use the `sizeT` combinators rather than fixing the size to 64 bits.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
```diff
@@ -1854,6 +1874,16 @@ callPuts sym mvar
...
 
+callStrlen
+  :: (IsSymInterface sym, HasPtrWidth wptr)
+  => sym
+  -> GlobalVar Mem
+  -> RegEntry sym (LLVMPointerType wptr)
+  -> OverrideSim p sym (LLVM arch) r args ret (RegValue sym (BVType 64))
+callStrlen sym mvar (regValue -> strPtr) = do
+  mem <- readGlobal mvar
+  len <- liftIO $ length <$> loadString sym mem strPtr Nothing
```
robdockins: It's worth noting that `loadString` only works for concrete strings (all bytes are symbolic), which means this `strlen` will only work on concrete strings.  In principle, we could make a `strlen` that works with symbolic data, but it would be quite a bit more complicated.

siddharthist: Indeed, though I get a fairly reasonable error message:
```javascript
m <- llvm_load_module "./test-strlen.bc";

let alloc_init ty v = do {
    p <- crucible_alloc ty;
    crucible_points_to p v;
    return p;
};

let ptr_to_fresh n ty = do {
    x <- crucible_fresh_var n ty;
    p <- alloc_init ty (crucible_term x);
    return (x, p);
};

let test_spec = do {
    (x, p) <- ptr_to_fresh "x" (llvm_array 10 (llvm_int 8));
    crucible_execute_func [p];
    crucible_return (crucible_term {{ 10 : [64] }});
};

crucible_llvm_verify m "_Z7test_mePc" [] true test_spec z3;
```
```cpp
#include <string.h>

size_t test_me(char *x) {
  return strlen(x);
}

int main() {
  return 0;
}
```
I get this error from SAW (which is decently self-explanatory):
```
Loading file "/home/siddharthist/code/fizz-hkdf/cpp-tests/test-strlen.saw"
user error ("crucible_llvm_verify" (/home/siddharthist/code/fizz-hkdf/cpp-tests/test-strlen.saw:21:1-21:21):                                                 
Symbolic execution failed.
  Unsupported feature: Symbolic value encountered when loading a string
  in _Z7test_mePc at :4:10
Stack frame
  Allocations:

  Writes:

Base memory
  Allocations:
    HeapAlloc 5 0xa:[64] Mutable <malloc>
    GlobalAlloc 4 0x0:[64] Immutable strlen
    GlobalAlloc 3 0x0:[64] Immutable main
    GlobalAlloc 2 0x0:[64] Immutable llvm.dbg.value
    GlobalAlloc 1 0x0:[64] Immutable _Z7test_mePc
  Writes:
    *(5, 0x0:[64]) := [c@2:bv
                      ,c@3:bv
                      ,c@4:bv
                      ,c@5:bv
                      ,c@6:bv
                      ,c@7:bv
                      ,c@8:bv
                      ,c@9:bv
                      ,c@10:bv
                      ,c@11:bv]
```

PR #135, opened 2019-01-09
Better error messages for undefined behavior

siddharthist:
Errors now cite the relevant documents, and generally have more information about exactly what went wrong.

Here's an example program and the associated error message:
```c
#include <stdlib.h>

int undefined(long* x) {
  int *y;
  y = malloc(1*sizeof(int));
  if (x + 7 <= y + 3) {
    return 1;
  }
  return 0;
}

int main() {
  long *x;
  x = malloc(1*sizeof(long));
  return undefined(x);
}
```
```
Symbolic execution failed.
  Undefined behavior:
  Reference: C99: 6.5.6 Additive operators
  Addition of an offset to a pointer that resulted in a pointer to an address outside of the allocation.                                                     
  (4, 0x0:[64])
  0x38:[64]
  in main at internal
```

TODO:
 - [x] Import and use this module in `MemModel.Pointer`
 - [ ] Use `UB.Config` to toggle what kinds of undefined behaviors are checked for
 - [x] Bring this information into the translation phase?

### DIFF
--- a/crucible-c/src/Overrides.hs
+++ b/crucible-c/src/Overrides.hs
@@ -21,7 +21,7 @@ import Data.Parameterized.Context(pattern Empty, pattern (:>))
 
 
 import What4.FunctionName(functionNameFromText)
-import What4.Symbol(userSymbol)
+import What4.Symbol(userSymbol, emptySymbol)
 import What4.Interface
           (freshConstant, bvLit, bvEq, asUnsignedBV,notPred
           , getCurrentProgramLoc)
@@ -43,6 +43,7 @@ import Lang.Crucible.Simulator.OverrideSim
         , registerFnBinding
         , getOverrideArgs
         , readGlobal
+        , writeGlobal
         )
 import Lang.Crucible.Simulator.SimError (SimErrorReason(..))
 import Lang.Crucible.Backend
@@ -53,9 +54,11 @@ import Lang.Crucible.LLVM.Translation
         , symbolMap
         , llvmMemVar
         )
+import Lang.Crucible.LLVM.DataLayout
+  (noAlignment)
 import Lang.Crucible.LLVM.MemModel
   (Mem, LLVMPointerType, pattern LLVMPointerRepr,loadString,HasPtrWidth,
-   llvmPointer_bv, projectLLVM_bv)
+   llvmPointer_bv, projectLLVM_bv, doArrayStore)
 
 import Lang.Crucible.LLVM.Extension(LLVM)  
 import Lang.Crucible.LLVM.Extension(ArchWidth)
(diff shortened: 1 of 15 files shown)


siddharthist:
@robdockins See what you think about the latest commits re: adding UB to the translation phase. There are a few other places it could happen, notably in `raw_bitop`, `translateConversion`, and in the whole `Constant` module.

robdockins:
Looks good to me.  We should do the same for `raw_bitop`.  I'm less concerned about the constant evaluation.  There's a few odd one-off things to look at as well, e.g. the overflow checks in `getElementPointer` and the bounds checks in `extractElt`.  We should also double check a few things.  For example, `bitcast` currently raises an error at runtime if asked to cast between incompatible types; maybe we should report that error at translation time instead.

siddharthist:
@robdockins I'm confused about `getelementpointer`. [The LLVM documentation says](https://llvm.org/docs/LangRef.html#id225)
> If the inbounds keyword is not present, the offsets are added to the base address with silently-wrapping two’s complement arithmetic.

Wouldn't that make it _not_ undefined behavior to have multiplication overflow? Or do we assume that the resulting pointer is out-of-bounds of the allocation it's supposed to point to? Or is it the case that the `inbounds` keyword is usually present?

By the way, here's an example of an even more verbose message from the recent commits:
```
  Undefined behavior encountered:
  Addition of an offset to a pointer resulted in a pointer to an address outside of the allocation.                                                          
  Reference:  The C language standard, version C99 §6.5.6 Additive operators, ¶8                                                                             
  Document URL: http://www.iso-9899.info/n1570.html
  Pointer: (4, 0x0:[64])
  Offset: 0x38:[64]
  in main at internal
```

robdockins:
Yeah, we are currently translating every GEP as though it had the `inbounds` flag set.  We can probably relax this somewhat.  We'll just have to convince ourselves that the arithmetic all works out across the relationship between the segmented memory model of the simulator and the flat memory model of the implementation.  I think it should...

siddharthist:
@robdockins I've added all the examples you pointed out. Any others off the top of your head? 

It doesn't look like we currently check for poison in `fptoui`, `fptosi`, ... (in `translateConversion`), so I'll make a separate ticket for that.

robdockins:
Just searching `crucible-llvm` for `assertExpr`, looks like we missed some in `insertElt` and `pointerCmp`.

There's also a pretty pervasive `assertExpr` inside `pointerAsBitvectorExpr`.  Not sure what to do about that one...

siddharthist:
I'm not sure we're doing the right thing for `insertElt`. From the LLVM spec, it looks like that's a poison value, whereas we are asserting its validity immediately (more like `assertUndefined` than `poisonSideCondition`). Is that your impression as well?

robdockins:
That... is probably true yeah.  The distinction between those things is actually pretty academic, but I suppose we should try to do the right thing.

siddharthist:
I handled `insertElt`, `extractElt`, and `pointerCmp`. What is the `pointerAsBitvectorExpr` assertion actually asserting? 

Is it the case that `pointerAsBitvectorExpr` is actually saying that "the given pointer was a (non-pointer) bitvector all along"?

robdockins:
Yeah, we use that coercion in situations where we know that we must have a raw (i.e., nonpointer) integer value.

siddharthist:
Hmm. I'm also not sure what to do about that one. What say we merge this for now and figure that out later?

robdockins:
Agree.  I don't see any reason to hold up the merge.

PR #141, opened 2019-01-22
Extend LLVM memory model to handle allocations and arrays of unbounded size

jpaykin:
This branch extends the LLVM memory model to be able to allocate spaces of unbounded size, which can be used to simulate one unbounded virtual memory region with reads and writes that may overlap. This was useful in simulating the behavior of machine code instructions in SemMC.

This pull request adds too new operations to the API in `LLVM.MemModel` but otherwise does not change the interface.

In the internal model `LLVM.MemModel.Generic`, the constructors `Malloc` and `MemArrayStore` now encode the size not  as a bitvector, but as a `Maybe` bitvector, where `Nothing` represents an unbounded region/array. Since `IntExpr`s can refer to the size of an array, their interpretation as `SymBV`s must now return a `Maybe SymBV`.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -453,8 +453,9 @@ ptrMessage msg ptr ty =
           , "  at type " ++ show (G.ppType ty)
           ]
 
--- | Load a 'RegValue' from memory. Also assert that the pointer is
--- valid and aligned, and that the loaded value is defined.
+-- | Load a 'RegValue' from memory.
+-- 
+-- Precondition: the pointer is valid and aligned, and the loaded value is defined.
 doLoad ::
   (IsSymInterface sym, HasPtrWidth wptr) =>
   sym ->
@@ -468,8 +469,9 @@ doLoad sym mem ptr valType tpr alignment = do
   --putStrLn "MEM LOAD"
   unpackMemValue sym tpr =<< loadRaw sym mem ptr valType alignment
 
--- | Store a 'RegValue' in memory. Also assert that the pointer is
--- valid and points to a mutable memory region.
+-- | Store a 'RegValue' in memory.
+--
+-- Precondition: the pointer is valid and points to a mutable memory region.
 doStore ::
   (IsSymInterface sym, HasPtrWidth wptr) =>
   sym ->
@@ -506,7 +508,8 @@ sextendBVTo sym w w' x
 
 
 -- | Allocate and zero a memory region with /size * number/ bytes.
--- Also assert that the multiplication does not overflow.
+--
+-- Precondition: the multiplication /size * number/ does not overflow.
 doCalloc ::
   (IsSymInterface sym, HasPtrWidth wptr) =>
   sym ->
@@ -551,7 +554,7 @@ doMallocUnbounded
   -> IO (LLVMPtr sym wptr, MemImpl sym)
 doMallocUnbounded = doMallocSize Nothing
 
-doMallocSize 
+doMallocSize
   :: (IsSymInterface sym, HasPtrWidth wptr)
   => Maybe (SymBV sym wptr) {- ^ allocation size -}
   -> sym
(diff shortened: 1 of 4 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -130,52 +133,92 @@ data MemWrite sym
...
+  OffsetExpr ->
+  IO (SymBV sym w)
+genOffsetExpr sym w f@(ExprEnv load store _size) expr =
   case expr of
     OffsetAdd pe ie -> do
       pe' <- genOffsetExpr sym w f pe
       ie' <- genIntExpr sym w f ie
-      bvAdd sym pe' ie'
+      case ie' of
+        Nothing -> error "Cannot construct an offset that references the size of an unbounded region"
```
robdockins: We should use `panic` here instead of `error` if we believe that this is a programming bug/error for this situation to occur.  If, instead, we believe that a user input could elicit this situation, we need some more robust error handling.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -315,6 +358,16 @@ evalMuxValueCtor sym w end vf subFn (MuxTable a b m t) =
...
 
+-- | Throws an error if passed an unbounded/@Nothing@ bitvector.
+maybePtrAdd :: (1 <= w, IsExprBuilder sym)
+            => sym
+            -> NatRepr w
+            -> LLVMPtr sym w
+            -> Maybe (SymBV sym w)
+            -> IO (LLVMPtr sym w)
+maybePtrAdd sym w ptr (Just bv) = ptrAdd sym w ptr bv
+maybePtrAdd _   _ _   Nothing   = error "Cannot use an unbounded bitvector as an offset"
```
robdockins: Similar here, user `panic` or something more robust

robdockins:
Looks good to me, except the uses of `error` throughout.  We should decide to wether to treat these as internal errors (call `panic`) or potentially-user visible errors and throw an appropriate IO exception or somesuch.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
jpaykin: I think it should be a bug for this to occur; there are specific times that `StoreSize` is referenced in an `OffsetBV`, for example in `Crucible.LLVM.MemModel.Common.symbolicRangeLoad`, but this function is only invoked (in `MemModel.Generic`) when the size is bounded. Because these `OffsetExpr`s are constructed in a separate place from where they are interpreted, however, this property is hard to verify.

I will change this to a `panic`.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1037,6 +1096,8 @@ unpackMemValue sym tpr (LLVMValZero tp) = unpackZero sym tp tpr
 unpackMemValue _sym (LLVMPointerRepr w) (LLVMValInt blk bv)
   | Just Refl <- testEquality (bvWidth bv) w
   = return $ LLVMPointer blk bv
+unpackMemValue _sym (BVRepr w) (LLVMValInt _ bv)
+  | Just Refl <- testEquality w (bvWidth bv) = return bv
```
robdockins: I just noticed this, and I'm skeptical.  What's the motivation here? Unpacking a load directly toward a `BVType` is something we I don't think we should be doing; and if we do decide to do that for some reason, we should be asserting that the block number is 0.

jpaykin: Ok, I did not understand this use-case correctly. As we just talked about offline, we should never be calling `unpackMemValue` (through e.g. `doLoad`) and so this case should instead be throwing a panic. Instead, I will add some documentation on top of `doLoad` to explain the correct usage.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Types.hs
```diff
@@ -111,7 +111,7 @@ llvmTypeAsRepr xs f = go (llvmTypeToRepr xs)
        go []       = f UnitRepr
        go [Some x] = f x
 
-       go _ = error $ unwords ["llvmTypesAsRepr: expected a single value type", show xs]
+       go _ = error $ unwords ["llvmTypeAsRepr: expected a single value type", show xs]
```
siddharthist: This looks like a place we could use `panic`

robdockins: Agreed, but I don't think it should hold up this PR.

PR #144, opened 2019-01-23
Add an undef constructor to LLVMConst and LLVMVal

### LINKED ISSUE #139: undef global constants
`libc++` uses a global constant value [`piecewise_construct`](https://en.cppreference.com/w/cpp/utility/piecewise_construct) of [an empty struct type](https://en.cppreference.com/w/cpp/utility/piecewise_construct_t) to essentially "tag" calls to [the constructor for `pair`](https://en.cppreference.com/w/cpp/utility/pair/pair). This value shows up in the bitcode of `libc++` as an [`undef` constant](https://llvm.org/docs/LangRef.html#undefined-values):
```llvm
@_ZNSt3__1L19piecewise_constructE = internal constant %"struct.std::__1::piecewise_construct_t" undef, align 1, !dbg !17944
```
(modulo debug info, that exact definition shows up in a bunch of files). It is not defined to be any specific value in `libc++` or `libc++abi`. 

Since this is a constant, `saw-script` will try to initialize it with [`LLVM.Globals.populateConstGlobals`](https://github.com/GaloisInc/crucible/blob/2c1f943b57fca52d32ca2915e582faef07f5ee8f/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs#L244), which [will fail](https://github.com/GaloisInc/crucible/blob/2c1f943b57fca52d32ca2915e582faef07f5ee8f/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs#L361). 

The fact that Clang outputs an `undef` value means that the behavior of the program is not supposed to rely on the specific bit pattern used:
> Undefined values are useful because they indicate to the compiler that the program is well defined no matter what value is used.

I think there are two reasonable approaches to taking care of this:
 - Don't fail when initializing a constant global fails (probably emit a warning)
 - Initialize `undef` values to fully symbolic values (to guarantee that the behavior of the program is sound regardless of the bit pattern used)

siddharthist:
This allows for initializing `undef` globals, though it will still fail if the simulation does very much with them. It's a first step, we can see later if more support for manipulating `undef` values is necessary. 

Fixes #139, see motivation there.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -1033,7 +1033,7 @@ unpackMemValue ::
   LLVMVal sym ->
   IO (RegValue sym tp)
 
-unpackMemValue sym tpr (LLVMValZero tp) = unpackZero sym tp tpr
+unpackMemValue sym tpr (LLVMValZero tp)  = unpackZero sym tp tpr
 
 unpackMemValue _sym (LLVMPointerRepr w) (LLVMValInt blk bv)
   | Just Refl <- testEquality (bvWidth bv) w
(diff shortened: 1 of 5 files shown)


siddharthist:
Agreed. I think it raises a lot of tricky questions, and we can see if they end up popping up.

PR #145, opened 2019-01-23
 Support for LLVM's llvm.global_ctors 

siddharthist:
This is necessary for C++ support, specifically `<iostream>` uses this. 

See https://llvm.org/docs/LangRef.html#the-llvm-global-ctors-global-variable

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Ctors.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Ctors.hs
@@ -1,0 +1,180 @@
+ ------------------------------------------------------------------------
+-- |
+-- Module           : Lang.Crucible.LLVM.Ctors
+-- Description      : Extract and manipulate the @llvm.global_ctors@ variable
+-- Copyright        : (c) Galois, Inc 2018
+-- License          : BSD3
+-- Maintainer       : Rob Dockins <rdockins@galois.com>
+-- Stability        : provisional
+------------------------------------------------------------------------
+{-# LANGUAGE DataKinds             #-}
+{-# LANGUAGE DeriveDataTypeable    #-}
+{-# LANGUAGE DeriveGeneric         #-}
+{-# LANGUAGE FlexibleContexts      #-}
+{-# LANGUAGE GADTs                 #-}
+{-# LANGUAGE ImplicitParams        #-}
+{-# LANGUAGE OverloadedStrings     #-}
+{-# LANGUAGE RankNTypes            #-}
+{-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TypeOperators         #-}
+
+module Lang.Crucible.LLVM.Ctors
+  ( Ctor(..)
+  , globalCtors
+  , callCtors
+  , callAllCtors
+  , callCtorsCFG
+  ) where
+
+import           Data.Data (Data)
+import           Data.Typeable (Typeable)
+import           GHC.Generics (Generic)
+
+import           Control.Monad.Except as Except
+import           Data.List (find, sortBy)
+import           Data.Ord (comparing, Down(..))
+import           Data.Maybe (fromMaybe)
+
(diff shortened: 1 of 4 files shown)


siddharthist:
Hmm, the generated CFG doesn't seem to be doing what I'd expect on an LLVM module where this variable is non-null:
```
llvm_global_ctors:
  $0 = emptyApp() % internal
llvm_global_ctors:
  return $0 % internal
```
There's a bit more debugging to be done before this is merged.

siddharthist:
Okay, the issue was that I meant to `fail` in the `LLVMGenerator` monad, but I called `fail` in the list monad, which silently ate it. The latest commit fixes the issue, I've tested this with SAW:
```
llvm_global_ctors:
  $0 = resolveGlobal llvm_memory "_GLOBAL__sub_I_HkdfTest.cpp" % internal
llvm_global_ctors:
  $1 = loadFunctionHandle llvm_memory $0 as FunctionHandleRepr [] UnitRepr % internal
llvm_global_ctors:
  $2 = call $1(); % internal
_GLOBAL__sub_I_HkdfTest.cpp:
  $0 = pushFrame llvm_memory % internal
```

crucible-llvm/src/Lang/Crucible/LLVM/Ctors.hs
```diff
@@ -0,0 +1,178 @@
...
+        fromMaybe
+          (throwError $ unlines $ [ "Ill-typed value in llvm.global_ctors: "
+                                  , show v
+                                  ])
+          (pure <$> extractCtors v)
+
+      -- Sort the values by priority, highest to lowest.
+      pure (sortBy (comparing (Down . ctorPriority)) vs')
+
+    Nothing -> throwError "Couldn't find global llvm.global_ctors"
```
robdockins: Is this an error condition? If there aren't any ctors, does clang emit this variable at all?

crucible-llvm/src/Lang/Crucible/LLVM/Ctors.hs
```diff
@@ -0,0 +1,178 @@
...
+----------------------------------------------------------------------
+-- ** callCtors
+
+-- | Call some or all of the functions in @llvm.global_ctors@
+callCtors :: (Ctor -> Bool) -- ^ Filter function
+          -> L.Module
+          -> LLVMGenerator h s arch UnitType (Expr (LLVM arch) s UnitType)
+callCtors select mod_ = do
+  ctors <- either fail (pure . filter select) (globalCtors mod_)
+  _ <- forM ctors $ \ctor ->
```
robdockins: Probably should prefer `forM_` here

robdockins:
There's a strange comment in the LLVM docs about only running `ctors` if the associated `data` pointer points to something that isn't discarded... I'm not sure exactly what that means for our setting. Do you?  If we get a fully linked `.bc` file that points to an `extern` constant, is that "discarded"?

crucible-llvm/src/Lang/Crucible/LLVM/Ctors.hs
siddharthist: Good question. The motivation here is that a user can toggle whether this function is getting called in saw-script, and it won't be by default. Thus, if the user enables `global_ctors` and they aren't present, that seems like a user error. 

We could change this so that the caller passes in an `L.Global`, but we would have to catch the case where they pass in the wrong one... What do you think?

siddharthist:
@robdockins Good point! The following line handles the case that it is null, but this code doesn't currently check if the third pointer points to a function or global in the module. I can fix that. 

https://github.com/GaloisInc/crucible/pull/145/files#diff-e32f54b4cec6dd8247464f3ab92d18adR98

robdockins:
Should we go ahead and merge this? Maybe open a corresponding ticket about the discarded symbols thing...

siddharthist:
Yeah, let's do that. This should work for a reasonable number of cases.

PR #156, opened 2019-02-12
libc++ overrides

siddharthist:
Some refactoring of how we register overrides (breaking into a few modules), and our first `libc++` overrides.


siddharthist:
This doesn't quite seem to do what I want: the override is certainly being registered for the correct function, but that function is then called just like normal (as if there were no override).

siddharthist:
Looks like the problem is that SAWscript registers the overrides in the opposite order I'd want them:
https://github.com/GaloisInc/saw-script/blob/cfdd7ec878ffb6594d0d17a1ac12b70fa5c249c5/src/SAWScript/CrucibleBuiltins.hs#L705-L710

PR #161, opened 2019-02-13
Allowing address space overflow in the LLVM memory model

jpaykin:
In the LLVM memory model, in the case that we've allocated an unbounded memory region, it was previously impossible to load from the last addressable byte in memory, or in general to read memory from addresses that overflow the address space. (For example if I want to read 4 bytes starting at address (-2).) 

This branch changes the checks in `fixedOffsetRangeLoad` so that if this situation occurs---when the load offset plus the size of data to read is greater than the load offset itself---then we just create this value range. However, it was also necessary to remove the overflow check in the allocation check in `Generic.hs`, so before merging I wanted to get get feedback about if these checks are truly necessary and, if so, if there is another approach I could consider.


robdockins:
Do any ISAs actually let you read from the very top of the address space and wrap around? I'd kind of be shocked... certainly such a read would not be aligned.

I'd expect reading the last byte in memory should be possible, but only if you read a single byte... which still wouldn't overflow.

jpaykin:
I have no reason to think they do, but it's more like in our situation (synthesis) it's easier not to rule it out. If we can read up to the last byte, it might be possible for us to manage the overflow in the semmc code rather than in the memory model, if you think it's better not to allow that overflow in the memory model itself. We could just detect that case early and make two calls to read, for example: one at the top chunk of the address space and one at the bottom. Is that a better compromise?

Overflow doesn't currently run into any alignment issues if `noAlignment` is chosen; do more realistic alignment strategies rule this case out? If so, maybe the alignment checks already enforce this property?

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Common.hs
```diff
@@ -244,11 +244,18 @@ rangeLoad lo ltp s@(R so se)
...
+-- when the load and store offsets are concrete and the store size is bounded
 fixedOffsetRangeLoad :: Addr
+                     -- ^ Address of load
                      -> StorageType
+                     -- ^ Type to load
                      -> Addr
+                     -- ^ Address of store
                      -> Mux (ValueCtor (RangeLoad Addr Addr))
 fixedOffsetRangeLoad l tp s
+  | le < l = -- the load crosses a boundary
```
robdockins: OK, we need to think about this situation a bit more.  Right now, we're using `Word64` to represent concrete addresses, and relying on wrap-around behavior of those bakes in a 64-bit assumption on the memory model.

So, we should figure out how to generalize this the right way, and also see if we need to additionally update the symbolic load operations.

jpaykin: Changing Bytes to natural numbers means that this extra logic is not necessary any more

jpaykin:
After some discussion, it seems like the best choice for now is to disallow overflows in both bounded and unbounded cases, but to make sure the overflow check doesn't rule out addressing up to the last byte. If others are happy with the state of this code, I think it is ready to be merged with master.

PR #168, opened 2019-02-26
Replce Natural with Integer in LLVM Bytes.

andreistefanescu:
This avoids the arithmetic underflow error in the s2n saw proofs.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Bytes.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Bytes.hs
@@ -27,20 +27,20 @@ import Numeric.Natural
 -- | A newtype for expressing numbers of bytes.
 --   This newtype is explicitly introduced to avoid confusion
 --   between widths expressed as numbers of bits vs numbers of bytes.
-newtype Bytes = Bytes { unBytes :: Natural }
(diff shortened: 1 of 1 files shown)


joehendrix:
`Bytes` is definitely used in places where the negative values do not make sense (e.g. number of bytes a type uses in memory or offsets of fields in structs).  I'd be inclined to keep `Bytes` as a `Natural` and stop using `Bytes` in places where you want to allow negative values.  One could just use `Integer` directly.

robdockins:
I'm with Joe, we really shouldn't be doing arithmetic on `Bytes` values that might be negative.  We should figure out the root cause here and fix it.  I'm inclined to delete the `Num` instance for `Bytes` and see what breaks.

robdockins:
Ah, but perhaps `Offset` needs to become a separate type that _is_ backed by integer.

robdockins:
Ugh, OK, doing this right is going to take some work.  There are a lot of offset calculations down inside the memory model which do not obviously preserve the positiveness of `Bytes` values.

I think we should do this work, but let's not leave downstream proofs broken in the meantime.  I'm in favor of merging, for now, but we should really get this straightened out.

PR #171, opened 2019-03-01
 Allow globals to recursively populate one another

### LINKED ISSUE #170: Symbols are looked up while constant globals are initialized
SAWscript initializes constant globals in memory. During this process, `populateConstGlobal` calls `constToLLVMVal`:
https://github.com/GaloisInc/crucible/blob/d0c8f1e1a712e779f9995363689b352ce10b17dd/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs#L279
which, for `SymbolConst`s, calls `doResolveGlobal`:
https://github.com/GaloisInc/crucible/blob/d0c8f1e1a712e779f9995363689b352ce10b17dd/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1298-L1302
However, if the symbol in question is another constant global, it won't already reside in memory! We need to have `populateConstGlobal` recursively call itself when it encounters `Symbol`s.

siddharthist:
Fixes #170 

Also adds some tests.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
@@ -73,6 +73,7 @@ import           Lang.Crucible.LLVM.TypeContext
 import           Lang.Crucible.Backend (IsSymInterface)
 import           Lang.Crucible.Panic (panic)
 
+import           GHC.Stack
 
 ------------------------------------------------------------------------
 -- GlobalInitializerMap
@@ -270,7 +271,9 @@ populateGlobal :: forall sym wptr.
   ( ?lc :: TypeContext
   , 16 <= wptr
   , HasPtrWidth wptr
-  , IsSymInterface sym ) =>
+  , IsSymInterface sym
+  , HasCallStack
+  ) =>
   sym ->
   L.Global {- ^ The global to populate -} ->
   MemType {- ^ Type of the global -} ->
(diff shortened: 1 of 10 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
```diff
@@ -260,21 +267,57 @@ populateConstGlobals = populateGlobals f
...
+         populateRec symbol = do
+           memimpl0 <- get
+           memimpl <-
+            case Map.lookup symbol (memImplGlobalMap mem) of
+              Just _  -> pure memimpl0 -- We already populated this one
+              Nothing ->
+                -- For explanations of the various modes of failure, see the
+                -- comment on 'GlobalInitializerMap'.
+                case Map.lookup symbol giMap of
+                  Nothing -> panic "populateGlobal" $
```
robdockins: Should these error conditions actually be `panic`s? We usually reserve those for things that can only occur as a result of programming bugs, but these seem like they could be caused by malformed inputs.

siddharthist: Oh, I suppose that's right. I was being generous in assuming LLVM wouldn't give us something malformed :-) What should we use? Maybe `fail`? It seems pretty unrecoverable.

robdockins: Yeah, `fail` is probably the best thing for now... I think that's what we do when we encounter type errors and such in LLVM translation.  The difference is that panic will crash all the way out of e.g. SAW, whereas a `fail` will get caught and dump you back at the repl.

PR #176, opened 2019-03-16
What4 refactoring

robdockins:
The logical representation for bitvectors and boolean values is enhanced, as is the semiring product representation for all semirings.

In particular, bitvectors now participate in the semiring machinery in two ways: the arithmetic semrinng with addition and multiplication, and the bitwise semring with XOR and AND.  This allows the gathering of constants, rearranging of terms, etc. and automatically implements associative and commutative rewriting for all the affected operations.

In addition, I've enhanced the representation of boolean predicates to use negation and n-way AND and OR operations instead of just negation and binary AND, as previously.  This automatically implements associative, commutative and idemptoency laws for AND and OR.  Additionally, we get absorption and resolution rules without much effort.  In addition, the direct encoding of disjunctions means that we no longer have a noisy deMorgan translation of ORs, which is nice for human-readability purposes.


robdockins:
This is a fairly fundamental change, so I'd like to have some additional eyes on it before merging.

what4/src/What4/Expr/BoolMap.hs
```diff
@@ -0,0 +1,170 @@
...
+--   either a conjunction or a disjunction of expressions.  The
+--   implementation uses a map from expression values to their
+--   polarities, and thus automatically implements the associative,
+--   commutative and idempotencey laws common to both conjunctions and
+--   disjunctions.  Moreover, if the same expression occurs in the
+--   collection with opposite polarities, the entire collection
+--   collapses via a resolution step to an \"inconsistent\" map.  For
+--   conjunctions this corresponds to a contradiction and
+--   represents false; for disjunction, this corresponds to the law of
+--   the exlcuded middle and represents true.
```
siddharthist:
```suggestion
--   the excluded middle and represents true.
```

what4/src/What4/Expr/BoolMap.hs
```diff
@@ -0,0 +1,170 @@
...
+instance Hashable Polarity where
+  hashWithSalt s Positive = hashWithSalt s (0::Int)
+  hashWithSalt s Negative = hashWithSalt s (1::Int)
+
+-- | Swap a polarity value
+negatePolarity :: Polarity -> Polarity
+negatePolarity Positive = Negative
+negatePolarity Negative = Positive
+
+newtype Wrap (f :: k -> Type) (x :: k) = Wrap { unWrap:: f x }
```
siddharthist: This seems like something that might belong in parameterized-utils, in that it has little to do with SMT solvers, and is more generally about working with type-level stuff.

robdockins: Yeah, it's true... but I don't  have a good sense of what API it should have, etc. and the name really sucks.

PR #179, opened 2019-04-01
Breakpoints

andreistefanescu:


crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
```diff
@@ -1090,3 +1091,6 @@ transConstantExpr expr = case expr of
  where
  badExp :: String -> m a
  badExp msg = throwError $ unlines [msg, show expr]
+
+testBreakpointFunction :: String -> Maybe String
```
robdockins: Seems a little weird to put this here?

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
```diff
@@ -1706,11 +1706,40 @@ callFunctionWithCont tailCall_ fnTy fn args assign_f k
...
+     , Just bpNm <- testBreakpointFunction nm = do
+        some_val_args <- mapM typedValueAsCrucibleValue args
+        case Ctx.fromList some_val_args of
+          Some val_args -> do
+            addBreakpointStmt (Text.pack bpNm) val_args
+            k
+
      | otherwise = callFunction tailCall_ fnTy fn args assign_f >> k
 
+typedValueAsCrucibleValue
```
robdockins: Let's pick a better name here, since this is specific to breakpoints

crucible/src/Lang/Crucible/CFG/ExtractSubgraph.hs
```diff
@@ -53,6 +53,7 @@ extractSubgraph (CFG{cfgBlockMap = orig, cfgEntryBlockID = _origEntry}) cuts bi
             { cfgBlockMap = bm
             , cfgEntryBlockID = entryID
             , cfgHandle = hn
+            , cfgBreakpoints = undefined
```
robdockins: :-(

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
```diff
@@ -273,6 +278,20 @@ extSwitchInfo :: SwitchInfo s blocks tp -> SwitchInfo s (blocks::>args) tp
...
 
+extBlockInfo
+  :: BlockInfo ext s ret blocks
+  -> BlockInput ext s (blocks ::> args) ret args
+  -> BlockInfo ext s ret (blocks ::> args)
+extBlockInfo bi binput = do
+  let blocks' = extBlockInputAssignment $ biBlocks bi
+  let jump_info' = extJumpInfoMap $ biJumpInfo bi
+  let switch_info' = extSwitchInfoMap $ biSwitchInfo bi
+  let breakpoints' = Bimap.mapR (mapSome C.extendBlockID) $ biBreakpoints bi
```
robdockins: This seems like it will rebuild the breakpoint map frequently. Is this a case where we can `unsafeCoerce` the whole map when `UNSAFE_OPS` is enabled?

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
```diff
@@ -377,46 +398,79 @@ inferBlockInfo blocks = seq input_map $ go bi0 blocks
...
+                  let bi' = extBlockInfo bi binput
                   let si = SwitchInfo block_id crepr ra
-                  let bi' = BI { biBlocks = extend blocks' binput
-                               , biJumpInfo = jump_info'
-                               , biSwitchInfo = insertSwitchInfo l si switch_info'
-                               }
-                  go bi' rest
+                  let bi'' = bi' { biSwitchInfo = insertSwitchInfo l si (biSwitchInfo bi') }
+                  gogo bi'' rest
+        gogo :: BlockInfo ext s ret blocks
```
robdockins: More than one `go` function probably means we need better names for both of them...

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
```diff
@@ -126,24 +131,24 @@ data BlockInput ext s blocks ret args
              -- | Arguments expected by block.
            , binputArgs       :: !(Assignment (Value s) args)
            , binputStmts      :: !(Seq (Posd (Stmt ext s)))
-           , binputTerm       :: !(Posd (TermStmt s ret))
+           , binputTerm       :: !(Posd (Either (JumpInfo s blocks) (TermStmt s ret)))
```
robdockins: This `Either` is pretty inscrutible... let's define a new datatype instead with more descriptive constructor names.

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
```diff
@@ -826,8 +883,15 @@ resolveStmts nm bi sz reg_map bindings appMap (Posd p s0:rest) t = do
                            (resolveAtom reg_map m))
                  (resolveStmts nm bi sz reg_map bindings appMap rest t)
 
+    Breakpoint{} -> error $
```
robdockins: Let's have a comment explaining why we don't expect breakpoints to make it here

crucible/src/Lang/Crucible/Simulator/Breakpoint.hs
```diff
@@ -0,0 +1,74 @@
+-----------------------------------------------------------------------
+-- |
+-- Module           : Lang.Crucible.Simulator.Breakpoint
+-- Description      : Support for symbolic execution breakpoints
+-- Copyright        : (c) Galois, Inc 2019
+-- License          : BSD3
+-- Maintainer       : Andrei Stefanescu <andrei@galois.com>
+-- Stability        : provisional
+--
+-- This module provides an execution feature for bounding the
```
robdockins: Correct this module comment

crucible/src/Lang/Crucible/Simulator/Breakpoint.hs
```diff
@@ -0,0 +1,74 @@
...
+        return $ C.ExecutionFeature $ \case
+          C.RunningState (C.RunPostBranchMerge block_id) state
+            | frame <- state ^. C.stateCrucibleFrame
+            , C.SomeHandle cfgHandle == C.frameHandle frame
+            , Just Refl <- testEquality
+                (fmapFC C.blockInputs cfgBlockMap)
+                (fmapFC C.blockInputs $ C.frameBlockMap frame)
+            , Just Refl <- testEquality breakpoint_block_id block_id
+            , Just Refl <- testEquality ret_type (C.frameReturnType frame) -> do
+              putStrLn $ "breaking on: " ++ show breakpoint_block_id
```
robdockins: Sanitize this, we shouldn't be writing directly to `stdout`.

crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
```diff
@@ -534,6 +534,7 @@ data RunningStateInfo blocks args
     -- | This indicates that we are in a @RunningState@ because we
     --   returned from calling the named function.
   | RunReturnFrom !FunctionName
+  | RunPostBranchMerge !(BlockID blocks args)
```
robdockins: Comment for this constructor

crucible/src/Lang/Crucible/Analysis/Postdom.hs
```diff
@@ -86,15 +91,19 @@ postdomMap m = r
...
+postdomAssignment m breakpointIds = fmapFC go m
+  where pd = postdomMap m breakpointIds
         go :: Block ext blocks ret c -> Const [Some (BlockID blocks)] c
         go b = Const $ fromMaybe [] (Map.lookup (Some (blockID b)) pd)
 
 -- | Compute posstdom information for CFG.
 postdomInfo :: CFG ext b i r -> CFGPostdom b
-postdomInfo g = postdomAssignment (cfgBlockMap g)
+postdomInfo g =
+  postdomAssignment (cfgBlockMap g) (Bimap.keysR $ cfgBreakpoints g)
```
robdockins: As we discussed, I think this is the wrong default... we ought to be performing the post-dominator analysis normally as though we are simply going to step through breakpoints, and only pass in the ones here that we are actually going to use an execution feature to break-and-return from.

crucible/src/Lang/Crucible/CFG/ExtractSubgraph.hs
robdockins: :-(

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
robdockins: I think we should probably have a fresh datatype here with more descriptive constructor names rather than just using `Either`

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
robdockins: I think this will reconstruct the breakpoint `Bimap` at every statement.  Is this a case where we can `unsafeCoerce` the whole map when `UNSAFE_OPS` is enabled?

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
robdockins: More than one `go` function probably means we need better names for them both.

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
robdockins: Let's add a comment explaining why we don't expect to see any `Breakpoint` statements here.

crucible/src/Lang/Crucible/Simulator/Breakpoint.hs
```diff
@@ -0,0 +1,74 @@
...
+import qualified Lang.Crucible.CFG.Expr as C
+import qualified Lang.Crucible.Simulator.CallFrame as C
+import qualified Lang.Crucible.Simulator.EvalStmt as C
+import qualified Lang.Crucible.Simulator.ExecutionTree as C
+import qualified Lang.Crucible.Simulator.Operations as C
+import qualified Lang.Crucible.Simulator.OverrideSim as C
+import qualified Lang.Crucible.Simulator.RegValue as C
+import qualified What4.FunctionName as W
+
+breakAndReturn
```
robdockins: In general, I'm guessing you actually need a mapping from breakpoint names to overrides?

crucible/src/Lang/Crucible/Simulator/CallFrame.hs
```diff
@@ -151,8 +152,18 @@ mkCallFrame :: CFG ext blocks init ret
...
                -- ^ Initial arguments
             -> CallFrame sym ext blocks ret init
-mkCallFrame g pdInfo args = do
-  let bid@(BlockID block_id) = cfgEntryBlockID g
+mkCallFrame g = mkBlockFrame g (cfgEntryBlockID g)
+
+-- | Create a new call frame.
+mkBlockFrame :: CFG ext blocks init ret
+                -- ^ Control flow graph
+             -> BlockID blocks args
```
robdockins: This is the entry point, right?  Let's add that comment.

crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
robdockins: Document what this run state is.

crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
```diff
@@ -399,11 +400,18 @@ callCFG ::
   CFG ext blocks init ret {- ^ Function to run -} ->
   RegMap sym init {- ^ Arguments to the function -} ->
   OverrideSim p sym ext rtp a r (RegEntry sym ret)
-callCFG cfg args =
-  Sim $ StateContT $ \c -> runReaderT $
-    let f = mkCallFrame cfg (postdomInfo cfg) args in
-    ReaderT $ return . CallState (ReturnToOverride c) (CrucibleCall (cfgEntryBlockID cfg) f)
+callCFG cfg = callBlock cfg (cfgEntryBlockID cfg)
 
+callBlock ::
```
robdockins: Haddock description for the function

crucible/src/Lang/Crucible/Simulator/Breakpoint.hs
robdockins: Sanitize this, we shouldn't be writing directly to `stdout`

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
andreistefanescu: 👍

crucible/src/Lang/Crucible/CFG/SSAConversion.hs
andreistefanescu: 👍

crucible/src/Lang/Crucible/Simulator/Breakpoint.hs
andreistefanescu: ups, thanks!

crucible/src/Lang/Crucible/Simulator/Breakpoint.hs
andreistefanescu: Currently, each execution feature handles one breakpoint. This is ok if there are only a few breakpoints. I can add a map later.

andreistefanescu:
Let's talk about the changes to `SSAConversion` when you have time.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
robdockins: I'm not sure... maybe in the main Translation module? It just seems strange to have it defined, but not used at all here.

crucible/src/Lang/Crucible/CFG/ExtractSubgraph.hs
robdockins: That's a possibility... I don't remember offhand what it was used for in the first place.

PR #190, opened 2019-04-10
Fix SMTLib2 syntax for floating-point conversions 

robdockins:
The bugfix for SMTLib syntax is contained in 6ee9bcb.

The remainder of this pull request is basically about adding enough support to crucible-syntax and crucible proper to execute a test case verifying the bugfix.  This involves adding a syntax for the floating-point types, rounding modes, and a selection of the floating-point cast operations.

Further floating-point operations ought to be easier to add in the future, given this infrastructure.

### DIFF
--- a/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
@@ -896,7 +896,7 @@ synthExpr typeHint =
              return $ SomeE StringRepr $ EApp $ ShowFloat fi e
            _ | AsBaseType bt <- asBaseType t1 ->
              return $ SomeE StringRepr $ EApp $ ShowValue bt e
-           _ -> describe ("base or floating point type, but got " <> T.pack (show t1)) empty
+           _ -> later $ describe ("base or floating point type, but got " <> T.pack (show t1)) empty
 
 data NatHint
   = NoHint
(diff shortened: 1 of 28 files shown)


robdockins:
Shouldn't be anything too controversial in here, I think.

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -680,14 +793,16 @@ synthExpr typeHint =
            cons (synthExpr typeHint) $
            emptyList
          matchingExprs typeHint et ef $ \tp t f ->
-           case asBaseType tp of
-             NotBaseType ->
-               let msg = T.concat [ "conditional where branches have base type (got "
+          case tp of
+            FloatRepr fi ->
+               return $ SomeE tp $ EApp $ FloatIte fi c t f
+            (asBaseType -> AsBaseType bty) ->
```
kquick: I'm curious about this notation, especially in comparison to line 897.  I think I understand the notation on 897, but not this one.  Is it a pattern notation?  I don't necessarily think it's wrong, I just don't know what/how it works.

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -680,14 +793,16 @@ synthExpr typeHint =
...
+               return $ SomeE tp $ EApp $ BaseIte bty c t f
+            _ ->
+               let msg = T.concat [ "conditional where branches have base or floating point type (got "
```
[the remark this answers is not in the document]
robdockins: Good catch.

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
robdockins: Yeah, this is using an extension called "view patterns", and it's functionally the same as the pattern on line 897.  View patterns pass a value through a function before pattern matching on the result.

Now that you've pointed it out, I like the view pattern style better here.

Issue #180, opened 2019-04-01
Cryptol not found (crucible-server)


WeeknightMVP:
When trying to invoke `crucible-server`, I get the following error message:

```
Failed to start simulator: scFindModule: module Cryptol not found!
CallStack (from HasCallStack):
  error, called at src/Verifier/SAW/SharedTerm.hs:378:10 in saw-core-0.1-J9um5kGQYo52kWyRW8wN8t:Verifier.SAW.SharedTerm
`````

WeeknightMVP:
Thanks!  My recent `crucible-server` build finds the Cryptol prelude.

Now when attempting SAWCore verification, I see messages to the effect that `Prelude.and_True` was not found.  `saw-core/src/Verifier/SAW/Rewriter.hs` shows this split into `Prelude.and_True1` and `Prelude.and_True2`.  My old client will have to be updated to reflect that and similar changes, but I think this issue can be closed.

PR #191, opened 2019-04-10
Make Cryptol prelude available in crucible-server

atomb:
This requires bumping several dependencies, as well.


robdockins:
Thanks for looking into this @atomb.  I'm verifying if this fixes #180 in context.

atomb:
Thanks, @robdockins!

PR #195, opened 2019-04-16
[WIP] Check solver version bounds

siddharthist:
See https://github.com/GaloisInc/saw-script/issues/390

Sometimes, What4 (or higher-level tools) require specific features of a given SMT solver. This PR implements enough functionality to effectively check the version of any SMTLIB2-compliant solver.


what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -912,3 +937,128 @@ shutdownSolver _solver p = do
...
+mkChunks :: [Word] -> [Versions.VChunk]
+mkChunks = map ((:[]) . Versions.Digits)
+
+-- | The minimum (inclusive) version bound for a given solver.
+--
+-- The keys come from @'smtWriterName'@ in @'WriterConn'@.
+-- See also https://github.com/GaloisInc/crucible/issues/194
+solverMinVersions :: Map String Version
+solverMinVersions =
+  [ -- TODO: Why is this verion required?
```
robdockins: Yices 2.6.1 provided the `print-success` command line option, which causes yices to print an acknowledgement message after every successful command.  We require this option be set to ensure that we maintain synchronization with the solver.

The reason for this is that Yices will sometimes respond with `unsat` to an `assert` command if the context becomes "obviously" unsatisfiable.  In order to reliably detect this condition, we require all commands to respond with some acknowledgement.

what4/src/What4/Protocol/SMTLib2.hs
```diff
@@ -912,3 +937,128 @@ shutdownSolver _solver p = do
...
+    , Version { _vEpoch = Nothing, _vChunks = mkChunks [2, 6, 1], _vRel = []}
+    )
+  ]
+
+-- | The maximum (non-inclusive) version bound for a given solver.
+--
+-- The keys come from @'smtWriterName'@ in @'WriterConn'@.
+solverMaxVersions :: Map String Version
+solverMaxVersions = []
+
```
robdockins: We might also want a version blacklist, in addition to the min and max bounds...

siddharthist:
@robdockins I was thinking that to get the Yices version, it would be enough to spin up `yices-smt2` just to run `(get-info :version)`. Are you thinking this would be too heavyweight of a solution?

robdockins:
Maybe... but in the most general case, you might have `yices` installed but not `yices-smt2`, or they might be different versions, or installed in different places.  I really would not want to use a fragile system for version checking if we're going to present the user with warnings/errors.

what4/src/What4/Protocol/SMTLib2.hs
siddharthist: Okay, I'll put this in a comment once we decide on an overall approach. Thanks for explaining this in more detail, I only vaguely remembered our Mattermost conversation after two weeks away :)

robdockins:
I decided to go ahead and merge this PR, even though it doesn't completely address the issue with yices versioning.  The ability to query SMTLib2-compliant solvers is independently useful and seems to be working correctly, so I don't see any good reason to continue delaying this PR.

I think I'd still like to see the ability to blacklist individual solver versions in addition to upper and lower version bounds, but that can be added in a separate PR.

PR #196, opened 2019-04-17
Update built-in rewrite rules for crucible-server

atomb:
The previous set had some duplicates, and at least one entry that didn't exist in the SAWCore prelude.


robdockins:
Pretty straightforward.  Looks good to me.

PR #200, opened 2019-04-17
Test LLVMVals for equality

siddharthist:
I believe this is necessary for https://github.com/GaloisInc/saw-script/issues/392. 

If `allOf` and `oneOf` seem useful, they should probably be put in `IsExprBuilder`.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
@@ -9,7 +9,6 @@
 ------------------------------------------------------------------------
 
 {-# LANGUAGE DataKinds #-}
-{-# LANGUAGE EmptyCase #-}
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE KindSignatures #-}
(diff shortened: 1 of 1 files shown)


siddharthist:
I tested this in conjunction with my solution to the above SAWscript issue, and it appeared to solve the problem/make progress.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
```diff
@@ -133,3 +140,125 @@ instance IsExpr (SymExpr sym) => Show (LLVMVal sym) where
...
+-- values, rather than chaining a bunch of 'andPred's.
+allOf :: (IsExprBuilder sym)
+      => sym
+      -> [Pred sym]
+      -> IO (Pred sym)
+allOf sym xs =
+  if and (mapMaybe asConstantPred xs)
+  then foldM (andPred sym) (truePred sym) xs
+  else pure (falsePred sym)
+
```
robdockins: Are you confident that this is more efficient than `andAllOf` from What4.Interface?  `andPred` itself already has constant-folding behavior...

siddharthist: Consider the list `[<symbolic>, <symbolic>, ..., <concretely-true>]`. `andAllOf` will have to apply `andPred` _n_-1 times to symbolic predicates before returning true, whereas this will return true immediately.

I guess it depends on how cheap of an operation `andPred` is. I don't really know much about its implementation. 

It might be best to use `andAllOf` in the `then` branch here, to reduce duplication.

PR #202, opened 2019-04-18
Add a conditional memory write to LLVM model for macaw.

joehendrix:


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -748,7 +747,24 @@ addMatchingWrite w v = ReadMem (modify (coerce ((w,v):)))
...
+--
+-- A common use case is that the symbolic simulator will branch into
+-- two execution states based on a symbolic branch, make different
+-- memory modifications on each branch, and then need to merge the two
+-- memories to resume execution along a single path using the branch
+-- condition.  To support this, our memory representation supports
+-- "branch frames", at any point one can insert a fresh branch frame
+-- (see `branchMem`), and then at some later point merge two memories
+-- back into a single memory (see `mergeMem`).  Our `mergeMem`
+-- implementation is abl to efficiently merge memories, but requires
```
siddharthist: abl -> able

robdockins:
Looks OK to me.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
siddharthist:
```suggestion
-- implementation is able to efficiently merge memories, but requires
```

PR #205, opened 2019-04-19
What4: New LabeledPred module

siddharthist:
- Take the LabeledPred type from Crucible.Backend and put it in What4
- Add a utility function
- Derive more classes

This is going to be useful in SAWscript for override matching: we would
like to be able to tell the user when an override failed because a specific
precondition was concretely false. By saving preconditions as `LabeledPred`s
and using `partitionLabeledPreds`, this will be easy.


robdockins:
Probably need to add the new module to `what4.cabal`

siddharthist:
@robdockins Good catch, done!

PR #208, opened 2019-04-19
Make bvConcat rewrites fire more often

robdockins:
Be slightly more aggressive about deciding that terms are equal, so the rewrite rule fires in cases involving floating point casts that arise from LLVM memory model operations.

### DIFF
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -3833,8 +3833,8 @@ foldIndicesInRangeBounds sym f0 a0 bnds0 = do
 #endif
 
 
--- | A slightly more aggressive syntaxtic equality check than testEquality,
---   `sameTerm` will recurse through a small collection syntax formers.
+-- | A slightly more aggressive syntactic equality check than testEquality,
+--   `sameTerm` will recurse through a small collection of known syntax formers.
 sameTerm :: Expr t a -> Expr t b -> Maybe (a :~: b)
 
 sameTerm (asApp -> Just (FloatToBinary fppx x)) (asApp -> Just (FloatToBinary fppy y)) =
@@ -5263,7 +5263,7 @@ instance IsExprBuilder (ExprBuilder t st fs) where
     let BaseFloatRepr fpp = exprType x in sbMakeExpr sym $ FloatFMA fpp r x y z
   floatEq sym x y
     | x == y = return $! truePred sym
-    | otherwise = (floatIEEELogicBinOp FloatEq) sym x y
+    | otherwise = floatIEEELogicBinOp FloatEq sym x y
   floatNe sym x y = notPred sym =<< floatEq sym x y
   floatFpEq sym x y
     | x == y = notPred sym =<< floatIsNaN sym x
--- a/what4/src/What4/Expr/Builder.hs
+++ b/what4/src/What4/Expr/Builder.hs
@@ -3832,15 +3832,17 @@ foldIndicesInRangeBounds sym f0 a0 bnds0 = do
     _ -> error "internal: invalid index to foldIndicesInRangeBounds"
 #endif
 
-{-
--- | Compute the weighted sum of two bitvectors.
-bvSum :: (1 <= w)
-      => sym
-      -> NatRepr w
-      -> WeightedSum Integer (SymBV sym w)
-      -> IO (SymBV sym w)
-bvSum = undefined
--}
+
+-- | A slightly more aggressive syntaxtic equality check than testEquality,
+--   `sameTerm` will recurse through a small collection syntax formers.
+sameTerm :: Expr t a -> Expr t b -> Maybe (a :~: b)
+
+sameTerm (asApp -> Just (FloatToBinary fppx x)) (asApp -> Just (FloatToBinary fppy y)) =
+  do Refl <- testEquality fppx fppy
+     Refl <- sameTerm x y
+     return Refl
+
+sameTerm x y = testEquality x y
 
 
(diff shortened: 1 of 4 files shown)


what4/src/What4/Expr/Builder.hs
```diff
@@ -5263,10 +5265,18 @@ instance IsExprBuilder (ExprBuilder t st fs) where
     | x == y = return $! truePred sym
     | otherwise = (floatIEEELogicBinOp FloatEq) sym x y
   floatNe sym x y = notPred sym =<< floatEq sym x y
-  floatFpEq = floatIEEELogicBinOp FloatFpEq
-  floatFpNe = floatIEEELogicBinOp FloatFpNe
-  floatLe = floatIEEELogicBinOp FloatLe
-  floatLt = floatIEEELogicBinOp FloatLt
+  floatFpEq sym x y
+    | x == y = notPred sym =<< floatIsNaN sym x
```
kquick: I'm thinking that `floatFpEq` is not the dual of `floatFpNe` for NaN:

    let x = NaN
        y = NaN
    in assert $ (floatFpEq sym x y) /= (floatFpNe sym x y)

Should `floatFpEq` return a `truePred` if `x` and `y` are **both** `NaN`, or should `floatFpNe` return `truePred` for those inputs?

what4/src/What4/Expr/Builder.hs
```diff
@@ -3832,15 +3832,17 @@ foldIndicesInRangeBounds sym f0 a0 bnds0 = do
...
--- | Compute the weighted sum of two bitvectors.
-bvSum :: (1 <= w)
-      => sym
-      -> NatRepr w
-      -> WeightedSum Integer (SymBV sym w)
-      -> IO (SymBV sym w)
-bvSum = undefined
--}
+
+-- | A slightly more aggressive syntaxtic equality check than testEquality,
```
kquick: ... aggressive synta **c** tic equality...  collection **of known** syntax formers.

what4/src/What4/Interface.hs
```diff
@@ -1799,9 +1796,12 @@ class (IsExpr (SymExpr sym), HashableF (SymExpr sym)) => IsExprBuilder sym where
   --
   --   NOTE! This test returns false if either value is @NaN@; in particular
   --   @NaN@ is not equal to itself!  Moreover, positive and negative 0 will
-  --   compare equal, despite having different bit patterns. This test is most
-  --   appropriate for interpreting the equalty tests of typical languages using
-  --   floating point.
+  --   compare equal, despite having different bit patterns.
+  --
+  --   This test is most appropriate for interpreting the equalty tests of
```
kquick: "equality"

what4/src/What4/Expr/Builder.hs
kquick: OK, I didn't read ahead to the Interface changes... I see the issue now as described there.

Do we expect that `notPred . floatFpEq` is typically the more appropriate operation than `floatFpNe`?  If so, maybe we want to add another (ignored) parameter to `floatFpNe` to prevent people from using it without having to figure out why it isn't expressed naturally and therefore making it more likely they've seen the doc note?

what4/src/What4/Expr/Builder.hs
```diff
@@ -5263,10 +5265,18 @@ instance IsExprBuilder (ExprBuilder t st fs) where
...
-  floatFpNe = floatIEEELogicBinOp FloatFpNe
-  floatLe = floatIEEELogicBinOp FloatLe
-  floatLt = floatIEEELogicBinOp FloatLt
+  floatFpEq sym x y
+    | x == y = notPred sym =<< floatIsNaN sym x
+    | otherwise = floatIEEELogicBinOp FloatFpEq sym x y
+  floatFpNe sym x y
+    | x == y = return $ falsePred sym
+    | otherwise = floatIEEELogicBinOp FloatFpNe sym x y
+  floatLe sym x y
```
kquick: Might want to add a doc note for `floatLe` in the Interface file about the NaN equality stuff (maybe just "see `floatFpEq` for details on NaN comparisions").

what4/src/What4/Expr/Builder.hs
```diff
@@ -5263,10 +5265,18 @@ instance IsExprBuilder (ExprBuilder t st fs) where
...
-  floatFpEq = floatIEEELogicBinOp FloatFpEq
-  floatFpNe = floatIEEELogicBinOp FloatFpNe
-  floatLe = floatIEEELogicBinOp FloatLe
-  floatLt = floatIEEELogicBinOp FloatLt
+  floatFpEq sym x y
+    | x == y = notPred sym =<< floatIsNaN sym x
+    | otherwise = floatIEEELogicBinOp FloatFpEq sym x y
+  floatFpNe sym x y
+    | x == y = return $ falsePred sym
+    | otherwise = floatIEEELogicBinOp FloatFpNe sym x y
```
siddharthist: Is the reason for the new logic here that `==` is faster than `floatIEEELogicBinOp`?

robdockins: `floatIEEELogicBinOp` doesn't evaluate the semantics of the operation at all... it literally just applies the given syntax former.  This just captures some common special cases.

siddharthist: @robdockins Gotcha. Thanks!

what4/src/What4/Expr/Builder.hs
robdockins: I think the documentation is probably sufficient.  I'd like to think that when presented with four variants of equality/non-equality tests, developers will read the docs to figure out which one to use.

PR #209, opened 2019-04-20
No MonadFail

kquick:
Change from implicit uses of MonadFail to explicit MonadThrow (from the exceptions package), throwing specific data types.


robdockins:
Seems like a good cleanup to me 👍

PR #212, opened 2019-04-22
Pretty-printing of LLVMVals

### LINKED ISSUE #210: Pretty-printing of LLVMVal
SAWscript's `matchArg` attempts to compare `LLVMVal`s with SAW's `SetupValues`. It would be nice if we could print `LLVMVal`s when this comparison fails (at least the concrete parts).

siddharthist:
Fixes #210

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
@@ -135,21 +135,17 @@ instance IsExpr (SymExpr sym) => PP.Pretty (LLVMVal sym) where
         (LLVMValInt blk w) ->
           case asNat blk of
             Just 0 ->
-              case (asUnsignedBV w, asSignedBV w) of
-                (Just unsigned, Just signed) -> PP.text $ unwords $
+              case (asUnsignedBV w) of
+                (Just unsigned) -> PP.text $ unwords $
                   [ "literal integer:"
                   , "unsigned value = " ++ show unsigned ++ ","
-                  , "signed value = " ++ show signed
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
```diff
@@ -40,6 +40,7 @@ import           Control.Monad (foldM, join)
 import           Data.Foldable (toList)
 import           Data.Maybe (mapMaybe)
 import           Data.List (intersperse)
+import qualified Text.PrettyPrint.ANSI.Leijen as PP
 
 import           Data.Parameterized.Classes
```
siddharthist:
```suggestion
import           Data.Parameterized.Class
```

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
```diff
@@ -123,6 +124,56 @@ zeroInt sym bytes k
...
+  pretty =
+    let typed doc tp = PP.text doc PP.<+> PP.text ":" PP.<+> PP.text (show tp)
+    in
+      \case
+        (LLVMValZero tp) -> PP.angles (typed "zero" tp)
+        (LLVMValUndef tp) -> PP.angles (typed "undef" tp)
+        (LLVMValInt blk w) ->
+          case asNat blk of
+            Just 0 ->
+              case (asUnsignedBV w, asSignedBV w) of
```
robdockins: Probably better to just case on `asUnsignedBV` and use `toSigned` to get the signed value instead of casing separately.

PR #217, opened 2019-04-23
Uniformly add variants of llvm intrinsics

robdockins:
Uniformly add variants of `llvm.memcpy.*`, `llvm.memmove.*` and `llvm.memset.*`
to deal with the removal of the `alignment` parameter to these functions that
occurred in recent LLVM (version 7, I think).


robdockins:
This merge should fix the current instance of #108

robdockins:
Yeah, we could probably rig up a way to unify the 32/64 bit variants, and maybe also to automatically determine which signature is being requested and such... I'm not sure it would be worth the trouble.

PR #218, opened 2019-04-24
Constant folding for assertion trees

siddharthist:
When we use the `AssertionTree` datastructure to present errors to the user, they usually contain redundant information. I'm still brainstorming how to fix this, but these operations should be helpful. 

When displaying assertion trees to users we could: 
 - Throw out/collapse constantly/trivially true parts of the tree (since they won't have contributed to the problem the user needs to solve)
 - Traverse the tree, making lists of concretely false subtrees that cause the overall collapsed tree to be false. These assertions can be specially emphasized.

The first should at least be made possible by `constantFoldMAT_`.


robdockins:
I don't love the names of some of these operations, but *shrug*, I'm not too worried about the color of this bikeshed.

siddharthist:
I'm going to close this until I have a better end-to-end idea of how to fix the problem of too-verbose messages.

PR #220, opened 2019-04-25
Find allocations with a given number

siddharthist:
The purpose of this is to display better error messages when loads through a pointer fail. For instance, it may fail because there are no matching allocations, or the matching allocation(s) didn't have the right size. 

This is complementary information to #178, which exposed information about _writes_, whereas here we expose information about _allocations_, which should be a little easier to digest, if a little less revealing.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
@@ -58,6 +58,10 @@ module Lang.Crucible.LLVM.MemModel.Generic
   , branchAbortMem
   , mergeMem
 
+  , SomeAlloc(..)
+  , possibleAllocs
+  , ppSomeAlloc
+
     -- * Pretty printing
   , ppType
   , ppPtr
@@ -81,6 +85,7 @@ import           GHC.Generics (Generic, Generic1)
 import           Numeric.Natural
 import           Text.PrettyPrint.ANSI.Leijen hiding ((<$>))
 import           Lang.Crucible.Panic (panic)
+
 
 import           Data.Parameterized.Classes
 import qualified Data.Parameterized.Context as Ctx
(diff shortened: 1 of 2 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1292,6 +1297,47 @@ mergeMem c x y =
...
+--   call of @helper@ on a 'MemAlloc' returns anything nonempty, that can just
+--   be returned (?).
+possibleAllocs ::
+  forall sym .
+  (IsSymInterface sym) =>
+  Natural              ->
+  Mem sym              ->
+  [SomeAlloc sym]
+possibleAllocs n =
+  let helper m =
```
robdockins: Why bind `m` here? It looks like it is always equal to `n`...

robdockins: Also, I usually prefer to use a `where` binding for this sort of pattern rather than a `let`

siddharthist: @robdockins I was also confounded by the need to bind `m`. Here's what happens when I get rid of the binding an parameter:
```
src/Lang/Crucible/LLVM/MemModel/Generic.hs:(1330,9)-(1339,64): error:
    * Couldn't match type `t0 (MemAlloc sym0)' with `Natural'
      Expected type: Natural -> [MemAlloc sym0] -> [SomeAlloc sym1]
        Actual type: t0 (MemAlloc sym0) -> [SomeAlloc sym1]
    * In the expression:
        foldMap
          $ \case
              MemFree _ -> []
              Alloc atp base sz mut alignment loc
                -> if base == n then
                       [SomeAlloc atp base sz mut alignment loc]
                   else
                       []
              AllocMerge (asConstantPred -> Just True) as1 as2 -> helper n as1
              AllocMerge (asConstantPred -> Just False) as1 as2 -> helper n as2
              AllocMerge _ as1 as2 -> helper n as1 ++ helper n as2
```

PR #221, opened 2019-04-29
Update a few submodules, improve Travis build

siddharthist:
SAWscript is already on this version: https://github.com/GaloisInc/saw-script/tree/master/deps


siddharthist:
Not really sure why Travis is failing to build `parameterized-utils`: https://travis-ci.org/GaloisInc/crucible/jobs/526167340#L5255
```
src/Data/Parameterized/TH/GADT.hs:107:18: error:
    • Couldn't match type ‘TyVarBndr’ with ‘Type’
      Expected type: DatatypeInfo -> [Type]
        Actual type: DatatypeInfo -> [TyVarBndr]
    • In the expression: datatypeVars
      In an equation for ‘dataParamTypes’: dataParamTypes = datatypeVars
    |
107 | dataParamTypes = datatypeVars
    |                  ^^^^^^^^^^^^
```

travitch:
It looks like the travis build for crucible is using an old commit of parameterized-utils (via submodules) - version 1.0.8.  2.0 is released - we probably just want to update the submodule.

siddharthist:
@travitch Thanks for the hint, I updated the submodule in the latest commit. Hopefully that'll work.

siddharthist:
The problem is the `--allow-newer` flag, which causes parameterized-utils' version bound on th-abstraction to be ignored: https://github.com/GaloisInc/parameterized-utils/issues/39

I'm going to split the Travis build in two: one with `--allow-newer` which can fail, and the other without (which can't fail).

PR #222, opened 2019-04-30
travis: Use Brew addon, rather than manually installing packages

siddharthist:
This should probably also happen in SAWscript.


atomb:
Yeah, I agree, we should do this for SAWScript (and for Cryptol).

PR #224, opened 2019-05-01
Run hlint in Travis

siddharthist:
I'd eventually like to fix/enable more warnings, but this is a start.

I tried to limit fixes to what I thought were uncontroversial changes, but we can easily drop any of these commits and just add another `ignore` line to `.hlint.yaml`.


crucible-llvm/src/Lang/Crucible/LLVM.hs
```diff
@@ -1,5 +1,5 @@
-{-# LANGUAGE EmptyCase #-}
```
andreistefanescu: is possible to configure `hlint` to delete the line (rather than truncate it)?

siddharthist: I'm happy to manually fix these, I doubt it's configurable. 

By the way, `hlint` by default only provides warnings, I used the `--refactor` flag to apply a few auto-fixes.

Issue #155, opened 2019-02-12
Speed up override matching


siddharthist:
Currently, the crucible-llvm registers overrides for declarations with no corresponding definition, which generally occurs for C standard library functions like `strlen` or `memcpy`. The algorithm is basically: for each declaration in the module, search through a list of overrides and see if the name matches. This is O(_nm_) where _n_ is the number of declarations and _m_ is the number of overrides in the list.

This could (and should) be improved to O(_n_ + _m_*log(_m_)) by first making a `Map L.Symbol (LLVMOverride ...)` and looking the declarations up in the map.  

The relevant code: https://github.com/GaloisInc/crucible/blob/7a9786196e378e002e8f6da26e9408d7922cf70e/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs#L105-L113

robdockins:
Unfortunately, it isn't quite that simple, as some of the override selection requires to look at a prefix of the symbol rather than the whole thing.  We could use a search trie, I suppose, which has similar complexity advantages.

siddharthist:
This might be a good choice: http://hackage.haskell.org/package/data-stringmap-1.0.1.1/docs/Data-StringMap.html

robdockins:
Yeah, that might work.  I think we'd have to invert the algorithm, but it should be OK.

That is, put all the declared and defined LLVM symbols in the map... then walk though the overrides we want to install, do a lookup or prefix find, install any overrides that match.

robdockins:
I think we should consider raising the priority of this ticket, as it actually seems to be causing problems in practice.  Loading even modestly-sized bitcode files (~5MB) causes significantly slowdown (~45seconds) that appears to be mostly spent in the override matching/installation phase.  Profiling indicates the name-demangling algorithm is occupying the bulk of program runtime, and demangling is only done during override installation.

A better algorithmic approach will probably go a long way, in addition to chasing down potential space-leaks in the underlying name demangling algorithm.

robdockins:
PR #228 has significantly improved this situation, to the point that bitcode parsing is now the dominant factor for moderately-sized bitcode inputs.  There may still be room for more sophisticated pre-filtering algorithms (perhaps based on RE matching, for example), but further work on this should probably be delayed until and unless further scaling-up causes override matching to dominate startup time again.

PR #228, opened 2019-05-10
Speed up the override registration phase

robdockins:
by prefiltering possible overrides first.

This significantly reduces the amount of startup time that is consumed
by installing overrides, especially C++ overrides that perform name demangling.

Each override that wants to be registered now provides a "matcher" which examines
the name of a symbol for an approximate test as to weather this override applies.
These can either be exact matches, prefix matches, or substring matches.  Any
override that passes the prefilter test is considered in order, as before, to
determine more precicely if it applies.

On one modestly-sized example, this reduced overall verification time from
105 seconds to 32 seconds.  The algorithm used here for pre-filtering is
quite naive, so additional improvements are perhaps possible.  However,
at the momemnt, startup time is dominated by parsing the bitcode file itself.

Partially addresses #155

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
@@ -21,6 +21,8 @@ module Lang.Crucible.LLVM.Intrinsics.Common
   , SomeLLVMOverride(..)
   , RegOverrideM
   , llvmSizeT
+  , OverrideTemplate(..)
+  , TemplateMatcher(..)
     -- ** LLVMContext
   , LLVMHandleInfo(..)
   , LLVMContext(..)
@@ -29,6 +31,9 @@ module Lang.Crucible.LLVM.Intrinsics.Common
   , symbolMap
   , mkLLVMContext
     -- ** register_llvm_override
+  , basic_llvm_override
+  , polymorphic1_llvm_override
+
   , build_llvm_override
   , register_llvm_override
   , register_1arg_polymorphic_override
@@ -50,6 +55,7 @@ import qualified Data.List as List
 import qualified Data.Text as Text
 import           Numeric (readDec)
 
+import qualified ABI.Itanium as ABI
 import qualified Data.Parameterized.Context as Ctx
 import           Data.Parameterized.Some (Some(..))
 import           Data.Parameterized.TraversableFC (fmapFC)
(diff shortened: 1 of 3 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -99,8 +105,20 @@ data SomeLLVMOverride p sym arch =
 llvmSizeT :: HasPtrWidth wptr => L.Type
 llvmSizeT = L.PrimType $ L.Integer $ fromIntegral $ natValue $ PtrWidth
 
+data OverrideTemplate p sym arch rtp l a =
+  OverrideTemplate
+  { overrideTemplateMatcher :: TemplateMatcher
+  , overrideTemplateAction :: RegOverrideM p sym arch rtp l a ()
+  }
+
```
siddharthist:
```suggestion

-- | This type controls whether an override is installed for a given name found in a module.
--  See 'filterTemplates'.
```

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libcxx.hs
```diff
@@ -70,33 +70,41 @@ import           Lang.Crucible.LLVM.Intrinsics.Common
...
+           Nothing -> empty
+           Just (SomeLLVMOverride override) -> register_llvm_override override
+
 
 -- type CPPOverride p sym arch args ret =
 --   L.Declare -> LLVMContext arch -> Maybe (LLVMOverride p sym arch args ret)
 
-type SomeCPPOverride p sym arch =
-  L.Declare -> LLVMContext arch -> Maybe (SomeLLVMOverride p sym arch)
+data SomeCPPOverride p sym arch =
```
siddharthist:
```suggestion
-- | We can only tell whether we should install a C++ override after demangling
--  the function name, which is expensive. As a first approximation, we ask whether
--  the function's name contains a few substrings, in order. 
data SomeCPPOverride p sym arch =
```

siddharthist:
Huh, the CI failure looks unrelated:
```haskell
Configuring boomerang-1.3.1...
Preprocessing library for boomerang-1.3.1..
Building library for boomerang-1.3.1..
[1 of 9] Compiling Text.Boomerang.HStack ( Text/Boomerang/HStack.hs, dist/build/Text/Boomerang/HStack.o )
[2 of 9] Compiling Text.Boomerang.Pos ( Text/Boomerang/Pos.hs, dist/build/Text/Boomerang/Pos.o )
[3 of 9] Compiling Text.Boomerang.Prim ( Text/Boomerang/Prim.hs, dist/build/Text/Boomerang/Prim.o )
Text/Boomerang/Prim.hs:57:10: error:
    • No instance for (Applicative (Parser e tok))
        arising from the superclasses of an instance declaration
    • In the instance declaration for ‘Monad (Parser e tok)’
   |
57 | instance Monad (Parser e tok) where
   |          ^^^^^^^^^^^^^^^^^^^^
```

robdockins:
That CI failure is really strange... it's trying to build a really old version of boomerang.

PR #240, opened 2019-05-21
Bug/what4 simplifier loop

robdockins:
Not sure if I actually want to merge this as-is, but @travitch has identified a bad interaction where the what4 rewriter loses sharing and does exponential work.  Memoizing the rewrite rules fixes the immediate problem, but leaves us with the bigger question about how aggressively we want to do this rewriting in the first place.

Also, if we do decide to use memoization as the solution to this problem, we should probably do it in a different way that lets us memoize more subterms; currently we can only memoize along paths where the sizes and other parameters to the function are constant.


robdockins:
I decided not to take this memoization approach, so I'm closing this pull request.

PR #241, opened 2019-05-22
Pretty-printing of LLVMVals with names of global variables

siddharthist:
This allows us to have error messages like this in SAWscript, which give users feedback in terms of the source language (C/LLVM) (note the presence of `@update256`):
```
could not match specified value with actual value:
  actual (simulator) value: @update256
  specified value: concrete pointer: allocation = 646, offset = 40
  type of actual value: i32(%struct.env_md_ctx_st*, i8*)**
  type of specified value: i32(%struct.env_md_ctx_st*, i8*)**
```
(c.f. https://github.com/GaloisInc/saw-script/pull/450). [This paste](https://clbin.com/jzjvI) has another example.

We should probably also consider using this in `readMem'`, so that if the memory model failed to load through a pointer to a global, the user doesn't have to look through the memory dump to find out which one it was.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -168,7 +168,7 @@ import           Data.IORef
 import qualified Data.List as List
 import           Data.Map (Map)
 import qualified Data.Map as Map
-import           Data.Maybe (mapMaybe, listToMaybe)
+import           Data.Maybe (mapMaybe)
 import           Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import           Data.Set (Set)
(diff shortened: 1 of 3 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1501,3 +1504,55 @@ allocGlobal sym mem (g, aliases, sz, alignment) = do
...
+                 -> IO (Maybe L.Symbol)
+isGlobalPointer' sym globalMap needle =
+  case testLeq (knownNat :: NatRepr 1) (ptrWidth needle) of
+    Nothing       -> pure Nothing
+    Just LeqProof -> isGlobalPointer sym globalMap needle
+
+-- | Pretty-print an 'LLVMVal', but replace pointers to globals with the name of
+--   the global when possible. Probably pretty slow on big structures.
+--
+-- TODO(lb): This is in this module instead of "MemModel.Value" because
```
siddharthist: I could use some help here if someone wouldn't mind taking a look. Thanks in advance!

robdockins:
Looks OK to me

siddharthist:
Thanks for the review!

PR #243, opened 2019-05-22
Fix andAlso in LLVMVal testEqual

andreistefanescu:


robdockins:
Oy... is there a test case that exposed this bug that we can add to the test suite?

siddharthist:
I think the only place this is currently used is in `SAWScript.CrucibleOverride.matchArg` in the case for `SetupGlobalInitializer`.

It should be relatively easy to add an artificial test case for, by just creating two concrete `LLVMVal`s and comparing them.

andreistefanescu:
I was experimenting with this in `saw-script` when I noticed the bug, but I ended up not using it. I don't  see an easy way of adding a unit test.

PR #250, opened 2019-05-24
Add a crux primitive: crucible_string

travitch:
This primitive allocates a fixed-length string with symbolic contents.  The
string has the user-specified number of symbolic bytes plus a concrete NUL
terminator.

Note that the symbolic bytes are unconstrained and could also be NUL.

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Overrides.hs
+++ b/crux-llvm/src/Crux/LLVM/Overrides.hs
@@ -1,12 +1,13 @@
-{-# Language ImplicitParams #-}
-{-# Language RankNTypes #-}
-{-# Language TypeFamilies #-}
-{-# Language TypeApplications #-}
-{-# Language TypeOperators #-}
-{-# Language DataKinds #-}
-{-# Language PatternSynonyms #-}
+{-# LANGUAGE FlexibleContexts #-}
 {-# Language ConstraintKinds #-}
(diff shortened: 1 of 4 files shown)


crux-llvm/src/Crux/LLVM/Overrides.hs
```diff
@@ -270,6 +277,44 @@ lib_fresh_i64 ::
...
+      -- the NUL terminator.
+      contentsName <- case userSymbol (name ++ "_contents") of
+        Left err -> fail (show err)
+        Right nm -> return nm
+      let arrayRep = BaseArrayRepr (Empty :> BaseBVRepr ?ptrWidth) (BaseBVRepr (knownNat @8))
+      initContents <- liftIO $ freshConstant sym contentsName arrayRep
+      mem2 <- liftIO $ doArrayConstStore sym mem1 ptr noAlignment initContents maxLenBV
+
+      -- Write the NUL terminator to the last byte
+      zeroByte <- liftIO $ bvLit sym (knownNat @8) 0
```
andreistefanescu: `arrayUpdate`  is an alternative way to write the final `0` in the array, and avoid a second write

travitch:
That is a good point - I don't think there is a need for it to be immutable.  I just started out thinking about it as for string literals.  I'll switch it to mutable.

PR #251, opened 2019-05-27
Allow prime as what4 symbol character

andreistefanescu:


robdockins:
Is prime an allowed character for Yices native languge also?

andreistefanescu:
@robdockins yes, I've tested this in native Yices, and it works as well

PR #255, opened 2019-06-03
Add what4 option to pass `--mcsat` to Yices

atomb:
The development version of Yices supports a `--mcsat` flag to enable the use of MC-SAT instead of DPLL(T) as the guiding search strategy.

### DIFF
--- a/crux/src/Crux/CruxMain.hs
+++ b/crux/src/Crux/CruxMain.hs
@@ -39,6 +39,7 @@ import What4.InterpretedFloatingPoint (IsInterpretedFloatExprBuilder)
 import What4.Interface (getConfiguration)
 import What4.FunctionName (FunctionName)
 import What4.Protocol.Online (OnlineSolver)
+import What4.Solver.Yices (yicesEnableMCSat)
 import What4.Solver.Z3 (z3Timeout)
 
 -- crux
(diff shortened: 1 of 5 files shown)


robdockins:
Looks pretty straightforward.  The only question I have is: how does the `mcsat` option interact with other options.  E.g., is it compatible with the `ef-solver` or `push-pop` mode?

atomb:
Yeah, that's something I was wondering and attempted to find out. From what I can tell, it does seem compatible with `push-pop` but not `ef-solver` (or `interactive`). Overall, I think there are various potential clashing options with Yices. Should we check for them all up-front or just let it fail? I'm not sure we can do anything better to respond if we detect the problem early.

robdockins:
Yeah, that's a good point.  The interactions might be pretty complicated and change over time.

I guess we should just make sure that the failure mode is OK, i.e., that the user is presented with an appropriate error message that indicates the underlying problem.

atomb:
Yep, I think that sounds good. I'll test that before merging this.

atomb:
Unfortunately, I'm having trouble getting this to work with any input, much less conflicting flags, so let's leave it unmerged for now. I think we may be feeding it problems that are outside the scope of what MC-SAT supports, but the result when trying to send essentially any problem to the MC-SAT engine through Crux is the super-helpful `hFlush: resource vanished (Broken pipe)` message.

atomb:
The issue seems to be that Yices doesn't support labeled assertions in MC-SAT mode.

atomb:
The latest commit disables the use of unsat cores in Crux when the Yices MC-SAT solver is turned on. We should probably make What4 produce an error if both are selected at the same time, too.

atomb:
Alright, now we check for the combination of MC-SAT and Unsat cores. I think this is ready to merge. Any other thoughts before I do that, @robdockins ?

PR #258, opened 2019-06-05
Use array stride

andreistefanescu:


siddharthist:
@andreistefanescu Out of curiosity, what does this PR do?

PR #261, opened 2019-06-05
Smt array memory model

andreistefanescu:
Depends on #258.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
@@ -647,6 +647,10 @@ readMem sym w l tp alignment m = do
   p2         <- isAligned sym w l alignment
   maybe_allocation_array <- asMemAllocationArrayStore sym w l m
   part_val <- case maybe_allocation_array of
+    -- if this read is inside an allocation backed by a SMT array store,
+    -- then decompose this read into reading the individual bytes and
+    -- assembling them to obtain the value, without introducing any
+    -- ite operations
     Just (arr, _arr_sz) -> do
       let loadArrayByteFn :: Offset -> IO (PartLLVMVal arch sym)
           loadArrayByteFn off = do
@@ -1144,6 +1148,12 @@ writeConstMem ::
   IO (Mem sym, Pred sym, Pred sym)
 writeConstMem = writeMemWithAllocationCheck isAllocated
 
+-- | Write a value to memory.
+--
+-- The returned predicates assert (in this order):
+--  * the pointer satisfies the checks specified by
+--    the @is_allocated@ function
+--  * the pointer's alignment is correct
 writeMemWithAllocationCheck ::
   forall sym w .
   (IsSymInterface sym, 1 <= w) =>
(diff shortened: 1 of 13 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1366,6 +1433,49 @@ possibleAllocs n = helper . memAllocs
               AllocMerge (asConstantPred -> Just False) as1 as2 -> helper as2
               AllocMerge _ as1 as2 -> helper as1 ++ helper as2
 
+asMemAllocationArrayStore ::
```
siddharthist: This could probably use a comment

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -643,16 +645,28 @@ readMem sym w l tp alignment m = do
...
-      PartLLVMVal p v ->
-        let ub1 = UB.ReadUnallocated  (UB.pointerView l)
-            ub2 = UB.ReadBadAlignment (UB.pointerView l) alignment
-            p'  = W4AT.And (p :| [ W4AT.Leaf (undefinedBehavior ub1 p1)
-                                 , W4AT.Leaf (undefinedBehavior ub2 p2)
-                                 ])
-        in return $ PartLLVMVal p' v
+  maybe_allocation_array <- asMemAllocationArrayStore sym w l m
+  part_val <- case maybe_allocation_array of
+    Just (arr, _arr_sz) -> do
```
robdockins: Is is correct to be ignoring the array size here? I don't recall exactly where we check that reads are in bounds, but this looks suspicious.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1132,11 +1142,60 @@ writeConstMem ::
...
   Mem sym       ->
   IO (Mem sym, Pred sym, Pred sym)
-writeConstMem sym w ptr tp alignment v m =
-  do sz <- bvLit sym w (bytesToInteger (typeEnd 0 tp))
-     p1 <- isAllocated sym w alignment ptr (Just sz) m
-     p2 <- isAligned sym w ptr alignment
-     return (memAddWrite (MemWrite ptr (MemStore v tp alignment)) m, p1, p2)
+writeConstMem = writeMemWithAllocationCheck isAllocated
+
+writeMemWithAllocationCheck ::
```
robdockins: Definitely need some documentation of the function behavior and arguments here

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1339,7 +1398,15 @@ mergeMem c x y =
 -- to the solver and get (overapproximate) concrete answers.
 
 data SomeAlloc sym =
-  forall w. SomeAlloc AllocType Natural (Maybe (SymBV sym w)) Mutability Alignment String
+  forall w. IsSymInterface sym => SomeAlloc AllocType Natural (Maybe (SymBV sym w)) Mutability Alignment String
```
robdockins: If it works to do so, I'd rather see this constraint propagated to the `Eq` instance below.  It's usually better to avoid packaging up typeclass dictionaries into existential types unless it's really necessary.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
robdockins: Agree, documentation here is important.  I'd also like a brief discussion of the performance implications of this @andreistefanescu, as we talked about in person the other day.  In case we later decide to go forward with the caching ideas we talked about, it would be good to record our thoughts on the high-level ideas.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
andreistefanescu: it is safe to ignore the array size, this code runs after the bounds checks.

PR #262, opened 2019-06-06
record program locations for symbolic branch points

### LINKED ISSUE #260: Record program locations for symbolic branches
It's my intuition (and don't hesitate to correct me) that every symbolic branch has a conditional statement in the source program associated to it. Because of this, we should be able to report the branch location in an `AbortedBranch` result.

I've started working on this in a branch, but there are a few places where I can't figure out where I would get a position from. I'm not at all familiar with this part of the code. https://github.com/GaloisInc/crucible/compare/master...siddharthist:ppmerge?expand=1

siddharthist:
When both branches of symbolic simulation fail, it's useful to know
where in the source program that branching happened. (e.g. in SAWscript,
we can tell users which "if" statement in their C program was the branch
point).

Fixes #260. I've tested this with SAW and it does indeed report branch points in my program, haven't confirmed that they're the appropriate ones.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/Operations.hs
+++ b/crucible/src/Lang/Crucible/Simulator/Operations.hs
@@ -704,7 +704,6 @@ resumeValueFromValueAbort ctx0 ar0 =
            (stateTree .~ ActiveTree ctx (er & partialValue.gpValue .~ frm))
            (resumeValueFromFrameAbort ctx ar0)
     VFVPartial ctx loc pred ay -> do
-      -- resumeValueFromValueAbort ctx (AbortedBranch (Posd (plSourceLoc loc) pred) ar0 ay)
       resumeValueFromValueAbort ctx (AbortedBranch loc pred ar0 ay)
     VFVEnd ->
       do res <- view stateContext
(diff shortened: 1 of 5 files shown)


crucible/src/Lang/Crucible/Simulator/Operations.hs
```diff
@@ -697,8 +703,9 @@ resumeValueFromValueAbort ctx0 ar0 =
          withReaderT
            (stateTree .~ ActiveTree ctx (er & partialValue.gpValue .~ frm))
            (resumeValueFromFrameAbort ctx ar0)
-    VFVPartial ctx p ay -> do
-      resumeValueFromValueAbort ctx (AbortedBranch p ar0 ay)
+    VFVPartial ctx loc pred ay -> do
+      -- resumeValueFromValueAbort ctx (AbortedBranch (Posd (plSourceLoc loc) pred) ar0 ay)
```
robdockins: stray comment line

travitch:
I think this kind of change is a great idea.  I'm curious about what level of granularity we can achieve.  If you see something like `if(a && b)`, do we have granular enough position information to capture difference between the test of `a` and `b` at the Crucible level? Crucible should translate that single C conditional into two conditionals to capture the short circuiting behavior.

siddharthist:
@travitch Great question, and I would also love to pursue that as a follow-up. Immediately upon testing this PR, I encounted a C conditional with five `&&`-ed clauses, making it difficult to determine exactly what test was being branched on.

robdockins:
The default behavior is for clang to introduce a branch for every `&&`.  I'd expect the source location to include the a column number in addition to the line number indicating which part of the test caused the branch.  Is that not happening?

PR #266, opened 2019-06-11
Add a Nonce-based TestEquality instance for ModuleTranslation

siddharthist:
... and a few other instances that I considered along the way before settling on the `Nonce`-based approach

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
@@ -100,6 +100,7 @@ import qualified Text.LLVM.AST as L
 
 import           Data.Parameterized.NatRepr as NatRepr
 import           Data.Parameterized.Some
+import           Data.Parameterized.Nonce
 
 import           What4.FunctionName
 import           What4.ProgramLoc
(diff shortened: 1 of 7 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
```diff
@@ -369,9 +376,10 @@ insDeclareHandle halloc ctx decl = do
 -- Note: We may want to add a map from symbols to existing function handles
 -- if we want to support dynamic loading.
 translateModule :: HandleAllocator s -- ^ Generator for nonces.
+                -> (NonceGenerator (ST s) t)
                 -> L.Module          -- ^ Module to translate
-                -> ST s (Some ModuleTranslation)
-translateModule halloc m = do
+                -> ST s (Some (ModuleTranslation t))
+translateModule halloc nonceGen m = do
```
robdockins: I don't think we need a separate nonce generator here, you can get the generator out of the handle allocator using the record accessor `haCounter`.

siddharthist:
@robdockins That's an interesting point I hadn't considered, I just figured those instances were usually a good idea because there's only ever one valid instance for any type. I'll revert those changes.

crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
siddharthist: Ah, I missed that because I saw that the constructor wasn't exported. Thanks!

PR #267, opened 2019-06-11
Fix Yices hanging when asserting in unsat context

atomb:
Yices doesn't like to see `(assert ...)` commands in an unsat context,
and prints a message about this to stderr instead of stdout. Since we're
already tracking whether the context is in an early unsat state, this
commit makes it so we don't emit any more assertions until we've left
the current solver frame.

### DIFF
--- a/what4/src/What4/Solver/Yices.hs
+++ b/what4/src/What4/Solver/Yices.hs
@@ -426,6 +426,10 @@ newConnection stream ack reqFeatures bindings = do
                  , supportQuantifiers = efSolver
                  }
 
+-- | This data type bundles a Yices command (as a Text Builder) with an
+-- indication as to whether it is safe to issue in an inconsistent
+-- context. Unsafe commands are the ones that Yices will complain about
+-- to stderr if issued, causing interaction to hang.
 data YicesCommand = YicesCommand
   { cmdEarlyUnsatSafe :: Bool
   , cmdCmd :: Builder
--- a/what4/src/What4/Solver/Yices.hs
+++ b/what4/src/What4/Solver/Yices.hs
@@ -70,6 +70,7 @@ import           Data.Bits
 
 import           Data.IORef
 import           Data.Foldable (toList)
+import           Data.Maybe
 import           Data.Parameterized.NatRepr
 import           Data.Parameterized.Some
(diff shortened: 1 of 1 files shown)


what4/src/What4/Solver/Yices.hs
```diff
@@ -434,36 +448,36 @@ instance SMTWriter (Connection s) where
...
-                       ]
+    safeCmd $ app "define" [Builder.fromText v <> "::"
+                             <> unType (fnType ((\(_,tp) -> viewSome yicesType tp) <$> args) (yicesType rtp))
+                           , renderTerm (yicesLambda args t)
+                           ]
 
   declareStructDatatype _ _ = return ()
 
-  writeCommand conn (Cmd cmd) =
+  writeCommand conn (YicesCommand earlyUnsatSafe cmd) =
```
robdockins: We didn't actually add the code here to skip sending the unsafe commands... oops

It happens to work, I guess, because we just skip the ACK phase for those commands, but we should avoid sending the commands also.

what4/src/What4/Solver/Yices.hs
```diff
@@ -415,13 +418,24 @@ newConnection stream ack reqFeatures bindings = do
...
+
+  earlyUnsatRef <- newIORef Nothing
+  conn <- newWriterConn stream (ack earlyUnsatRef) nm features' bindings (Connection earlyUnsatRef)
   return $! conn { supportFunctionDefs = True
                  , supportFunctionArguments = True
                  , supportQuantifiers = efSolver
                  }
 
-newtype YicesCommand = Cmd Builder
+data YicesCommand = YicesCommand
```
robdockins: Probably should add some code documentation here for what this safe/unsafe command stuff is all about.

what4/src/What4/Solver/Yices.hs
atomb: Oops! Good catch!

atomb: Fixed in c620a85c2a.

what4/src/What4/Solver/Yices.hs
atomb: Fixed in d70caf8e.

PR #269, opened 2019-06-11
OverrideSim: fail method -> MonadFail instance

siddharthist:
It's going away in GHC 8.8, and this works better for future-proof client code that has `MonadFail` constraints

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
+++ b/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
@@ -72,13 +72,16 @@ module Lang.Crucible.Simulator.OverrideSim
   , Lang.Crucible.Simulator.ExecutionTree.Override
   ) where
 
+import Prelude hiding (fail)
+
 import           Control.Exception
 import           Control.Lens
-import           Control.Monad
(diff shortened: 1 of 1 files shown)


crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
```diff
@@ -159,11 +162,14 @@ bindOverrideSim (Sim m) h = Sim $ unSim . h =<< m
 instance Monad (OverrideSim p sym ext rtp args r) where
   return = returnOverrideSim
   (>>=) = bindOverrideSim
-  fail msg = Sim $ StateContT $ \_c -> runGenericErrorHandler msg
```
robdockins: I don't know what the backward-compatibility story is here, but I don't want to force downstream clients to hide `Prelude.fail` and add `MonadFail` constraints just yet, so we should probably continue defining the `Monad.fail` method for now.

Should we use CPP to conditionally define this method on 8.6 and below?

siddharthist: CPP sounds like a good way to do it.

PR #273, opened 2019-06-20
Variety of Crux polish

atomb:


robdockins:
Looks like `crux-jvm` isn't building correctly

atomb:
After creating this PR I realized there were a few more things I wanted to do before merging it (updating `crux-jvm` being one of them). I'm going to leave it open until I've pushed a few more commits.

PR #274, opened 2019-06-20
Index writes by base pointer

andreistefanescu:


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -845,23 +852,130 @@ data MemState sym =
     -- of the list.
   | BranchFrame !Int !Int (MemChanges sym) (MemState sym)
 
-type MemChanges sym = ([MemAlloc sym], [MemWrite sym])
+type MemChanges sym = ([MemAlloc sym], MemWrites sym)
+
+newtype MemWrites sym = MemWrites [MemWritesChunk sym]
+
+data MemWritesChunk sym =
```
siddharthist: Could definitely use a comment

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -845,23 +852,130 @@ data MemState sym =
...
+    MemWritesChunkFlat [MemWrite sym]
+  | MemWritesChunkIndexed (IntMap [MemWrite sym])
+
+instance Semigroup (MemWrites sym) where
```
siddharthist: Also could probably use a comment, just a high-level description of how these are combined

PR #279, opened 2019-06-28
Handle empty writes when merging with indexed writes

chameco:
Extends #274 .

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
@@ -923,23 +923,27 @@ muxChanges c (left_allocs, lhs_writes) (rhs_allocs, rhs_writes) =
   )
 
 muxWrites :: Pred sym -> MemWrites sym -> MemWrites sym -> MemWrites sym
-muxWrites c lhs_writes rhs_writes
-  | MemWrites [MemWritesChunkIndexed lhs_indexed_writes] <- lhs_writes
-  , MemWrites [MemWritesChunkIndexed rhs_indexed_writes] <- rhs_writes =
-    MemWrites
-      [ MemWritesChunkIndexed $
-          mergeMemWritesChunkIndexed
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -923,23 +923,27 @@ muxChanges c (left_allocs, lhs_writes) (rhs_allocs, rhs_writes) =
...
+                     (MemWrites [MemWritesChunkFlat lhs])
+                     (MemWrites [MemWritesChunkFlat rhs])
+                 ])
+              lhs_indexed_writes
+              rhs_indexed_writes
+        ]
   | otherwise =
-    MemWrites [MemWritesChunkFlat [WriteMerge c lhs_writes rhs_writes]]
+    MemWrites [MemWritesChunkFlat [WriteMerge c l r]]
+  where indexed_chunk_map [MemWritesChunkIndexed m] = Just m
```
andreistefanescu: I would suggest: `asIndexedChunkMap :: MemWrites sym -> Maybe (IntMap [MemWrite sym])`

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -923,23 +923,27 @@ muxChanges c (left_allocs, lhs_writes) (rhs_allocs, rhs_writes) =
...
-              [ WriteMerge
-                  c
-                  (MemWrites [MemWritesChunkFlat lhs])
-                  (MemWrites [MemWritesChunkFlat rhs])
-              ])
-            lhs_indexed_writes
-            rhs_indexed_writes
-      ]
+muxWrites _ (MemWrites []) (MemWrites []) = MemWrites []
+muxWrites c l@(MemWrites lhs_writes) r@(MemWrites rhs_writes)
```
andreistefanescu: in general, a one character variable makes code harder to read (and to grep).

Issue #185, opened 2019-04-08
Bounded execution feature interacts strangely with symbolic branching


robdockins:
The `BoundedExec` execution feature of crucible, which underlies the `-i` option of `crux-llvm`, behaves in a rather strange way when it encounters loops in the context of previous symbolic branches.

The data structure containing information about how many times backedges have been traversed is maintained in a lookup table outside the execution tree, which means that it behaves "non-chronologically" with respect to the program, and executions along one branch are counted against the iteration count of other sibling branches, which is extremely nonintuitive.

The fix is to sink the accounting datastructure for iteration counts down into the simulator state so it is managed in program-chronological order.  This raises some interesting questions about what to do when iteration counts differ along paths that later merge...

PR #280, opened 2019-06-29
Path splitting

robdockins:
This PR implements an execution feature for path splitting, and a variety of refactorings and fixes to make it work properly.  It also fixes an issue with the bounded execution feature (CF #185) by recording the relevant data structures in a crucible global variable instead of alongside the simulator in an IORef.

The profiling simulator feature exhibits a similar problem, but it is less clear how to fix it.  As such, this PR is probably not ready to merge yet.


robdockins:
I've now piped through options for changing the `path-strategy` option, which can be either `always-merge` (the previous behavior) or `split-dfs` which performs path splitting and exploration in depth-first manner.

Rather than solve the problems that arise from the interaction of path splitting and profiling (which seem rather difficult), I've made it so that the profiling option is incompatible with path splitting in crux. At a later time, we can revisit this.

crucible/src/Lang/Crucible/Simulator/EvalStmt.hs
```diff
@@ -551,6 +552,40 @@ singleStepCrucible verb exst =
...
+  ExecutionFeatureModifiedState ::
+     ExecState p sym ext rtp -> ExecutionFeatureResult p sym ext rtp
+
+  -- | This exeuction feature result indicates that the state was modified
+  --   in an essential way that transforms it into new state altogether.
+  --   When this result is returned, it preempts any later execution
+  --   features and the main simulator loop and instead returns to the head
+  --   of the execution feature stack.
+  --
+  --   NOTE: In particular, the exeuction feature will encounter the
```
atomb: Minor typo: `exeuction`.

atomb: Also the same typo below.

crucible/src/Lang/Crucible/Simulator/EvalStmt.hs
```diff
@@ -551,6 +552,40 @@ singleStepCrucible verb exst =
...
+  --   state again before the simulator loop.  It is therfore very
+  --   important that the exeuction feature be prepared to immediately
+  --   encounter the same state again and make significant execution
+  --   progress on it, or ignore it so it makes it to the main simulator
+  --   loop.  Otherwise, the execution feature will loop back to itself
+  --   infinetely, starving out useful work.
```
atomb: Typo: `infinetely`.

crucible/src/Lang/Crucible/Simulator/PathSplitting.hs
```diff
@@ -0,0 +1,174 @@
...
+  IO (ExecState p sym ext rtp)
+restoreWorkItem (WorkItem branchPred loc frm st assumes) =
+  do let sym = st ^. stateSymInterface
+     setCurrentProgramLoc sym loc
+     restoreAssumptionState sym assumes
+     addAssumption sym (LabeledPred branchPred (ExploringAPath loc (pausedLoc frm)))
+     let ctx = st ^. stateTree . actContext
+     runReaderT (resumeFrame frm ctx) st
+
+{-
```
atomb: Is there any reason to keep this commented-out code?

crux/src/Crux/Log.hs
```diff
@@ -59,6 +59,9 @@ outputColored c msg =
 sayOK :: Logs => String -> String -> IO ()
 sayOK = sayCol Green
 
+sayWarn :: Logs => String -> String -> IO ()
```
atomb: Good addition.

PR #283, opened 2019-07-09
Recursion bound

robdockins:
WIP branch for adding a feature to bound recursion depth, similar to how the bounded execution feature bounds iteration depth.

This branch requires additional work to pipe options for enabling it through crux and for testing, so is not ready to merge yet.


robdockins:
This should be ready to merge now, and seems to be working correctly.

PR #287, opened 2019-07-23
Fp80 support

kquick:


what4/src/What4/Expr/Builder.hs
```diff
@@ -5744,6 +5744,7 @@ instance IsInterpretedFloatExprBuilder (ExprBuilder t st (Flags FloatReal)) wher
   iFloatLit sym _ = realLit sym
   iFloatLitSingle sym = realLit sym . toRational
   iFloatLitDouble sym = realLit sym . toRational
+  iFloatLitLongDouble sym = undefined
```
robdockins: Looks like we need an actual implementation of something here...

what4/src/What4/Expr/Builder.hs
```diff
@@ -5832,6 +5833,7 @@ instance IsInterpretedFloatExprBuilder (ExprBuilder t st (Flags FloatUninterpret
   iFloatLitDouble sym x =
     iFloatFromBinary sym DoubleFloatRepr
       =<< (bvLit sym knownNat $ toInteger $ IEEE754.doubleToWord x)
+  iFloatLitLongDouble sym x = undefined
```
robdockins: Need to implement this too...

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -395,7 +395,8 @@ evalStmt sym = eval
              | Just Refl <- testEquality handleTp expectedTp -> return (HandleFnVal h)
              | otherwise -> failedAssert
                  $ unwords ["Expected function handle of type " <> show expectedTp
-                           ,"but found handle of type " ++ show handleTp]
+                           ," for call to function " <> show (handleName h)
```
siddharthist: The `unwords` above is also going to add spaces between these list elements, did you intend to have two spaces here?

robdockins:
@kquick, where are we on this? Do you feel like it's ready to merge?

kquick:
I'm not quite ready yet, still wanting to do some verification and add the ability to specify which floating implementation is used via crux instead of the current hard-coded approach.

robdockins:
@kquick, the ability to select floating-point modes in crux is being added separately.  Is there a particular reason not to merge this?

PR #290, opened 2019-07-26
Some additional LLVM intrinsics

robdockins:
This PR overlaps with some others currently in-flight, probably not ready to merge just yet.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/.#Libc.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/.#Libc.hs
@@ -1,1 +1,0 @@
-rdockins@000288-rdockins-4.local.736
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/.#Libc.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/.#Libc.hs
@@ -1,0 +1,1 @@
+rdockins@000288-rdockins-4.local.736
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
@@ -489,36 +489,27 @@ callRealloc sym mvar alignment (regValue -> ptr) (regValue -> sz) =
      ptrNull <- liftIO (ptrIsNull sym PtrWidth ptr)
      symbolicBranches emptyRegMap
        -- If the pointer is null, behave like malloc
-       [ (ptrNull
-         , do mem <- readGlobal mvar
-              (newp, mem') <- liftIO $ doMalloc sym G.HeapAlloc G.Mutable "<realloc>" mem sz alignment
-              writeGlobal mvar mem'
-              return newp
-         ,
-         Nothing)
+       [ ( ptrNull
+         , modifyGlobal mvar $ \mem -> liftIO $ doMalloc sym G.HeapAlloc G.Mutable "<realloc>" mem sz alignment 
+         , Nothing
+         )
 
        -- If the size is zero, behave like malloc (of zero bytes) then free
        , (szZero
-         , do mem <- readGlobal mvar
-              (newp, mem') <- liftIO $
-                do (newp, mem1) <- doMalloc sym G.HeapAlloc G.Mutable "<realloc>" mem sz alignment
-                   mem2 <- doFree sym mem1 ptr
-                   return (newp, mem2)
-              writeGlobal mvar mem'
-              return newp
+         , modifyGlobal mvar $ \mem -> liftIO $
+              do (newp, mem1) <- doMalloc sym G.HeapAlloc G.Mutable "<realloc>" mem sz alignment
+                 mem2 <- doFree sym mem1 ptr
+                 return (newp, mem2)
          , Nothing
          )
 
        -- Otherwise, allocate a new region, memcopy `sz` bytes and free the old pointer
        , (truePred sym
-         , do mem  <- readGlobal mvar
-              (newp, mem') <- liftIO $
-                do (newp, mem1) <- doMalloc sym G.HeapAlloc G.Mutable "<realloc>" mem sz alignment
-                   mem2 <- uncheckedMemcpy sym mem1 newp ptr sz
-                   mem3 <- doFree sym mem2 ptr
-                   return (newp, mem3)
-              writeGlobal mvar mem'
-              return newp
+         , modifyGlobal mvar $ \mem -> liftIO $
+              do (newp, mem1) <- doMalloc sym G.HeapAlloc G.Mutable "<realloc>" mem sz alignment
+                 mem2 <- uncheckedMemcpy sym mem1 newp ptr sz
+                 mem3 <- doFree sym mem2 ptr
+                 return (newp, mem3)
          , Nothing)
        ]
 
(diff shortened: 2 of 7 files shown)


robdockins:
I think this is ready to merge now.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -495,6 +523,29 @@ callRealloc sym mvar alignment (regValue -> ptr) (regValue -> sz) =
...
+  :: (IsSymInterface sym, HasPtrWidth wptr, wptr ~ ArchWidth arch, ?lc :: TypeContext)
+  => sym
+  -> GlobalVar Mem
+  -> RegEntry sym (LLVMPointerType wptr)
+  -> RegEntry sym (BVType wptr)
+  -> RegEntry sym (BVType wptr)
+  -> OverrideSim p sym (LLVM arch) r args ret (RegValue sym (BVType 32))
+callPosixMemalign sym mvar (regValue -> outPtr) (regValue -> align) (regValue -> sz) =
+  case asUnsignedBV align >>= toAlignment . toBytes of
+    Nothing -> fail $ unwords ["posix_memalign: invalid alignment value:", show (printSymExpr align)]
```
langston-barrett: Does this indicate an invalid alignment value? Or a symbolic one? It seems like this is a case like `strlen` where we can't handle symbolic data for more fundamental reasons.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -495,6 +523,29 @@ callRealloc sym mvar alignment (regValue -> ptr) (regValue -> sz) =
...
+    Just a ->
```
langston-barrett: This could be replaced by something like:
```haskell
p <- callMalloc sym mvar alignment sz
mem <- readGlobal mvar
writeGlobal mvar =<< storeRaw sym mem' outPtr (bitvectorType (dl^.ptrSize)) (dl^.ptrAlign) (ptrToPtrVal p)
```
Both this and the implementation of `callMalloc` make me feel like we need a `modifyGlobal` combinator for the `readGlobal`/`writeGlobal` pairing.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -823,3 +873,52 @@ llvmAssertRtnOverride =
        do let err = AssertFailureSimError "Call to __assert_rtn"
           liftIO $ assert sym (falsePred sym) err
   )
+
+llvmGetenv
```
langston-barrett: This doesn't seem to be LLVM-specific, maybe just `getenv`?

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -823,3 +873,52 @@ llvmAssertRtnOverride =
...
+    , L.decArgs    = [ L.PtrTo $ L.PrimType $ L.Integer 8
+                     ]
+    , L.decVarArgs = False
+    , L.decAttrs   = []
+    , L.decComdat  = mempty
+    }
+  )
+  (Empty :> PtrRepr)
+  PtrRepr
+  (\_ sym _args -> liftIO $ mkNullPointer sym PtrWidth)
```
langston-barrett: This seems like a way to introduce unsoundness... If we go ahead and make this available we should probably document that it is a feature of the simulator. 

Imagine:
```c
char *should_launch_missiles = getenv("DO_BAD_STUFF");
if (should_launch_missiles)
  launch_missiles();
normal_operation();
```
Crucible would say code never launches missiles.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
robdockins: Yeah, this requires a concrete alignment value.  We reject if the value is symbolic, or if it isn't a power of 2.  I suppose we could have more fine-grained error reporting for the two different situations.

langston-barrett: I think that might be nice, because if someone sees this due to a symbolic alignment, it implies that they are doing something wrong as a user of Crucible (and perhaps that they could fix it), whereas the situation is less tractable than that.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
robdockins: It's true, this is a bit of a hack we introduced to get through some initialization code.  A more complete solution would be to allow the C environment variables to be somehow specified...

langston-barrett: Perhaps we could have an override that checks for whichever variables are being accessed during initialization and returns `nullptr` for those but then doesn't work for others? It would still be unsound, but a little less so :skull:

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
robdockins: Yeah, a `modifyGlobal` combinator is a good idea.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/.#Libc.hs
```diff
@@ -0,0 +1 @@
+rdockins@000288-rdockins-4.local.736
```
langston-barrett: Accidentally committed Emacs file

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -461,55 +489,65 @@ callRealloc sym mvar alignment (regValue -> ptr) (regValue -> sz) =
...
+  :: (IsSymInterface sym, HasPtrWidth wptr, wptr ~ ArchWidth arch, ?lc :: TypeContext)
+  => sym
+  -> GlobalVar Mem
+  -> RegEntry sym (LLVMPointerType wptr)
+  -> RegEntry sym (BVType wptr)
+  -> RegEntry sym (BVType wptr)
+  -> OverrideSim p sym (LLVM arch) r args ret (RegValue sym (BVType 32))
+callPosixMemalign sym mvar (regValue -> outPtr) (regValue -> align) (regValue -> sz) =
+  case asUnsignedBV align of
+    Nothing -> fail $ unwords ["posix_memalign: alignment valuem must be concrete:", show (printSymExpr align)]
```
langston-barrett:
```suggestion
    Nothing -> fail $ unwords ["posix_memalign: alignment value must be concrete:", show (printSymExpr align)]
```

langston-barrett:
A lot of the overrides look a lot nicer with `modifyGlobal`! :smiley:

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
robdockins: For now, I'm going to leave the dummy `getenv`.  I've opened ticked #302 to track the issue of a more realistic implementation for `getenv`, `putenv`, etc.

PR #292, opened 2019-07-31
Add a call stack to the UnresolvableFunction exception

langston-barrett:
Example:
```
Could not resolve function: _ZN6google4base21CheckOpMessageBuilderC1EPKc
Called at: :692:31
In _ZN6google17MakeCheckOpStringImmEEPNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEERKT_RKT0_PKc at :692:31
In _ZN6google12Check_GEImplImmEEPNSt3__112basic_stringIcNS1_11char_traitsIcEENS1_9allocatorIcEEEERKT_RKT0_PKc at :721:190
In _ZN4fizz3ShaINS_6Sha256EE4hmacEN5folly5RangeIPKhEERKNS3_5IOBufENS4_IPhEE at :18:43
```

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/Operations.hs
+++ b/crucible/src/Lang/Crucible/Simulator/Operations.hs
@@ -299,27 +299,31 @@ abortPartialResult s tgt pr =
 --   a callable function.  This usually indicates a programming error,
 --   but might also be used to allow on-demand function loading.
 --
---   The ProgramLoc argument references the call site for the unresolved
+--   The 'ProgramLoc' argument references the call site for the unresolved
 --   function call.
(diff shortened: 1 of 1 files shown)


robdockins:
We should probably demangle names (or have an option to) before presenting them to the user.

langston-barrett:
> We should probably demangle names (or have an option to) before presenting them to the user.

Unfortunately, we don't really have the capability to reliably do this. The `itanium-abi` library can demangle some C++ names, but not any with templates. We could just try to demangle each name and just present the ones for which we get `Left` as-is.

PR #293, opened 2019-08-01
Fix/vectorized llvm comparisons

robdockins:
Allows the LLVM instructions `icmp` and `select` to be vectorized. `fcmp` should probably be vectorized as well before we merge.


robdockins:
I think this is ready to merge now.

robdockins:
@langston-barrett ping

PR #294, opened 2019-08-01
Cxx/lax memory

robdockins:
This PR relaxes some of the rules regarding when pointers can be compared in ways that are disallowed by the C standard, but are nonetheless common and appear in `libcxx`.

This PR exists as a reminder to

1. Setup up configuration knobs for being able to selectively enable this, and similar, relaxations.
2. Investigate if we can/should equip pointers with a more principled total order by e.g., abstracting over all possible allocation block orderings using an arbitrary permutation.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM.hs
@@ -8,6 +8,7 @@
 -- Stability        : provisional
 ------------------------------------------------------------------------
 
+{-# LANGUAGE ImplicitParams #-}
 {-# LANGUAGE TypeFamilies #-}
 
 module Lang.Crucible.LLVM
@@ -53,8 +54,9 @@ llvmGlobals
 llvmGlobals ctx mem = emptyGlobals & insertGlobal var mem
   where var = llvmMemVar $ ctx
 
-llvmExtensionImpl :: HasPtrWidth (ArchWidth arch) => ExtensionImpl p sym (LLVM arch)
-llvmExtensionImpl =
+llvmExtensionImpl :: (HasPtrWidth (ArchWidth arch)) => MemOptions -> ExtensionImpl p sym (LLVM arch)
+llvmExtensionImpl mo =
+  let ?memOpts = mo in
   ExtensionImpl
   { extensionEval = llvmExtensionEval
   , extensionExec = llvmStatementExec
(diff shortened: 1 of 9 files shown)


robdockins:
I think this is ready to go in now.  The two pointer relaxations this branch considers are now controlled by a pair of boolean flags in a `MemOptions` record, which must be passed in when setting up the simulator.

crucible-llvm/src/Lang/Crucible/LLVM.hs
```diff
@@ -53,8 +54,9 @@ llvmGlobals
 llvmGlobals ctx mem = emptyGlobals & insertGlobal var mem
   where var = llvmMemVar $ ctx
 
-llvmExtensionImpl :: HasPtrWidth (ArchWidth arch) => ExtensionImpl p sym (LLVM arch)
-llvmExtensionImpl =
+llvmExtensionImpl :: (HasPtrWidth (ArchWidth arch)) => MemOptions -> ExtensionImpl p sym (LLVM arch)
+llvmExtensionImpl mo =
+  let ?memOpts = mo in
```
kquick: I'm assuming that this is an implicit param to avoid a much more intrusive implementation that would have threaded this explicitly through lots of places to make it available where it was needed?  If so, you might want to mention that in the PR description for posterity.

kquick:
Although deferring the control to the client is a nice option, we would still need to have some sort of indicator for what was being attempted to allow the client to decide what type of undefined operation was being attempted and at this stage the only real options are "fail" or "continue", so I don't think it would simplify the implementation very much.  Like you said though, we could transition to that later on if it turns out to be useful.

crucible-llvm/src/Lang/Crucible/LLVM.hs
robdockins: Yeah... at the moment it doesn't make much difference, but if similar options become more prevalent, I expect it will be a win.

robdockins:
I'd be open to that kind of refactoring at some point, but I don't think it would simplify things much now.  Also, we actually change the definition of pointer comparison depending on the flag, so a system for filtering UB isn't quite enough.

siddharthist:
@kquick I don't understand your comment. I'm suggesting a similar mechanism to what is in this MR, the difference would be allowing clients to enable or disable checking for _any_ kind of undefined behavior, whereas this MR addresses two specific instances. 

@robdockins Doesn't the "lax" definition of pointer equality subsume the default one? It seems to me we just need to perform the lax check and toggle whether the assertion about the base pointer equality is made.

kquick:
I was suggesting that we would need to define various `UndefinedBehavior` instances (I'm assuming `e` is a parameter over the language) and my _assumption_ is that the net changes for the current needs would probably be similar or possibly more for that approach.  I like the aspect of your proposal adding more ability for the client to control things, I'm just being conservative and agree that those types of generalizations could be a future direction.  I'll come by to chat though in case I've misunderstood.

siddharthist:
@kquick Thanks for clarifying!

PR #295, opened 2019-08-01
Support for reporting CallStack in SimError.

robdockins:


robdockins:
I'm not entirely happy with the technical approach here, and it's fallen rather behind master. See #424 for further discussion of the underlying issue.

PR #299, opened 2019-08-08
Add execution feature to observe argument buffer sizes

chameco:


crucible-llvm/src/Lang/Crucible/LLVM/ArraySizeProfile.hs
```diff
@@ -0,0 +1,153 @@
...
+ptrStartsAlloc _ = Nothing
+
+ptrAllocSize ::
+  forall sym w. W4.IsExpr (W4.SymExpr sym) =>
+  [G.MemAlloc sym] ->
+  C.LLVMPtr sym w ->
+  Maybe Int
+ptrAllocSize mem (C.llvmPointerView -> (blk, _)) = msum $ inAlloc <$> mem
+  where inAlloc :: G.MemAlloc sym -> Maybe Int
+        inAlloc (G.Alloc _ _ Nothing _ _ _) = Nothing
```
andreistefanescu: I prefer
```
inAlloc mem_alloc
  | (G.Alloc _ a (Just sz) _ _ _) <- mem_alloc
  , Just a ==  W4.asNat blk ->
    Just $ fromIntegral $ W4.asUnsignedBV sz
  | otherwise -> Nothing
```

crucible-llvm/src/Lang/Crucible/LLVM/ArraySizeProfile.hs
```diff
@@ -52,40 +53,41 @@ import qualified Lang.Crucible.LLVM.MemModel as C
...
-            then fromIntegral <$> asUnsignedBV size
+            then fromIntegral <$> W4.asUnsignedBV sz
             else Nothing
         inAlloc _ = Nothing
 
-ptrIsArray :: IsExpr (SymExpr sym) =>
-              [G.MemAlloc sym] ->
-              C.LLVMPtr sym w ->
-              Maybe Int
+ptrIsArray ::
```
siddharthist: I usually think of functions with names containing the word `Is` as predicates, i.e. returning booleans. The name doesn't make the significance of the return value clear.

siddharthist: This same comment applies below. Since all of these functions are used only once by the functions after them, it might be more appropriate to put them in a `where` clause.

chameco: Agreed entirely, these names have been irritating me for the same reason. It's kind of a weird situation where the functions are basically predicates, just with some extra data attached to true. Wanting to use the `Monad` instance for `Maybe` has led to some similarly-weird design choices, e.g. `ptrStartsAlloc` returning `Maybe ()` rather than `Bool`.

I think I'll change the names to be less predicate-like and just have `ptrStartsAlloc` return `Bool`, unless you have another idea?

siddharthist: @chameco That makes sense to me!

PR #304, opened 2019-08-13
fix cmd line arg for yices mcsat

pnwamk:
After tinkering a bit and checking the docs (http://yices.csl.sri.com/doc/context-operations.html#creation-and-configuration), it appears `--mcsat` is not a valid flag for the `yices` cmd line executable. Instead, setting the logic flag appropriately  enables MCSAT usage in Yices.

E.g., on my machine:
```
$ yices --mode=push-pop --print-success --mcsat
yices: invalid option: --mcsat
Try 'yices --help' for more information
```

whereas:
```
$ yices --mode=push-pop --print-success --logic=QF_NRA
yices> 
```

### DIFF
--- a/what4/src/What4/Solver/Yices.hs
+++ b/what4/src/What4/Solver/Yices.hs
@@ -608,7 +608,7 @@ yicesStartSolver features auxOutput sym = do -- FIXME
   yices_path <- findSolverPath yicesPath cfg
   enableMCSat <- getOpt =<< getOptionSetting yicesEnableMCSat cfg
   let args = ["--mode=push-pop", "--print-success"] ++
-             if enableMCSat then ["--mcsat"] else []
+             if enableMCSat then ["--logic=QF_NRA"] else []
       hasNamedAssumptions = features `hasProblemFeature` useUnsatCores ||
                             features `hasProblemFeature` useUnsatAssumptions
   when (enableMCSat && hasNamedAssumptions) $


kquick:
Can you document the version of yices here (just in case this changes over time)?

pnwamk:
On my machine where I observed this behavior I've got the following:
```
yices --version
Yices 2.6.1
Copyright SRI International.
Linked with GMP 6.1.2
Copyright Free Software Foundation, Inc.
Build date: 2019-04-03
Platform: x86_64-apple-darwin18.5.0 (release)
```

(Should I document it anywhere else?)

kquick:
I don't have anyplace else in mind, although feel free to if you find someplace.  We do validate we are using a supported version (https://github.com/GaloisInc/crucible/commit/0d9c3294ab44aaafd8be0b7677c83bc62cf99bda#diff-3fdf8fdc4781e09ccebfb35bd993a00bR949), and it's not likely that this UI will change much, I just thought it would be a good idea to indicate which version we are currently sync'd to for this change.

This change looks good to me: feel free to merge it unless you'd like input from rdockins or someone else first.

pnwamk:
Ah! I did more digging and it seems there is a command-line interface (`yices-smt2`) designed to handle SMT-LIB 2.5 input, which _does_ support a simple `--mcsat` flag (see [sec 5.1](http://yices.csl.sri.com/papers/manual.pdf) of the manual).

So... I'm not sure if it's better to just use that interface, or if it's better to pass a particular `--logic=LOGIC` flag to the "standard" `yices` interface s.t. `LOGIC`will use MCSAT. (One concern is there's many selectable logics with that flag, and I'm not sure which one would be ideal if we did go that route. The supported logics are listed in table 1.2 of the [manual](http://yices.csl.sri.com/papers/manual.pdf) on page 6.)

Anyway, I'm not not at all confident this particular "fix" is correct/ideal. Calling @robdockins =)

atomb:
The latest version of the `yices` executable built from the code on the `mcsat-bv` branch of the `yices2` repository does support the `--mcsat` flag (though it's not documented). It's admittedly not what most people will be using yet, so it has a limited audience, but it's useful for some of my current work (and will be useful to everyone before long).

pnwamk:
Cool -- okay I'll close this for now then :+1:

PR #312, opened 2019-09-27
Support memory invalidation in LLVM memory model

chameco:
Adds a new `WriteSource` called `MemInvalidate`. Reading from invalidated memory yields an error with a user-specified message.

This facilitates a `saw-script` change that should fix the compositional verification soundness bug described in GaloisInc/saw-script#30.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
@@ -28,6 +28,7 @@
 
 module Lang.Crucible.LLVM.Globals
   ( initializeMemory
+  , initializeMemoryConstGlobals
   , populateGlobal
   , populateGlobals
   , populateAllGlobals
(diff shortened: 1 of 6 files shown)


langston-barrett:
This breaks `macaw-symbolic`, so can't yet be tested against S2N/SAWscript on Fryingpan: http://fryingpan.dev.galois.com/hydra/jobset/s2n/PR-explicit-invalid-write.submodules-ghc865#tabs-jobs

If you create a branch in `macaw-symbolic` with the same name ("explicit-invalid-write") and update it, the tests will automatically pick up on the changes you make there and build SAW+S2N against both those updates together.

atomb:
With the naming changes, the `crux-llvm` build is failing. It should be easy to fix, though: all the references `initializeMemory` in `CruxLLVMMain` should become `initializeAllMemory`. I'd make the change myself, except that the branch is in @chameco's fork of the repo.

Once that's fixed, are there any remaining objections to merging this?

PR #314, opened 2019-10-04
Add option for per-goal timeout with Yices

atomb:
This adds an option to the What4 Yices backend to support per-goal timeouts, and updates Crux to use that option.

It's slightly dissatisfying because in Yices it can't be set globally and it's necessary to emit a `(set-timout n)` command before every `(check)` command. Because of the current architecture of `SMTWriter`, we need to expose a separate command to set the timeout for the current goal. However, this command does not exist in standard SMTLib2. So the method in the `SMTWriter` class returns a `Maybe Command`. This isn't ideal, and I'd love thoughts on a better way to do it.

### DIFF
--- a/what4/src/What4/Protocol/SMTWriter.hs
+++ b/what4/src/What4/Protocol/SMTWriter.hs
@@ -786,6 +786,10 @@ class (SupportTermOps (Term h)) => SMTWriter h where
   -- | Reset the solver state, forgetting all pushed frames and assertions
   resetCommand  :: f h -> Command h
 
+  -- | Set the timeout for the next goal. Returns @Nothing@ if setting
+  -- per-goal timeouts is not supported.
+  setGoalTimeoutCommand  :: f h -> Integer -> Maybe (Command h)
+
   -- | Check if the current set of assumption is satisfiable
   checkCommand  :: f h -> Command h
 
--- a/what4/src/What4/Protocol/SMTWriter.hs
+++ b/what4/src/What4/Protocol/SMTWriter.hs
@@ -67,6 +67,7 @@ module What4.Protocol.SMTWriter
   , Command
   , addCommand
   , addCommandNoAck
+  , addCommands
   , mkFreeVar
   , TypeMap(..)
   , freshBoundVarName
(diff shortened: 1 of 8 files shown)


what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -786,6 +786,10 @@ class (SupportTermOps (Term h)) => SMTWriter h where
   -- | Reset the solver state, forgetting all pushed frames and assertions
   resetCommand  :: f h -> Command h
 
+  -- | Set the timeout for the next goal. Returns @Nothing@ if setting
+  -- per-goal timeouts is not supported.
+  setGoalTimeoutCommand  :: f h -> Integer -> Maybe (Command h)
+
```
travitch: My only request is that we either use a more structured type for the timeout, or at the very least document the units here (in case some solvers use milliseconds vs seconds or something).

atomb: Good point. Maybe it should be a `DiffTime`?

travitch: That is kind of appealing, but maybe it isn't worth an extra dependency... on the other hand, it seems highly likely that what4 already depends on something that indirectly depends on `time`

atomb: From a quick look, `crucible`, `saw-core`, `cryptol`, `llvm-pretty-bc-parser`, and `crux` all depend on `time`.

atomb: Oh, and so does `directory`, which `what4` directly depends on. And it's a base package included with GHC.

atomb:
I've further investigated the option of having the `checkCommand` function return a list of commands. Unfortunately, it doesn't receive an argument that could include configuration options (such as a timeout). Changing this would require changing all uses of the `SMTWriter` class.

robdockins:
> I've further investigated the option of having the `checkCommand` function return a list of commands. Unfortunately, it doesn't receive an argument that could include configuration options (such as a timeout). Changing this would require changing all uses of the `SMTWriter` class.

Could you make it so that `Command` (for Yices, at least) is a newtype over a function that takes a parameter?  Then, I think you could unwrap it in `writeCommand` and apply the parameters.

atomb:
I've refactored the Yices solver interface to have a global goal timeout option, using the approach @robdockins suggested of making the `Command` type for a Yices connection be `Connection s -> YicesCommand`. This still requires the `checkCommand` and `checkWithAssumptions` methods in `SMTWriter` to return a list of commands, but that works out pretty cleanly. Unfortunately, this precludes using `DiffTime` for time intervals because What4 doesn't allow that type in configuration options.

atomb:
Okay, I'm pretty happy with the state of things now. Any objections to merging, @travitch or @robdockins ?

travitch:
No objections from me

PR #319, opened 2019-10-14
issue198

robdockins:
I think I finally tracked down a pesky bug causing out-of-bounds errors when switching `Bytes` and the number of elements in an array type to be backed by the `Natural` type, which was fixed in ae9defcfd.


robdockins:
Well, apparently I've still not found all the places where switching to Natural causes arithmetic underflow, so this PR isn't ready to go in.

PR #329, opened 2019-10-23
explicit float modes; add a crux argument for float modes

pnwamk:


pnwamk:
@robdockins it's now being passed down into the ExprBuilder record :+1:

crucible-saw/src/Lang/Crucible/Backend/SAWCore.hs
```diff
@@ -193,7 +193,7 @@ bindSAWTerm sym bt t = do
 newSAWCoreBackend ::
   SC.SharedContext ->
   NonceGenerator IO s ->
-  IO (SAWCoreBackend s (Yices.Connection s) fs)
+  IO (SAWCoreBackend s (Yices.Connection s) (Flags FloatReal))
 newSAWCoreBackend sc gen = do
```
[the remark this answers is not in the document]
pnwamk: Fixed via 8b2b77fefad474562c4977b6dc929160fd237ce3.

crucible-saw/src/Lang/Crucible/Backend/SAWCore.hs
```diff
@@ -191,10 +191,11 @@ bindSAWTerm sym bt t = do
   return sbVar
 
 newSAWCoreBackend ::
```
pnwamk: @travitch am I right assuming `newSawCoreBackend` should take the float mode flag too? I ask because the current API is fixing the SAWCore solver as Yices, so it wouldn't be _crazy_ to keep it also tied to the real interpretation mode for now I guess... I could be convinced either way ¯\\_(ツ)\_/¯

pnwamk:
@tristan do we need to worry about version-bumping any of the libs in this repository w/ these changes? (E.g., I noticed `saw-script` depends on `what4 >= 0.4`... seems like it needs to be `> 0.4.2` (i.e., the current version) to build though after these changes...)

pnwamk:
The version bump question _may_ be less relevant now that the changes to `newSAWCoreBackend` in this PR have been reverted (https://github.com/GaloisInc/crucible/pull/329/commits/4dd3fdde61657472eb2ad03dea2e1bdc8086476f)... at least as far as it concerns saw-script.

crucible-saw/src/Lang/Crucible/Backend/SAWCore.hs
pnwamk: Oops, I think I misspoke a little here, sorry. Although this function is pinned at Yices, you could still annotate whatever float flag you wanted previously, so pinning it at `FloatReal` is _not_ backwards compatible =\ I think the simplest/sanest thing is to just have the flag as I originally had it.

atomb: It's probably worth noting that whatever you do here is unlikely to break anything. We currently have no way to write specifications of programs that use floats (though that's coming soon!), so we aren't really using this aspect of What4 in any visible way. Symbolic floats or reals can't be created in the normal SAW workflow.

kquick:
> @tristan do we need to worry about version-bumping any of the libs in this repository w/ these changes? (E.g., I noticed `saw-script` depends on `what4 >= 0.4`... seems like it needs to be `> 0.4.2` (i.e., the current version) to build though after these changes...)

@pnwamk (cc: @travitch), you can version bump if you'd like to have that help with the cabal compatibility checking.  For unreleased dev updates, we are adding a 5th field to the version, and starting at `.99` for that field.  This allows the dev versions to fall "between" official release versions and still be sensibly used as constraints in downstream cabal files.  This is relatively new, and so we haven't officially utilized outside of parameterized-utils (e.g. https://github.com/GaloisInc/parameterized-utils/blob/master/parameterized-utils.cabal), but in this case, you could update what4's version to "0.4.2.0.99" and then anything downstream could specify "what4 > 0.4.2.0" to require at least that version (and what4's version would be changed to 0.4.2.1 or 0.4.3.0 or 0.5 at actual release time, depending on the contents at release).  Using submodules references is also fine and this is only a partially implemented new versioning scheme, but I wanted to followup so you had this info.

PR #352, opened 2019-11-22
WIP SV-COMP

robdockins:
Make crux into an SV-COMP competetor


robdockins:
#452 is a better take on this

PR #359, opened 2019-12-03
minor fixes

robdockins:
Fix some minor issues found while working on other things, and add a "fail fast" option to crux that skips attempting to prove further goals once one is disproved


crux/src/Crux/Goal.hs
```diff
@@ -116,16 +116,18 @@ proveGoals opts _ctxt Nothing =
...
   do let sym = ctxt ^. ctxSymInterface
      sp <- getSolverProcess sym
-     goalNum <- newIORef (0,0) -- total, proved
+     goalNum <- newIORef (0,0,0) -- total, proved, disproved
      nameMap <- newIORef Map.empty
      unless hasUnsatCores $
       sayWarn "Crux" "Warning: skipping unsat cores because MC-SAT is enabled."
      res <- inNewFrame sp (go sp goalNum gs0 nameMap)
-     (tot,proved) <- readIORef goalNum
+     (tot,proved,disproved) <- readIORef goalNum
```
langston-barrett: Should `tot = proved + disproved`? If so, it might be nice to add an assertion to that effect, just in case we forget to add to the counters somewhere, etc.

what4/src/What4/Interface.hs
```diff
@@ -537,7 +537,6 @@ class (IsExpr (SymExpr sym), HashableF (SymExpr sym)) => IsExprBuilder sym where
 
   -- | Equality of boolean values
   eqPred  :: sym -> Pred sym -> Pred sym -> IO (Pred sym)
-  eqPred sym x y = notPred sym =<< xorPred sym x y
```
langston-barrett: What's the problem with this default implementation?

robdockins: Nothing, really, but `xor` is defined as the negation of equality now, and eqPred is actually the primitive that is defined.  The only real problem is that the definition is misleading

PR #365, opened 2019-12-05
Add inNewFrameWithVars to Protocol.Online

danmatichuk:
This pushes a new assumption frame where the
given bound variables are considered
free variables for the duration of the frame.

### DIFF
--- a/what4/src/What4/Protocol/Online.hs
+++ b/what4/src/What4/Protocol/Online.hs
@@ -20,6 +20,7 @@ module What4.Protocol.Online
   , pop
   , reset
   , inNewFrame
+  , inNewFrameWithVars
   , check
   , checkAndGetModel
   , checkWithAssumptions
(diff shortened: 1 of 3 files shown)


danmatichuk:
This works by adding bindings to fresh free variables that are only live for the current frame. Here we are relying on the cache to enforce our scoping, but I'm not sure if these invented variables can still escape the scope (e.g. by being bound in a cache that's persistent across frames).

robdockins:
This looks right to me, the `DeleteOnPop` mode ought to have the scoping behavior you want.  It would be nice to see some additions to the test suite that exercises this behavior.

what4/src/What4/Solver/Yices.hs
```diff
@@ -436,7 +436,7 @@ newConnection stream ack reqFeatures timeout bindings = do
   let nlSolver = reqFeatures `hasProblemFeature` useNonlinearArithmetic
   let features | efSolver  = useLinearArithmetic
                | nlSolver  = useNonlinearArithmetic .|. useIntegerArithmetic
-               | otherwise = yicesDefaultFeatures
+               | otherwise = reqFeatures
```
robdockins: This looks like the right fix, but we should probably also add `useIntegerArithmetic` to the default features.

danmatichuk: This snuck in by accident, this is orthogonal to the bound variable issue.

robdockins:
Looks good to me

PR #375, opened 2019-12-13
Many crux and crux-llvm improvements

atomb:
This includes a variety of improvements to `crux-llvm`, including:

* Changed how results are summarized (including proved, disproved, unknown, and incomplete results).
* Add override for `__VERIFIER_nondet_bool`.
* Include side conditions for resource exhaustion.
* FIx a bug in source rendering.
* Add short error description to `AssertFailureSimError`.

There's one horrible hack in `summarizeTree` that I'd be interested in better solutions to if anyone has ideas.


crucible/src/Lang/Crucible/CFG/Extension/Safety.hs
```diff
@@ -262,6 +263,18 @@ class HasStructuredAssertions (ext :: Type) where
                , "else " <$$> indent 2 doc2
                ])
 
+  summarizeTree :: (IsExprBuilder sym, IsExpr (SymExpr sym))
+                => proxy ext
+                -> AssertionClassifierTree ext (RegValue' sym)
+                -> String
+  summarizeTree proxyExt =
+    intercalate ", " . nub . filter (not . ("muxing" `isPrefixOf`)) .
```
robdockins: Is this the hack you are worried about?

atomb: Yep, that's the one. There are some messages that start with `muxing` that just aren't useful in the output (and there's a TODO comment in the source noting that already). It might be that they could be made more useful, or it might be that they could somehow never be generated? That might require some refactoring, though.

what4/src/What4/Protocol/SMTWriter.hs
```diff
@@ -2035,6 +2036,8 @@ appSMTExpr ae = do
     RealDiv xe ye -> do
       x <- mkBaseExpr xe
       case ye of
+        SemiRingLiteral SR.SemiRingRealRepr r _ | r == 0 -> do
+          freshBoundTerm RealTypeMap $ rationalTerm infinity
```
robdockins: I don't think I'm really comfortable with this change.  What's the reasoning?

atomb: It is a somewhat unsatisfactory change, I agree. It came up from an SV-COMP benchmark that involves verification of the `atan` and `atan2` functions on floats. In this code, there's a branch that does a division by a value that's initialized to zero. It may be (and may not be?) the case that that particular situation is unreachable. However, we still wind up building and sending to the SMT solver a division by zero. Without this case, the program just crashes with a `zero denominator` message from a call to `error`.

When working with integers, we've usually made division total and added a side condition to ensure that the denominator is non-zero. With reals, something similar is probably sensible. With floats, the "correct" result depends on what mode you're in. Infinity is one of the legal values, as is NaN.

robdockins: It looks like we're avoiding using the division operator for some reason.  SMTLib2 and Yices both support a division operator, so we don't actually need to do that.  We should treat this just the same way as integer division and use the division operator.

crux/src/Crux.hs
```diff
@@ -287,3 +289,25 @@ runSimulator lang opts@(cruxOpts,_) =
        writeProf profInfo
 
      readIORef gls
+
+reportStatus :: (?outputConfig::OutputConfig) => ProvedGoals a -> IO ()
```
robdockins: We report "Overall status: Valid" even if simulation timed out.  It seems like incomplete would be a better verdict.

crux/src/Crux.hs
```diff
@@ -278,7 +278,9 @@ runSimulator lang opts@(cruxOpts,_) =
                 mgt <- provedGoalsTree ctx' proved
                 case mgt of
                   Nothing -> return ()
-                  Just gt -> modifyIORef gls (Seq.|> gt)
+                  Just gt ->
+                    do reportStatus gt
```
robdockins: If there is more than one path (e.g., in the SplitDFS mode) we'll report the proof status separately for each path... probably it would be better to compile all the results and give a single report at the end?

what4/src/What4/Protocol/SMTWriter.hs
atomb: Oh, yeah, we should just use the SMTLib division operator (making sure to avoid concrete division). I'll change that.

crux/src/Crux.hs
atomb: Good point. I'll change that.

crux/src/Crux.hs
atomb: Yep, good point.

atomb: Fixed in 6b3ca68

what4/src/What4/Protocol/SMTWriter.hs
atomb: On closer inspection, this is a little harder. I think the reason for the `x * rationalTerm (recip r)` is that the `SupportTermOps` class uses a `Num` superclass for numeric operations, which includes `*` but not `/` or `div`. And it's the `recip` operation (on a concrete value) that led to the error. The `SupportTermOps` class has named `intDiv` and `floatDiv` functions, but nothing for reals.

robdockins: OK, let's do this: the immediate fix is just to add a guard that `r /= 0` before calling `recip` and allow it to fall into the second case if it is.  Separately, let's make a ticket for adding direct support for real division.

atomb: Turns out it was easy to add real division. See  28feb2e.

crux/src/Crux.hs
atomb: Fixed in e4dd45f.

crucible/src/Lang/Crucible/CFG/Extension/Safety.hs
robdockins: It looks like this is the one remaining piece holding up this PR... I'm inclined to say we should leave this as-is for now, maybe with a comment describing why it's a hack and what we'd like to do better.

Perhaps a more comprehensive look at how these assertion trees are used should be a separate ticket.

atomb: I think that sounds reasonable. Something should probably change, but it's unclear exactly what that should be yet. And, for now, the `summarizeTree` function does what its single caller needs it to do.

PR #377, opened 2019-12-16
Add an online backend for Boolector

travitch:
The Boolector devs (very) recently added all of the features we need to support
an online backend (including the global scoping option).  This commit adds the
necessary interface bits to use Boolector in online mode.

### DIFF
--- a/crucible/src/Lang/Crucible/Backend/Online.hs
+++ b/crucible/src/Lang/Crucible/Backend/Online.hs
@@ -50,6 +50,9 @@ module Lang.Crucible.Backend.Online
     -- ** Z3
   , Z3OnlineBackend
   , withZ3OnlineBackend
+    -- ** Boolector
+  , BoolectorOnlineBackend
+  , withBoolectorOnlineBackend
     -- ** CVC4
   , CVC4OnlineBackend
   , withCVC4OnlineBackend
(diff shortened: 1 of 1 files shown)


robdockins:
Does boolector support a per-goal timeout?  That's the other nice feature to have in an online backend.

Otherwise what's here looks about right to me.

travitch:
I don't see any options for per-goal timeouts (and it doesn't look like the parser has support for anything like that)

kquick:
How feasible is it to update the tests to have one that invokes the Boolector back-end?

travitch:
It is slightly annoying right now because some of the features we need are only supported in an unreleased version of boolector.  We could add something but disable them until a release lands

andreistefanescu:
@travitch or even better, let's use this opportunity to add a few tests with queries for which `Boolector` is the solver of choice.

robdockins:
Since the developers seem receptive, it might be nice to also ask for a per-goal timeout option.  I don't think there is SMTLib2 syntax for that, hence Z3, CVC4, etc. make it a command-line option.

travitch:
There aren't really any tests we've seen where Boolector is the solver of choice, but it is roughly comparable to yices right now for bitvectors.

robdockins:
@travitch, can you follow up with the Boolector devs about the possibility of a per-goal timeout option? At that point, it would be basically on-par with the other solvers in terms of online functionality.

PR #380, opened 2019-12-18
Use both default and requested features in the What4 Yices backend

### LINKED ISSUE #379: Yices does not have necessary features enabled
Since #372, some SAWScript regression tests using `w4_unint_yices` have been failing - see https://travis-ci.org/GaloisInc/saw-script/builds/625777088. It looks like the `useLinearArithmetic` feature flag is not being set. Previously, it was always enabled (along with the rest of `yicesDefaultFeatures`).

It would be easy to always enable the default features in addition to the requested features, but I'm unsure if this goes against the intent of #372.

ping @danmatichuk

chameco:
Fixes #379


robdockins:
I'm not sure I like this fix.  Part of the intent here is to (eventually) allow solvers to chose the minimal logic that can be used to solve the problem, and always ORing in the default features keeps us from ever computing a logic smaller than the default one.

I think the right fix for this is to make sure that `predicateVarInfo` includes the linear arithmetic feature when needed.

PR #386, opened 2019-12-24
[crux-llvm] Report version of clang used in testing.

kquick:


crux-llvm/test/Test.hs
```diff
@@ -1,26 +1,41 @@
...
+  -- Determine which version of clang will be used for these tests.
+  -- An exception (E.g. in the readProcess if clang is not found) will
+  -- result in termination (test failure).  Uses partial 'head' but
+  -- this is just tests, and failure is captured.
+  clangBin <- lookupEnv "CLANG" >>= \case
+    Just x -> return x
+    Nothing -> return "clang"
+  let isVerLine = isPrefixOf "clang version"
+      getVer = head . drop 2 . words . head . filter isVerLine . lines
+  ver <- getVer <$> readProcess "clang" [ "--version" ] ""
```
robdockins: You compute `clangBin`, but don't use it here

kquick: Doh!  Thanks for catching that.  I just update the PR.

PR #400, opened 2020-01-14
Fix(?) online solver process termination

travitch:
The `interruptProcessGroupOf` function from System.Process sends SIGINT, which
is the equivalent of Ctrl+C.  I'm not sure if Ctrl+C behavior is quite the right
notion we want to capture; I think SIGTERM (signal 15 on x86/Linux and most
other platforms) is the "soft kill" that is still handleable.

Selfishly, I want to make this change because yices (in native non-SMT mode) has
special handling for SIGINT that doesn't work when MCSAT is enabled.  yices-smt2
has the correct behavior.

Switching to SIGTERM (via `terminateProcess`) has the expected behavior on
yices, z3, and cvc4.

Note that this is not the uncatchable SIGKILL (usually signal 9).

### DIFF
--- a/what4/src/What4/Protocol/Online.hs
+++ b/what4/src/What4/Protocol/Online.hs
@@ -47,7 +47,7 @@ import           System.Exit
 import           System.IO
 import qualified System.IO.Streams as Streams
 import           System.Process
-                   (ProcessHandle, interruptProcessGroupOf, waitForProcess)
+                   (ProcessHandle, terminateProcess, waitForProcess)
 import           Text.PrettyPrint.ANSI.Leijen hiding ((<$>), (<>))
 
 import           What4.Expr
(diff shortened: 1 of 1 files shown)


travitch:
Note that the relevant yices code is at https://github.com/SRI-CSL/yices2/blob/master/src/frontend/yices/yices_reval.c#L712 (which calls https://github.com/SRI-CSL/yices2/blob/master/src/context/context.c#L6264)

robdockins:
Humm, I think you're right.  The semantics on *nix systems for `terminateProcess` seems more clearly what we want.  On Windows, it's a bit less clear (just because I understand the process lifetime less well there), but I think it is fine.

robdockins:
Also, probably should file a yices bug.  If they aren't going to do anything in MCSAT mode on SIGINT, they should just install the default handler instead.

travitch:
Thanks - I have a note to submit a yices bug

PR #406, opened 2020-01-16
GHC 8.8 updates

robdockins:
Update to build on GHC 8.8.  I suspect some of these fixes will break builds on older GHC.  Fixing may require some additional CPP magic. As such, I'm marking this a draft PR for now.

### DIFF
--- a/crucible-jvm/src/Lang/Crucible/JVM/Simulate.hs
+++ b/crucible-jvm/src/Lang/Crucible/JVM/Simulate.hs
@@ -6,6 +6,7 @@ License          : BSD3
 Maintainer       : sweirich@galois.com
 Stability        : provisional
 -}
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE EmptyCase #-}
 {-# LANGUAGE FlexibleContexts #-}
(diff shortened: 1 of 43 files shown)


robdockins:
This now builds clean on 8.4 and 8.6 in CI and locally on 8.8.  Once we fix some upstream dependency issues on the CI system, I expect it will build clean on 8.8 also.

kquick:
Still seeing errors with GHC8.8 (http://fryingpan.dev.galois.com/hydra/build/672259):

```
[24 of 51] Compiling Lang.Crucible.Utils.RegRewrite ( src/Lang/Crucible/Utils/RegRewrite.hs, dist/build/Lang/Crucible/Utils/RegRewrite.o )

src/Lang/Crucible/Utils/RegRewrite.hs:185:18: error:
    • Couldn't match type ‘Type’ with ‘CrucibleType’
        arising from a functional dependency between:
          constraint ‘MonadState
                        (MapF (Nonce s0) (Nonce s))
                        (StateT (MapF (Nonce s0) (Nonce s)) (ST h))’
            arising from a use of ‘get’
          instance ‘MonadState s1 (StateT s1 m)’ at <no location info>
    • In a stmt of a 'do' block: mapping <- get
      In the expression:
        do mapping <- get
           case MapF.lookup n mapping of
             Just n' -> return n'
             Nothing -> do ...
      In an equation for ‘rename’:
          rename ng n
            = do mapping <- get
                 case MapF.lookup n mapping of
                   Just n' -> return n'
                   Nothing -> ...
    |
185 |       mapping <- get
    |                  ^^^
```

robdockins:
Ugh, I though d010d3c1a69ac359535d00400a489564a95985c7 would fix that.  I'll see if I can figure out why that's still happening.

PR #412, opened 2020-01-17
CBMC intrinsics / mode

robdockins:
Let's see if we can push some CBMC proofs through crux.


robdockins:
I haven't really tested these yet, but CI says their presence isn't breaking anything, so I'm going to merge these now to tidy up the PR list and prevent future merge conflicts.

PR #418, opened 2020-01-25
Squash warnings

robdockins:
fix a variety of warnings and also remove some obsolete CPP for older GHCs

### DIFF
--- a/what4/src/What4/Utils/StringLiteral.hs
+++ b/what4/src/What4/Utils/StringLiteral.hs
@@ -171,9 +171,23 @@ stringLitIndexOf (Char16Literal x) (Char16Literal y) k =
     Just n  -> toInteger n + toInteger k
 
 stringLitIndexOf (Char8Literal x) (Char8Literal y) k =
-  case BS.findSubstring y (BS.drop (fromIntegral k) x) of
+  case bsFindSubstring y (BS.drop (fromIntegral k) x) of
     Nothing -> -1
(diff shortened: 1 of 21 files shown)


what4/src/What4/Utils/StringLiteral.hs
```diff
@@ -171,10 +171,24 @@ stringLitIndexOf (Char16Literal x) (Char16Literal y) k =
...
 stringLitIndexOf (Char8Literal x) (Char8Literal y) k =
-  case BS.findSubstring y (BS.drop (fromIntegral k) x) of
+  case bsFindSubstring y (BS.drop (fromIntegral k) x) of
     Nothing -> -1
     Just n  -> toInteger n + toInteger k
 
+-- | Get the first index of a substring in another string,
+--   or 'Nothing' if the string is not found.
+--
+--   Copy/pasted from the old `bytestring` implementation because it was
```
langston-barrett: They have a suggestion in the deprecation warning to use `breakSubstring` instead, is that not appropriate here? https://github.com/haskell/bytestring/blob/a80c4261fafe5365d91e14e09e0f05f4b28684cc/Data/ByteString.hs#L1414

robdockins: Well, `breakSubstring` doesn't compute the desired index, so it has to be postprocessed, and it's not completely trivial, as seen below.  I think the deprecation is silly, CF https://github.com/haskell/bytestring/issues/180

PR #421, opened 2020-01-27
Implement “durable assertions”

atomb:
These will not be simplified away and will always show up in the final list of goals to prove. This commit modifies the overrides for the various assertion functions supported by `crux-llvm` to use durable assertions, so that users will always see a goal in the result for each assertion they explicitly specify in the input source code.

### DIFF
--- a/crucible/src/Lang/Crucible/Backend.hs
+++ b/crucible/src/Lang/Crucible/Backend.hs
@@ -48,6 +48,7 @@ module Lang.Crucible.Backend
 
     -- * Utilities
   , addAssertion
+  , addDurableAssertion
   , addAssertionM
   , addFailedAssertion
   , assertIsInteger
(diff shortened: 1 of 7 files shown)


robdockins:
You'll probably want to make, e.g., `__VERIFIER_error()` also produce durable assertions, as many benchmarks implement `assert` as `if (!p) { __VERIFIER_error(); }`

atomb:
I would think that the proof obligations generated by `__VERIFIER_error()`, which asserts `False`, would never be simplified to `True` and removed. Am I missing anything?

robdockins:
That is a really good point; yeah, you're right.

On the other hand, error calls guarded by an `if` will definitely be filtered out if `p` is concretely true.  Not sure there's much we can do about that.

atomb:
Yeah, perhaps the best approach to that case is to recommend that people use the API in `crucible.h` when doing new verifications, and make it clear that the SV-COMP API support exists primarily for backward compatibility with a large suite of existing benchmarks, for which explanations aren't as relevant because they're not part of the competition evaluation process.

crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -208,6 +209,13 @@ class IsBoolSolver sym where
   -- See also 'addAssertion'.
   addProofObligation :: sym -> Assertion sym -> IO ()
```
robdockins: I think I'd prefer to see a default implementation of `addProofObligation` here in terms of `addDurableProofObligation` rather than have the same default behavior copied into each of the implementations.

Let's also add a comment explaining that "obvious" (concretely true) proof obligations may be discarded.

atomb: 👍

atomb:
It's worth noting that I fixed the `crux-llvm` test suite with Apple's `clang` along with these changes. It's not entirely related, but showed up in the process of testing them.

PR #427, opened 2020-01-29
update with fixes for crucible/HEAD update

pnwamk:


pnwamk:
oops - ignore. Wrong target.

PR #428, opened 2020-01-30
Tr/crux separate online solver

travitch:
This set of commits adds options to crux to let the user choose different solvers for path satisfiability checking and proving goals.  Changes:

- Users can select a separate solver for path satisfiability checking (`--path-sat-solver`)
- Goals can be proved using solvers that do not have an online mode (e.g., boolector and dreal)
- The goal solver can be forced into offline mode even if it supports online solving (useful for eventually solving in parallel)
- The default behavior (if only `--solver` is specified) should be the same as before
- Adds support for the STP solver in both online and offline modes

### DIFF
--- a/crux/src/Crux.hs
+++ b/crux/src/Crux.hs
@@ -1,3 +1,6 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE NoMonoLocalBinds #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 {-# Language RankNTypes, ImplicitParams, TypeApplications, MultiWayIf #-}
 {-# Language OverloadedStrings, FlexibleContexts, ScopedTypeVariables #-}
 module Crux
@@ -11,33 +14,42 @@ module Crux
   ) where
 
 import Control.Lens
+import Control.Monad ( unless )
 import Data.Text (Text)
 import qualified Data.Text as Text
 import Data.Foldable
 import Data.IORef
 import Data.List(intercalate)
+import Data.Maybe ( fromMaybe )
+import Data.Proxy ( Proxy(..) )
 import qualified Data.Sequence as Seq
+import qualified Data.Text as T
 import Control.Exception(catch, displayException)
 import Control.Monad(when)
 import System.Exit(exitSuccess, ExitCode(..))
 import System.Directory(createDirectoryIfMissing)
 import System.FilePath((</>))
 
-import Data.Parameterized.Nonce(withIONonceGenerator, NonceGenerator)
+import Data.Parameterized.Classes
+import Data.Parameterized.Nonce(newIONonceGenerator, NonceGenerator)
+import Data.Parameterized.Some ( Some(..) )
 
 import Lang.Crucible.Backend
 import Lang.Crucible.Backend.Online
+import qualified Lang.Crucible.Backend.Simple as CBS
 import Lang.Crucible.Simulator
 import Lang.Crucible.Simulator.BoundedExec
 import Lang.Crucible.Simulator.BoundedRecursion
 import Lang.Crucible.Simulator.Profiling
 import Lang.Crucible.Simulator.PathSatisfiability
 
-import What4.Config (Opt, ConfigOption, setOpt, getOptionSetting, verbosity)
+import What4.Config (Opt, ConfigOption, setOpt, getOptionSetting, verbosity, extendConfig)
 import What4.InterpretedFloatingPoint (IsInterpretedFloatExprBuilder)
 import What4.Interface (IsExprBuilder, getConfiguration)
+import qualified What4.Expr.Builder as WEB
 import What4.FunctionName (FunctionName)
 import What4.Protocol.Online (OnlineSolver)
+import qualified What4.Solver as WS
 import What4.Solver.CVC4 (cvc4Timeout)
 import What4.Solver.Z3 (z3Timeout)
 import What4.Solver.Yices (yicesEnableMCSat, yicesGoalTimeout)
@@ -49,6 +61,7 @@ import Crux.Goal
 import Crux.Model
 import Crux.Report
 import qualified Crux.Config.Load as Cfg
+import qualified Crux.Config.Solver as CCS
 import Crux.Config.Common
 import Crux.Config.Doc
 import Crux.Extension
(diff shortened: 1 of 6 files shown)


crux/src/Crux.hs
```diff
@@ -117,53 +128,93 @@ showVersion l = outputLn ("crux version: " ++ Crux.version ++ ", " ++
...
+            Just s -> symCfg sym cvc4Timeout (floor (s * 1000))
+            Nothing -> return ()
+          k floatRepr sym
+        CCS.Z3Online -> withZ3OnlineBackend floatRepr nonceGen ProduceUnsatCores $ \sym -> do
+          case goalTimeout cruxOpts of
+            Just s -> symCfg sym z3Timeout (floor (s * 1000))
+            Nothing -> return ()
+          k floatRepr sym
+        CCS.STPOnline ->
+          -- We don't have a timeout option for STP
```
robdockins: Issue a warning, perhaps, if `goalTimeout` is configured but we can't accomidate?

crux/src/Crux/Goal.hs
```diff
@@ -129,24 +135,121 @@ proveToGoal _ allAsmps p pr =
...
+                   . (?outputConfig :: OutputConfig, sym ~ ExprBuilder t st fs)
+                  => WS.SolverAdapter st
+                  -> CruxOptions
+                  -> SimCtxt sym p
+                  -> Maybe (Goals (LPred sym asmp) (LPred sym ast))
+                  -> IO (Maybe (Goals (LPred sym asmp) (LPred sym ast, ProofResult (Either (LPred sym asmp) (LPred sym ast)))))
+proveGoalsOffline _adapter _opts _ctx Nothing = return Nothing
+proveGoalsOffline adapter opts ctx (Just gs0) = do
+  goalNum <- newIORef (0, 0, 0)
+  unless hasUnsatCores $ do
```
robdockins: Probably don't need this warning for offline solving, which doesn't appear to support unsat cores at all.

crux/src/Crux/Goal.hs
```diff
@@ -129,24 +135,121 @@ proveToGoal _ allAsmps p pr =
...
+          let sym = ctx ^. ctxSymInterface
+          assumptions <- WI.andAllOf sym (L.folded . id) (fmap (^. labeledPred) (mconcat assumptionsInScope))
+          goal <- notPred sym (p ^. labeledPred)
+          -- NOTE: We don't currently provide a method for capturing the output
+          -- sent to offline solvers.  We would probably want a file per goal.
+          let logData = WS.defaultLogData
+          mres <- withTimeout $ WS.solver_adapter_check_sat adapter sym logData [assumptions, goal] $ \satRes ->
+            case satRes of
+              Unsat _ -> do
+                atomicModifyIORef' goalNum (\(x, u, s) -> ((x, u, s+1), ()))
```
robdockins: I think you're incrementing the wrong counter here, and below

crux/src/Crux/Goal.hs
travitch: Thanks

crux/src/Crux/Goal.hs
travitch: Whoops, thanks.  I changed it to use a record with named fields to make this mistake harder to repeat

travitch:
That would be pretty easy, but seems like it would probably be a breaking API change - where do you think is the right place to slot that in?

robdockins:
Well, I admit, I haven't thought through the details.  But I'd think just return the `ProcessedGoals` record along with the `Goals` structure.

what4/src/What4/Utils/Process.hs
```diff
@@ -75,6 +75,6 @@ withProcessHandles path args mcwd action = do
               (\(_ :: SomeException) -> return ())
   let onError (_,_,_,ph) = do
         -- Interrupt process; suppress any exceptions that occur.
-        catch (interruptProcessGroupOf ph) (\(_ :: SomeException) -> return ())
+        catch (terminateProcess ph) (\(_ :: SomeException) -> return ())
```
atomb: What's the rationale for this change? The documentation suggests that it might be an improvement, but potentially not on Windows. Have you done any experiments there?

crux/src/Crux.hs
```diff
@@ -233,83 +290,188 @@ data ProgramCompleteness
...
+        case mgt of
+          Nothing -> return ()
+          Just gt -> modifyIORef glsRef (Seq.|> gt)
+  when (simVerbose cruxOpts > 1) $ do
+    say "Crux" "Simulation complete."
+  when (isProfiling cruxOpts) $ writeProf profInfo
+  (,) <$> readIORef compRef <*> readIORef glsRef
+
+isProfiling :: CruxOptions -> Bool
+isProfiling cruxOpts = profileCrucibleFunctions cruxOpts || profileSolver cruxOpts
```
atomb: All of the simulator refactoring above is great, and just makes things cleaner overall.

crux/src/Crux.hs
```diff
@@ -117,53 +130,97 @@ showVersion l = outputLn ("crux version: " ++ Crux.version ++ ", " ++
...
+    unsatCores | yicesMCSat cruxOpts = NoUnsatFeatures
+               | otherwise           = ProduceUnsatCores
+
+    withOnlineBackendFM floatRepr =
+      case selectedSolver of
+        CCS.Yices -> withYicesOnlineBackend floatRepr nonceGen unsatCores $ \sym -> do
+          symCfg sym yicesEnableMCSat (yicesMCSat cruxOpts)
+          case goalTimeout cruxOpts of
+            Just s -> symCfg sym yicesGoalTimeout (floor s)
+            Nothing -> return ()
```
atomb: I wish we had a way to reduce some of the duplication related to setting timeouts here. It makes me wonder whether a more global setting, like `solverInteractionFile`, would be appropriate. It could have a fixed set of units (seconds?) and solvers would internally have to calculate their own timeouts based on that.

what4/src/What4/Utils/Process.hs
travitch: This is catching a missed case from #400, which made this change more widely.  Briefly, `interruptProcessGroupOf` is basically Ctrl+C, while `terminateProcess` is `kill -15` (soft terminate, not `kill -9`). I made the change because some solvers have special (and incorrect) handling of Ctrl+C that makes it not terminate the process.  The one that bit me was that yices tries to be fancy with Ctrl+C handling, but actually doesn't support Ctrl+C interrupts in MC-SAT mode.

crux/src/Crux.hs
travitch: Oh my yes, that would be nice.

PR #430, opened 2020-02-03
What4 exprtests

kquick:
Adds testing framework for What4 expressions.  Uses Hedgehog (a quickcheck-style tester) Generator to create What4 expressions to test.  At this point not all expression types can be generated, but this gives good coverage for SymPred, SymNat, and SymBV.  There are a couple of concerns revealed, but those should be handled by a separate effort.

### DIFF
--- a/what4/test/GenWhat4Expr.hs
+++ b/what4/test/GenWhat4Expr.hs
@@ -875,7 +875,20 @@ bvExprs bvTerm conTE projTE teSubCon expr width toWord =
   if width <= 16
   then
     [
-      -- Something is wrong with these shifty instructions.
+      -- Currently, shifts must be handled very carefully.  The
+      -- underlying datatype for a BV is Integer, which is not
+      -- Bounded, and the shift amount is the same BV size as the
+      -- shifted source.  Shifting a 64-bit value left by something
+      -- between 32 and 64-bits requires a very large Haskell
+      -- allocation for storing those huge integers.
+      --
+      -- Properly, the shift source should return 0 (or -1) if the
+      -- shift amount is greater than the current BV size, but this is
+      -- not currently implemented.  For the time being, the
+      -- constraint above limits these tests to be run for shift sizes
+      -- that are supportable within a couple of GB of system memory.
+      --
+      -- Observed results:
       --
       --   8bit & 16bit BV : test time < 8 sec, tests pass
       --   32-bit BV : memory utilization goes up to ~4-5GB
--- a/what4/test/GenWhat4Expr.hs
+++ b/what4/test/GenWhat4Expr.hs
@@ -1,0 +1,933 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE PolyKinds #-}
+{-# LANGUAGE RankNTypes #-}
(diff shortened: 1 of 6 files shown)


what4/test/GenWhat4Expr.hs
```diff
@@ -0,0 +1,933 @@
...
+           (fromIntegral $ countTrailingZeros $ toWord $ uBV $ mask $ testval x)
+           (\sym -> bvCountTrailingZeros sym =<< expr x sym))
+
+  -- TBD: carrylessMultiply
+
+  ] ++
+  if width <= 16
+  then
+    [
+      -- Something is wrong with these shifty instructions.
```
robdockins: This is probably due to the `shiftL` and `shiftR` instructions creating _huge_ integer values when the shift amount gets large (for shiftL) or large and negative (for `shiftR`).

kquick: Good catch, I had forgotten we were using unconstrained `Integer` as the base type.  Some quick testing in GHCI reveals that this is the case for `shiftL`.  This doesn't happen for `shiftR`, however, tests still get heap overflow for `bvLshr`.

Do we want to add some special handling to our `bvAshr`, `bvLshr`, and `bvShl` to protect against these larger shift values, or do we just want to say "don't do that"?  For machine-based modelling, I can see some reasonable constraints being applied, but perhaps not for general crucible use (e.g. Matlab, Java, etc.), so maybe we can't apply constraints, but on the other hand, the "don't do that" will potentially result in some hidden costs if the shift amounts are ever big, but not big enough to fail (as in the 32-bit tests).   If we just say "don't do that" then I'll just ensure we have some smallish limits here in the tests; if we want to have barriers, I'll leave these as special cases until we get the barriers.  What do you think?

robdockins: Looks like we currently don't have any special-case handling for large shift amounts, and we definitely should.  For any shifts larger than word-width, the semantics should just be to return the 0 value.  The rotate operations already to range-reduction via computing remainders, so that's why they don't cause problems.

I don't know exactly what's going on with `bvLshr`, but it's probably similar.

For LLVM, I'm pretty sure it's UB to pass a shift amount larger than the word width, so that would filter out large constants; maybe for Java too.  That might explain why we haven't seen this occur in the wild yet.

kquick:
I'd hesitate to make this part of the library itself, primarily because this is a testing thing and adding `Tasty` and `Hedgehog` dependencies for runtime code doesn't seem right.  The typical way to do something like this seems to be to create another package, something like "what4-test" that provides this library.  This can be awkward, particularly because this means that the actual what4 testing isn't performed in the what4 package build itself, but if there is significant downstream use of this anticipated, this is probably the direction to take.  We can add this split later if it becomes needed, or we can do it now.  What do you think of this approach?

Adding variables can be done at the generated level, but I'm not yet sure how to handle that in the parallel Haskell expression level.  The current testing takes advantage of the Concrete values to get immediately verifiable results, so adding variables would probably be a parameter or a different set of non-recursive generators supplying IsSymExprBuilder vars.  I'd like to undertake this as a separate effort if that's OK with you.

robdockins:
Yeah, that makes sense.  Let's leave it as-is for now, but let's try to keep any potentially-reusable parts in separate modules in case we want to do that switch later.  Seems to me like it currently is that way already.

As to variables, yeah, don't hold this up, I was just curious what your thinking about how similar or different that would be to what's already here.

what4/test/GenWhat4Expr.hs
robdockins: Per our offline discussion, let's deal with oversized shifts in a separate pass.

PR #437, opened 2020-02-06
[WIP] term annotations

robdockins:
Fairly major refactoring to replace the `AssertionTree` and related data structures with inline term annotations.  This hopefully will fix performance problems, and also allow more sophisticated proof condition explanations.

This will have relatively significant client package impacts, so I'm spinning up this PR before its fully done to have CI assist with figuring out where updates need to be made.


robdockins:
I'm thinking this isn't the correct approach anymore.

PR #440, opened 2020-02-12
let Crux extensions use the online solver backend

spernsteiner:
This branch provides Crux extensions with the ability to use the online solver backend, if one exists.  The extension's `SimulateCallback` can now include a new `MaybeOnlineSolver sym` constraint and use it to obtain a `Maybe (SomeOnlineSolver sym)` value - either `Just` a proof that `sym` is an online backend, or `Nothing`.  After matching on the proof, the extension can call functions like `withSolverProcess` that require an online backend.

The Rust frontend needs this functionality to support [custom printing of counterexamples](https://github.com/GaloisInc/mir-verifier/pull/10).

### DIFF
--- a/crux/crux.cabal
+++ b/crux/crux.cabal
@@ -58,6 +58,7 @@ library
     Crux.Config.Common,
     Crux.Config.Solver,
     Crux.Extension,
+    Crux.Online,
     Crux.Version
 
   other-modules:
--- a/crux/src/Crux.hs
+++ b/crux/src/Crux.hs
@@ -66,6 +66,7 @@ import Crux.Config.Common
 import Crux.Config.Doc
 import Crux.Extension
 import qualified Crux.Version as Crux
+import Crux.Online
 
 -- | Throws 'ConfigError' (in addition to whatever the crucible and the
 -- callbacks may throw)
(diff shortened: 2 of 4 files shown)


robdockins:
This seems OK to me.  One thing I don't understand though: why add the `MaybeOnlineClass` constraint instead of simply adding a `Maybe SomeOnlineSolver` argument to the callback type? Is that just to avoid having to touch downstream clients?

spernsteiner:
>why add the `MaybeOnlineClass` constraint instead of simply adding a `Maybe SomeOnlineSolver` argument to the callback type? Is that just to avoid having to touch downstream clients?

Yes, compatibility.  It seemed unfortunate to force a change to every extension, when most don't care whether they have an online solver and would just ignore the new argument.

robdockins:
On thinking about this a bit, I think we're better off with just adding another parameter.  And, actually, it's probably better for me just just fold a `Maybe SomeOnlineSolver` parameter into the new callback structure from #446, rather than merge this, because that refactor significantly alters the structure of this part of the code.

spernsteiner:
> And, actually, it's probably better for me just just fold a `Maybe SomeOnlineSolver` parameter into the new callback structure from #446

That works for me.  It looks like #446 will solve some other problems for crux-mir as well.  (Specifically, making it easier to run several independent test cases with independent success/failure/counterexample reports in a single invocation of crux-mir.)

robdockins:
Alright, that's a plan.  Let me know if there are other changes that would support `crux-mir` also.

PR #441, opened 2020-02-13
[WIP] Revamp LLVM error messages using annotations

robdockins:
A second attempt at the refactoring from #437.


robdockins:
Rebasing this got a bit messy, so see #453 instead

PR #446, opened 2020-02-17
Crux refactor

robdockins:
Refactor crux to remove a confusing inversion-of-control situation.

Individual crux tools are now more directly in control of the main execution path, with `crux` itself instead providing its functionality as large-grain combinators.  This reduces unintended coupling between the tools and is generally easier to understand.


crucible-jvm/tool/Main.hs
```diff
@@ -110,40 +111,34 @@ defaultOptions =
...
+cruxJVMConfig = Crux.Config
+  { Crux.cfgFile = pure defaultOptions
+  , Crux.cfgEnv =
+      [ Crux.EnvVar "JDK_JAR"
+        "Path to .jar file containing the JDK"
+        $ \p opts -> Right $ opts { jarList = p : jarList opts }
+      ]
+  , Crux.cfgCmdLineFlag =
+      [ Crux.Option ['c'] ["classpath"]
+        "TODO"
```
atomb: Heh. We should probably document these properly at some point. I'll create an issue.

PR #449, opened 2020-02-19
Allow bvZext, bvSext, and bvTrunc to be no-ops if inp size == out size

kquick:


what4/src/What4/Expr/Builder.hs
```diff
@@ -590,12 +590,12 @@ data App (e :: BaseType -> Type) (tp :: BaseType) where
         -> !(e (BaseBVType w))   -- rotate amount
         -> App e (BaseBVType w)
 
-  BVZext :: (1 <= w, w+1 <= r, 1 <= r)
```
robdockins: I'd prefer we not change the syntax for BVZext and BVSext; that is, still require that they be non-trivial extensions.  Since we are doing a test for the noop case in `bvSext` and `bvZext` operations anyway, the syntax should only be applied in non-trivial cases anyway.

In other code (e.g., SMTWriter) we rely on the extension amount being nonzero, so it's better to preserve that invariant, I think.

robdockins:
@kquick, I'm in the middle of splitting what4 out into a separate repository, so let's not merge this PR just yet.  Go ahead and finish what you want to do with it and let me know when you're done.  I'll manually manage splitting the patches.

robdockins:
Closing for now.  Feel free to reopen when you're ready to look at this again @kquick

PR #451, opened 2020-02-19
What4 process tweaks

robdockins:
Fiddle with the process handling code in what4 to more reliably clean up when processes are killed.

Because GHC signal/process/thread handling is a pretty complicated beast, this doesn't completely guarantee good behavior, but seems to do a better job in common cases.

### DIFF
--- a/crux-llvm/exe/Main.hs
+++ b/crux-llvm/exe/Main.hs
@@ -7,7 +7,9 @@ import System.Posix.Process
 
 import CruxLLVMMain (mainWithOutputConfig, defaultOutputConfig)
 
--- Convert a SIGTERM signal into a SIGINT signal for the entire process group
+-- Rebroadcast SIGTERM to the entire process group.  The CatchOnce
+-- handler keeps this from handling and retransmitting SIGTERM to
+-- itself over and over.
 installSIGTERMHandler :: IO ()
 installSIGTERMHandler =
   do gid <- getProcessGroupID
--- a/crux-llvm/exe/Main.hs
+++ b/crux-llvm/exe/Main.hs
@@ -13,7 +13,7 @@ installSIGTERMHandler =
   do gid <- getProcessGroupID
      void $ installHandler sigTERM (CatchOnce (handler gid)) Nothing
  where
- handler gid = signalProcessGroup sigINT gid
+ handler gid = signalProcessGroup sigTERM gid
 
 
 main :: IO ()
--- a/crux-llvm/exe/Main.hs
+++ b/crux-llvm/exe/Main.hs
@@ -1,3 +1,23 @@
 module Main (main) where
 
-import CruxLLVMMain (main)
+import Control.Monad
+import System.Exit
+import System.Posix.Signals
+import System.Posix.Process
+
+import CruxLLVMMain (mainWithOutputConfig, defaultOutputConfig)
+
+-- Convert a SIGTERM signal into a SIGINT signal for the entire process group
+installSIGTERMHandler :: IO ()
+installSIGTERMHandler =
+  do gid <- getProcessGroupID
+     void $ installHandler sigTERM (CatchOnce (handler gid)) Nothing
+ where
+ handler gid = signalProcessGroup sigINT gid
+
+
+main :: IO ()
+main =
+  do installSIGTERMHandler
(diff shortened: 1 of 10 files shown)


crux-llvm/exe/Main.hs
```diff
@@ -1,3 +1,23 @@
...
 
-import CruxLLVMMain (main)
+import Control.Monad
+import System.Exit
+import System.Posix.Signals
+import System.Posix.Process
+
+import CruxLLVMMain (mainWithOutputConfig, defaultOutputConfig)
+
+-- Convert a SIGTERM signal into a SIGINT signal for the entire process group
```
travitch: Question here: why do we need to convert the SIGTERM into a SIGINT? It seems like the goal is to just broadcast the signal we got to the rest of the process group (which we have to do manually because the GHC RTS/process API have special handling for SIGINT).  If we just rebroadcast the SIGTERM, wouldn't that accomplish basically the same goal?

what4/src/What4/Utils/Process.hs
```diff
@@ -59,7 +60,8 @@ withProcessHandles path args mcwd action = do
           { std_in  = CreatePipe
           , std_out = CreatePipe
           , std_err = CreatePipe
-          , create_group = True
+          , create_group = False
+          , delegate_ctlc = True
           , cwd = mcwd
```
travitch: I just noticed that these options are duplicated in the SMTLib2 module - would it be worth consolidating them to make sure they get updated together since this seems to be fragile? I'm not sure that the extra effort is really worth it, just a thought.

what4/src/What4/Utils/Process.hs
```diff
@@ -68,13 +70,25 @@ withProcessHandles path args mcwd action = do
...
         -- Interrupt process; suppress any exceptions that occur.
-        catch (terminateProcess ph) (\(_ :: SomeException) -> return ())
+        catchJust filterAsync (terminateProcess ph) (\(ex :: SomeException) ->
+          hPutStrLn stderr $ displayException ex)
 
   bracket startProcess cleanup (\hs -> onException (action hs) (onError hs))
+
+
+-- | Filtering function for use with `catchJust` or `tryJust`
+--   that filteres out asynch exceptions so they are rethrown
```
travitch: filteres -> filters

what4/src/What4/Utils/Process.hs
```diff
@@ -68,13 +70,25 @@ withProcessHandles path args mcwd action = do
...
-                  void $ waitForProcess ph)
-              (\(_ :: SomeException) -> return ())
+                  hClose err_h)
+              (\(ex :: SomeException) -> hPutStrLn stderr $ displayException ex)
+        void $ waitForProcess ph
+
   let onError (_,_,_,ph) = do
         -- Interrupt process; suppress any exceptions that occur.
-        catch (terminateProcess ph) (\(_ :: SomeException) -> return ())
+        catchJust filterAsync (terminateProcess ph) (\(ex :: SomeException) ->
```
travitch: Just so I understand here: we are propagating Async exceptions from the call to `terminateProcess`, but just swallowing failures.  Is this code trying to make sure that we don't accidentally swallow a ctrl+c?

crux-llvm/exe/Main.hs
robdockins: I'll try that and see what happens.  The reason I converted to SIGINT was because I wanted to activate GHC's special handling (and the delegate_ctlc behavior), but maybe it's not necessary?

what4/src/What4/Utils/Process.hs
robdockins: Yeah, the overall goal is to make sure we don't accidentally swallow ctrl+c or other deliberate thread-killing async exceptions, like the ones I believe the `async` package uses for killing threads.

what4/src/What4/Utils/Process.hs
robdockins: Yup, and also in the `Yices` module, where I forgot to update them. That suggests it is worth consolidating.

crux-llvm/exe/Main.hs
robdockins: Turns out, rebroadcasting SIGTERM also does the trick, and I think it's a better approach.

PR #452, opened 2020-02-20
SV-COMP mode, take 2

robdockins:


robdockins:
Yeah, the crux-llvm changes are pretty much just splitting up modules.

Yeah, `setResourceLimit` is pretty easy to use; it's in the `unix` package, though, so it isn't totally portable.

robdockins:
OK, there are still things I want to do do improve the SV-COMP mode, but I think it's better to get what we have now merged than to wait.

Issue #425, opened 2020-01-29
Replace AssertionTree with predicate annotations


robdockins:
The AssertionTree datastructure does not have explicit sharing, and is probably the reason for the slowdown described in #407.

Rather than rework this datastructure to preserve sharing, I think it makes more sense to allow what4 predicates to be labeled with annotations.  These annotations can later be traversed to build the kind of user-visible explanations that AssertionTree is currently used for.  This will both simplify the API and, hopefully, alleviate a performance problem.

robdockins:
CF #441.

robdockins:
Annotations are now used instead of assertion trees.  Next we need to implement #462 to actually use the embedeed annotations.

PR #453, opened 2020-02-25
[WIP] LLVM term annotations

robdockins:
Another, more cleaned-up take on addressing #425


robdockins:
I'm not convinced about 0622138 at all; `panic` should only be used for situations where the situation should never happen, regardless of user input.  That really doesn't seem to be the case here.

andreistefanescu:
`muxLLVMVal` is used on two `LLVMVal`s that are obtained by reading from the memory the same `MemType` on the two branches of a `WriteMerge`. I don't think that `panic` can happen.

robdockins:
Ah, I see.  Yeah, tracing it back, it looks like all call sites come from places where it is plausible that the values should be structurally compatible.  I'll add a comment to that effect.

PR #456, opened 2020-02-28
Support conditional memory write operations.

andreistefanescu:
Required by GaloisInc/saw-script#652.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -160,6 +160,7 @@ module Lang.Crucible.LLVM.MemModel
   , G.possibleAllocs
   , G.ppSomeAlloc
   , doConditionalWriteOperation
+  , mergeWriteOperations
 
     -- * PtrWidth (re-exports)
   , HasPtrWidth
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1090,6 +1092,41 @@ storeRaw sym mem ptr valType alignment val = do
...
+  => sym
+  -> MemImpl sym
+  -> Pred sym
+  -> (MemImpl sym -> IO (MemImpl sym))
+  -> (MemImpl sym -> IO (MemImpl sym))
+  -> IO (MemImpl sym)
+mergeWriteOperations sym mem cond true_write_op false_write_op = do
+  frame_id <- pushAssumptionFrame sym
+  loc <- getCurrentProgramLoc sym
+  addAssumption sym $ LabeledPred cond $
```
robdockins: I don't think you want to add this assumption here.  They way it's scoped at the moment, it will apply on both the if and the else branches.  I think you should just leave it off altogether, actually; and I don't think the assumption frame is necessary either.  If you do want to do this, you'll need a separate frame for each side.

However, because this is basically an atomic operation, I don't think we need all this assumption frame management at all.  If we instead had a generic if/then/else operation we needed in SAWScript, we ought to do that in the `OverrideSim` monad, and use the full symbolic simulator infrastructure to handle the necessary branching and potential failure situations.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1090,6 +1092,41 @@ storeRaw sym mem ptr valType alignment val = do
...
+  :: (IsSymInterface sym, HasPtrWidth wptr)
+  => sym
+  -> MemImpl sym
+  -> Pred sym
+  -> (MemImpl sym -> IO (MemImpl sym))
+  -> IO (MemImpl sym)
+doConditionalWriteOperation sym mem cond write_op =
+  mergeWriteOperations sym mem cond write_op return
+
+mergeWriteOperations
```
robdockins: Let's add some haddock comments here.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1090,6 +1092,41 @@ storeRaw sym mem ptr valType alignment val = do
     assert sym p2 (err $ ptrMessage errMsg2 ptr valType)
     return mem{ memImplHeap = heap' }
 
+doConditionalWriteOperation
```
robdockins: Haddocks, please

andreistefanescu: done

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
andreistefanescu: done

PR #460, opened 2020-03-26
Add assertThenAssume option.

### LINKED ISSUE #423: Change "assert" behavior to not assume proof goals
Currently, the common method of asserting proof obligations and side-conditions in crucible first records a proof obligation and then assumes the asserted fact into the current path condition.

This can sometimes be helpful, as later proof obligations may be easier to prove under the assumption that previous verifications succeed.  However, in many cases the conditions are not useful, and simply cause the later problems to grow. This can especially be a problem if the proof obligations are large.  In some observed cases, verification time is dominated simply by streaming the problem to the solvers.

We should perhaps change this behavior, or make it configurable.  The tradeoffs are not immediately clear.  See #323 for some additional discussion.

andreistefanescu:
Resolves #423.

### DIFF
--- a/crucible/src/Lang/Crucible/Backend.hs
+++ b/crucible/src/Lang/Crucible/Backend.hs
@@ -54,8 +54,8 @@ module Lang.Crucible.Backend
   , assertIsInteger
   , readPartExpr
   , ppProofObligation
+  , backendOptions
   , assertThenAssumeConfigOption
-  , assertThenAssumeOption
   ) where
 
 import           Control.Exception(Exception(..), throwIO)
@@ -258,6 +258,9 @@ assertThenAssumeOption = mkOpt
   boolOptSty
   (Just (PP.text "Assume a predicate after asserting it."))
   (Just (ConcreteBool False))
+
+backendOptions :: [ConfigDesc]
+backendOptions = [assertThenAssumeOption]
 
 -- | Add a proof obligation for the given predicate, and then assume it
 -- (when the assertThenAssume option is true).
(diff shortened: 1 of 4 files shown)


crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -244,26 +249,43 @@ class IsBoolSolver sym where
...
 addAssertion ::
   (IsExprBuilder sym, IsBoolSolver sym) =>
   sym -> Assertion sym -> IO ()
 addAssertion sym a@(AS.LabeledPred p msg) =
   do addProofObligation sym a
-     addAssumption sym (AS.LabeledPred p (AssumingNoError msg))
+     assert_then_assume_opt <- getOpt
+       =<< getOptionSetting assertThenAssumeConfigOption (getConfiguration sym)
+     when assert_then_assume_opt $
+       addAssumption sym (AS.LabeledPred p (AssumingNoError msg))
```
atomb: We could probably benefit from sharing this code between `addAssertion` and `addDurableAssertion`.

robdockins:
Nitpick: I'd like to see something like

```
backendOptions :: [ConfigDesc]
backendOptions = [assertThenAssumeOption]
```

added to the `Backend` module.  Then we can refer to that in downstream modules.  That lets us easily add other options later if necessary.

crucible/src/Lang/Crucible/Backend.hs
andreistefanescu: 👍

Issue #461, opened 2020-03-31
Unify crux command line/config file handling


pnwamk:
## Overview
It would be nice to have a single, unified way to specify how `crux` (and `crux`-derived tools) can be customized, from which we can automatically generate:
 + command line arg/option parsing
 + configuration file parsing
 + command line help/docs
 + other docs (e.g., Tex)

## Motivation
The current approach of specifying/parsing/etc each method for customizing `crux`'s behavior leads to a lot of duplication in code/documentation and can lead to subtle inconsistencies. Presumably having a unified approach would make it easier for users/customers and would make it easier to generate consistent documentation for.

## Current State of Affairs
Currently it appears we're separately specifying/handling the following to generate a [Crux.Config](https://github.com/GaloisInc/crucible/blob/8b58dfe3d335096885a71be11cd5cc91e7345fff/crux/src/Crux/Config.hs#L33):

+ command line args/flags via [simple-get-opt](https://hackage.haskell.org/packages/search?terms=simple-get-opt) (which handles command-line documentation and parsing of the command line args),
+ configuration files via [config-schema](https://hackage.haskell.org/packages/search?terms=config-schema) (which handles configuration file documentation and parsing of the command line args), and
+ environment variables via a few [home-grown data/functions](https://github.com/GaloisInc/crucible/blob/8b58dfe3d335096885a71be11cd5cc91e7345fff/crux/src/Crux/Config.hs#L45) which are aware of how we're currently handling the command line interface

## Proposal
I'm interested in implementing this improvement if we can get some alignment on what it should look like and would love any comments/suggestions/etc to that end.

At the moment I'm imagining throwing together a `data`-type which specifies generically how each "crux option" can be thought about (i.e., it's type, valid values predicate (if needed), description, command line flags (if any), config file entry (if any), etc) from which it is easy to delegate to the aforementioned tools to handle the specifics of command line/config file/environment parsing/handling/etc.

## Risk
The only thing I can think of at the moment is this could easily lead to some consolidation of documentation and/or option naming/handling which could introduce some small backwards incompatibilities if we're not careful.

robdockins:
See also #457

pnwamk:
It's probably worth noting there _are_ some packages that do this already... none jumped out to me as ideal candidates in my perusing, e.g.:

[...]

Issue #457, opened 2020-02-29
Unify configuration specification


travitch:
Currently, command line options and config file options are specified separately but with substantial overlap.  It would be really useful to have a single specification produce both sets of options uniformly. There are a few improvements that this would enable/benefits it would provide:

- Less chance of inconsistency between the two specs
- Automatic printing of default values (currently, the description strings have them hard coded, which makes it difficult to change defaults on a per-frontend basis)
- We could extract a user manual section (in e.g., .tex format) describing each option in detail, letting us keep user manuals up-to-date with the tool

@robdockins suggested that we could also use this improved framework to expose simulation-time variables to frontends (e.g., https://github.com/GaloisInc/what4/blob/6189c160b9477fb2bfeb07566d99881eb1c9a202/what4/src/What4/Solver/Yices.hs#L1013).

robdockins:
Yeah there's a pretty sophisticated system for describing and updating configuration options internal to what4.  It's mostly used for configuration options for individual solvers, but is also used for, e.g., the setting describing if path-satisfiability should be on (which means it can be twiddled at runtime)

https://github.com/GaloisInc/what4/blob/master/what4/src/What4/Config.hs

langston-barrett:
Closing in favor of #461

PR #466, opened 2020-04-08
WIP: unify crux config option specification

pnwamk:
# Summary
This WIP PR is an attempt to unify how we think about and specify the options to `crux`-derived tools. (See https://github.com/GaloisInc/crucible/issues/461 and https://github.com/GaloisInc/crucible/issues/457)

It takes a slightly "command line centric" view, i.e. that for every command line option which can affect the behavior of a `crux`-tool:
+ there is a corresponding configuration file entry, 
+ there is _possibly_ an associated environment variable,
+ there is a single description of the kinds of values one can provide for that option
+ there should be a short description of the option and often a longer documentation-style description for what the option does suitable for inclusion in a PDF or similar.

I tried to keep the design close to how things generally work now. E.g., we essentially already assume there exists a default/initial configuration which is then updated via the provided options (see the [comment here](https://github.com/GaloisInc/crucible/blob/8eb3988fc43b7310de1f89e3a50f3416b3288b6c/crux/src/Crux/Config.hs#L27), how command line flags/options work, etc), so that became the shared model for handling all three of the ways of updating the config options.

# Specifics

Here are the key components currently:

+ a [CmdLineOptions](https://github.com/pnwamk/crucible/blob/2bff3413d444442229012c79a38895074d1b0811/crux/src/Crux/Config.hs#L152) is the top-level configuration, specifying how free-form/positional command line args are treated and containing a list of all of the `CmdLineOpt`s which include all flags/option specifications;

+ a [CmdLineOpt](https://github.com/pnwamk/crucible/blob/2bff3413d444442229012c79a38895074d1b0811/crux/src/Crux/Config.hs#L177) specifies how a particular command line flag/option is handled, along with which config file section and environment variable (if any) correspond to that setting as well;

+ a [ValSpec](https://github.com/pnwamk/crucible/blob/2bff3413d444442229012c79a38895074d1b0811/crux/src/Crux/Config.hs#L263) is a generic/unified description of what kind of values can be given for an option. These are then "compiled" into the appropriate parsers/etc for the command line, config file, and environment variable handling.

+ the code in [Crux.Config](https://github.com/pnwamk/crucible/blob/2bff3413d444442229012c79a38895074d1b0811/crux/src/Crux/Config.hs#L314) below those data structures is concerned with compiling the unified representation into the representations understood by the command line parser, config file parser, and environment variable parser;

+ in [Crux.Config.Common](https://github.com/pnwamk/crucible/blob/2bff3413d444442229012c79a38895074d1b0811/crux/src/Crux/Config/Common.hs#L353) there are a few examples of specifying options in this unified way.

# Next Steps?

- [ ] Decide if this is a reasonable approach and if we want to refine and adopt it.
   + If we want to refine and adopt it:
      - [ ] specify the remaining common options in the new (possibly refined) format
      - [ ] write some tests to find bugs or any problems with current assumptions/design
  + If this is not a reasonable approach... what do people think would be better?

Misc: Apologies on any code style/quality issues. This was a fun learning experience for me in the Haskell space ;-)

### DIFF
--- a/crux/src/Crux/Config.hs
+++ b/crux/src/Crux/Config.hs
@@ -1,4 +1,11 @@
-{-# Language MultiWayIf, OverloadedStrings #-}
+{-# LANGUAGE MultiWayIf #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE ApplicativeDo #-}
+{-# LANGUAGE ExistentialQuantification #-}
 -- | This module deals with loading configurations.
 module Crux.Config
   ( -- * Writing configurations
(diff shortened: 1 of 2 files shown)


robdockins:
@pnwamk, I'm sorry this fell off my radar for so long.  Are you still working on this? I like this idea generally, and I'm planning to look into the details soon.

travitch:
I can help finish it off (it has been on my list to get back to it)

pnwamk:
I am not actively working on it at the moment, no. I've been thinking on and off I should re-visit it and see what can/should be done so it's not lost. I vaguely remember trying to make the interface more "combinator like" per a request from @travitch and still having some work to do tidying things, but it's been a while.

langston-barrett:
Since @pnwamk left Galois a while back and @travitch is leaving soon, and this has merge conflicts, I'm closing it for now. Feel free to re-open.

langston-barrett:
Since @pnwamk left Galois a while back and this has merge conflicts, I'm closing for now. Feel free to re-open.

PR #469, opened 2020-04-10
Fix the semantics of `strlen` on symbolic strings.

### LINKED ISSUE #468: strlen
I'm really not sure what is wrong here. Is something wrong with `saw`'s notion of `strlen`? Or, am I doing something wrong? Any help would be appreciated.

strlen_test.c
```
#include <string.h>

size_t strlen_test (char *string) {
  size_t nLength = strlen(string);
  return nLength;
}
```

strlen_test.saw
```
let alloc_init ty v = do {
  p <- crucible_alloc ty;
  crucible_points_to p v;
  return p;
};

let ptr_to_fresh n ty = do {
  x <- crucible_fresh_var n ty;
  p <- alloc_init ty (crucible_term x);
  return (x, p);
};

let strlen_test_spec = do {
  (string, pstring) <- ptr_to_fresh "string" (llvm_array 10 (llvm_int 8));

  crucible_precond {{ string!0 == 0 }}; //NULL terminated string

  crucible_execute_func [pstring];
};

strlen_test_bc <- llvm_load_module "strlen_test.bc";

strlen_test_result <- crucible_llvm_verify strlen_test_bc "strlen_test" [] false strlen_test_spec z3;
```

```
$ clang --version
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin19.4.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
$ clang -emit-llvm strlen_test.c -c -o strlen_test.bc -g
$ saw --version
0.4.0.99 (d87d0d73)
$ saw strlen_test.saw
[13:45:56.549] Loading file "strlen_test.saw"
[13:45:56.576] Verifying strlen_test ...
[13:45:56.578] Simulating strlen_test ...
[13:45:56.579] "crucible_llvm_verify" (strlen_test.saw:23:23-23:43):
Symbolic execution failed.
  Error during memory load
  in strlen_test at strlen_test.c:4:20
  Details:
  No previous write to this location was found
    No matching writes to this memory were found.
Stack frame
  Allocations:
    StackAlloc 6 0x8:[64] Mutable 8-byte-aligned "strlen_test.c:4:10"
    StackAlloc 5 0x8:[64] Mutable 8-byte-aligned "strlen_test.c:3:27"
  Writes:
    Indexed chunk:
      5 |->
        *(5, 0x0:[64]) := (4, 0x0:[64])
Base memory
  Allocations:
    HeapAlloc 4 0xa:[64] Mutable 1-byte-aligned strlen_test.saw:2:8
    GlobalAlloc 3 0x0:[64] Immutable 1-byte-aligned strlen_test
    GlobalAlloc 2 0x0:[64] Immutable 1-byte-aligned strlen
    GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned llvm.dbg.declare
  Writes:
    Indexed chunk:
      4 |->
        *(4, 0x0:[64]) := [c@6:bv
                          ,c@7:bv
                          ,c@8:bv
                          ,c@9:bv
                          ,c@10:bv
                          ,c@11:bv
                          ,c@12:bv
                          ,c@13:bv
                          ,c@14:bv
                          ,c@15:bv]
```

robdockins:
Previously, `strlen` would recurse down a string until it
ran into a concrete zero.  This is fine, except that it would
assert that each load along the way would succeed, without taking
into account the previous zero tests that had to fail.  To correct
this we need to emulate a path condition such that to load
at position `n` each of the `n-1` previous locations must have
been non-zero, and the loads are only required to succeed under
that condition.  This allows one, for example, to assume that
some value in a string is 0 (but not specify which one).  Previously,
`strlen` would index off the end of the allocation of such a string
and fail.  Now it will succeed because the path condition required
to index out-of-bounds is inconsistent.

Fixes #468

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -215,6 +215,7 @@ import           Lang.Crucible.Simulator.SimError
 
 import           Lang.Crucible.LLVM.DataLayout
 import           Lang.Crucible.LLVM.Extension
+import           Lang.Crucible.LLVM.Extension.Safety
 import           Lang.Crucible.LLVM.Bytes
 import           Lang.Crucible.LLVM.MemType
 import           Lang.Crucible.LLVM.MemModel.Type
@@ -227,6 +228,7 @@ import qualified Lang.Crucible.LLVM.Extension.Safety.UndefinedBehavior as UB
 import           Lang.Crucible.LLVM.Translation.Constant
 import           Lang.Crucible.LLVM.Types
 import           Lang.Crucible.Panic (panic)
+
 
 
 import           GHC.Stack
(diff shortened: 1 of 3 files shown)


weaversa:
> This allows one, for example, to assume that some value in a string is 0 (but not specify which one).

Oh I like that!

I hate to ask, but is `strlen` the only function that has this kind of problem?

robdockins:
> I hate to ask, but is `strlen` the only function that has this kind of problem?

I think so.  `memcpy`, `memmove` and `memset` are all memory model primitives, so they don't have this issue.  All the other functions that deal with strings require the bytes to be fully concrete

If we want to support more of "string.h" we'd arguably be better off providing a library of naive C implementations to avoid running into this issue again.

PR #473, opened 2020-04-27
Add support for LLVM 9/10 `objectsize`

atomb:
This version adds another boolean parameter. We currently ignore it.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
@@ -175,8 +175,11 @@ declare_overrides =
   , basic_llvm_override LLVM.llvmObjectsizeOverride_32
   , basic_llvm_override LLVM.llvmObjectsizeOverride_64
 
-  , basic_llvm_override LLVM.llvmObjectsizeOverride_32'
-  , basic_llvm_override LLVM.llvmObjectsizeOverride_64'
+  , basic_llvm_override LLVM.llvmObjectsizeOverride_32_null
+  , basic_llvm_override LLVM.llvmObjectsizeOverride_64_null
+
+  , basic_llvm_override LLVM.llvmObjectsizeOverride_32_null_dynamic
+  , basic_llvm_override LLVM.llvmObjectsizeOverride_64_null_dynamic
 
   , basic_llvm_override LLVM.llvmStacksave
   , basic_llvm_override LLVM.llvmStackrestore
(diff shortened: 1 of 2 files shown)


robdockins:
Pretty straightforward, as far as it goes. However, the new flag changes the semantics of the LLVM intrinsic in a way that's hard for me to understand.

Previously, we assumed that this intrinsic appearing in the code stream meant the compiler was not able to fold away it's value (maybe we should spend some time evaluating this hypothesis, actually), which meant that we could always return the unknown value at runtime.  Now, however, the dynamic version of the intrinsic might still be able to compute answers sometimes.

In pathological cases, I worry that we could mis-verify programs that use this information in a way other than intended (as a safety check).  I guess we could nondeterministicly return either the correct answer or the "unknown" answer when the dynamic flag is set.  I'm not sure how much it is worthwhile to be paranoid about this sort of thing.

atomb:
Yeah, I wonder about this one as well as the other one we added a while back and are also ignoring. The code I've seen generated sets it to `false`, which I think corresponds to what we're doing now (and what LLVM used to expect). We could always require it to be `false` (and maybe do the same with the previous argument).

robdockins:
Oh, that `null` flag also changes the semantics somewhat; hunh... I guess we should probably get that one right.

I dunno, what do you think about the idea of just always returning a nondeterministic choice between unknown and the correct value (even in cases where we used to always return unknown)?  That seems like a more accurate rendering of the intended semantics here.

atomb:
Wouldn't that make it hard to prove anything that depended on getting the right value? My understanding is that when both arguments are `false` we get the sort of straightforward behavior we know how to handle. If we just required that for now, then we could think a little more about how to handle the `true` case for both arguments.

robdockins:
My understanding is that this intrinsic is pretty much exclusively used to do best-effort bounds checks.  We already are doing precise bounds checks, so don't think we'd encounter anything we couldn't already handle, although we might do some redundant proofs.

For now, I'm happy with requiring those arguments to be false. If we encounter examples with those arguments set in the wild, we can decide what to do then.

atomb:
I added a check requiring the `dynamic` argument to be `false`, but it turns out the `nullunknown` parameter is `true` in the case I was looking at. It might be that we should open that as a separate issue, since it's about pre-existing code, not the code added in this PR.

robdockins:
Yeah, let's open a separate issue for that.  This looks good to me, as far as the `dynamic` flag goes.

atomb:
New issue opened: #474.

PR #489, opened 2020-05-28
Bv sized

robdockins:
Incorporate the bv-sized changes in what4 into crucible.

This seems like it's getting close to merge-ready.


cabal.project
```diff
@@ -18,12 +18,12 @@ packages:
 optional-packages:
   dependencies/abcBridge/
   dependencies/aig/
+  dependencies/bv-sized
```
robdockins, line 21: Should remove this if we're depending on hackage bv-sized

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
```diff
@@ -626,101 +623,97 @@ evalIarith' ::
...
-         when (sx == minSigned w && sy == -1)
+         when (x == BV.minSigned w && y == BV.mkBV w (-1))
               (throwError "Signed division overflow in constant arithmetic operation")
-         let r = sx `rem` sy
+         let r = BV.srem w x y
          return r
 
     _ -> throwError "Floating point operation applied to integer arguments"
 
+-- BGS: Leave this alone for now, as we don't have a good way to
```
robdockins, line 686: Is this comment still accurate? Do we still need to do something here?

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
```diff
@@ -626,101 +623,97 @@ evalIarith' ::
...
+  BV.BV w -> BV.BV w -> m LLVMConst
 evalBitwise op w x y = IntConst w <$>
-  do let ux = toUnsigned w x
-     let uy = toUnsigned w y
-     let sx = toSigned w x
-     case op of
-       L.And -> return (ux .&. uy)
-       L.Or  -> return (ux .|. uy)
-       L.Xor -> return (ux `xor` uy)
+  let yshf = fromInteger (BV.asUnsigned y)
```
robdockins, line 695: Probably shouldn't be so cavalier about `fromInteger` here

crucible/crucible.cabal
```diff
@@ -30,6 +30,7 @@ library
     base >= 4.8 && < 4.15,
     ansi-wl-pprint,
     bimap,
+    bv-sized >= 1.0.0,
```
robdockins, line 33: Probably should put an upper bound here.

robdockins:
A few pretty minor comments.  Otherwise, I'm happy with where this is.

PR #495, opened 2020-06-09
[crux] Concretize the type of the symbolic backend in the SimulatorCa…

travitch:
…llback

This allows backends to learn the concrete type of the symbolic backend.  While it is nice to keep this abstract in as much code as possible, some things (e.g., traversing symbolic terms) are only really possible with this type concretized.

I'm writing an execution feature that needs to traverse symbolic terms, so I unfortunately need this extra constraint.  At least - I haven't figured out a better way - suggestions welcome.


atomb:
This seems pretty reasonable to me. I don't expect we're likely to change the backend we use.

PR #499, opened 2020-06-11
Expr builder reorg

robdockins:
Major reorganization of the ExprBuilder


robdockins:
This won't be ready for a little while, closing for now.

PR #501, opened 2020-06-16
Generalize the personality parameter in crux

travitch:
Crux previously had the personality field fixed to the crux `Model` type.  This change relaxes that to be a type that contains a `Model` so that clients of crux can use their own types to store extra information. It doesn't have many changes in core crux, but did have a bunch of type signature updates in the LLVM overrides.

I wanted to float this for discussion - I am pretty sure I can hack around it in my application, but this seemed like a reasonable approach.


robdockins:
In general, I'm in favor of this style of parameterization. The main downside seems to be the additional complexity that library users have to grapple with both in terms of pure cognitive complexity and gnarly type signatures.

I think we should combat this with "user guides" that explain the dimensions of customizability that users need to grapple with and a collection of reasonable default choices to cases where they don't need the provided flexibility.

travitch:
That is a good point - I'll at least add some documentation about the parameters

travitch:
I'll include the migration instructions in the squashed commit.  I have another crux tool that I'll update separately.  I'm taking a look at mir-verifier.  Are there any other objections to merging this?

PR #505, opened 2020-07-03
Cryptol RecordMap type update

kquick:
From Cryptol PR#786


kquick:
Build failures are 8.4, which is not supported by cryptol's new libBF.  The travis config needs to be updated for that, but I'll do that separately.

PR #506, opened 2020-07-04
Add reverse cache in SAWCoreBackend.

andreistefanescu:
Use it `bindSAWTerm` to return the same What4 variable
when binding the same SAWCore term multiple times.


andreistefanescu:
@atomb this results in more accurate What4 during the symbolic execution. For example, without this change
```crucible_execute_func [(crucible term x), (crucible_term x)]```
would results in the creation of two different What4 variables. This can be a problem for the symbolic size allocations, when the SAWCore term representing the symbolic size is passed both to `crucible_symbolic_alloc` and to `crucible_array_prefix_points_to`, but ends up translated to two different What4 variables.

atomb:
Ah, right. That could definitely address some issues we've seen with symbolic termination.

PR #507, opened 2020-07-04
Generalize asMemAllocationArrayStore to symbolic size allocations.

andreistefanescu:
This change enables `asMemAllocationArrayStore` to work on symbolic size allocations and writes in the simple case when the allocation size and the write size are determined equal by `bvEq`. This is sufficient to handle the combination of `crucible_symbolic_alloc` (GaloisInc/saw-script#760) and `crucible_array_prefix_points_to` (GaloisInc/saw-script#762) when the size of the allocation and the size of the points-to are the same SAWCore term. Deciding at What4 level that the array write covers the entire allocation prevents a possible ite explosion when reading. Supporting array writes that partially cover the allocation is more complicated, and may be addressed in the future as needed.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
@@ -55,6 +55,7 @@ module Lang.Crucible.LLVM.MemModel.Generic
   , branchMem
   , branchAbortMem
   , mergeMem
+  , asMemAllocationArrayStore
 
   , SomeAlloc(..)
   , possibleAllocs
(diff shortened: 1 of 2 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1678,11 +1678,12 @@ asMemAllocationArrayStore sym w ptr mem
...
               | Just write_blk_no <- asNat (llvmPointerBlock write_ptr)
               , blk_no == write_blk_no
-              , MemArrayStore arr (Just arr_store_sz_bv) <- write_source
-              , Just arr_store_sz <- BV.asUnsigned <$> asBV arr_store_sz_bv
-              , sz == arr_store_sz
-              , Just Refl <- testEquality w (ptrWidth write_ptr) ->
-                return $ Just arr
+              , MemArrayStore arr (Just arr_store_sz) <- write_source
+              , Just Refl <- testEquality w (ptrWidth write_ptr) -> do
+                sz_eq <- bvEq sym sz arr_store_sz
```
robdockins, line 1683: Do we need to generalize this to a partial value?  Checking that the equality is concretely true is a bit limiting... I guess it's still sound if we just return `Nothing`, because instead we will fall back on the regular memory model operations, I think.

It would really help if you would add some description to the PR about why you're doing this generalization and what the expected results will be.

andreistefanescu: it's sounds, but potentially (very) incomplete. There are going to be some incomplete interactions between the symbolic size allocations/writes and the concrete size allocations/writes, but that's unavoidable

PR #521, opened 2020-07-28
An an option to CruxOptions for supplying additional problemFeatures

abakst:
This PR adds an option to supply a set of `problemFeatures` to be `OR`'d in with the default `problemFeatures` when configuring online solvers. This allows us to force the array theory to be enabled, for example.

- `crux/src/Crux/Config/Common.hs` modified to extend `CruxOptions` and set `problemFeatures` to the empty set of features by default
- `crucible/src/Lang/Crucible/Backend/Online.hs` modified to `OR` in the additional features
- `crux/src/Crux.hs` modified to pass in the additional features to the `with...OnlineBackend` functions

### DIFF
--- a/crux/src/Crux.hs
+++ b/crux/src/Crux.hs
@@ -244,12 +244,12 @@ withSelectedOnlineBackend cruxOpts nonceGen selectedSolver maybeExplicitFloatMod
             Just s -> symCfg sym yicesGoalTimeout (floor s)
             Nothing -> return ()
           k floatRepr sym
-        CCS.CVC4 -> withCVC4OnlineBackend floatRepr nonceGen ProduceUnsatCores extraFeatures $ \sym -> do
+        CCS.CVC4 -> withCVC4OnlineBackend floatRepr nonceGen unsatCoreFeat extraFeatures $ \sym -> do
           case goalTimeout cruxOpts of
             Just s -> symCfg sym cvc4Timeout (floor (s * 1000))
             Nothing -> return ()
           k floatRepr sym
-        CCS.Z3 -> withZ3OnlineBackend floatRepr nonceGen ProduceUnsatCores extraFeatures $ \sym -> do
+        CCS.Z3 -> withZ3OnlineBackend floatRepr nonceGen unsatCoreFeat extraFeatures $ \sym -> do
           case goalTimeout cruxOpts of
             Just s -> symCfg sym z3Timeout (floor (s * 1000))
             Nothing -> return ()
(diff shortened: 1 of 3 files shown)


crux/src/Crux.hs
```diff
@@ -234,20 +234,22 @@ withSelectedOnlineBackend cruxOpts nonceGen selectedSolver maybeExplicitFloatMod
...
       case selectedSolver of
-        CCS.Yices -> withYicesOnlineBackend floatRepr nonceGen unsatCoreFeat $ \sym -> do
+        CCS.Yices -> withYicesOnlineBackend floatRepr nonceGen unsatCoreFeat extraFeatures $ \sym -> do
           symCfg sym yicesEnableMCSat (yicesMCSat cruxOpts)
           case goalTimeout cruxOpts of
             Just s -> symCfg sym yicesGoalTimeout (floor s)
             Nothing -> return ()
           k floatRepr sym
-        CCS.CVC4 -> withCVC4OnlineBackend floatRepr nonceGen unsatCoreFeat $ \sym -> do
+        CCS.CVC4 -> withCVC4OnlineBackend floatRepr nonceGen ProduceUnsatCores extraFeatures $ \sym -> do
```
robdockins, line 247: It's true that the way `unsatCoreFeat` is written is Yices-specific, but we should still honor the `unsatCores` option, and only ask the solver to produce unsat corses if that option is set.

crux/src/Crux/Config/Common.hs
```diff
@@ -233,6 +237,8 @@ cruxOptions = Config
             section "proof-goals-fail-fast" yesOrNoSpec False
             "If true, stop attempting to prove goals as soon as one of them is disproved"
 
+          onlineProblemFeatures <- pure noFeatures
```
robdockins, line 240: Is the intention here to add a way to configure this, or just to allow downstream tools to twiddle this setting?

crux/src/Crux.hs
abakst: Yikes, good catch! I misread the diff post-merge!

crux/src/Crux/Config/Common.hs
abakst: The latter -- I'm not sure there would be a meaningful way to expose this feature except in clients of the library.

robdockins: I could imagine individual options for things like "enable nonlinear arithmetic during path-sat checking"... I'm not sure.  I guess this is fine, although it feels a little weird not to have it hooked up to any user settable options.

PR #532, opened 2020-08-10
crux: emit JSON version of the report

Ptival:
For some purposes (e.g. VSCode extension), having the report be a
Javascript file registering a global variable is awful.

This commit makes the script also emit a JSON copy.  Due to CORS, the
redundancy seems unavoidable: one cannot load the JSON from the
Javascript file without using a server, which I assume we are trying to
avoid.


robdockins:
I've included this in #531

PR #533, opened 2020-08-10
crux: add meta charset

Ptival:
Browser was unhappy.


robdockins:
I've included this patch in #531

PR #535, opened 2020-08-11
Try many solvers

abakst:
This is a proposal for throwing multiple solvers at a problem in offline mode.

The changes are:

- Allow the user to specify a _list_ of solvers (or the string "all") to try for each goal
- For each goal, `forkIO` a thread for each solver. If a thread returns a definite result, then use that as the result for the goal. Otherwise, each thread has either timed out or thrown an exception, in which case we might want to report those exceptions, for example (this is what the PR does currently).

At this point I'm looking for comments/suggestions.

### DIFF
--- a/crux/src/Crux/Config/Solver.hs
+++ b/crux/src/Crux/Config/Solver.hs
@@ -63,9 +63,6 @@ data SolverConfig = SingleOnlineSolver SolverOnline
                   -- ^ Use one online solver connection for path satisfiability
                   -- checking and a separate online solver connection for goal
                   -- discharge
-                  | OnlyOfflineSolver SolverOffline
-                  -- ^ Use only an offline solver with no support for path
-                  -- satisfiability checking
                   | OnlyOfflineSolvers [SolverOffline]
                   -- ^ Try any of a number of offline solvers with no support
                   -- for path satisfiability checking
@@ -202,7 +199,7 @@ parseSolverConfig cruxOpts = validatedToEither $
       -- discharged with an offline solver.
       OnlineSolverWithOfflineGoals <$> asOnlineSolver onSolver <*> asAnyOfflineSolver (CCC.solver cruxOpts)
   where
-    tryAnyOffline = OnlyOfflineSolver <$> asAnyOfflineSolver (CCC.solver cruxOpts)
-    tryOnlyOffline = OnlyOfflineSolver <$> asOnlyOfflineSolver (CCC.solver cruxOpts)
+    tryAnyOffline = OnlyOfflineSolvers . pure <$> asAnyOfflineSolver (CCC.solver cruxOpts)
+    tryOnlyOffline = OnlyOfflineSolvers . pure <$> asOnlyOfflineSolver (CCC.solver cruxOpts)
     trySingleOnline = SingleOnlineSolver <$> asOnlineSolver (CCC.solver cruxOpts)
     tryManyOffline = OnlyOfflineSolvers <$> asManyOfflineSolvers (CCC.solver cruxOpts)
--- a/crux/src/Crux/Config/Solver.hs
+++ b/crux/src/Crux/Config/Solver.hs
@@ -13,6 +13,7 @@ module Crux.Config.Solver (
   ) where
 
 import           Control.Applicative ( (<|>), empty, Alternative )
+import           Data.List.Split (wordsBy)
 import qualified Data.Foldable as F
 import qualified Data.Traversable as T
 import qualified Lang.Crucible.Backend.Simple as CBS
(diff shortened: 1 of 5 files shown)


robdockins:
You might take a look at how I did portfolio solving in cryptol:
https://github.com/GaloisInc/cryptol/blob/3bec6cdbb4e01701c467a76931dbb0cb00d558e9/src/Cryptol/Symbolic/What4.hs#L448

I think using the `async` library for this simplifies things a lot.

abakst:
Yes that looks much simpler

crux/src/Crux/Goal.hs
```diff
@@ -237,33 +239,66 @@ proveGoalsOffline adapter opts ctx explainFailure (Just gs0) = do
...
+          mapM_ kill as'
+          return $ Right r
+        Right r@(NotProved _ (Just _)) -> do
+          mapM_ kill as'
+          return $ Right r
+        Right _ ->
+          await as' es
+
+    -- `cancel` from async blocks until the canceled thread has terminated.
+    kill a = throwTo (asyncThreadId a) AsyncCancelled
```
robdockins, line 301: Don't use an async exception here.  This generally bypasses exception handling code; but we want exception-handling to send signals to shutdown external solvers and have them shutdown.  `AsyncCancelled` will leave the solvers running and eating CPU time.  I tend to use `ExitSuccess`.

crux/src/Crux/Goal.hs
```diff
@@ -237,33 +239,66 @@ proveGoalsOffline adapter opts ctx explainFailure (Just gs0) = do
...
+                    Sat (evalFn, _) -> do
+                      let model = ctx ^. cruciblePersonality . personalityModel
+                      vals <- evalModel evalFn model
+                      explain <- explainFailure evalFn p
+                      return (NotProved explain (Just (ModelView vals)))
+                    Unknown -> do
+                      return (NotProved mempty Nothing)
+                case mres of
+                  Just res -> return res
+                  Nothing -> return (NotProved mempty Nothing)
```
robdockins, line 260: The nothing case corresponds to timeout, right? Probably worth mentioning that in the explanation.

crux/src/Crux/Goal.hs
```diff
@@ -237,33 +239,66 @@ proveGoalsOffline adapter opts ctx explainFailure (Just gs0) = do
...
-                return (Prove (p, gt))
-              Unknown -> do
-                end
-                modifyIORef' goalNum (updateProcessedGoals p (NotProved mempty Nothing))
-                return (Prove (p, NotProved mempty Nothing))
-          case mres of
-            Just res -> return res
-            Nothing -> return (Prove (p, NotProved mempty Nothing))
 
+          let runOneSolver adapter = do
```
robdockins, line 243: I'm not a big fan of complicated nestings of `where` and `let`.  If we can do it without too much difficulty, I'd like to see this lifted to the same `where` block as `go` (which could really use a better name).

crux/src/Crux/Config/Solver.hs
```diff
@@ -65,6 +66,9 @@ data SolverConfig = SingleOnlineSolver SolverOnline
                   | OnlyOfflineSolver SolverOffline
                   -- ^ Use only an offline solver with no support for path
                   -- satisfiability checking
+                  | OnlyOfflineSolvers [SolverOffline]
```
robdockins, line 69: Does `OnlyOfflineSolver` need to be a separate case?

crux/src/Crux/Goal.hs
abakst: Do you happen to know if the timeout corresponds to wall clock or actual CPU time? That is, is a per-thread timeout the same behavior as a a timeout on the `waitAnyCatch`? Not terribly important, but I think `waitAnyCatch` ends up catching the `Timeout` exception, which means that if `withTimeout` returns, it only returns a `Just` value.

crux/src/Crux/Config/Solver.hs
abakst: Not necessarily, I'm ambivalent about whether or not this should be a separate case or to just use pattern matching. I can remove the constructor.

crux/src/Crux/Goal.hs
robdockins: I think it's wall-clock time, but I'm not sure.  Given the way this is written, I'd expect the `withTimeout` combinator to handle the `Timeout` exception before it bubbles out, but I don't know offhand where that combinator comes from or how it works, exactly.

robdockins:
I'm worried this is starting to fall behind master.  Is there still work to do before merging this?

abakst:
I'm not planning on making any more changes, so if it looks good I think we can merge it.

robdockins (approved):
I'm happy with this as long as CI gives the thumbs up

PR #536, opened 2020-08-12
Crux tweaks

robdockins:
Some additional tweaks to crux, mostly to do with error message reporting and some finer points of the HTML report UI

### DIFF
--- a/crux/src/Crux.hs
+++ b/crux/src/Crux.hs
@@ -104,7 +104,7 @@ newtype SimulatorCallback
         forall sym t st fs. (IsSymInterface sym, Logs, sym ~ WEB.ExprBuilder t st fs) =>
           sym ->
           Maybe (SomeOnlineSolver sym) ->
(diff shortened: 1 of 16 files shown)


crux/src/Crux.hs
```diff
@@ -605,8 +605,7 @@ printFailedGoals opts (CruxSimulationResult _cmpl allGls)
     = sayFail "Crux" (show $ vcat [ "Found counterexample for verification goal", ex ])
 
     | otherwise
-    = sayFail "Crux" (show $ vcat [ "Failed to prove verification goal", ex, ppSimError err, text msg])
-
+    = sayFail "Crux" (show $ vcat [ "Failed to prove verification goal", ex ])
```
atomb, line 608: Is this change because the extra information above isn't really user-comprehensible?

robdockins: No, the change is because we now pass "unknown" goals into the same explanation callback as "disproved" goals.  This gives clients more control over what gets printed.

atomb: Oh, got it. That makes sense.

PR #537, opened 2020-08-12
Add GitHub Action for crux-llvm

atomb:
This builds `crux-llvm` on Linux, macOS, and Windows. It also builds `crux-llvm-svcomp` on the former two, but omits Windows due to its `unix` dependency.


robdockins (approved):
I don't feel qualified to review the content of the github actions changes, but the results seem OK.

atomb:
Yeah, this was more to give you a heads-up that this existed than anything else.

PR #541, opened 2020-09-15
Add Go frontend to Crucible

atomb:
Adds a `crucible-go` package that translates Go into Crucible, with the help of the [`goblin`](https://github.com/galoisinc/goblin) tool. This package includes a `crux-go` executable that makes basic Crux functionality available for Go.

Some Go language constructs are still unsupported, and some Crux features have not yet been implemented, but this does not change other packages in the repository in significant ways, so merging will help ensure that it stays up-to-date with the latest Crucible changes.

The one significant change it _does_ make to Crucible is the addition of a taint analysis based on abstract interpretation. This requires a few small changes to the general abstract interpretation framework. It also adds a few utility functions to the CFG generator infrastructure.

Apologies for the huge PR! I don't expect anyone to carefully review the added stuff. I'm hoping @travitch can check that the abstract interpretation changes are reasonable, though.


robdockins:
Neat!

crucible-go/src/Lang/Crucible/Go/Overrides.hs
```diff
@@ -0,0 +1,169 @@
...
+
+mkFresh :: IsSymInterface sym
+        => String
+        -> BaseTypeRepr ty
+        -> Crux.OverM p sym ext (RegValue sym (BaseToType ty))
+mkFresh nm ty =
+  do sym  <- C.getSymInterface
+     name <- case W4.userSymbol nm of
+               Left err -> fail (show err)
+               Right a  -> return a
```
travitch, line 71: Is it worth using `safeSymbol` here instead? It doesn't seem that important if the name is mangled slightly

crucible-go/src/Lang/Crucible/Go/Overrides.hs
```diff
@@ -0,0 +1,169 @@
...
+  if pkgName == pkg && fName == overrideName override then
+    Just o
+  else
+    lookupOverride pkgName fName overrides
+
+lookupOverride' :: Maybe (Text, FunctionName)
+                -> [SomeOverride p sym ext]
+                -> Maybe (SomeOverride p sym ext)
+lookupOverride' nm overrides =
+  nm >>= flip (uncurry lookupOverride) overrides
```
travitch, line 169: Is it worth a note to make these into a `Map` or `HashMap` one day?

crucible-go/src/Lang/Crucible/Go/Simulate.hs
```diff
@@ -0,0 +1,179 @@
...
+mkFunctionBindings overrides ((ident, AnyCFG cfg) : cfgs) =
+  let f = case lookupOverride' ident overrides of
+        Just (SomeOverride _pkg argsRepr retRepr override) ->
+          case (testEquality (cfgArgTypes cfg) argsRepr,
+                testEquality (cfgReturnType cfg) retRepr) of
+            (Just Refl, Just Refl) -> UseOverride override
+            _ -> error $ "mkFunctionBindings: override " ++ show ident ++
+                 " type mismatch. expected " ++ show (cfgArgTypes cfg) ++
+                 " -> " ++ show (cfgReturnType cfg) ++ ", got " ++
+                 show argsRepr ++ " -> " ++ show retRepr
```
travitch, line 101: Maybe now isn't the time, but we have been trying to move to `panic` instead of `error`

crucible-go/src/Lang/Crucible/Go/Types.hs
```diff
@@ -0,0 +1,326 @@
...
+import           Lang.Crucible.FunctionHandle
+import           Lang.Crucible.Simulator.ExecutionTree
+import           Lang.Crucible.Simulator.Intrinsics
+import           Lang.Crucible.Syntax
+import           Lang.Crucible.Types
+
+----------------------------------------------------------------------
+-- | Go extension tag.
+
+newtype Go = Go ()
```
travitch, line 53: Oh, odd - would `data Go = Go` make more sense here?  Or even just an empty data decl?

crucible-go/src/Lang/Crucible/Go/Types.hs
```diff
@@ -0,0 +1,326 @@
...
+
+-- | Translator state.
+data TransState =
+  TransState
+  { machineWordWidth :: PosNat -- ^ Bit width of unqualified "int" and
+                               -- "uint" types
+  , retRepr :: Some TypeRepr -- ^ Return type of current function
+  , namespaces :: HashMap Text Namespace
+  , pkgName :: Text -- ^ Name of package curently being translated
+  , sng :: Maybe (Some (NonceGenerator IO))
```
travitch, line 139: Just a note to myself - I'm surprised to see this `NonceGenerator` have its state thread type parameter quantified away.  I wonder how it is used such that this is okay.  It seems like any use wouldn't really be possible unless one didn't actually need a nonce.

crucible-go/src/Lang/Crucible/Go/Types.hs
```diff
@@ -0,0 +1,326 @@
...
+type TranslateM' a = StateT TransState (ExceptT String IO) a
+
+-- | A newtype around TranslateM' indexed by NodeTypes so it can be
+-- used as a functor in recursion constructions (see
+-- Lang.Crucible.Go.Rec).
+newtype TranslateM (tp :: NodeType) =
+  TranslateM { runTranslateM :: TranslateM' (Translated tp) }
+
+liftIO' :: IO a -> TranslateM' a
+liftIO' = lift . lift
```
travitch, line 247: Why not just inline the definition of `TranslateM'` and then derive `MonadIO`?

crucible/src/Lang/Crucible/Analysis/Taint.hs
```diff
@@ -0,0 +1,158 @@
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE ExistentialQuantification #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE TupleSections #-}
+
+module Lang.Crucible.Analysis.Taint (cfgTaintAnalysis
+                                    ,taintDomain
+                                    ,taintInterp
+                                    ,Tainted(..)) where
```
travitch, line 10: I don't think this is actually used anywhere.  It is neat, but maybe we should just not include it since there are no clients?

crucible/crucible.cabal
```diff
@@ -67,6 +67,7 @@ library
     Lang.Crucible.Analysis.Fixpoint.Components
     Lang.Crucible.Analysis.Postdom
     Lang.Crucible.Analysis.Reachable
+    Lang.Crucible.Analysis.Taint
```
travitch, line 70: See the comment in the module - I don't think this is actually used anywhere.

crucible/src/Lang/Crucible/CFG/Generator.hs
```diff
@@ -360,6 +365,20 @@ mkAtom :: (Monad m, IsSyntaxExtension ext) => Expr ext s tp -> Generator ext s t
...
+                    AtomExpr <$> freshAtom (NewRef a)
+
+readReference :: (Monad m, IsSyntaxExtension ext) => Expr ext s (ReferenceType tp) -> Generator ext s t ret m (Expr ext s tp)
+readReference ref = do a <- mkAtom ref
+                       AtomExpr <$> freshAtom (ReadRef a)
+
+writeReference :: (Monad m, IsSyntaxExtension ext) => Expr ext s (ReferenceType tp) -> Expr ext s tp -> Generator ext s t ret m ()
+writeReference ref val = do aref <- mkAtom ref
+                            aval <- mkAtom val
+                            undefined -- Generator $ addStmt $ WriteRef aref aval
```
travitch, line 380: I'm a bit worried about leaving this undefined if we are commiting it.  I don't think it is used anywhere at the moment, so we could just delete it.

crucible-go/src/Lang/Crucible/Go/TransUtil.hs
```diff
@@ -0,0 +1,756 @@
...
+asType' repr e k = k $ asType repr e
+
+-- | asTypeEither lifted to assignments.
+asTypesEither :: CtxRepr ctx'
+              -> Assignment (Gen.Expr Go s) ctx
+              -> Either (Some TypeRepr, Some (Gen.Expr Go s))
+              (Assignment (Gen.Expr Go s) ctx')
+asTypesEither Empty Empty = Right Empty
+asTypesEither (reprs :> repr) (es :> e) =
+  pure (:>) <*> asTypesEither reprs es <*> asTypeEither repr e
```
travitch, line 98: I think that is the same as `(:>) <$> asTypesEither reprs es <*> asTypeEither repr e`

crucible-go/src/Lang/Crucible/Go/TransUtil.hs
```diff
@@ -0,0 +1,756 @@
...
+  _repr -> b
+
+tryAsPointer :: Gen.Expr Go s tp
+             -> (forall a. TypeRepr a -> Gen.Expr Go s (PointerType a) -> b)
+             -> b
+             -> b
+tryAsPointer e k b = case exprType e of    
+  MaybeRepr (VariantRepr
+             (Ctx.Empty :> ReferenceRepr repr :> StructRepr
+               (Ctx.Empty :> ReferenceRepr (VectorRepr repr') :> NatRepr))) ->
```
travitch, line 155: Later tech debt note: I think there is a `PointerRepr` defined in the Encodings module that might make this cleaner.  Or perhaps that pattern synonym could be improved to make this easier (e.g., like the LLVMPointerRepr)

crucible-go/src/Lang/Crucible/Go/Types.hs
travitch: Ah, I see - it is passed to the Crucible Generator runner, which uses an arbitrary nonce generator and quantifies away the `s`.  We could just get rid of this and pass the `globalIONonceGenerator` in the Translation module

travitch (approved):
I made a few small suggestions.  It basically looks fine to me.  The one thing I want to call out is that we could just remove the taint analysis module.

crucible-go/src/Lang/Crucible/Go/Overrides.hs
atomb: Good idea.

crucible-go/src/Lang/Crucible/Go/Overrides.hs
atomb: Yeah, good suggestion.

crucible-go/src/Lang/Crucible/Go/Types.hs
atomb: Yep, I think so.

crucible/src/Lang/Crucible/CFG/Generator.hs
atomb: Oh, and it's one of three functions added that are all unnecessary. This one is the same as `writeRef`. I removed them all.

crucible-go/src/Lang/Crucible/Go/TransUtil.hs
atomb: Fixed.

crucible/src/Lang/Crucible/Analysis/Taint.hs
atomb: It's also unfinished, now that I look more closely. I'm inclined to leave the file there, so we can finish it at some point, but to remove it from the exported modules list (as in your comment below).

crucible/crucible.cabal
atomb: Removed.

crucible-go/src/Lang/Crucible/Go/TransUtil.hs
atomb: Comment added.

crucible-go/src/Lang/Crucible/Go/Types.hs
atomb: It seems that `TranslateM'` is used elsewhere, too. Maybe that could be cleaned up, but it would be a bunch of changes.

crucible-go/src/Lang/Crucible/Go/Types.hs
atomb: Fixed.

crucible-go/src/Lang/Crucible/Go/Simulate.hs
atomb: Done.

PR #542, opened 2020-09-16
Add `crux-mir` for analysis of Rust MIR code

atomb:
This merges the contents (and history) of the [mir-verifier](https://github.com/galoisinc/mir-verifier)  repo into the `crux-mir` subdirectory, with a few small changes to update that code to work with the latest version of Crucible.


crux-mir/.travis.yml
```diff
@@ -0,0 +1,100 @@
+dist: xenial
```
robdockins, line 1: Probably should remove this `.travis.yml`

crux-mir/.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "deps/crucible"]
```
robdockins, line 1: probably remove this too

crux-mir/src/Mir/Language.hs
```diff
@@ -205,11 +208,13 @@ runTests (cruxOpts, mirOpts) = do
             setSimulatorVerbosity (Crux.simVerbose cruxOpts) sym
             let simCtx = C.initSimContext sym mirIntrinsicTypes halloc outH
                     C.emptyHandleMap mirExtImpl Crux.emptyModel
-            return $ Crux.RunnableState $
-                C.InitialState simCtx C.emptyGlobals C.defaultAbortHandler C.UnitRepr $
-                    C.runOverrideSim C.UnitRepr $ simTest symOnline fnName
+            return (Crux.RunnableState $
+                    C.InitialState simCtx C.emptyGlobals C.defaultAbortHandler C.UnitRepr $
+                     C.runOverrideSim C.UnitRepr $ simTest symOnline fnName
+                   , undefined
```
spernsteiner, line 214: A number of symbolic tests are failing on this `undefined`.  One example:
```
cd crux-mir
./translate_libs.sh
cabal new-run -- crux-mir test/symb_eval/num/checked_add.rs
```

crux-mir/.travis.yml
spernsteiner: Is there an easy way to keep the crux-mir CI running after the merge?  I'd rather not disable it if we can avoid it

robdockins: I don't see why not; it should just move into the top level travis config, or get ported to github actions.

atomb: I'll look at moving/porting it. I want to set up something for producing binaries people can download for the release, and have already set up an Action for that for `crux-llvm`.

crux-mir/.gitmodules
atomb: Yup. Good catch.

crux-mir/src/Mir/Language.hs
atomb: Oof. I didn't mean to leave that in there! It was a just a temporary "how many other compile errors are there?" workaround that I forgot I'd made.

crux-mir/.travis.yml
atomb: I think I'm going to do this as part of a separate PR, to avoid a messy merge. So this PR is now actually ready for review.

crux-mir/src/Mir/Intrinsics.hs
```diff
@@ -818,7 +818,7 @@ writeMirVectorWithSymIndex sym _ (MirVector_Array a) i val = do
 data MIR
 type instance ExprExtension MIR = EmptyExprExtension
 type instance StmtExtension MIR = MirStmt
-type instance AssertionClassifier MIR = NoAssertionClassifier
+--type instance AssertionClassifier MIR = NoAssertionClassifier
```
robdockins, line 821: can just delete this, I think

atomb: Good catch. Done.

robdockins (approved):
Glad to see this getting merged in.

atomb:
Me too!

PR #543, opened 2020-09-17
Generalize crux-llvm workflows to crux-mir

atomb:


.github/package-crux-mir.sh
```diff
@@ -0,0 +1,26 @@
...
+
+rm -rf $PKG
+
+mkdir $PKG
+mkdir $PKG/bin
+mkdir $PKG/doc
+
+cabal v2-build exe:crux-mir
+cp $EXE $PKG/bin
+cp `which cargo-mir-json` $PKG/bin
```
spernsteiner, line 16: `cargo install` has a `--root` option - I think `cargo install --root=$PKG` would install all the binaries into `$PKG/bin`.  That way we get all of them (`cargo-crux-test` is missing from this list), and we don't have to update this if we add/remove some in the future.

.github/package-crux-mir.sh
```diff
@@ -0,0 +1,26 @@
...
+cp `which mir-json` $PKG/bin
+cp `which mir-json-callgraph` $PKG/bin
+cp `which mir-json-dce` $PKG/bin
+cp `which mir-json-rustc-wrapper` $PKG/bin
+cp crux-mir/README.md $PKG/doc
+cp -r crux-mir/rlibs $PKG
+
+tar cf $PKG.tar.gz $PKG
```
spernsteiner, line 24: Should be `tar czf`, to enable gzip compression.  (I was wondering how on earth github managed to fit a 300MB .tar.gz into a 40MB .zip file...)

spernsteiner:
I downloaded the crux-mir Linux artifacts, and they don't work due to having the wrong rpath.  `readelf -d mir-json` shows:
```
Library runpath: [/usr/share/rust/.rustup/toolchains/nightly-2020-03-22-x86_64-unknown-linux-gnu/lib]
```
But of course my rustup install is in `~/.rustup`, not `/usr/share/rust/.rustup`, so `mir-json` can't find `librustc_driver.so` or `libstd.so`.  I think we may need to either ship the rust compiler libraries in the crux-mir package so we can use a `$ORIGIN`-relative rpath, or else ship some wrappers that figure out the right toolchain path and either set `$LD_LIBRARY_PATH` or do something clever with `dlopen`.

.github/package-crux-mir.sh
atomb: Given your comment below, I'm thinking of just leaving out the `mir-json` binaries. Given that users will be Rust programmers, having them run `cargo install` in the `mir-json` repo themselves seems not too onerous.

.github/package-crux-mir.sh
atomb: Good catch! I should fix that for `crux-llvm`, too.

PR #544, opened 2020-09-21
Add testing and more to CI scripts

atomb:


spernsteiner (approved):
The crux-mir parts look good to me.  I tested the Linux artifacts locally (specifically `crux-mir-Linux-8.8.4.zip`) and it seemed to work: after running `export PATH=$PWD/bin:$PATH CRUX_RUST_LIBRARY_PATH=$PWD/rlibs` in the extracted directory, I ran `cargo crux-test` on an example and it worked as expected.

atomb:
Glad to hear it's working on Linux! I just tried it on macOS using the same procedure, and it seems good here, too.

PR #548, opened 2020-09-28
Add crux-test-gen

spernsteiner:
This branch adds a new tool for generating test cases from a grammar.  It's designed to allow exhaustive symbolic testing of an API, but the design is flexible enough that it may be useful in other situations as well.  See the [README][readme] for details of its functionality, and see [example/vec1.txt][vec1] and [example/vec2.txt][vec2] for examples of test generation grammars.

[README]: https://github.com/GaloisInc/crucible/blob/46f93ec2fd835f1d069fd502360947424b71f270/crux-test-gen/README.md
[vec1]: https://github.com/GaloisInc/crucible/blob/46f93ec2fd835f1d069fd502360947424b71f270/crux-test-gen/example/vec1.txt
[vec2]: https://github.com/GaloisInc/crucible/blob/46f93ec2fd835f1d069fd502360947424b71f270/crux-test-gen/example/vec2.txt


atomb (approved):
This seems even more useful than I'd realized, likely way beyond Crux. I haven't carefully reviewed all of the Rust code, but could if you'd like. It seems like you have a good set of tests included, so I'm fairly confident that it's working correctly.

PR #551, opened 2020-10-02
crux-mir: properly create translation.json when --branch-coverage is set

spernsteiner:
This was causing crux-mir coverage tests to fail when run from a clean checkout.


atomb (approved):
Thanks for the quick fix!

atomb:
All the GitHub workflows have completed successfully, so we can probably merge.

Issue #552, opened 2020-10-05
`cabal v2-install` fails with various errors


spernsteiner:
`cabal v2-install` begins by running `cabal v2-sdist` on each package `cabal.project`, regardless of what target is being installed.  Currently, this fails on three packages:

 * [ ] `cabal v2-sdist pkg:crucible-go` fails with `cabal: Main.hs doesn't exist`.  The test suite main file `tests/Main.hs` is not present in the repo.
 * [ ] `cabal v2-sdist pkg:crucible-server` fails with `cabal: Error: Could not find module: Lang.Crucible.Proto with any suffix: ["gc","chs","hsc","x","y","ly","cpphs","hs","lhs","hsig","lhsig"]. If the module is autogenerated it should be added to 'autogen-modules'.`
 * [ ] `cabal v2-sdist pkg:crux-llvm` fails with `cabal: WinMain.hsMain.hs doesn't exist`.  Cabal appears to have concatenated two different conditional `main-is` directives into one.

Since the `sdist` step fails, nothing in the entire `crucible` repo can be installed using `cabal v2-install`.

I'm using `cabal-install version 3.0.0.0, compiled using version 3.0.1.0 of the Cabal library`

atomb:
I would be in support of removing `crucible-server` from `cabal.project`, at least for the moment, which would fix the second one. The `crucible-go` issue should be easily fixable. The `crux-llvm` one, though, is really awkward (and seems clearly like a bug in Cabal).

robdockins:
If we want to EOL crucible-server, I think that's a potentially reasonable step.  However, I think the suggested fix about would probably work for the immediate problem.  `Lang.Crucible.Proto` is a generated module.

atomb:
Ah, yeah, if adding `Lang.Crucible.Proto` to `autogen-modules` will fix it, and it still builds, let's just do that.

atomb:
I'm going to fix the `crux-llvm` part of this, and update the installation instructions to match.

atomb:
See #555

atomb:
Now #555 should entirely resolve this issue.

PR #553, opened 2020-10-05
crux-mir: update installation instructions

spernsteiner:
Updates the README to account for the recent move from a separate `mir-verifier` repo to part of `crucible`, and adds a workaround for #552.

### DIFF
--- a/crux-mir/README.md
+++ b/crux-mir/README.md
@@ -11,26 +11,29 @@ functionality.
 
 ## Preliminaries
 
-You must have the most recent version of the `mir-json` executable in your
-path.  See [the `mir-json` README][mir-json-readme] for installation
-instructions.
-
 `crux-mir` uses several submodules, so make sure they're initialized:
 
     $ git submodule update --init
+
+Next, navigate to the `crucible/dependencies/mir-json` directory and install
+`mir-json` according to the instructions in [the `mir-json`
+README][mir-json-readme].
 
 [mir-json-readme]: https://github.com/GaloisInc/mir-json#readme
 
 
 ## Installation
 
-Use ghc-8.4.4 or ghc-8.6.5.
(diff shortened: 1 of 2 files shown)


crux-mir/README.md
```diff
@@ -11,27 +11,30 @@ functionality.
...
+`mir-json` according to the instructions in [the `mir-json`
+README][mir-json-readme].
+
 [mir-json-readme]: https://github.com/GaloisInc/mir-json#readme
 
 
 ## Installation
 
-Use ghc-8.4.4 or ghc-8.6.5.
+Use ghc-8.4.4 or ghc-8.6.5.  From the `crux-mir` directory, run:
```
atomb, line 27: Oh, I think 8.8.4 will also work. It's in the CI now.

crux-mir/install.project
```diff
@@ -0,0 +1,8 @@
+packages:
+  ../crucible/
+  ../crux/
+  ./
+
+optional-packages:
+  ../dependencies/aig/
+  ../dependencies/what4/what4/
```
atomb, line 8: Are these the only packages needed even to run the test suite?

atomb (approved):
This looks good! The comments I left could potentially suggest changes, but I'm not sure they need to.

crux-mir/install.project
atomb: Running the test suite with this is probably not actually necessary, given what it's used for, so never mind.

crux-mir/README.md
spernsteiner: I added 8.8.4, and also removed 8.4.4, which isn't covered by CI any more

crux-mir/install.project
spernsteiner: Right, though it turns out the test suite does actually work with `cabal v2-test --project-file=install.project`

spernsteiner:
Rebased and removed `install.project`, since #555 makes it unnecessary

PR #554, opened 2020-10-06
Line up Crux version numbers

atomb:
This commit bumps the version numbers for `crux`, `crux-llvm`, and
`crux-mir` with the goal of presenting "Crux" as a unified entity.


spernsteiner (approved):
No objections here.

`mir-json` has a version number too, but I'm not sure it matters, since it's not displayed anywhere.

robdockins (approved):

PR #555, opened 2020-10-06
Update installation instructions

atomb:
Also update Cabal configuration to make these instructions work.


crux-llvm/README.md
```diff
@@ -34,16 +34,21 @@ The `crux-llvm` tool can be built by doing the following:
...
-        ./build-stack.sh
+        cabal v2-build
 
-This will compile `crux-llvm` and supporting libraries and install them
-in the directory `crucible/crux-llvm/bin`.
+This will compile `crux-llvm` and supporting libraries such that they
+can be executed with `cabal v2-run`. To install the binaries in the
+standard Cabal binary path, run the following:
+
+        cabal v2-install exe:crux-llvm --project-file=install.project --overwrite-policy=always
```
spernsteiner, line 43: This branch already fixes the trickiest of the three errors in #552 - if you fixed the other two, then `install.project` wouldn't be needed, right?

atomb: Heh. That's true. Maybe I should just fix the others and remove it.

yav: You may want to recommend a version of GHC/Cabal to install (e.g., 8.8.4 and 3.2).  I wonder if things work with the `2.*` version?

Also, for installing, it might be useful to mention `--installdir` this is where `cabal` will put the sym-link to the executable.

crux-llvm/README.md
```diff
@@ -34,16 +34,21 @@ The `crux-llvm` tool can be built by doing the following:
...
+standard Cabal binary path, run the following:
+
+        cabal v2-install exe:crux-llvm --overwrite-policy=always
 
 # Invocation
 
-In the `crux-llvm` directory, to analyze `file.c`, run
+In the `crux-llvm` directory (either in the repository or the root of
+the directory extracted from a distribution tarball), to analyze
+`file.c`, run
```
yav, line 49: Is there a reason why you have to be in the repository?  Also, you may want to mention that `cruc-llvm` should be in the path.

yav (approved):

crux-llvm/README.md
atomb: Unfortunately there is, for the moment. The `c-src` directory needs to be a subdirectory of your current directory. We really should fix that, but I don't think it'll happen today.

crux-llvm/install.project
```diff
@@ -0,0 +1,11 @@
+packages:
```
spernsteiner, line 1: This `install.project` file is no longer used

spernsteiner (approved):
Once this is merged, I'll remove the `install.project` workaround from #553 and then merge that

crux-llvm/install.project
atomb: Fixed.

PR #561, opened 2020-10-06
Fix an indexing bug in the What4 -> SAWCore translation

robdockins:
The What4 `testBit` operation counts bits by their positional value
(i.e., little endian) wheresas the SAWCore `scAt` operation indexes
bits from the most significant bit (big endian).


brianhuffman (approved):
Quite right. Nice catch.

PR #570, opened 2020-10-20
Path sat fix

robdockins:
This PR fixes path satisfiability handling for SAW, and installs a new configuration option to the Online backend for enabling and disabling communication with the solver.  The new configuration option prevents the system from wasting time communicating with the solver if no queries will be issued.

### DIFF
--- a/crux/src/Crux/Goal.hs
+++ b/crux/src/Crux/Goal.hs
@@ -24,7 +24,8 @@ import           System.Exit (ExitCode(ExitSuccess))
 import qualified System.Timeout as ST
 import           Text.PrettyPrint.ANSI.Leijen hiding ((<$>))
 
-import What4.Interface (notPred, printSymExpr,asConstantPred)
+import What4.Interface (notPred, printSymExpr,asConstantPred,getConfiguration)
+import What4.Config (setOpt, getOptionSetting)
 import qualified What4.Interface as WI
 import What4.SatResult(SatResult(..))
 import What4.Expr (ExprBuilder, GroundEvalFn(..), BoolExpr)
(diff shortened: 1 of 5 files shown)


crux/src/Crux/Goal.hs
```diff
@@ -351,7 +352,12 @@ proveGoalsOnline sym opts ctxt explainFailure (Just gs0) =
...
        else
          prepStatus "Checking: " (countGoals gs0)
-     res <- withSolverProcess sym $ \sp ->
+
+     -- make sure online features are enabled
+     enableOpt <- getOptionSetting enableOnlineBackend (getConfiguration sym)
+     _ <- setOpt enableOpt True
+
+     res <- withSolverProcess sym (fail "Online solving not enabled!") $ \sp ->
               inNewFrame sp (go (start,end) sp goalNum gs0 nameMap)
```
atomb, line 361: This should never happen, but it could be useful to clarify that it's an internal error if it ever does.

atomb (approved):
This looks great. Thanks for fixing it!

robdockins:
@spernsteiner, the changes in this PR break the way `concretize` works in `crux-mir`, and I'm not sure what it the best way to fix it.  Basically, the online backend will no longer actively maintain a solver connection during simulation unless path satisfiability checking is enabled, but the concretize function relies on this connection being active.

I guess there are a couple of options: have an alternate code path for doing concretization in "offline" mode, which would first collect the path conditions and to a one-shot SAT check, and use that if online solving isn't enabled for some reason.  Alternately, we could force the online solver to be enabled.

What do you think?

spernsteiner:
It sounds like the easiest fix for now is to force-enable the online solver in crux-mir.

I don't think it makes sense to do extra work to support `concretize` in offline mode.  In practice, `concretize` is only used for panic messages, as a way of viewing (parts of) counterexamples.  There are better ways we could support that for offline mode if needed, such as adding executable counterexample support to crux-mir.

PR #572, opened 2020-10-28
Mem alloc array check

robdockins:
Generalize and improve the `asMemAllocationArrayStore` check.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
@@ -1419,7 +1419,7 @@ freeMem :: forall sym w .
   NatRepr w ->
   LLVMPtr sym w {- ^ Base of allocation to free -} ->
   Mem sym ->
-  String {- ^ Source location -} -> 
+  String {- ^ Source location -} ->
   IO (Mem sym, Pred sym, Pred sym)
 freeMem sym w (LLVMPointer blk off) m loc =
   do p1 <- bvEq sym off =<< bvLit sym w (BV.zero w)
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -689,7 +689,7 @@ readMem sym w gsym l tp alignment m = do
             return $ Partial.totalLLVMVal sym $ LLVMValInt blk0 byte
       genValueCtor sym (memEndianForm m) mop
         =<< loadTypedValueFromBytes 0 tp loadArrayByteFn
-    Nothing -> readMem' sym w (memEndianForm m) gsym l m tp alignment (memWrites m)
+    _ -> readMem' sym w (memEndianForm m) gsym l m tp alignment (memWrites m)
```
andreistefanescu, line 692: is it possible to keep `Nothing`? I don't like matching on `_`

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1258,7 +1258,9 @@ writeMemWithAllocationCheck is_allocated sym w gsym ptr tp alignment val mem = d
               _ -> return acc_arr
       res_arr <- foldM storeArrayByteFn arr [0 .. (sz - 1)]
       return $ memAddWrite ptr (MemArrayStore res_arr (Just arr_sz)) mem
-    Nothing -> return $ memAddWrite ptr (MemStore val tp alignment) mem
+
+    _ -> return $ memAddWrite ptr (MemStore val tp alignment) mem
```
andreistefanescu, line 1262: `_` -> `Nothing`?

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1535,42 +1538,67 @@ asMemAllocationArrayStore ::
...
+                    (Nothing, Nothing) -> pure Nothing
+
+              -- Only traverse the tail if necessary, and be careful
+              -- only to traverse it once
               case (lhs_result, rhs_result) of
-                (Just lhs_arr, Just rhs_arr) ->
-                  Just <$> arrayIte sym cond lhs_arr rhs_arr
-                _ -> return Nothing
+                (Nothing, Nothing) -> findArrayStore tail_mem_writes
+                (_, Nothing) ->
```
andreistefanescu, line 1590: let's make it consistent with `(Just _, Just _)`: `_` -> `Just _`

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1535,42 +1538,67 @@ asMemAllocationArrayStore ::
...
+                         pure (Just (ok, lhs_arr))
+
+                    (Nothing, Just (rhs_ok, rhs_arr)) ->
+                      do cond' <- notPred sym cond
+                         ok <- andPred sym cond' rhs_ok
+                         pure (Just (ok, rhs_arr))
+
+                    (Nothing, Nothing) -> pure Nothing
+
+              -- Only traverse the tail if necessary, and be careful
```
andreistefanescu, line 1586: thanks for fixing this, I forgot about it!

andreistefanescu (approved):

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
robdockins: No, because the previous `Just` clause has a guard.  We could add an `otherwise` and repeat the RHS, but I think that's worse than a wildcard match.

Issue #478, opened 2020-05-08
Bad simulation performance on `if ... { if ... { return } }`


spernsteiner:
`good` and `bad` do the same thing, but simulation of `good` runs fast (under a second even for N = 100) while `bad` is very slow (4 seconds for N = 12, and grows exponentially).

```C
#include <stdint.h>
#include <crucible.h>

#define N 12

int good() {
    for (int i = 0; i < N; ++i) {
        uint8_t x = crucible_uint8_t("x");
        // One symbolic branch
        if (100 < x) {
            return 1;
        }
    }
    return 0;
}

int bad() {
    for (int i = 0; i < N; ++i) {
        uint8_t x = crucible_uint8_t("x");
        // Two symbolic branches
        if (100 < x) {
            if (100 < x) {
                return 1;
            }
        }
    }
    return 0;
}

int main() {
    //return good();
    return bad();
}
```

The problem seems related to the number of symbolic branches leading up to the `return`.  For example, `if (100 < x && x < 200) return 1;` (which has 2 branches) performs like `bad`, while `if ((100 < x) & (x < 200)) return 1;` (1 branch; note use of `&`) performs like `good`.  Adding more branches, as in `x < 100 && x < 100 && x < 100`, makes performance even worse (4.5s at N = 8, growing as 3^N instead of 2^N).

Crux reports "All goals discharged through internal simplification", so the slowdown is entirely in Crucible, not the solver.

robdockins:
You see an almost identical performance issue between the following functions:

```
#include <stdint.h>
#include <crucible.h>

#define N 12

int good2() {
  int r = 0;

    for (int i = 0; i < N; ++i) {
        uint8_t x = crucible_uint8_t("x");
        // Two symbolic branches
        if (100 < x) {
            if (100 < x) {
	      r = 1;
            }
        }
    }

    return r;
}

int bad2() {
  int r = 0;

    for (int i = 0; i < N; ++i) {
        uint8_t x = crucible_uint8_t("x");
        // Two symbolic branches
        if (100 < x) {
            if (100 < x) {
	      r = 1; break;
            }
        }
    }

    return r;
}


int main() {
  return good2();
  //return bad2();
}
```

[...]

PR #588, opened 2020-11-24
Early merges

abakst:
This change introduces a new transform aimed to address issues observed in #478, caused by loops with "early exits", i.e. loops whose branches have postdominators outside the loop. The transform includes two components:

- A pass that identifies loops and any branches to blocks outside the loop;
- A pass that, given a set of branches ("to" and "from" blocks), inserts a single 'router' block that acts as the destination of all of the "to" blocks and then branches to the correct "from" block. That is, given pairs of blocks (bin_i, bout_j), conceptually the transformation:
  1. Invents a register `r`, and in each block `bin_i` sets `r := inj_j()` where `inj_j()` is the `j-th` injection in a sum type
  2. Invents a block `route` a modifies each `bin_i` to jump to `route`. `route` then switches on the value of `r`: if `r == inj_j()` then control flow jumps to `bout_j` .

This PR exposes this transform as an option in `crux-llvm`.


abakst:
I've marked this as WIP as I still need to properly integrate some tests, but the implementation should be ready for review

travitch:
I don't know if this is entirely relevant to you, but I saw that you have some loop finding code in here.  There is another implementation of loop finding in crucible already https://github.com/GaloisInc/crucible/blob/master/crucible/src/Lang/Crucible/Analysis/Fixpoint/Components.hs. It might not be as useful to you if you need more granularity on the sub-components of a loop.

crucible/src/Lang/Crucible/CFG/Transforms.hs
```diff
@@ -0,0 +1,730 @@
...
+       let bmap'            =  Map.insert (blockID blk) blk' bmap
+       Fold.foldlM (go penv' (Set.insert (blockID blk) seen)) (genv', bmap') next
+  where
+    next = fromMaybe [] (termNextLabels (pos_val (blockTerm blk)))
+
+    go penv' seen' (genv', bmap') blkId
+      | Just blk' <- Map.lookup blkId bmap'
+      = traverseStep f bmap' genv' penv' seen' blk'
+      | otherwise
+      = undefined
```
robdockins, line 683: Don't like to see `undefined`.  Let's `panic` if we beleive this cannot happen

crucible/src/Lang/Crucible/CFG/Transforms.hs
```diff
@@ -0,0 +1,730 @@
...
+errorPath :: Map.Map (BlockID s) (Block ext s ret) -> CFGEdge s -> Bool
+errorPath bmap (_, bid) =
+  case Map.lookup bid bmap of
+    Just (blockTerm -> Posd _ (ErrorStmt _)) -> True
+    _ -> False
+
+newtype PartialValue s tp = PartialValue { getPartial :: Reg s (MaybeType tp) }
+type ValueToPartialMap s  = MapF.MapF (Value s) (PartialValue s)
+  
+traverseCFG :: ( Monad m, TraverseExt ext )
```
robdockins, line 647: This seems like it might be generally reusable... maybe worth pulling into a more generic module?

robdockins (changes requested):
This is a hard algorithm to review, but I'll say that I don't see anything obviously wrong, so that's a good start :-)

I'm probably not going to be happy merging this until I see some kind of unit tests that show at least some simple examples of this algorithm working and producing the expected results.  I'll also want to see a lot more documentation on the internal functions, as a lot of what's going on in here is pretty subtle.

Finally, the git history for this PR has a lot of messy intermediate commits.  I'd like to see it rebased into something more organized.

crucible/src/Lang/Crucible/CFG/Transforms.hs
```diff
@@ -0,0 +1,730 @@
...
+  , liFooterIn   :: ![CFGEdge s]
+  , liDominators :: ![BlockID s]
+  } deriving (Eq, Show)
+type CFGEdge s = (BlockID s, BlockID s)
+
+instance Ord (LoopInfo s) where
+   li1 <= li2 = isNested li1 li2
+
+-- | Detect all loops in a cfg.
+-- The assumption is that two backedges in a cfg will have distinct destination blocks.
```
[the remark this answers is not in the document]
abakst: You're right, I'm not sure that it should be in general. While I'm not convinced it will matter in practice if we discard that case (i.e., don't apply the transformation), it might be worth double checking if this assumption is critical. For now, I think this assumption ought to be checked.

crucible/src/Lang/Crucible/CFG/Transforms.hs
```diff
@@ -0,0 +1,730 @@
...
+      LambdaID (lambdaAtom -> a)
+        | Just (getPartial -> pr) <- MapF.lookup (AtomValue a) pvals ->
+          do pa <- freshAtom ng (atomPosition a) (MaybeRepr (typeOfAtom a))
+             let setPa = Posd (atomPosition a) (DefineAtom pa (EvalApp (JustValue (typeOfAtom a) a)))
+                 setPr = Posd (atomPosition a) (SetReg pr pa)
+             return $ Seq.fromList [ setPa, setPr ]
+      _ -> return mempty
+
+    lowerBlockOutput = case blockTerm blk of
+      Posd p (Output ll a) ->
```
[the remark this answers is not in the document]
abakst: I'll write a proper explanation. However, if I understand correctly, these are the only term-statements that output a value (i.e. an `Atom`) to the destination block. This Atom is the (potential) target of the "lowering".

crucible/src/Lang/Crucible/CFG/Transforms.hs
```diff
@@ -0,0 +1,730 @@
...
+      mkBlock (LambdaID (mapping Ctx.! blkIdx))
+              mempty
+              mempty
+              (Posd internal (bidToTerm blkId rename)) : blks
+
+-- | Given a set of edges (i, j) in E,
+-- Create a single block that merges all paths before
+-- continuing on to the respective j's
+--
+-- Create a unique block F and multiple blocks  i' j' such
```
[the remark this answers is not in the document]
abakst: Yes, that's right. It feels a bit cumbersome -- if there's a better way to do it I'm interested.

crucible/src/Lang/Crucible/CFG/Transforms.hs
abakst: Does it make sense to move it into `Reg` (which seems to be the main module for mutable-register based CFGs)?

crucible/src/Lang/Crucible/CFG/Transforms.hs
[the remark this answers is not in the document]
robdockins: Humm... not offhand.

crucible/src/Lang/Crucible/CFG/Transforms.hs
robdockins: Yeah, it might make sense there.

crucible/src/Lang/Crucible/CFG/Transforms.hs
```diff
@@ -0,0 +1,730 @@
...
+    members    = [ x | x <- fromHeader, header `elem` G.reachable x g, headerDominates x ]
+    headerDominates n
+      | Just ds <- Map.lookup n doms
+      = header `elem` ds
+      | otherwise
+      = False
+
+------------------------------------------------------------------------
+-- Fixup Pass
+              -- HEREHEREHERE need to account for lambda block lowered atoms
```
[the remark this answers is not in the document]
abakst: Eek, that one was leftover, thanks.

crucible/src/Lang/Crucible/CFG/Transforms.hs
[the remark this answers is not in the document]
abakst: Actually on further reflection, there should be a few others... thanks :)

abakst:
I believe I've addressed your comments @robdockins. I also added some tests and tweaked how the test suite works: basically the test suite can be organized into subdirectories and the tests will be grouped accordingly. This seemed like a convenient way to organize any potential future test suites, but I'm not especially committed to this structure.

crucible/src/Lang/Crucible/CFG/EarlyMergeLoops.hs
```diff
@@ -0,0 +1,839 @@
...
+  -- is not the footer node.
+  , liFooterIn   :: ![CFGEdge s]
+  -- ^ All the edges to the footer
+  , liDominators :: ![BlockID s]
+  -- ^ The dominators of the loop header.
+  } deriving (Eq, Show)
+type CFGEdge s = (BlockID s, BlockID s)
+
+-- | Detect all loops in a cfg.
+-- The assumption is that two backedges in a cfg will have distinct destination blocks.
```
robdockins, line 87: If this assumption is violated, then we return the empty list of loop info and do nothing, is that correct?  Let's state that in this comment.

crucible/src/Lang/Crucible/CFG/EarlyMergeLoops.hs
```diff
@@ -0,0 +1,839 @@
...
+           | blockID blk == LabelID rootLabel
+           = return root'
+           | otherwise
+           = lowerBlock ng pvals blk
+
+     blks' <- mapM lower (cfgBlocks cfg)
+     let cfg' = cfg { cfgBlocks = blks' }
+     return cfg'
+  where
+    root = fromMaybe (error "Unknown root") $ findBlock cfg rootLabel
```
robdockins, line 221: Let's use `panic` here instead.

crux-llvm/src/Crux/LLVM/Config.hs
```diff
@@ -107,6 +108,9 @@ llvmCruxConfig =
          optLevel <- Crux.section "opt-level" Crux.numSpec 1
                            "Optimization level to request from `clang`"
 
+         loopMerge <- Crux.section "opt-loop-merge" Crux.yesOrNoSpec False
+                        "Insert merge blocks in loops with early exits"
```
robdockins, line 112: Maybe add a short sentence here about why you'd want to do this.  "may speed up simulation", etc.

robdockins:
This is in much better shape than before, thanks for working on this @abakst!

I have some minor comments above, and a slightly more serious comment here.  I'm still a little concerned about the algorithmic structure of `earlyMergeLoops`.  I think I understand why we need to reanalyze the structure of the loops after every iteration, but are we pretty sure we can't accidentally get into an infinite loop?  We're adding new blocks, but I guess they can never be the target of a back-edge, and that's enough to ensure that the keeping track of the seen blocks means we eventually terminate?

Second, I wonder about running `lowerUndefPass` after each iteration inside `earlyMergeLoop`... isn't it possible that the same register could be lowered multiple times, resulting in an unnecessary tower of `Maybe` types and projections?  Would it work to run `lowerUndefPass` just once after all the loops have been defunctionalized?

robdockins (changes requested):
Minor changes and some questions; see above.  Otherwise, I think this is ready to go.

abakst:
@robdockins I believe I've addressed your concerns at this point, with one additional change -- I've checked in the optimization tests as LLVM bitcode (disassembled) so that the tests remain stable across clang versions.

It's worth recording here that for some of the tests, clang version 10 produces CFGs whose performance with and/or without the optimization is quite different from the CFG produced by version 9. It might be worth investigating this more in the future to tune how we decide which loops should be transformed.

I think the remaining CI failures (on macOS only) are unrelated to this change.

crucible/src/Lang/Crucible/CFG/EarlyMergeLoops.hs
```diff
@@ -541,22 +542,51 @@ earlyMergeLoops :: ( TraverseExt ext, Monad m, Show (CFG ext s init ret) )
...
+  = do cfg' <- earlyMergeLoop ng cfg l
+
+       -- Check if we should proceed: in particular, if the new CFG
+       -- renamed or produced new loops, then we need to bail as we
+       -- might not terminate in that case.
+       let ls'      = cfgLoops cfg'
+       let seen'    = (Set.insert (liHeader l) seen)
+       let nextStep = candidates seen' ls'
+       -- The termination transition invariant is that 
+       when (length thisStep <= length nextStep) $
```
robdockins, line 573: Nitpick, this should be `<` for termination.

robdockins (approved):
Let's keep the original `.c` files that produced the `.ll` around instead of deleting them (so we can update them in the future if necessary), but otherwise this looks good!

PR #592, opened 2020-12-17
Make the crucible-llvm tests more configurable

travitch:
This change adds options to choose the clang and llvm-as binaries used by the
test suite (the defaults are the same as before). This is useful for Linux
distros where those tools are installed with versioned suffixes.


kquick (approved):
Nice update!

PR #594, opened 2020-12-19
WebAssembly support for crucible

robdockins:
This PR implements a rough first cut at providing WebAssembly translation support for Crucible.  Support is currently incomplete in a variety of ways.  For reference, I have been following the WebAssembly 1.0 specification, which may be found here:
https://www.w3.org/TR/2019/REC-wasm-core-1-20191205 (there is a newer 1.1 spec, but it doesn't seem to have much uptake yet, and the changes seem relatively small).

What works:

- Parsing and loading modules via "Script" format from the `haskell-wasm` package
- Much of the module instantiation/linking phase
- Instruction translations for integer value arithmetic, control flow, direct function calls and memory access
- Running `assert_return` clauses of the `haskell-wasm` test suite

What doesn't work yet:

- Most floating-point instructions
- Numeric conversions
- data segment initialization
- table segment initialization
- the indirect function call instruction

Other rough edges:

- This code doesn't compile as-is without exposing more of the abstract syntax data structures from `haskell-wasm`.  I have not yet tried to get these changes incorporated upstream
- Error handling is a bit of a mess.  Many `error` calls should be converted to `panic`, as they should be ruled out by module verification, and others should be reported in more appropriate ways.
- Comments are currently sparse on the ground
- No test suite yet
- I'm not confident we are doing the correct thing WRT partial numeric instructions.  We should ensure that we are making appropriate assertions regarding divide-by-zero, etc.

### DIFF
--- a/crucible-wasm/src/Lang/Crucible/Wasm.hs
+++ b/crucible-wasm/src/Lang/Crucible/Wasm.hs
@@ -1,0 +1,322 @@
+{-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE RecordWildCards #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE ViewPatterns #-}
+
+module Lang.Crucible.Wasm
+( ExternalType(..)
+, Address
+, Namespace
+, FuncTable
+, lookupFuncTable
+, Store(..)
+, ModuleInstance(..)
+, WasmGlobal(..)
+, instantiateModule
+, execScript
+, emptyScriptState
+, WasmExt
+, extImpl
+, wasmIntrinsicTypes
+) where
+
+import Control.Lens hiding (Empty, (:>),Index )
+import Control.Monad
+import Control.Monad.Trans
+
+import qualified Data.Map as Map
+import qualified Data.Sequence as Seq
+import qualified Data.Text.Lazy as TL
+import qualified Data.BitVector.Sized as BV
+import qualified Data.ByteString.Lazy as LBS
+import qualified Data.Parameterized.Map as MapF
+import           Data.Word
+
+import Data.Parameterized.Context as Ctx
+import Data.Parameterized.Some
+
+import qualified Language.Wasm as Wasm
+import qualified Language.Wasm.Structure as Wasm
+
+import Lang.Crucible.Backend
+import Lang.Crucible.FunctionHandle
+import Lang.Crucible.Simulator
+import Lang.Crucible.Types
+
+import qualified Lang.Crucible.LLVM.MemModel.Generic as G
+
+import Lang.Crucible.Wasm.Extension
+import Lang.Crucible.Wasm.Instantiate
+import Lang.Crucible.Wasm.Memory
+import Lang.Crucible.Wasm.Translate
+import Lang.Crucible.Wasm.Utils
+
+import What4.Interface
+
+wasmIntrinsicTypes :: IsSymInterface sym => IntrinsicTypes sym
+wasmIntrinsicTypes =
+  MapF.insert (knownSymbol @"Wasm_Mem") IntrinsicMuxFn $
+  MapF.empty
+
+
+data ScriptState =
+  ScriptState
+  { scriptStore :: Store
+  , scriptNamespace :: Namespace
+  , scriptLoadedModules :: Namespace
+  , scriptCurrentModule :: Maybe (Maybe Wasm.Ident, Wasm.ValidModule, ModuleInstance)
+  }
+
+emptyScriptState :: ScriptState
+emptyScriptState =
+  ScriptState
+  { scriptStore = emptyStore
+  , scriptNamespace = emptyNamespace
+  , scriptLoadedModules = emptyNamespace
+  , scriptCurrentModule = Nothing
+  }
+
+type WasmOverride p sym = OverrideSim p sym WasmExt (RegEntry sym UnitType)
+
+execScript :: [Wasm.Command] -> ScriptState -> WasmOverride p sym EmptyCtx UnitType ScriptState
+execScript script ss = foldM (flip execCommand) ss script
+
+execCommand :: Wasm.Command -> ScriptState -> WasmOverride p sym EmptyCtx UnitType ScriptState
+execCommand (Wasm.ModuleDef mdef) ss =
+  do halloc <- use (stateContext . to simHandleAllocator)
+     let st = scriptStore ss
+     let ns = scriptNamespace ss
+     ((nm,vm,im), st') <- runInstM st (defineModule halloc ns mdef)
+
+     initGlobals im st'
+     initMemories im st'
+
+     mapM_ registerFnBinding =<< liftIO (translateFunctions im st')
+     debug (show (instDataSegs im))
+     mapM_ writeDataSegment (instDataSegs im)
+
+     executeStart im
+
+     let lmods' = case nm of
+                    Nothing  -> scriptLoadedModules ss
+                    Just nm' -> namespaceInsertModule nm' im (scriptLoadedModules ss)
+
+     return ss{ scriptStore = st'
+              , scriptLoadedModules = lmods'
+              , scriptCurrentModule = Just (nm,vm,im)
+              }
+
+execCommand (Wasm.Assertion a) ss =
+  do execAssertion a ss
+     return ss
+
+execCommand (Wasm.Action a) ss =
+  do _ <- execAction a ss
+     return ss
+
+execCommand c ss =
+  do liftIO $ putStrLn $ unwords ["Command not implemented", show c] -- TODO!
+     return ss
+
+
+executeStart ::
+  ModuleInstance ->
+  WasmOverride p sym EmptyCtx UnitType ()
+executeStart im =
+  case instStartFunc im of
+    Nothing -> pure ()
+    Just startFn ->
+      do ctx <- getContext
+         ctxSolverProof ctx $ void $
+           callFnVal' (HandleFnVal startFn) Empty
+
+writeDataSegment ::
+  (GlobalVar WasmMem, Word32, LBS.ByteString) ->
+  WasmOverride p sym EmptyCtx UnitType ()
+writeDataSegment (mvar, offset, chunk) =
+  do ctx <- getContext
+     ctxSolverProof ctx $
+       do sym <- getSymInterface
+          mem <- readGlobal mvar
+          mem' <- liftIO $
+                  do offset' <- bvLit sym knownNat (BV.word32 offset)
+                     wasmStoreChunk sym offset' chunk mem
+          debug (show (G.ppMem (wasmMemHeap mem')))
+          writeGlobal mvar mem'
+
+initMemories ::
+  ModuleInstance ->
+  Store ->
+  WasmOverride p sym EmptyCtx UnitType ()
+initMemories im st =
+  do ctx <- getContext
+     sym <- getSymInterface
+     ctxSolverProof ctx $ forM_ (instMemAddrs im) \case
+       (Wasm.Memory lim, True, addr) ->
+         case Seq.lookup addr (storeMemories st) of
+           Nothing -> panic "initMemories" ["could not resolve memory address " ++ show addr]
+           Just gv ->
+             do mem <- liftIO (freshMemory sym lim)
+                writeGlobal gv mem
+
+       -- imported memory, no need to initialize
+       _ -> return ()
+
+initGlobals ::
+  ModuleInstance ->
+  Store ->
+  WasmOverride p sym EmptyCtx UnitType ()
+initGlobals im st =
+  do ctx <- getContext
+     sym <- getSymInterface
+     ctxSolverProof ctx $ forM_ (instGlobalAddrs im) \case
+       (_gty, Just cv, addr) ->
+          case Seq.lookup addr (storeGlobals st) of
+            Nothing -> panic "initGlobals" ["could not resolve global address " ++ show addr]
+            Just (ConstantGlobal _) -> return () -- TODO? shouldn't happen, error?
+            Just (MutableGlobal gv) ->
+              case cv of
+                I32Val w
+                  | Just Refl <- testEquality (globalType gv) (BVRepr (knownNat @32))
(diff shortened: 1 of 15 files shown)


robdockins:
Since the initial comment, I've added support for:

* table segments
* data segments
* start function execution
* indirect function-call support with concrete index argument

Finishing indirect function call support for symbolic arguments will probably require the same work as a fix for #10, which is support for truly symbolic function handles.

robdockins:
I'd like to get some resolution to https://github.com/SPY/haskell-wasm/pull/10 before we merge this, but I think this is otherwise ready to merge.  I don't want this to get too out of date before merging, even though it is incomplete.

crucible-wasm/src/Lang/Crucible/Wasm/Memory.hs
```diff
@@ -0,0 +1,277 @@
...
+checkLim sym (Wasm.Limit lo Nothing)
+  | lo <= bit 16
+  = do lo' <- bvLit sym knownNat (BV.mkBV knownNat (fromIntegral lo))
+       hi' <- bvLit sym knownNat maxPages
+       pure (lo',hi')
+
+checkLim sym (Wasm.Limit lo (Just hi))
+  | lo <= hi && hi <= bit 16
+  = do lo' <- bvLit sym knownNat (BV.mkBV knownNat (fromIntegral lo))
+       hi' <- bvLit sym knownNat maxPages
```
kquick, line 86: Should this be `hi` instead of `maxPages`?

crucible/src/Lang/Crucible/CFG/Expr.hs
```diff
@@ -765,6 +765,38 @@ data App (ext :: Type) (f :: CrucibleType -> Type) (tp :: CrucibleType) where
...
+        -> App ext f (BVType w)
+
+  -- count leading zeros
+  BVClz :: (1 <= w)
+        => !(NatRepr w)
+        -> !(f (BVType w))
+        -> App ext f (BVType w)
+
+  -- count trailing zeros
+  BVCtz :: (1 <= w)
```
kquick, line 789: Mnemonically `BVClz` initially made me think "clear with zeroes".  Since these are new, maybe something a little more suggestive like `BVCntLZ` and `BVCntTZ`

crucible-wasm/tool/Main.hs
```diff
@@ -0,0 +1,77 @@
+{-# LANGUAGE BlockArguments #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Main where
+
+--import Control.Monad
+--import Control.Monad.IO.Class
```
kquick, line 7: Remove unused imports.

crucible-wasm/src/Lang/Crucible/Wasm/Translate.hs
```diff
@@ -0,0 +1,861 @@
...
+--import Control.Monad.Trans.State
+--import Control.Monad.Trans.Except
+import Data.Foldable(toList)
+import Data.Word
+import Numeric.Natural
+import GHC.Stack
+
+--import qualified Data.Text.Lazy as TL
+--import qualified Data.Map as Map
+--import qualified Data.IntMap as IntMap
```
kquick, line 34: Might as well cleanup these unused imports.

crucible-wasm/src/Lang/Crucible/Wasm/Translate.hs
```diff
@@ -0,0 +1,861 @@
...
+     iteVals c (StackI32 x) (StackI32 y)
+iteVals c (StackI64 x) (StackBool b) =
+  do y <- mkAtom (App (BoolToBV knownNat (AtomExpr b)))
+     iteVals c (StackI64 x) (StackI64 y)
+iteVals c (StackBool b) (StackI32 y) =
+  do x <- mkAtom (App (BoolToBV knownNat (AtomExpr b)))
+     iteVals c (StackI32 x) (StackI32 y)
+iteVals c (StackBool b) (StackI64 y) =
+  do x <- mkAtom (App (BoolToBV knownNat (AtomExpr b)))
+     iteVals c (StackI64 x) (StackI64 y)
```
kquick, line 167: Should there be  promotion from I32 to I64 (and F32 to F64)?

crucible-wasm/src/Lang/Crucible/Wasm/Translate.hs
```diff
@@ -0,0 +1,861 @@
...
+
+translateFunction :: forall p sym args ret.
+  Wasm.FuncType ->
+  Wasm.Function ->
+  ModuleInstance ->
+  Store ->
+  FnHandle args ret ->
+  IO (Sim.FnBinding p sym WasmExt)
+translateFunction fty fn@Wasm.Function{ .. } im st hdl =
+  do liftIO $ putStrLn $ unlines ["Translating function", show fn]
```
kquick, line 305: Logging/debug?

crucible-wasm/src/Lang/Crucible/Wasm/Translate.hs
```diff
@@ -0,0 +1,861 @@
...
+     (SomeCFG g, []) <- defineFunction InternalPos (Some globalNonceGenerator) hdl defn
+     case toSSA g of
+       C.SomeCFG g' ->
+         do liftIO (print g') -- TODO remove debugging output
```
kquick, line 309: TODO

crucible-wasm/src/Lang/Crucible/Wasm/Translate.hs
```diff
@@ -0,0 +1,861 @@
...
+    -- I64ExtendSI32
+    -- I64ExtendUI32
+    -- FConvertIU {- Float Size -} BitSize {- Int Size -} BitSize
+    -- FConvertIS {- Float Size -} BitSize {- Int Size -} BitSize
+    -- F32DemoteF64
+    -- F64PromoteF32
+    -- IReinterpretF BitSize
+    -- FReinterpretI BitSize
+
+    _ -> error $ unwords ["Instruction not implemented", show instr]
```
kquick, line 775: `panic` here

crucible-wasm/src/Lang/Crucible/Wasm/Instantiate.hs
```diff
@@ -0,0 +1,531 @@
...
+     vm <- case Wasm.validate m of
+             Left verr -> throwE (ValidationError verr)
+             Right vm  -> pure vm
+     im <- instantiateModule halloc ns vm
+     pure (ident, vm, im)
+
+
+buildModule :: Wasm.ModuleDef -> InstM (Maybe Wasm.Ident, Wasm.Module)
+buildModule (Wasm.RawModDef ident m) = pure (ident, m)
+buildModule _ = bailOut "build module"
```
kquick, line 61: Just curious if the semantics here are different than for `panic`?

kquick: And if so, maybe document why.

crucible-wasm/src/Lang/Crucible/Wasm/Instantiate.hs
```diff
@@ -0,0 +1,531 @@
...
+installElemSegment ::
+  ModuleInstance ->
+  Wasm.ElemSegment ->
+  InstM ()
+installElemSegment im es@Wasm.ElemSegment{ .. } =
+  do I32Val tblOff <- evalConst im offset -- TODO, better error management
+     st <- lift get
+     liftIO (putStrLn "installing element segment")
+     liftIO (print es)
+     liftIO (print (instTableAddrs im))
```
kquick, line 421: logging/debug?

crucible-wasm/src/Lang/Crucible/Wasm/Instantiate.hs
```diff
@@ -0,0 +1,531 @@
...
+                    (ExternalTable (Wasm.TableType lim' tt'))
+  | tt == tt' && matchLim lim lim' = Just (ExternalTable (Wasm.TableType lim tt))
+
+matchImportDesc _ (Wasm.ImportMemory mt) (ExternalMem (Wasm.Memory mt'))
+  | matchLim mt mt' = Just (ExternalMem (Wasm.Memory mt))
+
+matchImportDesc _ _ _ = Nothing
+
+-- TODO double check that we got this limit checking stuff in the right order
+--  and it all makes sense
```
kquick, line 502: Looks reasonable to me.  Is the TODO still needed?  If it's still needed, can you provide more here about the context in which it needs to be validated?

kquick (approved):
This looks really clear and clean!

I realize it was a "pencils down" situation that stopped your work, so apologies for the comments on the todo-ish portions, but it would be good to get those into an actionable form.

Most of the rest of the comments are stylistic or suggestions, so feel free to ignore if you don't like them.  

I'm looking forward to seeing this merged into crucible master!

crucible-wasm/src/Lang/Crucible/Wasm/Memory.hs
```diff
@@ -0,0 +1,277 @@
...
+import Lang.Crucible.Backend
+import Lang.Crucible.Simulator
+import Lang.Crucible.Simulator.Intrinsics
+import Lang.Crucible.Types
+
+import qualified Language.Wasm.Structure as Wasm
+
+import           Lang.Crucible.LLVM.Bytes
+import           Lang.Crucible.LLVM.DataLayout (EndianForm(..), noAlignment)
+import           Lang.Crucible.LLVM.MemModel
```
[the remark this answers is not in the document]
robdockins: Agreed, I thought about breaking out just such a package as I was working on this, but I didn't want to take away time from the immediate goal at hand.  One of the interesting things that came out of this for me was a more clear idea of which parts of the memory model really are C/LLVM specific and which parts are more fully reusable.

crucible/src/Lang/Crucible/CFG/Expr.hs
robdockins: Sure, we can use longer names here.

crucible-wasm/src/Lang/Crucible/Wasm/Translate.hs
robdockins: No, this promotion isn't actually part of the semantics, it's just a way to avoid certain noisy round-tripping operations that encode bools into bitvectors and project them back.  There's no automatic promotion in the Wasm spec.

crucible-wasm/src/Lang/Crucible/Wasm.hs
```diff
@@ -0,0 +1,316 @@
...
+execCommand (Wasm.Assertion a) ss =
+  do execAssertion a ss
+     return ss
+
+execCommand (Wasm.Action a) ss =
+  do _ <- execAction a ss
+     return ss
+
+execCommand c ss =
+  do liftIO $ putStrLn $ unwords ["Command not implemented", show c]
```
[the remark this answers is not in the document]
robdockins: That's a good question.  I'm coming around to the idea that we should perhaps make a distinction between `panic` and `unimplemented`.  In my mind, `panic` means the programmer has thought about this and believes that various explicit or implicit invariants mean that the situation cannot occur, and it is a programming mistake bug if the situation occurs.  Whereas, `unimplemented` or `unsupported` is explicitly an acknowledgement that valid inputs can drive the program into this situation, but we just haven't written the necessary code to deal with it.

Operationally, I think `unimplemented` should probably raise an appropriate (potentially non-fatal) exception instead of printing the panic message and immediately exiting.

crucible-wasm/src/Lang/Crucible/Wasm/Extension.hs
```diff
@@ -0,0 +1,325 @@
...
+  Wasm_IndirectFunction ::
+    Wasm.FuncType {- ^ Type of the function to lookup -} ->
+    FuncTable {- ^ table in which to look up functions -} ->
+    TypeRepr (FunctionHandleType args ret) {- ^ Crucible type of the function to look up -} ->
+    !(f (BVType 32)) {- ^ computed table index -} ->
+    WasmStmt f (FunctionHandleType args ret)
+
+  Wasm_MemSize ::
+    GlobalVar WasmMem ->
+    WasmStmt f (BVType 32)
```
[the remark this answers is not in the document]
robdockins: Yeah, I like that idea.

crucible-wasm/src/Lang/Crucible/Wasm/Instantiate.hs
robdockins: I mostly used this for unimplemented stuff, which I'm starting to think should be different from `panic` as I discussed above.

crucible-wasm/src/Lang/Crucible/Wasm/Instantiate.hs
robdockins: There's some slightly complicated contravariance going on here, and I want to really check carefully against the spec that I got all the inequalities in the correct direction.

crucible-wasm/src/Lang/Crucible/Wasm.hs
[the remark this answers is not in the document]
kquick: I agree that there's definitely a semantic differentiation, but is there a runtime distinction (i.e. can it be a non-fatal/recoverable exception if it's truly unimplemented)?  If so, then yes, let's make an `unimplemented`, otherwise we can just make sure the panic message clarifies the distinction (or I guess we can make an `unimplemented` that does the latter).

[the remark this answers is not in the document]
robdockins: Well... for something like SAW, I guess I'd rather see some useful "unimplemented" message and end up back at the SAW prompt rather than baling out of the executable altogether like `panic` does.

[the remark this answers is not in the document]
kquick: My overall thinking relative to all the panic notes below is leaving the right breadcrumbs for both the user and the developer to know who to tell and where to fix it when these unimplemented/todo/whatever things get encountered. 
 
As context: Comments in the code, `print` statements that might get lost or don't suggest there needs to be a report, or `error` calls that might simply result in alternatives or quietly unsatisfiable branches are mostly passive and could result in invalid analyses or a lot of debugging to figure out why something isn't working right, which is why I tend to default to a fairly agressive reaction like `panic`.

crucible-wasm/src/Lang/Crucible/Wasm/Instantiate.hs
kquick: Can you add this context to the comments in the code?

crucible-wasm/src/Lang/Crucible/Wasm/Translate.hs
kquick: Gotcha.  Can you add that to the comments for this block of functions?

crucible-wasm/src/Lang/Crucible/Wasm/Memory.hs
robdockins: oh... good catch

PR #615, opened 2021-01-15
Start support for muxing function handles

travitch:
This change (which does not compile yet) generalizes the current `RegValue`
instance for function handles into a mux (the `PartitioningMuxTree`, ported from
the `FancyMuxTree` in crux-mir).

This change modifies the call sequence to call all possible functions indicated
by a muxed function handle and mux their results together.


langston-barrett:
Closing as this is very old and unlikely to merged in its current state.

PR #616, opened 2021-01-15
Crucible llvm testsplit

kquick:
* Splits out crucible llvm tests into separate modules to allow for easier reading/extension.

* Updates the file-based tests to use tasty-sugar for file detection instead of a hard-coded set of input files which will allow for easy future extension of both test types and test cases.

* Compilation and assembly of file-based tests are part of the tests themselves and not pre-steps that would block other testing on failure.

* Accepts `CLANG` and `LLVM_AS` environment variables as well as `--clang` and `--llvm-assembler` command-line options for more flexibility and compatibility with crux-llvm.


robdockins (approved):
I haven't looked thorough the contents of the test, I assume you just moved them around.  Looks like a good refactoring to me.

travitch (approved):
Looks good - it might just want to use the new SomeSym in parameterized-utils

Issue #439, opened 2020-02-12
per-goal timeout option doesn't seem to be working for yices


robdockins:
crux verification goals do not seem to be honoring the goal timeout setting.

atomb:
This seems to be working now.

PR #617, opened 2021-01-17
[crucible] Add deadman timeout for online solver actions

kquick:
Fixes https://github.com/GaloisInc/crucible/issues/439 by implementing
a parallel timeout operation (at 1 second longer than the timeout
requested of the solver itself) whose expiration causes the solver
process to be killed.

This helps with issue 439 because Yices encounters some internal
divergence issue that is not stopped by the specified timeout
value (demonstrable by running the crux-llvm
test-data/golden/strlen_test2 with the Yices solver option).

This also helps provide timeout management for solvers that do not
provide a built-in timeout capability (like Boolector).

### DIFF
--- a/crucible/src/Lang/Crucible/Backend/Online.hs
+++ b/crucible/src/Lang/Crucible/Backend/Online.hs
@@ -74,6 +74,9 @@ module Lang.Crucible.Backend.Online
   ) where
 
 
+import           Control.Concurrent ( threadDelay )
+import           Control.Concurrent.Async ( race )
+import           Control.Exception ( throwIO )
 import           Control.Lens
 import           Control.Monad
 import           Control.Monad.Catch
@@ -83,11 +86,11 @@ import           Data.Data (Data)
 import           Data.Foldable
 import           Data.IORef
 import           Data.Parameterized.Nonce
+import qualified Data.Text as Text
 import           Data.Typeable (Typeable)
 import           GHC.Generics (Generic)
-import           System.IO
-import qualified Data.Text as Text
 import qualified Prettyprinter as PP
+import           System.IO
 
 import           What4.Config
 import           What4.Concrete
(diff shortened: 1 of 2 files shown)


kquick:
Note that this requires https://github.com/GaloisInc/what4/pull/83

crucible/src/Lang/Crucible/Backend/Online.hs
```diff
@@ -422,17 +425,40 @@ withSolverProcess' getSolver sym def action = do
...
+                ContinueOnError ->
+                  action p
+                ImmediateExit ->
+                  onException (action p) stopTheSolver
+
+        if (getGoalTimeoutInSeconds $ solverGoalTimeout p) == 0
+          then doAction
+          else let deadmanTimeoutPeriod = (fromInteger $
+                                           getGoalTimeoutInMilliSeconds
+                                           (solverGoalTimeout p) + 1000) * 1000
```
robdockins, line 446: This  multiplication by 1000 is suspicious... is `threadDelay` in microseconds? If so, let's comment about that.

robdockins (approved):
I like the idea of this, although I'm always a little suspicious of any changes to the process handling, as it's easy to have unintended effects.  Have you tested that everything works as it should? Solvers get killed when they need to and not when they don't? Do keyboard interrupts still work correctly, etc?

atomb (approved):
I'm really happy to have this functionality, but would also second @robdockins concern. We should make sure we test this as  thoroughly as possible.

travitch:
A note we discussed on a bug triage call: last time I was working on timeouts for yices, I think I saw that it has different handling of signals if you are using the MCSAT backend vs the standard backend. I think that it could swallow signals sometimes, which might be something we need to report upstream.

kquick:
Just a note, the timeout should be handled at the what4 level, although there may be associated handling here.  There is a what4 change in progress, but not yet submitted.

langston-barrett:
I assume the What4 change was https://github.com/GaloisInc/what4/pull/128? Is there anything left to do here? Please reopen if so.

PR #619, opened 2021-01-18
Pin LLVM versions in github actions

jared-w:
This should prevent our builds from failing if LLVM changes underneath as previously we were relying on the first LLVM to be found on the path, which can change without warning.

### DIFF
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -112,29 +112,13 @@ build() {
 }
 
 test() {
-  # System-agnostic path
-  export PATH="$PATH:/usr/local/opt/llvm/bin:/c/Program Files/LLVM/bin"
   cabal v2-test "$@"
-}
-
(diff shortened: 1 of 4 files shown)


kquick:
With the merge of 624 this is less of an issue now that support for LLVM 11 has been merged in.  The remaining failure is a legitimate behavior change in LLVM 11 that I should be able to address in #618.

jared-w:
Good to know! It still seems like a good idea to make sure that we know what version of clang we're using. The fact that it worked for so long is a matter of luck more than anything else, and I'd prefer to save my luck for other things :)

kquick:
In 618, the clang version is detected and reported as part of the test output.

robdockins:
I think this is superseded by #753.  Feel free to reopen if I'm mistaken.

PR #622, opened 2021-01-22
crucible-jvm: Migrate never-updated fields from CodebaseState to Codebase

RyanGlScott:
`Lang.JVM.Codebase` has a `CodebaseState` data type that is used within an `IORef` in `Codebase`. However, only two of the four fields of `CodebaseState` are ever updated with `writeIORef`, so it doesn't make much sense to keep the other two fields in an `IORef`. This patch moves them from `CodebaseState` to `Codebase` to make this more obvious.

This is purely a refactoring and has no user-visible changes in behavior.


RyanGlScott:
This was superseded by #638.

PR #624, opened 2021-01-26
Update to include support for LLVM versions 11 and 12.

kquick:


kquick:
The  update to the llvm-pretty and llvm-pretty-bc-parser now cause the MacOS tests (running Clang 11) to run further.  

The single test failure is presumably not a crucible issue but a behavior difference in output flushing introduced in LLVM 11?  [I can get the same failure under Linux using Clang/LLVM 11 there as well.] If that is the case, the expected test results can be most easily parameterized for LLVM version in pending PR #618.

I therefore recommend merging this PR as an improvement and addressing the difference in expected test output via PR #618.

atomb (approved):
This seems like an improvement, and I agree that we can make further progress on the output flushing issue as part of a separate PR. Thanks!

robdockins (approved):

PR #635, opened 2021-02-02
attempt to fix gpg signing

jared-w:


RyanGlScott (approved):
This looks sensible to me. I suppose we won't know for sure if this fixes the problem I have (trying to submit a PR from a fork, rather than a branch of https://github.com/GaloisInc/crucible) until I rebase on top of this, but until then, I think it would be reasonable to land this.

Issue #634, opened 2021-02-02
Run crucible-jvm tests on CI


RyanGlScott:
Currently, the `tests.sh` script associated with `crucible-jvm` is not run on CI, which leaves it susceptible to bitrotting. Indeed, it's not surprising that's it isn't ran on CI at the moment, as it hardcodes the location of `rt.jar`:

https://github.com/GaloisInc/crucible/blob/91236c30878f17f7d3421c70b65508a0299be52c/crucible-jvm/tests/test.sh#L3

We should fix this and make sure that CI is capable of running this.

RyanGlScott:
It looks like the ashes test suite under `tests/ashes` (which also leverages `crucible-jvm`) isn't ran on CI either.

RyanGlScott:
Getting ashes to work at all (let alone on CI) will likely be more challenging, as it doesn't appear to mesh well with how Crux currently dumps information to the terminal:

```
$ ./runAshes.hs 
kaffeRegressionSuite/benchmarks/str                        Unexpect Fail
Expected:
12abcd34567890
Got:
[Crux] Checking "Str.java"
12abcd78345678
[Crux] Simulation complete.
[Crux] All goals discharged through internal simplification.
[Crux] Overall status: Valid.
initializeClass java/lang/String  (start)
initializeClass java/lang/Object  (start)
special java/lang/Object/<clinit>
initializeClass java/lang/Object  (finish)
special java/lang/String/<clinit>
initializeClass java/lang/String  (finish)
new ClassName "java/lang/StringBuffer" (start)
new ClassName "java/lang/StringBuffer" (finish)
initializeClass java/lang/StringBuffer  (start)
initializeClass java/lang/AbstractStringBuilder  (start)
initializeClass java/lang/Object  (start)
initializeClass java/lang/Object  (finish)
initializeClass java/lang/AbstractStringBuilder  (finish)
special java/lang/StringBuffer/<clinit>
initializeClass java/lang/StringBuffer  (finish)
initializeClass java/lang/String  (start)
initializeClass java/lang/Object  (start)
initializeClass java/lang/Object  (finish)
initializeClass java/lang/String  (finish)
invoke: java/lang/StringBuffer/insert
finish invoke:java/lang/StringBuffer/insert
initializeClass java/lang/System  (start)
initializeClass java/lang/Object  (start)
initializeClass java/lang/Object  (finish)
Initializing System.out static field
Finished initializing System.out
Initializing System.err static field
[...]

PR #636, opened 2021-02-02
crucible-jvm: Add --java-bin-dirs

### LINKED ISSUE #633: crucible-jvm: Allow specifying Java's location with --java-bin-dirs/PATH
Currently, the `crucible-jvm` executable allows specifying the locations of Java classes through the `-j` flag (for JAR files) and the `-c` flag (for `.class` files). This has two issues:

1. For JDK 8 and earlier, this requires finding where your JDK installation's `rt.jar` file is and manually passing it to `crucible-jvm`. This is tedious.
2. This approach doesn't work at all for JDK 9 and later—see GaloisInc/saw-script#861.

Elsewhere, SAW aims to fix these issues with GaloisInc/saw-script#1030, which adds a `-b` flag which points to the location where the `java` executable lives. (Alternatively, if `-b` isn't set, SAW can also find `java` if it is located on the `PATH`.) With this information, one doesn't need to specify where `rt.jar` lives at all, as it can be inferred from `java -XshowSettings:properties`. Moreover, knowing where `java` is located is essential to getting JDK 9 and later to work.

I propose we adopt a similar `-b`/`PATH` scheme for `crucible-jvm`.

RyanGlScott:
This mirrors a similar recent change introduced to SAW in GaloisInc/saw-script#1030.

While I was in town, I updated the two `crucible-jvm`–related test suites to use the `PATH` instead of the `-j` flag. I can't exactly confirm via CI if these changes work (see #634), but they appear to do the right thing locally.

Fixes #633.


crucible-jvm/tool/Main.hs
```diff
@@ -129,6 +158,11 @@ cruxJVMConfig = Crux.Config
         $ Crux.ReqArg "TODO"
         $ \p opts ->
             Right $ opts { jarList = jarList opts ++ splitSearchPath p }
+      , Crux.Option ['b'] ["java-bin-dirs"]
+        "TODO"
```
atomb, line 162: It would be nice to have a real description for this option.

crucible-jvm/tool/Main.hs
```diff
@@ -129,6 +158,11 @@ cruxJVMConfig = Crux.Config
...
+        $ Crux.ReqArg "TODO"
```
atomb, line 163: Likewise here.

atomb (approved):
This looks good to me, though I think real descriptions for the new (and old!) options in `crucible-jvm` might be worthwhile. I hope that once #635 is merged, and this is updated to include that, the CI problems will go away.

crucible-jvm/tool/Main.hs
RyanGlScott: Fair enough. I copied over this `TODO` from the existing `TODO`s above, but it's easy enough to just fill them in. I've done so in a commit that I just pushed.

PR #638, opened 2021-02-03
Lang.JVM.Codebase: experimental support for loading JIMAGE files

RyanGlScott:
This allows `crucible-jvm` to deal with JDK 9 or later, which packages its standard library not in a JAR file, but in a JIMAGE file. Extracting `.class` files from JIMAGE files proves to be surprisingly tricky, and I've carefully documented the intricacies of doing so in `Note [Loading classes from JIMAGE files]` in `Lang.JVM.Codebase`. This is part of a fix for GaloisInc/saw-script#861.

### DIFF
--- a/crucible-jvm/crucible-jvm.cabal
+++ b/crucible-jvm/crucible-jvm.cabal
@@ -53,10 +53,13 @@ library
     base >= 4 && < 5,
     aig,
     array,
+    base16-bytestring >= 0.1 && < 1.1,
     bv-sized >= 1.0.0,
+    bytestring,
     containers,
     crucible,
-    directory,
+    cryptohash-sha1 >= 0.11 && < 0.12,
+    directory >= 1.2.5,
     extra >= 1.6.4,
     filepath,
     haskeline >= 0.7,
@@ -88,6 +91,7 @@ library
     Lang.Crucible.JVM.Overrides
     Lang.JVM.Codebase
     Lang.JVM.JavaTools
+    Lang.JVM.ProcessUtils
   other-modules:
 
   ghc-options: -Wall -Wcompat -Werror=incomplete-patterns -Werror=missing-methods -Werror=overlapping-patterns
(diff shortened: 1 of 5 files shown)


RyanGlScott:
I'm marking this as a draft for now since there appears to be a lurking bug that is uncovered when running `crucible-jvm`'s test suite with JDK 11:

```
$ make clean && ./test.sh 
rm -f *.class
javac -g Str.java
Compiled from "Str.java"
class Str {
  Str();
    Code:
       0: aload_0
       1: invokespecial #1                  // Method java/lang/Object."<init>":()V
       4: return

  public static void main(java.lang.String[]);
    Code:
       0: ldc           #2                  // String 1234567890
       2: astore_1
       3: new           #3                  // class java/lang/StringBuffer
       6: dup
       7: aload_1
       8: invokespecial #4                  // Method java/lang/StringBuffer."<init>":(Ljava/lang/String;)V
      11: astore_2
      12: aload_2
      13: iconst_2
      14: ldc           #5                  // String abcd
      16: invokevirtual #6                  // Method java/lang/StringBuffer.insert:(ILjava/lang/String;)Ljava/lang/StringBuffer;
      19: pop
      20: getstatic     #7                  // Field java/lang/System.out:Ljava/io/PrintStream;
      23: aload_2
      24: invokevirtual #8                  // Method java/io/PrintStream.println:(Ljava/lang/Object;)V
      27: return
}
[Crux] Checking "Str.java"
starting executeCrucibleJVM
[Crux] Index 10 is not a method reference.
  error, called at src/Language/JVM/Parser.hs:387:12 in jvm-parser-0.3.0.0-inplace:Language.JVM.Parser
```

Amazingly, I haven't come across this before in the SAW test cases that I've ran, so I'm glad I thought to try this. I still need to figure out why this happens, however.

RyanGlScott:
Per the discussion in https://github.com/GaloisInc/saw-script/issues/861#issuecomment-774283950, I've unmarked this as a draft. The `crucible-jvm` tool now prints out a warning if you are using JDK 9 or later indicating that support for modern JDKs is still experimental, and the warning includes a pointer to #641. This warning should serve as a template for SAW.

atomb (approved):
This looks great. Thanks for including lots of design rationale documentation!

PR #639, opened 2021-02-04
Demote nat

robdockins:
Fix crucible uses of the `SymNat` type now that it is no longer a `BaseType`


crucible/src/Lang/Crucible/Backend/Online.hs
```diff
@@ -458,7 +457,7 @@ withSolverProcess ::
   IO a {- ^ Default value to return if online features are disabled -} ->
   (SolverProcess scope solver -> IO a) ->
   IO a
-withSolverProcess = withSolverProcess' (\sym -> readIORef (B.sbStateManager sym))
+withSolverProcess = withSolverProcess' (\sym -> pure (B.sbUserState sym))
```
kquick, line 460: The `withSolverProcess'` is also used in crucible-saw (src/Lang/Crucible/Backend/SAWCore.hs, line ~1178), which didn't get updated yet (unless I missed it in this PR).

Also, since `withSolverProcess'` is only called from here and in crucible-saw, where it doesn't need to be in the IO monad either, should the argument just be changed to be pure (and so for example, here the value is just `B.sbUserState` instead of `(\sym -> ...)`?

crucible/src/Lang/Crucible/CFG/Expr.hs
```diff
@@ -983,16 +977,16 @@ data App (ext :: Type) (f :: CrucibleType -> Type) (tp :: CrucibleType) where
   -- If no such position exists, return a negative value.
   StringIndexOf :: !(f (StringType si))
                 -> !(f (StringType si))
-                -> !(f NatType)
+                -> !(f IntegerType)
```
kquick, line 980: In crucible/src/Lang/Crucible/Simulator/Evaluation.hs, should the `StringIsIndexOf` evaluation verify or add an assertion that the index argument is positive?  I don't know if the solver automatically does this check but for a concrete value it seems like we'd need something along the lines of:

```
k'' <- evalSub k
k' <- case asInt k'' of
        Nothing -> return k''  -- symbolic index
        Just n -> if n < 0
                  then addFailedAssertion sym $ GenericSimError "negative string index supplied to StringIndexOf"
                  else return k''
```

crucible/src/Lang/Crucible/CFG/Expr.hs
```diff
@@ -983,16 +977,16 @@ data App (ext :: Type) (f :: CrucibleType -> Type) (tp :: CrucibleType) where
...
 
   -- @stringSubstring s off len@ extracts the substring of @s@ starting at index @off@ and
   -- having length @len@.  The result of this operation is undefined if @off@ and @len@
   -- do not specify a valid substring of @s@; in particular, we must have @off+len <= length(s)@.
   StringSubstring :: !(StringInfoRepr si)
                   -> !(f (StringType si))
-                  -> !(f NatType)
-                  -> !(f NatType)
+                  -> !(f IntegerType)
+                  -> !(f IntegerType)
```
kquick, line 989: Same evaluation concern here for these two arguments as with the `StringIndexOf` argument above.

crux-llvm/test-data/golden/golden/vector-cmp.good
```diff
@@ -1,10 +1,2 @@
-let -- internal
-    v10 = floatFromBinary 0x40a00000:[32]
-    -- internal
-    v11 = floatFromBinary 0x42280000:[32]
-    -- internal
-    v50 = ite (floatLe v11 v10) 0x0:[1] 0x1:[1]
-    -- internal
-    v52 = bvZext 32 (bvAnd (ite (floatLe v10 v11) 0x1:[1] 0x0:[1]) v50)
- in bvOr 0x1:[32] (bvShl v52 0x1:[32])
+0x3:[32]
```
kquick, line 1: As I understand it, crucible/what4 is now able to concretely evaluate both branches of this test case as opposed to just the one case, which is good, so this is a valid test, but do we want to create a todo for making more complicated vector tests that will ensure solver vector verification (it now looks like all the vector tests are concrete results), or is that no longer a viable consideration with the what4 changes?

kquick (changes requested):
I think crucible-saw needs to be changed, and perhaps the evaluation of `StringIndexOf` and `StringSubString` needs some updates.  Other than that, the rest are just suggestions and this looks fine.

crucible/src/Lang/Crucible/Backend/Simple.hs
```diff
@@ -71,7 +70,7 @@ newSimpleBackend floatMode gen =
      return sym
 
 getAssumptionStack :: SimpleBackend t fs -> IO (AssumptionStack (B.BoolExpr t) AssumptionReason SimError)
-getAssumptionStack sym = sbAssumptionStack <$> readIORef (B.sbStateManager sym)
+getAssumptionStack sym = pure (sbAssumptionStack (B.sbUserState sym))
```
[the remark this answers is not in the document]
robdockins: Mostly, it changes all the call sites.

crucible/src/Lang/Crucible/Backend/Online.hs
robdockins: `crucible-saw` is now obsolete and is being allowed to bit-rot.  It's been removed from the `cabal.project`.

The `withSolverProcess'` stuff was some awkward pipe-fitting that I think is not needed anymore.  Maybe I should just get rid of it.

crucible/src/Lang/Crucible/CFG/Expr.hs
robdockins: This is a good point.  I'll check on what is happening here and decide what to do.

crux-llvm/test-data/golden/golden/vector-cmp.good
robdockins: I think the aspects being tested here are really just the LLVM translation of vector values.  If I remember correctly, the vector components are treated as scalars by the What4 layer.  I don't think we need a separate test.

crucible/src/Lang/Crucible/CFG/Expr.hs
robdockins: Looking into it a bit more closely, the SMTLib functions being modeled here are actually total and have a well-defined result even for out-of-range and negative values of these indices. As such, I think I'm going to leave the crucible definitions here as-is and open a What4 ticket about ensuring these functions are faithful in the out-of-range cases.

kquick (approved):

crucible/src/Lang/Crucible/CFG/Expr.hs
kquick: Might be worth adding a comment to that effect here.

PR #642, opened 2021-02-07
[crux-llvm] only generate tests for the current available clang version.

kquick:
Also updates strlen_test2.c output for pre-clang11 build verification.


kquick:
There's more consolidation and updating that can be done, but this returns the CI results to green.

PR #643, opened 2021-02-09
[crux-llvm] Switch golden testing config files from symlinks to plain files

kquick:


abakst (approved):
Looks good! Just a minor comment -- is it worth removing `default.config`? It's probably not useful anymore.

PR #644, opened 2021-02-09
[crux-llvm] Add CI_TEST_LEVEL to test and github action CI config.

kquick:
A CI_TEST_LEVEL of 0 (the default if unspecified) will run a set of
tests expected to complete in under a minute, suitable for a normal
development environment.

A CI_TEST_LEVEL of 1 (specified for github actions) will run the full
set of tests, even those that can take an extended period of time.

### DIFF
--- a/.github/workflows/crux-llvm-build.yml
+++ b/.github/workflows/crux-llvm-build.yml
@@ -25,6 +25,8 @@ jobs:
     runs-on: ${{ matrix.os }}
     needs: [outputs]
     continue-on-error: ${{ matrix.allow-failure }}
+    env:
+      CI_TEST_LEVEL: "1"
     strategy:
       fail-fast: false
       matrix:
--- a/crux-llvm/test/Test.hs
+++ b/crux-llvm/test/Test.hs
@@ -5,7 +5,7 @@
 module Main where
 
 import           Control.Exception ( SomeException, catches, try, Handler(..), IOException )
-import           Control.Monad ( unless )
+import           Control.Monad ( unless, when )
 import           Data.Bifunctor ( first )
 import qualified Data.ByteString.Lazy as BSIO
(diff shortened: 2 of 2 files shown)


robdockins (approved):
This overall idea seems OK to me, but I wonder if we're better off splitting out querying the CI test level into a separate package that we can depend on so that we can ensure we do it consistently in different packages.  I guess I'd prefer the test level to be a datatype instead of just a number too, with slightly more descriptive names.

kquick:
We had a discussion on the devops channel about some of your points, I'll try to summarize that here but happy to discuss in more detail:  
 * the CI_TEST_LEVEL is intended to be a simple and easy selector between developer convenience and CI thoroughness, and doesn't preclude proving more targeted or different control vectors.  The idea here is that if a developer adds a test that takes a long time but isn't necessarily a critical core verification, that this is a well-known selector to hide it behind (the plan is to use this across all packages, not just crucible).  [As a datapoint, this reduces the crux-llvm test time from about 150s at level 1 to about 40 s at level 0, so a savings of ~2 minutes of developer time each time it's run].
 *  Other vectors could be handled by other env vars or arguments as the need for them is exposed
 * An env var was chosen since it's something that is easily controlled by a user or the CI environment.  Specifically, a command-line argument is not used because different languages have different methods (and `cabal test` doesn't allow passing arguments to tests).

tl;dr This is a bit of an evolving space and may (will probably) be refined, but this is a starting spot.

Re a common library of code: also something reasonable if there's enough to extract.  For this PR, most of it is specific to the tests involved, so there's not much to extract, but definitely as we start doing more of this we will want consistency and that suggests extracting whatever emerges as the common parts and putting it... somewhere.  As you noted, we'd like this to be a fairly universal convention across all our packages.

PR #647, opened 2021-02-11
Improve formatting of crux tools help output

kquick:
Improves the formatting of `--help` information from crux tools like `crux-llvm`.

* Options descriptions are wrapped at terminal width and stay in their column
* Terminal width for output is auto-detected and defaults to 80 for non-terminals.
* Options are displayed alphabetically (by short option, then long option).
* Option values are consistently UPPERCASE to signify a user-supplied value position.


atomb (approved):
Thanks for doing this! Things like this make a big difference in my first impressions of new tools, so I like seeing us make improvements in this direction.

PR #649, opened 2021-02-15
Derive Eq, Ord instances for MemType, SymType, and others

langston-barrett:
This makes the API easier to work with, e.g. I can derive `Ord` for a datatype containing a `MemType` and put it in a `Set`.

The new, derived `Eq` instance change for `StructInfo` is backwards incompatible, but the old instance didn't distinguish between `StructInfo`s with different `DataLayout`s. Since there was no comment, I assumed that no code is *relying* on that unconventional choice, but might be worth a changelog note?

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
@@ -87,7 +87,7 @@ alignmentToExponent :: Alignment -> Natural
 alignmentToExponent (Alignment n) = fromIntegral n
 
 newtype AlignInfo = AT (Map Natural Alignment)
-  deriving (Eq)
+  deriving (Eq, Ord)
 
 -- | Make alignment info containing no alignments.
 emptyAlignInfo :: AlignInfo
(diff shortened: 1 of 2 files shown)


robdockins:
I believe the assumption is that the datalayout should always be equal, and the Eq instance is optimized to not compare them.

I'm actually not entirely sure why we are capturing it; it looks like it only ever flows into `siDropLastField`, which doesn't have a call site inside crucible, so I don't know what it's used for.

langston-barrett:
> I believe the assumption is that the datalayout should always be equal, and the Eq instance is optimized to not compare them.

Should I restore the old instance and write a similar one for `Ord`, then?

robdockins:
I'm actually wondering if we can just remove the DataLayout field from the structure altogether.

langston-barrett:
> I'm actually wondering if we can just remove the DataLayout field from the structure altogether.

Okay, I'll revise and remove these! Thanks for looking into it!

langston-barrett:
> I'm actually wondering if we can just remove the DataLayout field from the structure altogether.

Done! Pinging @robdockins for another round of review.

robdockins (approved):

PR #650, opened 2021-02-15
Add targetArch to crux config; crux-llvm-svcomp use normal compile/link

kquick:
The `crux-llvm-svcomp` tool had an internal compile/link process that duplicated the compilation/linking process in the `crux-llvm` core library (and repeated a hard-coded internal filename). This removes the special case compilation to use the standard compilation process for `crux-llvm-svcomp`, adding the ability to specify the `crux-llvm` target architecture for compilation.


robdockins (approved):
Seems like a good addition and simplification.

PR #652, opened 2021-02-22
Add PartialResultFrame, newtype FunctionBindings update

kquick:
This introduces a couple of changes that will be useful for later polyglot changes, but which can stand on their own.
* Adds a PartialResultFrame type alias for brevity
* Change FunctionBindings from a type alias to a newtype.


robdockins (approved):

robdockins:
Oops, actually, looks like the newtype change broke some stuff in `crucible-mir`.

PR #658, opened 2021-03-05
[crucible-llvm] Use general ext type parameter instead of "LLVM arch"

kquick:
This separates the extension identification from the architecture (and corresponding pointer width) identification.

The identification of  the architecture and pointer width is now explicitly provided where needed to express that context or information in the crucible-llvm and crux-llvm code base without obtaining it from the `ext` parameter.

A significant portion of this PR is adjustment of type signatures to express `ext` as either `ext` or `LLVM` instead of `(LLVM arch)`, or to switch from, add, or remove `arch` or `ext` and corresponding `arch` constraints.  Significant changes include:
* The extension type no longer includes the architecture encoding
* ~~statement execution and evaluation in MemModel.hs are now provided an explicit `ArchRepr arch` argument.~~
*  In LLVM/Translation/Expr, the `ScalarView` and associated functions now thread a `Proxy#` representative of the `arch` parameter through the datastructure and callbacks to provide the necessary ptrwidth information.  This impacts other portions of the Translation (like Instruction.hs) but is not expected to have much impact outside of the Translation elements.
* Crux LLVM override initialiation is passed a `Proxy# arch` where needed (as provided from the `ModuleTranslation`), and where not needed, the `ArchOk arch` constraint is removed.
* Extends many of the `LLVMStmt` constructors to add the `HasPtrWidth` constraint to the constructor, carrying the constraint in the `LLVMStmt` object instead of the `ext` or the code path.

The net effect is to remove the `arch` parameterization from a large number of places where it isn't needed or used.

This is a precursor to changes for polyglot support of multiple active extensions.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -347,12 +347,11 @@ instance IsSymInterface sym => IntrinsicClass sym "LLVM_memory" where
 -- | Top-level evaluation function for LLVM extension statements.
 --   LLVM extension statements are used to implement the memory model operations.
 llvmStatementExec ::
-  (HasPtrWidth (ArchWidth arch), Partial.HasLLVMAnn sym, ?memOpts :: MemOptions) =>
-  ArchRepr arch ->
+  (Partial.HasLLVMAnn sym, ?memOpts :: MemOptions) =>
   EvalStmtFunc p sym LLVM
-llvmStatementExec arch stmt cst =
+llvmStatementExec stmt cst =
   let sym = cst^.stateSymInterface
-   in stateSolverProof cst (runStateT (evalStmt sym arch stmt) cst)
+   in stateSolverProof cst (runStateT (evalStmt sym stmt) cst)
 
 type EvalM p sym ext rtp blocks ret args a =
   StateT (CrucibleState p sym ext rtp blocks ret args) IO a
@@ -361,13 +360,12 @@ type EvalM p sym ext rtp blocks ret args a =
 --   The semantics are explicitly organized as a state transformer monad
 --   that modifies the global state of the simulator; this captures the
 --   memory accessing effects of these statements.
-evalStmt :: forall p sym ext arch rtp blocks ret args tp.
+evalStmt :: forall p sym ext rtp blocks ret args tp.
   (IsSymInterface sym, Partial.HasLLVMAnn sym, HasCallStack, ?memOpts :: MemOptions) =>
   sym ->
-  ArchRepr arch ->
   LLVMStmt (RegEntry sym) tp ->
   EvalM p sym ext rtp blocks ret args (RegValue sym tp)
-evalStmt sym _arch = eval
+evalStmt sym = eval
  where
   getMem :: GlobalVar Mem ->
             EvalM p sym ext rtp blocks ret args (MemImpl sym)
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -348,24 +348,26 @@ instance IsSymInterface sym => IntrinsicClass sym "LLVM_memory" where
 --   LLVM extension statements are used to implement the memory model operations.
 llvmStatementExec ::
   (HasPtrWidth (ArchWidth arch), Partial.HasLLVMAnn sym, ?memOpts :: MemOptions) =>
-  EvalStmtFunc p sym (LLVM arch)
-llvmStatementExec stmt cst =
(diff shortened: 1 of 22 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
```diff
@@ -447,21 +449,21 @@ evalGEP instr (GEPResult _lanes finalMemType gep0) = finish =<< go gep0
          traverse (\(x,i) -> calcGEP_array mt' x i) (Seq.zip xs idxs)
 
 
-calcGEP_array :: forall wptr arch s ret.
+calcGEP_array :: forall wptr s arch ret.
```
travitch, line 452: Is this to fix some type application ordering?

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -348,24 +348,26 @@ instance IsSymInterface sym => IntrinsicClass sym "LLVM_memory" where
...
-  (IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym, HasCallStack, ?memOpts :: MemOptions) =>
+evalStmt :: forall p sym ext arch rtp blocks ret args tp.
+  (IsSymInterface sym, Partial.HasLLVMAnn sym, HasCallStack, ?memOpts :: MemOptions) =>
   sym ->
-  LLVMStmt wptr (RegEntry sym) tp ->
+  ArchRepr arch ->
+  LLVMStmt (RegEntry sym) tp ->
   EvalM p sym ext rtp blocks ret args (RegValue sym tp)
-evalStmt sym = eval
+evalStmt sym _arch = eval
```
travitch, line 370: I think this is the most interesting part that I want to think through. The `arch` type parameter doesn't seem to have an explicit connection to anything else here. Is it possible to pass the wrong arch repr for the extension in use? What would happen if we did?

travitch (commented):
Small comments, except for the one in MemModel.hs

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
kquick: This change is not required and superfluous: I had initially added an `ext` here and subsequently removed it, but in the process of determining where to add it I had re-ordered this to match the order of introduction.  I can revert this change if desired.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
kquick: This was unnecessary and is removed in 00e47a7.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
```diff
@@ -119,23 +125,23 @@ instance Show (LLVMExpr s arch) where
...
+  = Scalar proxy# tp xs
 asScalar (ZeroExpr llvmtp)
   = let ?err = error
-     in zeroExpand llvmtp $ \tpr ex -> Scalar tpr ex
+     in zeroExpand proxy# llvmtp $ \archProxy tpr ex -> Scalar archProxy tpr ex
 asScalar (UndefExpr llvmtp)
   = let ?err = error
-     in undefExpand llvmtp $ \tpr ex -> Scalar tpr ex
+     in undefExpand proxy# llvmtp $ \archProxy tpr ex -> Scalar archProxy tpr ex
 asScalar _ = NotScalar
```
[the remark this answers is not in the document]
kquick: Added commentary in 08cd2d5

travitch (approved):
I approve, contingent on @robdockins

robdockins:
I haven't taken a close look at this yet, but I'm really not a fan of existentializing the `arch` type parameter from the statement type.  It no longer encodes the invariant that there is a single known architecture width.  That means we'll have to do dynamic tests somewhere else.

kquick:
Tristan raised a similar concern.  Some of the uses were vestigial from previous attempts that were no longer needed so I removed them.  I was able to adjust the others to use `Proxy# arch` instead of `ArchRepr arch` to clarify that it's only threading the arch type where necessary and that there's no runtime impact imposed by these changes.

I've updated the PR description to match these changes.  Please take a look when you get a chance @rdockins; it's a large patch but mostly removes arch at the type level where it isn't needed, and hopefully this addresses your concern above.

robdockins (approved):
Something odd is going on with the crux-mir test, but it seems like it is unrelated.

kquick:
I think the crux-mir test is timing out and just needs to be run with a longer timeout (tasty timer, solver timeout, ... somewhere).  This PR doesn't touch anything that would affect crux-mir.  

Thanks for the review!

PR #659, opened 2021-03-09
Compositional crux-mir support

spernsteiner:
This is the crucible/crux-mir component of GaloisInc/saw-script#1117.  This branch adds various infrastructure needed to support the main compositional crux-mir implementation in the saw-script repo.

Specifically, this branch adds new Crucible intrinsic types `MethodSpec` and `MethodSpecBuilder`, and adds functions for manipulating them to the `crucible` Rust library.  However, the intrinsics are defined as existential types, using typeclasses to provide the operations for each type, and no implementations of those typeclasses or overrides for the Rust API functions are included in `crux-mir`.  Instead, `saw-script/crux-mir-comp` defines concrete implementation types and extends `crux-mir` with the necessary overrides, so that these implementations can use `saw-script`'s `MethodSpec` and related SAW types.

This branch also adds some additional functions for manipulating `MirReference`s, which are used in `crux-mir-comp`'s handling of memory.

### DIFF
--- a/crux-mir/src/Mir/Intrinsics.hs
+++ b/crux-mir/src/Mir/Intrinsics.hs
@@ -1551,7 +1551,7 @@ mirRef_indexAndLenIO sym s (MirReferenceMux ref) = do
 
 mirRef_indexAndLenSim :: IsSymInterface sym =>
     MirReferenceMux sym tp ->
-    OverrideSim (Model sym) sym MIR rtp args ret
+    OverrideSim p sym MIR rtp args ret
         (PartExpr (Pred sym) (RegValue sym UsizeType, RegValue sym UsizeType))
 mirRef_indexAndLenSim ref = do
     sym <- getSymInterface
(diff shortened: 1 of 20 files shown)


crux-mir/src/Mir/Intrinsics.hs
```diff
@@ -1709,3 +1932,104 @@ getSlicePtr e = getStruct i1of2 e
 
 getSliceLen :: Expr MIR s (MirSlice tp) -> Expr MIR s UsizeType
 getSliceLen e = getStruct i2of2 e
+
+
+--------------------------------------------------------------------------------
+-- ** MethodSpec and MethodSpecBuilder
```
atomb, line 1938: My hope is that someday we can eventually, generalize the stuff from here down, move it to the `crux` package, and use it for other languages, too.

atomb (approved):
Looks great. I added one comment about eventually generalizing some of this stuff, but that's not something we need to do now.

PR #662, opened 2021-03-12
Update crucible-wasm for FnBindings change

kquick:


robdockins (approved):
Are we building crucible-wasm as part of CI?  It would be nice to do so to keep it from bitrotting.

kquick:
I'd like to get crucible-wasm under CI; there's some haskell-wasm support to be resolved first.

PR #665, opened 2021-03-12
[crucible] Remove unused functions.

kquick:


travitch (commented):
We can just delete this - it isn't especially deep and it is still in the history if we want it

robdockins (approved):
I'll defer to @travitch on this one.

travitch (approved):

PR #668, opened 2021-03-13
Add check of cabal config file version to github actions.

kquick:


robdockins:
This seems OK to me, but what is the concern being addressed?

kquick:
This verifies that the version of cabal required by the .cabal file is compatible with the cabal library for that GHC version.  The cabal library is a part of the GHC base package (https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history).

This wasn't previously caught because it's possible to build the Cabal-install tool on a more recent GHC version (thus getting support for a later cabal library version via the tool), even though the compilation is being performed by an older GHC compiler.  To actually ensure that the cabal file is compatible with the GHC + cabal library for that GHC version, this change avoids using the Cabal-install tool and uses the standard `Setup` build, which ensures the cabal library associated with the GHC version is used (which will detect any version compatibility issue).

robdockins (approved):

PR #675, opened 2021-03-15
crux-mir: implement overflow checks for Mul/Div/Rem

### LINKED ISSUE #612: implement overflow checks for `Mul`, `Div`, and `Rem`
Currently, our translation of `CheckedBinOp` returns a correct overflow flag for `Add`, `Sub`, and shifts, but always returns `false` (no overflow) for `Mul`, `Div`, and `Rem`.  These are a little more complicated to implement, but should be added for completeness.

spernsteiner:
We previously had overflow checks for Add and Sub, so `crux-mir` could detect panics due to arithmetic overflow in those operations.  This branch adds support for the remaining arithmetic operations: Mul, Div, and Rem.  `crux-mir` now supports overflow checks everywhere that `rustc` itself does.

The overflow rules are:
* `a * b` overflows if the truncated NxN -> N bit multiply produces a different result than the fully precise NxN -> 2N multiply.  Specifically, we check whether `ext(a) * ext(b) != ext(a * b)`, where `ext` is zero-extension to 2N bits for unsigned types and sign-extension for signed types.
* `a / b` and `a % b` "overflow" if `b == 0`, or (if `a` and `b` are signed) if `a` is the most negative possible value (such as `-128_i8`) and `b == -1`.

Fixes #612

### DIFF
--- a/crux-mir/src/Mir/Trans.hs
+++ b/crux-mir/src/Mir/Trans.hs
@@ -337,7 +337,7 @@ transBinOp bop op1 op2 = do
     fst <$> evalBinOp bop mat me1 me2
 
 -- Evaluate a binop, returning both the result and an overflow flag.
-evalBinOp :: M.BinOp -> Maybe M.ArithType -> MirExp s -> MirExp s ->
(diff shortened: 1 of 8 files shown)


atomb (approved):
Looks great. Thanks for including thorough testing, as always!

PR #677, opened 2021-03-16
[crucible-llvm] Use panic instead of error in buildIdentMap failure case.

kquick:
I think this represents and internal error and not something that would normally fail due to input or user error.


kquick:
Looks like more spurious crux-mir failures on MacOS; this PR only touches crucible-llvm so those failures can be ignored.

robdockins (commented):
I believe you are correct that this case should represent panic.

While you're in here, it might be nice to do something about the `Right` pattern match on line 220 also.  In that case, I think that is a failure condition that could possibly be caused by legitimate user input.

kquick:
I traced the other failure back to points of origin and it all seems to be based on the LLVM Module object obtained by translating the input file, so there doesn't seem to be any user error involved.  I've therefore updated line 220 with a panic and removed the irrefutable pattern match.

PR #680, opened 2021-03-17
[crucible-llvm] Use common handleType to get the handle type for a Fu…

kquick:
…nctionHandle


kquick:
Again, ignore the crux-mir test failure; changes here wouldn't affect mir.

robdockins (approved):

PR #681, opened 2021-03-17
Fix misspelling in error message.

kquick:


RyanGlScott (approved):
No tpyo shall go unnoticed.

Issue #627, opened 2021-01-29
crux-mir performance regression


spernsteiner:
`crux-mir`'s `symb_eval/scalar/test1.rs` test case has gotten slower: on my machine, it takes around 2m30s to run in commit 9bb4b78b0412698046837d8057b99adf6ab81459 (when `crux-mir` was merged into this repository), and now takes around 3m30s as of 09532645133d39cda983bca67c2dbb91d05e4373 (current `master`).  `git bisect` blames ced74d4a5e4d7eb01fdb9f3d6ada382a365fc22f "Add a configuration option that controls if the online backend methods for maintaining a connection with a solver are enabled" from #570.  On ced74d4a5e4d7eb01fdb9f3d6ada382a365fc22f the test runs in about 3m20s, while on 12a06e3d24d25b2016b7491d7c282dcfd634380b (its immediate ancestor) it runs in 2m25s.

I've been timing the test using the following command:
```sh
time cabal v2-run -- crux-mir --assert-false-on-error -s z3 test/symb_eval/scalar/test1.rs
```
(After building first, so the `time` doesn't include the time spent building)

Note that the two commits in question (ced74d4a5e4d7eb01fdb9f3d6ada382a365fc22f and 12a06e3d24d25b2016b7491d7c282dcfd634380b) don't build as-is, due to changes on Hackage.  Both require changing `what4.cabal`'s bound on the `versions` dependency from `versions >= 3.5.2` to `versions >= 3.5.2 && < 4.0` (otherwise you'll get a type error about `Versions.VChunk` and `Versions.VUnit`).  Furthermore, ced74d4a5e4d7eb01fdb9f3d6ada382a365fc22f requires the `crux-mir` build fix from d18fd5074433dad46c9917b159b8b9aaf6d8da24 - I applied it by running `git checkout d18fd50 -- crux-mir` in the top-level `crucible` directory.

robdockins:
Wow, that's pretty surprising.  When you do the run, is it shutting down and restarting solver instances or something like that?

PR #684, opened 2021-03-18
crux-mir: increase test timeout as a temporary CI fix

spernsteiner:
The performance regression in #627 has been causing crux-mir's `scalar/test1.rs` test to exceed its timeout on the macOS CI runners.  This branch increases the timeout (180s -> 600s) so CI will stop breaking.  Once #627 is fixed, we should revert this.


kquick (approved):
Looks good.  I'd love to see this merged and see if that eliminates the misleading failures we've been seeing.

PR #685, opened 2021-03-18
crucible-go: Remove unused dependencies

RyanGlScott:
This was originally motivated by discovering that the `lens-simple` dependency (which does not compile on GHC 9.0) goes unused in `crucible-go`. I then decided to leverage GHC's `-Wunused-packages` flag to discover other unused dependencies. This patch removes all such dependencies.

### DIFF
--- a/crucible-go/crucible-go.cabal
+++ b/crucible-go/crucible-go.cabal
@@ -17,27 +17,10 @@ executable crux-go
 
   build-depends:
     base >= 4 && < 5,
-    aig,
-    ansi-wl-pprint,
-    array,
-    containers,
     crucible,
-    directory,
-    filepath,
-    haskeline >= 0.7,
-    lens,
-    mtl >= 2.1,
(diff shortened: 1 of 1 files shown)


crucible-go/crucible-go.cabal
```diff
@@ -17,27 +17,10 @@ executable crux-go
...
-    mtl >= 2.1,
     parameterized-utils >= 1.0 && < 2.2,
-    pretty >= 1.1,
-    split >= 0.2,
-    text,
-    transformers >= 0.3,
-    transformers-compat,
-    vector >= 0.7,
     golang,
-    what4 >= 0.4,
```
kquick, line 38: Removing these from this section drops some version constraints that should be moved to the area below where they still apply (at least `vector` and `what4`, but probably should check all others.

kquick (commented):
Thanks for doing this.  Once we propagate the constraints I think this should be good to go.

PR #687, opened 2021-03-18
Wasm ci updates

kquick:
Mostly this is just CI stuff, with a few code tweaks for older GHC (8.6.5) for compatibility with the rest of crucible.

### DIFF
--- a/.github/workflows/crux-llvm-build.yml
+++ b/.github/workflows/crux-llvm-build.yml
@@ -1,4 +1,10 @@
 name: crux-llvm
+
+# Note: also tests crucible-wasm builds, which aren't technically
+# crux-llvm, but since crucible-wasm depends on crucible and
+# crucible-llvm, it was easier to add it here than to creati a
+# separate workflow duplicating builds.
+
 on:
   push:
     branches: [master, "release-**"]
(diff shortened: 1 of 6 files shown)


.github/workflows/crux-llvm-build.yml
```diff
@@ -1,4 +1,10 @@
 name: crux-llvm
+
+# Note: also tests crucible-wasm builds, which aren't technically
+# crux-llvm, but since crucible-wasm depends on crucible and
+# crucible-llvm, it was easier to add it here than to creati a
```
RyanGlScott, line 5:
```suggestion
# crucible-llvm, it was easier to add it here than to create a
```

.github/workflows/crux-llvm-build.yml
```diff
@@ -186,9 +192,16 @@ jobs:
...
       - shell: bash
+        name: Build crux-llvm SVComp tool
         run: .github/ci.sh build exe:crux-llvm-svcomp
         if: runner.os != 'Windows'
 
+      - shell: bash
+        name: Build crucible wasm support library
+        run: |
+          (cd dependencies/haskell-wasm; nix-shell -p hpack --run hpack)
+          ./github/ci.sh build crucible-wasm
```
RyanGlScott, line 203:
```suggestion
          ./.github/ci.sh build crucible-wasm
```

I think?

.github/workflows/crux-llvm-build.yml
kquick: oops

.github/workflows/crux-llvm-build.yml
kquick: Actually, I just need to remove the slash, but good catch.  The moral of the story is one shouldn't code while distracted.

PR #673, opened 2021-03-15
uc-crux-llvm: A tool for under-constrained symbolic execution

langston-barrett:
[Rendered README](https://github.com/langston-barrett/crucible/tree/lb/uc-crux-llvm/uc-crux-llvm)

PR #690, opened 2021-03-19
crucible-llvm: Add a 'gas' parameter to strlen override

### LINKED ISSUE #689: crucible-llvm: Loop bounds for strlen
The current `strlen` override [advertises](https://github.com/GaloisInc/crucible/blob/4fa48a2417bd03dc92e54ab9970046bc7234b7ec/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1019-L1024) that:

> this function will not terminate until it eventually reaches a concete null-terminator or a load error.

This makes it difficult to apply Crucible to code where the size and/or content of a string could be fully symbolic. For instance, `crux-llvm` appears to spin infinitely on this program:
```c
#include <stdlib.h>
#include <string.h>

#include "crucible.h"

char* mkstr() {
  int sz = crucible_int32_t("sz");
  char* x = malloc(sz);
  for(int i = 0; i < sz; i++) {
    x[i] = crucible_int8_t("x");
  }
  return x;
}

int main() {
  char *str = mkstr();
  size_t sz = strlen(str);
  check(sz >= 100);
}
```
For standard loops, one can use the bounded execution feature, but this doesn't apply inside overrides like this. Perhaps we could leverage memory model flags (like those for lax arithmetic) to configure a bound for `strlen` and similar overrides?

langston-barrett:
Fixes #689. If we merge #673, I can add a testcase in that tool by adding an `strlen` override that passes a non-`Nothing` value, and using a non-terminating program like the one from #689 .

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -1020,16 +1020,18 @@ isAllocatedAlignedPointer sym w alignment mutability ptr size mem =
 --
 --   The pointer to read from must be concrete and nonnull.  The contents
 --   of the string may be symbolic; HOWEVER, this function will not terminate
---   until it eventually reaches a concete null-terminator or a load error.
+--   until it eventually reaches a concete null-terminator or a load error,
+--   unless a maximum bound is provided via the @gas@ parameter.
 strLen :: forall sym wptr.
   (IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym) =>
   sym ->
+  Maybe Word {-^ bound on loop/string length -} ->
   MemImpl sym      {- ^ memory to read from        -} ->
   LLVMPtr sym wptr {- ^ pointer to string value    -} ->
   IO (SymBV sym wptr)
-strLen sym mem = go (BV.zero PtrWidth) (truePred sym)
+strLen sym gas mem = go (BV.zero PtrWidth) gas (truePred sym)
   where
-  go !n cond p =
+  go !n !g cond p =
     loadRaw sym mem p (bitvectorType 1) noAlignment >>= \case
       Partial.Err pe ->
         do ast <- impliesPred sym cond pe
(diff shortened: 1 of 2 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1020,16 +1020,18 @@ isAllocatedAlignedPointer sym w alignment mutability ptr size mem =
 --
 --   The pointer to read from must be concrete and nonnull.  The contents
 --   of the string may be symbolic; HOWEVER, this function will not terminate
---   until it eventually reaches a concete null-terminator or a load error.
+--   until it eventually reaches a concete null-terminator or a load error,
+--   unless a maximum bound is provided via the @gas@ parameter.
 strLen :: forall sym wptr.
   (IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym) =>
   sym ->
+  Maybe Word {-^ bound on loop/string length -} ->
```
langston-barrett, line 1028: We may want to consider making this non-optional (if rather high, by default). IMO it would be best if Crux and SAW never entered infinite loops without at least printing a warning.

travitch, line 1028: I'll be curious to hear what Rob thinks, but I like this idea in general. I'd say I'd like a newtype for `Gas` if we do this

robdockins:
I'm OK with the basic idea here.  Maybe we should populate the gas parameter iff the user has specified loop iteration bounds, which would mimic the behavior of the naive loop. I don't think it should be set by default. Related: I wonder if we should find a way to do path-sat checking in the loop if that is requested in the options.

In fact, I'm starting to wonder if we should just literally implement the naive C loop so all the usual options would just work. I had hoped (and still hope someday) to implement special support for null-terminated strings in the memory model, but in the meantime having this be a primitive just seems to lead to issues.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
robdockins: No strong opinion about a `Gas` newtype.

robdockins (approved):

travitch:
In the past we've talked about just compiling bug chunks of e.g., dietlibc or musl into bitcode and incorporating them into the target bitcode like we do with libc++. Does this `strlen` override provide anything over that approach?

langston-barrett:
> Maybe we should populate the gas parameter iff the user has specified loop iteration bounds

Yeah, this is what I intend to do in `uc-crux-llvm`.

> In fact, I'm starting to wonder if we should just literally implement the naive C loop so all the usual options would just work. I had hoped (and still hope someday) to implement special support for null-terminated strings in the memory model, but in the meantime having this be a primitive just seems to lead to issues.

This would also make it a lot easier to implement missing string manipulation overrides like `strcmp`, `strncmp`, `strdup`, `strchr`, `strrchr`...

robdockins:
> In the past we've talked about just compiling bug chunks of e.g., dietlibc or musl into bitcode and incorporating them into the target bitcode like we do with libc++. Does this `strlen` override provide anything over that approach?

The main advantage of the current method is that we don't have to produce and provide the necessary bitcode files, which is definitely a PITA.  It might be more efficient as a built-in, but I'm really not sure.  If we do eventually add special support for null-terminated strings, we'll need to make the string functions overrides anyway.

So, I'm a bit torn about it.  If we did decide to implement more of libc via musl or whatever, I think it would be easier to justify bundling in the necessary libc variants.  As it is... hard to say.

travitch:
If we end up at a better place with distributing libc++ bitcode, that might tip the balance towards providing more implementations that way (of course deferring to overrides with special handling for NUL-terminated strings when we get there).

Perhaps if we can just expose a few primitives for supporting NUL-terminated strings in the memory model, we can either provide C overrides that use those special features or implement them in the Crucible concrete syntax (one day extended to support language extensions).

langston-barrett:
Closing for now since `uc-crux-llvm` didn't end up needing this and the IR&D is ending.

PR #697, opened 2021-03-23
uc-crux-llvm: Add unsound overrides for getenv, gethostname

langston-barrett:
These overrides aren't suitable for verification, but they can help achieve higher coverage for finding bugs with `uc-crux-llvm`. The behavior without them is to just bail unconditionally when calls to these functions (and many others) are encountered.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Config.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Config.hs
@@ -42,8 +42,7 @@ data UCCruxLLVMOptions = UCCruxLLVMOptions
     exploreParallel :: Bool,
     entryPoints :: [String],
     skipFunctions :: [String],
-    verbosity :: Int,
-    unsoundOverrides :: Bool
+    verbosity :: Int
   }
 
 -- | Crucible will infinitely loop when it encounters unbounded program loops,
@@ -95,14 +94,6 @@ skipDoc = "List of functions to skip during exploration"
 verbDoc :: Text
 verbDoc = "Verbosity of logging. (0: minimal, 1: informational, 2: debug)"
 
-unsoundOverridesDoc :: Text
-unsoundOverridesDoc =
-  Text.unwords
-    [ "Apply unsound (underapproximate) overrides for these functions:",
-      "gethostname, and getenv. This can lead to more code coverage, but any",
-      "safety claims about functions that call these ones might not hold."
-    ]
-
 ucCruxLLVMConfig :: IO (Crux.Config UCCruxLLVMOptions)
 ucCruxLLVMConfig = do
   llvmOpts <- llvmCruxConfig
@@ -118,8 +109,7 @@ ucCruxLLVMConfig = do
             <*> Crux.section "explore-parallel" Crux.yesOrNoSpec False exploreParallelDoc
             <*> Crux.section "entry-points" (Crux.listSpec Crux.stringSpec) [] entryPointsDoc
             <*> Crux.section "skip-functions" (Crux.listSpec Crux.stringSpec) [] skipDoc
-            <*> Crux.section "verbosity" Crux.numSpec 0 verbDoc
-            <*> Crux.section "unsound-overrides" Crux.yesOrNoSpec False unsoundOverridesDoc,
+            <*> Crux.section "verbosity" Crux.numSpec 0 verbDoc,
         Crux.cfgEnv =
           map
             ( \envDescr ->
@@ -212,13 +202,6 @@ ucCruxLLVMConfig = do
                      Crux.parsePosNum
                        "LEVEL"
                        $ \v opts ->
-                         opts {verbosity = v},
-                 Crux.Option
-                   []
-                   ["unsound-overrides"]
-                   (Text.unpack unsoundOverridesDoc)
-                   $ Crux.NoArg $
-                     \opts ->
-                       Right opts {unsoundOverrides = True}
+                         opts {verbosity = v}
                ]
       }
--- a/uc-crux-llvm/src/UCCrux/LLVM/Config.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Config.hs
@@ -42,7 +42,8 @@ data UCCruxLLVMOptions = UCCruxLLVMOptions
     exploreParallel :: Bool,
     entryPoints :: [String],
     skipFunctions :: [String],
-    verbosity :: Int
+    verbosity :: Int,
+    unsoundOverrides :: Bool
   }
 
 -- | Crucible will infinitely loop when it encounters unbounded program loops,
@@ -94,6 +95,14 @@ skipDoc = "List of functions to skip during exploration"
 verbDoc :: Text
 verbDoc = "Verbosity of logging. (0: minimal, 1: informational, 2: debug)"
 
+unsoundOverridesDoc :: Text
+unsoundOverridesDoc =
+  Text.unwords
+    [ "Apply unsound (underapproximate) overrides for these functions:",
+      "gethostname, and getenv. This can lead to more code coverage, but any",
+      "safety claims about functions that call these ones might not hold."
+    ]
+
 ucCruxLLVMConfig :: IO (Crux.Config UCCruxLLVMOptions)
 ucCruxLLVMConfig = do
   llvmOpts <- llvmCruxConfig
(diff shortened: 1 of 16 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Config.hs
```diff
@@ -181,6 +191,18 @@ ucCruxLLVMConfig = do
...
+                   ["skip-functions"]
+                   "List of functions to skip during exploration"
+                   $ Crux.ReqArg "FUN" $
+                     \v opts -> Right opts {skipFunctions = v : skipFunctions opts},
+                 Crux.Option
+                   []
+                   ["unsound-overrides"]
+                   (Text.unpack unsoundOverridesDoc)
+                   $ Crux.NoArg $
+                     \opts -> Right opts {unsoundOverrides = True}
```
travitch, line 206: Should we default to True? It seems like there is little downside (in the abstract)

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
```diff
@@ -0,0 +1,184 @@
...
+        -- TODO Check for name size
+        -- Otherwise, return a canned name
+        ( What4.truePred sym,
+          Override.modifyGlobal mvar $ \mem ->
+            liftIO $
+              do
+                let val = LLVMMem.LLVMValString hostname
+                let ty = LLVMMem.llvmValStorableType val
+                mem1 <- LLVMMem.storeRaw sym mem ptr ty noAlignment val
+                bv0' <- What4.bvLit sym (knownNat @32) (BV.mkBV (knownNat @32) 0)
```
travitch, line 149: Is it possible to annotate the resulting value as unsound so that we can factor that into reporting?

travitch: Is it possible to learn constraints from observing how the synthesized names are used that might influence what we generate?

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
```diff
@@ -0,0 +1,184 @@
...
+                return (bv0', mem1),
+          Nothing
+        )
+      ]
+
+-- | Super unsound. Returns the same string value every time, but in a fresh
+-- allocation for each call. A more sound approximation would be to require
+-- that the variable name be concrete and then have a map from names to
+-- allocations holding values.
```
travitch, line 158: I think this is reasonable for now. In the future, it would be nice if we could just pull this functionality from DanM's work on symbolic IO https://github.com/GaloisInc/crucible/tree/feature/symio/crucible-symio.  That said, if we could feed information back into the values generated, that would become *very* interesting (and only suitable for uc-crux)

travitch (commented):
A few preliminary comments

uc-crux-llvm/src/UCCrux/LLVM/Config.hs
langston-barrett: I was worried because the reporting claims things like "this function is always safe", which wouldn't always be true if said function (transitively) called these overrides, but combined with tracking the use of these overrides so that we can report precisely when they were used, I agree that there is little downside. I'll just go ahead and remove this configuration option entirely.

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
langston-barrett: > Is it possible to annotate the resulting value as unsound so that we can factor that into reporting?

Yes, and we should do that!

> Is it possible to learn constraints from observing how the synthesized names are used that might influence what we generate?

Maybe! What would the goal be? Perhaps more branch coverage?

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
langston-barrett: > That said, if we could feed information back into the values generated

This is definitely a possibility, though as above I'm not exactly sure what we'd want to discover about the program and feed back in to this override.

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
travitch: Yes, the goal would be more branch coverage. I'm not sure that we need to do it right now, but it would be good to think about what that looks like to make sure we don't put ourselves in a bad spot for later.

I was imagining that you might annotate the returned string values and then inspect the path conditions produced by symbolic execution. If you see that there is a branch that depends on one of the annotated terms, we could inspect the condition to figure out what inputs would allow us to reach more paths.  I guess there are a few questions with that:
1. Does annotation prevent constant propagation (and e.g., simplification of branch conditions)
2. Do we want it to or not?
3. Can we improve branch coverage by returning strings of concrete length but symbolic contents? Would doing that be enough? If we did that, would we then be able to see the branching in the path condition to explore alternative values?

langston-barrett: > I'm not sure that we need to do it right now

I think there are a few other things that would lead to more coverage more easily in the short term, but I think this would be a great thing to circle back to. Here are some items that I would guess would have a bigger immediate impact:

- The ability to skip undefined functions including un-overriden libc functions and any library functions not included in the bitcode
- Additional underapproximate overrides
- Additional *sound* overrides (e.g. `strchr`)
- Support for generating/reasoning about global variables

All of those are just guesses, though.

> but it would be good to think about what that looks like to make sure we don't put ourselves in a bad spot for later.

Yeah, I think we'd be in a reasonably good place to support this: Registering overrides happens in each iteration of the "bugfinding loop" (the one depicted in the flowchart in the README), and at each new iteration the information learned from previous iterations (constraints) is available. So we could collect whatever data we like during simulation and feed it back to the overrides in the next iteration pretty easily.

> Does annotation prevent constant propagation (and e.g., simplification of branch conditions)

TBH I don't actually know the mechanism of annotations, so I can't answer this one.

> Do we want it to or not?

This is even further outside my understanding :)

> Can we improve branch coverage by returning strings of concrete length but symbolic contents? Would doing that be enough? If we did that, would we then be able to see the branching in the path condition to explore alternative values?

Seems likely, and that would be a pretty darn easy first step.

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
langston-barrett: See more discussion above.

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
langston-barrett: > Is it possible to annotate the resulting value as unsound so that we can factor that into reporting?

Done!

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
```diff
@@ -0,0 +1,202 @@
...
+  GlobalVar Mem ->
+  RegEntry sym (LLVMPointerType wptr) ->
+  RegEntry sym (BVType wptr) ->
+  OverrideSim p sym ext r args ret (RegValue sym (BVType 32))
+callGetHostName _proxy sym mvar (regValue -> ptr) (regValue -> len) =
+  do
+    let hostname = "hostname"
+    let lenLt bv = liftIO (What4.bvSlt sym len =<< What4.bvLit sym ?ptrWidth bv)
+    lenNeg <- lenLt (BV.mkBV ?ptrWidth 0)
+    -- TODO(lb): Panic if ?ptrWidth is like 2 or something crazy
```
travitch, line 144: `?ptrWidth` is at least 16 because `ArchOk` resolves to `HasPtrWidth`

travitch (approved):

uc-crux-llvm/src/UCCrux/LLVM/Overrides.hs
langston-barrett: Rad, I changed this from a TODO that a NOTE that a check isn't currently necessary. Thanks!

langston-barrett:
@kquick The builds for this PR [are failing](https://github.com/GaloisInc/crucible/pull/697/checks?check_run_id=2196654817) with some Nix-related errors. Do these seem related to/will these be taken care of by https://github.com/GaloisInc/crucible/pull/687/commits/f132a67477a3c07fbd7314bab30836ea1aef357a / https://github.com/GaloisInc/crucible/pull/687/commits/17cf37ab49c7fb158b4f19eb6e585b84d4f0cb9e ?

kquick:
Yes, sorry, that's the change that's needed.  I just added that change to master as well.

langston-barrett:
Looks like rebasing fixed this. Thanks @kquick!

PR #698, opened 2021-03-23
uc-crux-llvm: Configurable hard timeout and parallelism for exploration mode

langston-barrett:
These features help it scale to bigger codebases. The hard timeout is helpful for things like the strlen override (#689), or just generally being robust to nontermination/slowness bugs in Crucible/solvers.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Config.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Config.hs
@@ -35,8 +35,10 @@ data UCCruxLLVMOptions = UCCruxLLVMOptions
   { ucLLVMOptions :: LLVMOptions,
     doExplore :: Bool,
     reExplore :: Bool,
-    entryPoints :: [String],
     exploreBudget :: Int,
+    exploreTimeout :: Int,
+    exploreParallel :: Bool,
+    entryPoints :: [String],
     skipFunctions :: [String],
     verbosity :: Int
   }
@@ -85,15 +87,25 @@ ucCruxLLVMConfig = do
               False
               "Re-explore functions that have already been explored (i.e., have logs)"
             <*> Crux.section
-              "entry-points"
-              (Crux.listSpec Crux.stringSpec)
-              []
-              "Comma-separated list of functions to examine."
-            <*> Crux.section
               "explore-budget"
               Crux.numSpec
               8
               "Budget for exploration mode"
+            <*> Crux.section
+              "explore-timeout"
+              Crux.numSpec
+              5
+              "Hard timeout for exploration of a single function (seconds)"
+            <*> Crux.section
+              "explore-parallel"
+              Crux.yesOrNoSpec
+              False
+              "Explore different functions in parallel"
+            <*> Crux.section
+              "entry-points"
+              (Crux.listSpec Crux.stringSpec)
+              []
+              "Comma-separated list of functions to examine."
             <*> Crux.section
               "skip-functions"
               (Crux.listSpec Crux.stringSpec)
@@ -157,15 +169,6 @@ ucCruxLLVMConfig = do
                          opts {verbosity = v},
                  Crux.Option
                    []
-                   ["explore-budget"]
-                   "Budget for exploration mode"
-                   $ Crux.ReqArg
-                     "INT"
-                     $ Crux.parsePosNum
-                       "INT"
-                       $ \v opts -> opts {exploreBudget = v},
-                 Crux.Option
-                   []
                    ["skip-functions"]
                    "List of functions to skip during exploration"
                    $ Crux.ReqArg "FUN" $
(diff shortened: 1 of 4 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Config.hs
```diff
@@ -181,6 +184,30 @@ ucCruxLLVMConfig = do
                    ["re-explore"]
                    "Re-explore functions that have already been explored (i.e., have logs)"
                    $ Crux.NoArg $
-                     \opts -> Right opts {reExplore = True}
+                     \opts -> Right opts {reExplore = True},
+                 Crux.Option
+                   []
+                   ["explore-budget"]
+                   "Budget for exploration mode"
```
travitch, line 191: What are the units for the budget?

uc-crux-llvm/src/UCCrux/LLVM/Config.hs
```diff
@@ -84,16 +86,26 @@ ucCruxLLVMConfig = do
...
+              "Hard timeout for exploration of a single function (seconds)"
+            <*> Crux.section
+              "explore-parallel"
+              Crux.yesOrNoSpec
+              False
+              "Explore different functions in parallel"
+            <*> Crux.section
+              "entry-points"
+              (Crux.listSpec Crux.stringSpec)
+              []
```
travitch, line 107: Is there an implicit `main` default?

uc-crux-llvm/src/UCCrux/LLVM/Run/Explore.hs
```diff
@@ -79,30 +133,13 @@ explore appCtx modCtx cruxOpts ucOpts halloc =
...
-                        ( mempty
-                            { unimplementedFreq = Map.singleton (panicComponent unin) 1
-                            }
-                        )
-              else do
-                (appCtx ^. log) Med $ "Skipping already-explored function " <> Text.pack func
-                pure mempty
+      if Config.exploreParallel ucOpts
+        then
+          mapConcurrently
```
travitch, line 141: This could be amusingly dangerous. `mapConcurrently` immediately spawns a thread for each item in the collection. I think you might want something like https://hackage.haskell.org/package/scheduler to bound the number of parallel threads.

uc-crux-llvm/src/UCCrux/LLVM/Config.hs
langston-barrett: Functions to explore!

uc-crux-llvm/src/UCCrux/LLVM/Config.hs
langston-barrett: No, it will suggest you use `--entry-points` or `--explore` if neither is provided.

uc-crux-llvm/src/UCCrux/LLVM/Run/Explore.hs
langston-barrett: I've tried this out with a budget as high as 2000 on large projects like `libpng`, `liblzma`, etc. and not had any problems. This is probably in part due to the fact that we don't get very deep into a lot of functions, and so it finishes very quickly for a large percentage of the code.

Anyway, `scheduler` would be an easy addition and could improve performance as well as being less dangerous so I'll integrate it! Thanks for the pointer.

langston-barrett: Hmm, using the "scheduler" package results in a ~2x slowdown on `libpng` from `mapConcurrently`. What do you think, Tristan? Worth it to avoid problems in the future? Or just keep an eye on it and maybe revisit as the code evolves?

langston-barrett: This is how I'm using it BTW: https://github.com/GaloisInc/crucible/pull/698/commits/92b4b7624f33d4bd77ac1d6fcbff70e1de7630a1

uc-crux-llvm/src/UCCrux/LLVM/Config.hs
langston-barrett: Clarified in the documentation.

uc-crux-llvm/src/UCCrux/LLVM/Run/Explore.hs
travitch: Do you have 16 cores to support that? Passing 0 might be better... it would then just use all available capabilities (i.e., whatever was passed via `+RTS -N`)

langston-barrett: I do! I also tried with 0/just with the `Par` pattern.

langston-barrett: I didn't have the right combination of threads in my Cabal file, I wasn't linking the executable with `-threaded`. With that, `scheduler` goes down to 45s, and `mapConcurrently` goes down to *10s*. However, the results from `mapConcurrently` aren't what I'd expect - it seems to be exploring fewer functions that I ask for. It doesn't seem worth time time it would take to debug given that `scheduler` is also the morally correct approach, so I'll call this done for now.

travitch (approved):

PR #699, opened 2021-03-24
Modifications to eliminate partial fields and incomplete patterns.

kquick:
Individual commit comments describe the various changes, but the overall is to avoid using record fields with multi-constructor ADT types and eliminate partial and/or irrefutable pattern matches.

The first commit synchronizes crucible with PR 111 in the what4 repository (the identically named branch).

### DIFF
--- a/crucible/src/Lang/Crucible/Analysis/Postdom.hs
+++ b/crucible/src/Lang/Crucible/Analysis/Postdom.hs
@@ -89,7 +89,9 @@ postdomMap m breakpointIds = r
           [ (pd_id, mapMaybe f l)
           | (pd,_:l)  <- G.dom g 0
           , pd > 0
-          , let Just pd_id = Map.lookup pd idMap
+          -- Get the post dominator blkID, using a total version of:
+          --    let Just pd_id = Map.lookup pd idMap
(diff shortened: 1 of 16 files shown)


crucible/src/Lang/Crucible/Analysis/Postdom.hs
```diff
@@ -89,7 +89,9 @@ postdomMap m breakpointIds = r
           [ (pd_id, mapMaybe f l)
           | (pd,_:l)  <- G.dom g 0
           , pd > 0
-          , let Just pd_id = Map.lookup pd idMap
```
robdockins, line 92: I think this is better replaced by a guard `Just pd_id <- Map.lookup pd idMap`.  I think those can appear in comprehensions like this (maybe with an extension enabled).

robdockins (changes requested):
Looks OK, other than the issue mentioned in `Postdom`.  I'd rather not change from a map lookup to a linear list-scan just to squash a warning.

crucible/src/Lang/Crucible/Analysis/Postdom.hs
kquick: Good catch.  I think I'd been thinking of `List.lookup`, but I agree that a linear scan of map elements is the wrong direction here.

PR #701, opened 2021-03-26
uc-crux-llvm: Run hlint, ormolu in CI (and fix warnings)

langston-barrett:


kquick:
I'm not a big fan of adding linting and code format requirements to the CI process.  I think these are great to run as part of your personal development, but they tend to express personal preferences without a corresponding gain in reducing bugs.  They also tend to require a number of exceptions (as noted in your hlint file) and limit developer expressiveness which creates additional cognitive dissonance.

Relative to my general position on this described above, I would advise against having these in CI but I won't block it if you would prefer to have them.  I would prefer that they be setup as CI warnings but not actual failures (https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstepscontinue-on-error) however.

I do support the additional compilation warnings and associated code updates, and would recommend adding `-Wincomplete-uni-patterns` to the list.

uc-crux-llvm/uc-crux-llvm.cabal
```diff
@@ -12,9 +12,15 @@ Synopsis:      A bugfinding tool for C programs, using under-constrained symboli
 
 common bldflags
   ghc-options: -Wall
+               -Werror
```
travitch, line 15: I'm not super keen on blanket `-Werror` (I think the specific ones below seem fine, though I don't recognize missing home modules), which often makes things really exciting around the time when we need to add support for a new ghc version (or experiment with one to avoid compiler bugs)

langston-barrett:
> I think these are great to run as part of your personal development, but they tend to express personal preferences without a corresponding gain in reducing bugs.

I think you're describing a *mode of use* of linting tools, and one that I'm painfully familiar with :-). In general, my default is usually to disable *all* warnings from each tool, and to selectively enable them based on a consensus with the team I'm working on. Generally, we can arive at a small set of rules that have low/no "false positives", e.g. the rule included here about never using `error`. However, since I'm the only person working on this code at the moment I've elected to include a few more warnings than I might in a team setting. I'd definitely be happy to switch to the "whitelist" approach now, or at any later time.

As for formatting tools, I do hold the opinion that they're useful, and that they become useless if not enforced in CI. I don't necessarily want to hash out the reasoning here (I'm happy to discuss via chat or in a meeting), but I'll also say that I'm happy to drop this check as soon as anyone else works on this if we decide it's not appropriate.

> I do support the additional compilation warnings and associated code updates, and would recommend adding -Wincomplete-uni-patterns to the list.

Will do, thanks for the tip!

uc-crux-llvm/uc-crux-llvm.cabal
langston-barrett: It sounds like the concern is that `-Wall` combined with `-Werror` can lead to new build failures if a warning is added to `-Wall` between GHC versions - is that right? Would it address that concern if I inlined the set of warnings that `-Wall` consists of?

travitch: That is one possibility (at least picking the ones that we are actually worried about). Another is to do something like what macaw does and have a separate cabal.project file for ci https://github.com/GaloisInc/macaw/blob/master/.github/workflows/ci.yaml#L21

langston-barrett: Ah, that seems like a good approach! I'll add a `cabal.project` file like that to be used in CI.

langston-barrett: Okay, since I'm lazy I just removed `-Werror` for now, might revisit in the future and add the `cabal.project` file, though that would probably be best done over the whole repo.

PR #704, opened 2021-03-27
crux-llvm code re-arrangement

kquick:
This PR generally provides some code re-arrangement to split out some useful functionality into independent functions and make other entry points available for external re-use (e.g. Polyglot will make use of these).

The changes are generally described as:
* [crucible-llvm] `mkMemVar` now takes a name for the memory region instead of using a hard-coded internal name
* The `GlobalVar Mem` obtained from `mkMemVar` is now passed explicitly where needed instead of bundling it into llvm module translation and needing to extract it from the `LLVMContext`
  * [crucible-llvm] `translateModule` is now passed the `memVar` instead of allocating it internally.
  * [crucible-llvm] `setupSimCtxt` is passed the `memVar`, not the `llvmCtxt`.
  * [crux-llvm] `llvmMetrics` is renamed to `memMetrics` and is passed the `memVar`, not the `llvmCtxt`.
* [crux-llvm] the `simulateLLVMFile` sequence is re-arranged, and subsections are split out into new functions (allowing external re-use of these functions for other setup arrangements):
   1. Memory and handle initialization happens first, independent of the LLVM file Module handling
   2. a new `setupFileSim` function provides the simulation's `InitialState` setup after invoking `prepLLVMModule` (see next item).
   3. a new `prepLLVMModule` provides LLVM file parsing and translation into Crucible IR, returning a `PreppedLLVM` structure with the resulting components needed by other initialization functions.
   4. the default annotation failure explanation is extracted to the new `explainFailure` function (and the new `Explainer` type alias is created for referencing the explanation function for crucible.

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Simulate.hs
+++ b/crux-llvm/src/Crux/LLVM/Simulate.hs
@@ -126,7 +126,7 @@ registerFunctions llvm_module mtrans =
 
 simulateLLVMFile :: FilePath -> LLVMOptions -> Crux.SimulatorCallback
 simulateLLVMFile llvm_file llvmOpts = do
-  Crux.SimulatorCallback $ \sym _maybeOnline -> do
+  Crux.SimulatorCallback $ \sym maybeOnline -> do
     halloc <- newHandleAllocator
     bbMapRef <- newIORef (Map.empty :: LLVMAnnMap sym)
     let ?recordLLVMAnnotation = \an bb -> modifyIORef bbMapRef (Map.insert an bb)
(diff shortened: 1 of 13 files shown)


crux-llvm/src/Crux/LLVM/Simulate.hs
```diff
@@ -125,55 +125,78 @@ registerFunctions llvm_module mtrans =
...
+               checkFun (entryPoint llvmOpts) (cfgMap trans)
+
+
+
+
+data PreppedLLVM sym = PreppedLLVM { p_llvmMod :: LLVM.Module
+                                   , p_someTrans :: Some ModuleTranslation
+                                   , p_memVar :: GlobalVar Mem
+                                   , p_mem :: MemImpl sym
+                                   }
```
langston-barrett, line 171: The majority of record accessors in Crucible use `camelCase`, I'd recommend keeping that scheme as consistent as possible and removing the underscores in the accessors here.

langston-barrett (approved):
The changes to `uc-crux-llvm` LGTM.

crux-llvm/src/Crux/LLVM/Simulate.hs
kquick: Thanks!  I had initially used  underscores as a signal to myself that this was a tentative change and missed cleaning these up when I decided to go forward on this implementation.

crux-llvm/src/Crux/LLVM/Simulate.hs
```diff
@@ -192,12 +215,12 @@ checkFun nm mp =
...
-             . LLVMContext arch
+memMetrics :: forall p sym ext
+             . GlobalVar Mem
             -> Map.Map Text (Metric p sym ext)
-llvmMetrics llvmCtxt = Map.fromList [ ("LLVM.allocs", allocs)
-                                    , ("LLVM.writes", writes)
-                                    ]
+memMetrics memVar = Map.fromList [ ("LLVM.Mem.allocs", allocs)
+                                 , ("LLVM.Mem.writes", writes)
+                                 ]
```
travitch, line 223: I'm curious if anything depended on these. I can't think of anything, but I have some vague memory that we had hooked up a JS profiler frontend to this at some point. I'm not sure if there was any special dependency of this data or other data.

travitch: Ah, it was this https://github.com/GaloisInc/sympro-ui

travitch: And yes, those fields are referenced by name https://github.com/GaloisInc/sympro-ui/blob/master/ts/profile.js#L12

kquick: I did not know about sympro-ui, but these name changes were superfluous, so I've reverted them and added a note about the dependency.

travitch (approved):
Thanks, that addressed my concerns.

I'm curious what other clients will be affected. When you merge this, can you provide a brief migration guide saying what has changed?

PR #707, opened 2021-03-29
uc-crux-llvm: Some support for arrays of unbounded size

langston-barrett:
The included test program causes a panic before this commit, now it mostly works.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/FullType/MemType.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/FullType/MemType.hs
@@ -46,6 +46,7 @@ toMemType =
     FTNonVoidFuncPtrRepr retRepr argsRepr -> funType (Just retRepr) argsRepr
     FTOpaquePtrRepr _ident -> PtrType OpaqueType
     FTArrayRepr natRepr fullTypeRepr -> ArrayType (natValue natRepr) (toMemType fullTypeRepr)
+    FTUnboundedArrayRepr fullTypeRepr -> ArrayType 0 (toMemType fullTypeRepr)
     FTStructRepr structInfo _ -> StructType structInfo
   where
     funType :: Maybe (FullTypeRepr m ft) -> Ctx.Assignment (FullTypeRepr m) argTypes -> MemType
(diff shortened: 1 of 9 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Shape.hs
```diff
@@ -266,6 +275,7 @@ ppShapeSeekError =
 
 -- | Modify the 'Shape' at a given 'Cursor'
 modifyA ::
+  forall f m a tag inTy atTy.
```
travitch, line 278: Does anything introduced by this change actually require the type variables to be brought into scope?

uc-crux-llvm/src/UCCrux/LLVM/FullType/MemType.hs
```diff
@@ -46,6 +46,7 @@ toMemType =
     FTNonVoidFuncPtrRepr retRepr argsRepr -> funType (Just retRepr) argsRepr
     FTOpaquePtrRepr _ident -> PtrType OpaqueType
     FTArrayRepr natRepr fullTypeRepr -> ArrayType (natValue natRepr) (toMemType fullTypeRepr)
+    FTUnboundedArrayRepr fullTypeRepr -> ArrayType 0 (toMemType fullTypeRepr)
```
travitch, line 49: Is it always going to be a zero here? Is this just the only representation in Crucible-llvm memory types?

travitch (approved):
This looks fine to me. It would be helpful to note in the merge comment that:
1. The size of arrays in type specs is now `Maybe Nat`, with `Nothing` indicating unbounded
2. I believe only zero-sized unbounded arrays are currently supported (the rest is future work)?

uc-crux-llvm/src/UCCrux/LLVM/FullType/MemType.hs
langston-barrett: I'm not totally sure I understand the second part of your question, but yeah, this is how `MemType` represents these unbounded/unsized array types.

uc-crux-llvm/src/UCCrux/LLVM/Shape.hs
langston-barrett: Whoops, that's leftover from an earlier draft! Will remove.

langston-barrett:
Hmm, the CI failures seem unrelated:
```
+ [[ -x bin/yices ]]
+ command -v yices
Error: Process completed with exit code 1.
```

RyanGlScott:
I've experienced the Ubuntu failure as well in a `saw-script` PR. It appears that the site which hosts Yices bindists (https://yices.csl.sri.com/) is down, and trying to run something like `curl -o yices.tar.gz -sl "https://yices.csl.sri.com/releases/2.6.2/yices-2.6.2-x86_64-pc-linux-gnu-static-gmp.tar.gz"` will simply error out.

PR #708, opened 2021-03-30
uc-crux-llvm: Improve type translation

langston-barrett:
This is based on #707, which is only held up due to some CI difficulties.

Previously, the declaration for each function was separately translated into `FullType`s, now all declarations are done at once. This has a few benefits:

- Efficiency: there's no duplicated work when exploring multiple functions
- Parsimony: The ModuleTypes abstraction applies best at the module level
- Functionality: The big feature introduced here is `findFun`/`CFGWithTypes`: the `FullType`s for a declaration can be looked up in the `ModuleContext`. One spot this will be useful is in generating return values from skipped functions.

Additionally, this patch introduces support for variable-arity function types. This was necessary because the default is now to translate *all* function types in a module, which would fail on any module with a varargs function if these weren't supported.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Context/Module.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Context/Module.hs
@@ -6,53 +6,179 @@ License      : BSD3
 Maintainer   : Langston Barrett <langston@galois.com>
 Stability    : provisional
 -}
+{-# LANGUAGE GADTs #-}
 {-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE ScopedTypeVariables #-}
 
 module UCCrux.LLVM.Context.Module
   ( ModuleContext,
+    SomeModuleContext (..),
     moduleFilePath,
     makeModuleContext,
     llvmModule,
+    moduleTypes,
     moduleTranslation,
     dataLayout,
     withTypeContext,
+
+    -- * Looking up CFGs
+    CFGWithTypes (..),
+    findFun,
   )
 where
 
 {- ORMOLU_DISABLE -}
-import           Control.Lens ((^.), Simple, Getter, Lens, lens, to)
+import           Control.Lens ((^.), Simple, Getter, Lens, lens, to, at)
+import           Control.Monad (when)
+import           Data.Proxy (Proxy(Proxy))
+import           Data.Type.Equality ((:~:)(Refl), testEquality)
 
-import           Text.LLVM (Module)
+import           Text.LLVM (Module, Symbol(Symbol))
+
+import qualified Data.Parameterized.Context as Ctx
+import           Data.Parameterized.Some (Some(Some))
+
+import qualified Lang.Crucible.CFG.Core as Crucible
+import qualified Lang.Crucible.Types as CrucibleTypes hiding ((::>))
 
 import           Lang.Crucible.LLVM.DataLayout (DataLayout)
+import           Lang.Crucible.LLVM.Extension (LLVM)
 import           Lang.Crucible.LLVM.Translation (ModuleTranslation)
 import qualified Lang.Crucible.LLVM.Translation as LLVMTrans
 import           Lang.Crucible.LLVM.TypeContext (TypeContext, llvmDataLayout)
+
+import           Crux.LLVM.Overrides (ArchOk)
+
+import           UCCrux.LLVM.Errors.Panic (panic)
+import           UCCrux.LLVM.Errors.Unimplemented (unimplemented)
+import qualified UCCrux.LLVM.Errors.Unimplemented as Unimplemented
+import           UCCrux.LLVM.FullType.CrucibleType (DeclTypes, TranslatedTypes(..), TypeTranslationError, FunctionTypes(..), MatchingAssign(..), translateModuleDefines, testCompatibility, lookupDeclTypes)
+import           UCCrux.LLVM.FullType.Type (FullTypeRepr, ModuleTypes, MapToCrucibleType)
+import           UCCrux.LLVM.FullType.ReturnType (ReturnType(..), ReturnTypeToCrucibleType)
+import           UCCrux.LLVM.FullType.VarArgs (VarArgsRepr, varArgsReprToBool)
 {- ORMOLU_ENABLE -}
 
-data ModuleContext arch = ModuleContext
+-- | The @m@ type parameter represents a specific LLVM module
+data ModuleContext m arch = ModuleContext
   { _moduleFilePath :: FilePath,
     _llvmModule :: Module,
+    _moduleTypes :: ModuleTypes m,
+    _declTypes :: DeclTypes m arch,
     _moduleTranslation :: ModuleTranslation arch
   }
 
-moduleFilePath :: Simple Lens (ModuleContext arch) FilePath
+moduleFilePath :: Simple Lens (ModuleContext m arch) FilePath
 moduleFilePath = lens _moduleFilePath (\s v -> s {_moduleFilePath = v})
 
-llvmModule :: Simple Lens (ModuleContext arch) Module
+llvmModule :: Simple Lens (ModuleContext m arch) Module
 llvmModule = lens _llvmModule (\s v -> s {_llvmModule = v})
 
-moduleTranslation :: Simple Lens (ModuleContext arch) (ModuleTranslation arch)
+moduleTypes :: Simple Lens (ModuleContext m arch) (ModuleTypes m)
+moduleTypes = lens _moduleTypes (\s v -> s {_moduleTypes = v})
+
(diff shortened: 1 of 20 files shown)


langston-barrett:
Cancelled CI builds because I'm planning on merging #713 next, will rebase this onto the default branch once that's in.

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
```diff
@@ -81,55 +99,155 @@ toCrucibleType proxy =
...
+            ( Ctx.sizeInt (Ctx.size crucibleTypes)
+                - if isVarArgs
+                  then 1
+                  else 0
+            )
+            ( \idx ->
+                do
+                  Some fullTypeRepr <-
+                    withExceptT FullTypeTranslation $
+                      toFullTypeM (memTypes !! idx)
```
travitch, line 217: Is this indexing always safe? Are the sizes of `memTypes` and `crucibleTypes` related?

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
```diff
@@ -81,55 +99,155 @@ toCrucibleType proxy =
...
+        let crucibleTypes = Crucible.cfgArgTypes cfg
+        let memTypes = fdArgTypes liftedDecl
+        let isVarArgs = fdVarArgs liftedDecl
+        unless
+          ( let numCrucibleTypes = Ctx.sizeInt (Ctx.size crucibleTypes)
+             in length memTypes == numCrucibleTypes
+                  || ( isVarArgs
+                         && length memTypes + 1 == numCrucibleTypes
+                     )
+          )
```
travitch, line 204: Could we name this condition? It would make it easier to visually see that this can lead to an error

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
```diff
@@ -81,55 +99,155 @@ toCrucibleType proxy =
...
-assignmentToFullType proxy crucibleTypes memTypes =
+  L.Module ->
+  ModuleTranslation arch ->
+  Either TypeTranslationError (TranslatedTypes arch)
+translateModuleDefines llvmModule trans =
   case makeModuleTypes ?lc of
-    Some moduleTypes ->
+    Some initialModuleTypes ->
+      let Identity (maybeResult, modTypes) =
+            runStateT
```
travitch, line 169: If this is fixed to identity, would it be cleaner to just use `runState`?

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
```diff
@@ -81,55 +99,155 @@ toCrucibleType proxy =
...
+                  pure $ Some fullTypeRepr
             )
-            moduleTypes
-        case testCompatibilityAssign proxy fullTypes crucibleTypes of
-          Just Refl -> Right (SomeAssign fullTypes moduleTypes' Refl)
+        retType <-
+          case fdRetType liftedDecl of
+            Just retType ->
+              Just <$> withExceptT FullTypeTranslation (toFullTypeM retType)
+            Nothing -> pure Nothing
```
travitch, line 224: This might just be `traverse`?

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
```diff
@@ -81,55 +99,155 @@ toCrucibleType proxy =
...
+          Just Refl ->
+            pure
+              ( L.decName decl,
+                FunctionTypes
+                  { ftArgTypes = MatchingAssign fullTypes crucibleTypes',
+                    ftRetType = retType,
+                    ftIsVarArgs = isVarArgs
+                  }
+              )
           Nothing -> panic "Impossible" []
```
travitch, line 251: Could we say why this is impossible, in case it turns out to not be impossible?

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
```diff
@@ -81,55 +99,155 @@ toCrucibleType proxy =
...
+                  panic
+                    "translateModuleDefines"
+                    ["varargs function with no arguments"]
+                Ctx.AssignExtend rest vaRepr ->
+                  case testEquality vaRepr LLVMTrans.varArgsRepr of
+                    Just Refl -> Some rest
+                    Nothing ->
+                      panic
+                        "translateModuleDefines"
+                        ["varargs function with no varargs repr"]
```
travitch, line 240: Is this something we could meaningfully pull out into a function? It might not be reusable, but it might be easier to read

uc-crux-llvm/src/UCCrux/LLVM/Context/Module.hs
```diff
@@ -6,53 +6,171 @@ License      : BSD3
...
+      unimplemented "findFun" Unimplemented.VarArgsFunction
+
+    case testEquality (Crucible.cfgArgTypes cfg) argCTys of
+      Nothing -> panic "findFunc" ["Mismatched argument types"]
+      Just Refl ->
+        Just $
+          case Crucible.cfgReturnType cfg of
+            CrucibleTypes.UnitRepr ->
+              case retTy of
+                Just (Some _) -> panic "findFun" ["Extra return type"]
```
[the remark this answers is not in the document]
langston-barrett: Yeah, I'll expand on this a bit.

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
langston-barrett: Yeah, they're asserted to be related in the `unless` above. I'll add a comment to that effect.

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
langston-barrett: Yeah, will do!

uc-crux-llvm/src/UCCrux/LLVM/Main.hs
```diff
@@ -25,6 +25,7 @@ module UCCrux.LLVM.Main
     loopOnFunctions,
     translateLLVMModule,
     translateFile,
+    SomeModuleContext' (..),
```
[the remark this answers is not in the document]
langston-barrett: Because `SomeModuleContext` is already taken: https://github.com/GaloisInc/crucible/pull/708/files#diff-b78ee7aee50f1bf752da31af534521e16bc7a272aa4dbecbfa89205e9cd14e9aR125

It wasn't necessary before because `ModuleContext` only had one type parameter, but now that it has two there are four variants of which ones get existentially closed over...

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
langston-barrett: Indeed it is!

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
langston-barrett: Yeah, for sure! Just quickly here: `toCrucibleType . toFullTypeM` should be `id`.

uc-crux-llvm/src/UCCrux/LLVM/FullType/CrucibleType.hs
langston-barrett: Sure thing!

uc-crux-llvm/src/UCCrux/LLVM/FullType/MemType.hs
```diff
@@ -42,20 +43,22 @@ toMemType =
...
+    FTNonVoidFuncPtrRepr varArgs retRepr argsRepr ->
+      funType (Just retRepr) argsRepr (varArgsReprToBool varArgs)
     FTOpaquePtrRepr _ident -> PtrType OpaqueType
     FTArrayRepr natRepr fullTypeRepr -> ArrayType (natValue natRepr) (toMemType fullTypeRepr)
     FTUnboundedArrayRepr fullTypeRepr -> ArrayType 0 (toMemType fullTypeRepr)
     FTStructRepr structInfo _ -> StructType structInfo
   where
-    funType :: Maybe (FullTypeRepr m ft) -> Ctx.Assignment (FullTypeRepr m) argTypes -> MemType
-    funType maybeRetRepr argsRepr =
+    funType :: Maybe (FullTypeRepr m ft) -> Ctx.Assignment (FullTypeRepr m) argTypes -> Bool -> MemType
```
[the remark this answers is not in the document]
langston-barrett: I also replaced a few `Bool`s elsewhere with `Some VarArgsRepr`.

travitch (approved):

PR #714, opened 2021-03-31
Allow CRUCIBLE clang_flags in source to affect crux-llvm compilation.

kquick:
The source file can embed a comment of the form `/* CRUCIBLE clang_flags: .... */` (or in C++ single-line comment syntax) and the ellipsized portion will be added to the flags pass to `clang` for compilation.

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Compile.hs
+++ b/crux-llvm/src/Crux/LLVM/Compile.hs
@@ -1,9 +1,11 @@
 {-# LANGUAGE DataKinds #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
+
 module Crux.LLVM.Compile where
 
 import           Control.Exception ( SomeException(..), try, displayException )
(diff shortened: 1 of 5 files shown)


kquick:
ignore the macos build failures: these are unrelated and resolved by changes already on master

crux-llvm/src/Crux/LLVM/Compile.hs
```diff
@@ -163,6 +168,38 @@ genBitCodeToFile finalBCFileName files cruxOpts llvmOpts copySrc = do
   return finalBCFile
 
 
+-- | A C source file being compiled and evaluated by crux can contain
+-- zero or more lines matching the following:
+--
+-- >   /* CRUCIBLE clang_flags: {FLAGS} */
+-- >   // CRUCIBLE clang_flags: {FLAGS}
```
atomb, line 175: I've seen confusion from people I've shown Crux to about the difference between Crucible and Crux. I suspect this might be slightly clearer if it started with CRUX instead of CRUCIBLE.

There's part of me that wants `clang_flags` to be `CLANG_FLAGS` to be more similar to Make variables.

atomb (approved):
This looks good to me. My one comment is really a bike shedding question. The feature is definitely useful, and the implementation looks good to me.

crux-llvm/src/Crux/LLVM/Compile.hs
kquick: I wavered initially and went with CRUCIBLE because that's the same namespace as the `#include <crucible.h>` extensions, but I actually updated it to support both "CRUX" and "CRUCIBLE".

I also made the "clang_flags" portion be case-insensitive, so if you want CLANG_FLAGS you can have them. :-)

PR #713, opened 2021-03-30
uc-crux-llvm: Detect uninitialized stack reads

langston-barrett:

PR #715, opened 2021-03-31
uc-crux-llvm: Detect uninitialized heap reads

langston-barrett:
This is based on top of #713 to avoid merge conflicts, will rebase when that one is merged.

### DIFF
--- a/uc-crux-llvm/README.md
+++ b/uc-crux-llvm/README.md
@@ -279,6 +279,7 @@ Uncertain results:
       - [ ] `free` called on non-argument pointer with non-zero offset
       - [ ] Write of `const` memory
       - [ ] Illegal (un)signed wrap when both operands are concrete
+      - [x] Uninitialized stack reads
       - [x] Concretely null pointer dereference
     - [ ] Missing preconditions:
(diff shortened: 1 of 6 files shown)


travitch (approved):
I'm not a huge fan of the `mallocLocation` string, but I'm not sure that something more invasive is worthwhile at the moment. My understanding is that those allocations only occur during environment setup and not execution - is that correct?

langston-barrett:
> I'm not a huge fan of the `mallocLocation` string, but I'm not sure that something more invasive is worthwhile at the moment. 

My thoughts exactly!

> My understanding is that those allocations only occur during environment setup and not execution - is that correct?

Yeah, that's precisely what is being distinguished in `Classify` by comparing to `mallocLocation`.

langston-barrett:
The one CI test failure is

>         readCreateProcess: cargo "run" "--manifest-path" "report-coverage/Cargo.toml" "--quiet" "--" "--no-color" "test/coverage/out/coverage/3a1fbbbh::crux_test[0]/report_data.js" (exit -9): failed

in `crux-mir`. I haven't seen this before but I can't imagine that it's related to the changes here.

PR #718, opened 2021-04-01
uc-crux-llvm: Manufacture return values for skipped functions

langston-barrett:
Based on #708, that PR should be merged first.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Classify/Types.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Classify/Types.hs
@@ -214,6 +214,17 @@ ppUnfixable =
 data Unfixed
   = UnfixedArgPtrOffsetArg
   | UnfixedFunctionPtrInArg
+  | -- The following would be addressed by applying "missing precondition"
+    -- heuristics to return values from skipped functions
+    UnfixedRetReadUninitialized
+  | UnfixedRetWriteUnmapped
+  | UnfixedRetCall
+  | UnfixedRetMemset
+  | UnfixedRetPtrFree
+  | UnfixedRetPtrAddOffset
+  | UnfixedRetReadBadAlignment
+  | UnfixedRetWriteBadAlignment
+  | UnfixedRetDivRemByZero
   deriving (Eq, Ord, Show)
 
 ppUnfixed :: Unfixed -> Text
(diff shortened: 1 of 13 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Stats.hs
```diff
@@ -40,7 +40,7 @@ data Stats = Stats
   { missingAnnotation :: !Word,
     symbolicallyFailedAssert :: !Word,
     timeouts :: !Word,
-    truePositiveFreq :: Map TruePositiveTag Word,
+    truePositiveFreq :: Map Text Word,
```
travitch, line 43: Why switch to `Text` as the key type here?

uc-crux-llvm/src/UCCrux/LLVM/Cursor.hs
```diff
@@ -185,12 +185,17 @@ ppCursor top =
...
+--
+-- * an argument,
+-- * a global variable, or
+-- * the manufactured return value from a \"skipped\" function
 --
 -- For documentation of the type parameters, see the comment on 'Cursor'.
 data Selector m (argTypes :: Ctx (FullType m)) inTy atTy
   = SelectArgument !(Ctx.Index argTypes inTy) (Cursor m inTy atTy)
   | SelectGlobal !L.Symbol (Cursor m inTy atTy)
+  | SelectReturn !L.Symbol (Cursor m inTy atTy)
```
travitch, line 198: Is the symbol the name of the function that was skipped? If so, is it the case that we never try this with indirect calls, where we wouldn't be able to get that symbol?

uc-crux-llvm/src/UCCrux/LLVM/Classify/Types.hs
```diff
@@ -214,13 +227,36 @@ ppUnfixable =
...
+    -- heuristics to return values from skipped functions
+    UnfixedRetReadUninitialized
+  | UnfixedRetWriteUnmapped
+  | UnfixedRetCall
+  | UnfixedRetMemset
+  | UnfixedRetPtrFree
+  | UnfixedRetPtrAddOffset
+  | UnfixedRetReadBadAlignment
+  | UnfixedRetWriteBadAlignment
+  | UnfixedRetDivRemByZero
```
travitch, line 240: Would it be useful to capture the name of the function that did the returning for these?

travitch (approved):

uc-crux-llvm/src/UCCrux/LLVM/Classify/Types.hs
langston-barrett: That's definitely reasonable, but I'm also planning on fixing all these right after #721 by using all the same heuristics for arguments, globals, and returned values of skipped functions (in fact, I have a WIP branch where that's done) so these will go away soon!

uc-crux-llvm/src/UCCrux/LLVM/Cursor.hs
langston-barrett: Great question - This works just fine for indirect calls because these `Selector`s are generated *inside* the relevant override, together with the returned value:

https://github.com/GaloisInc/crucible/pull/718/files#diff-e76f47d5b800a8b1d75a4bed25c18ba59ed5181cc7b4538079604b32c7f063f1R198

uc-crux-llvm/src/UCCrux/LLVM/Stats.hs
langston-barrett: Hmm good question... I think it's not necessary. I'll try changing it back!

PR #730, opened 2021-05-04
Experimental concurrency support

abakst:
The main purpose of this PR is to:

1. Add `crucible-concurrency`, a library used to add concurrency support for crucible-based tools. Notes on designs, limitations, future work, etc are included in `DesignNotes.md` in the `crucible-concurrency` subdirectory.
2. Enable some concurrency support in `crux-mir` using the above. Notes on this implementation are included in `Concurrency.md` in the `crux-mir` subdirectory. The changes broadly consist of a new module `Mir.Concurrency`, as well as some changes to the Rust libraries shipped with the tool (i.e., those under `lib`).

`crucible-concurrency` ships with an implementation for `crucible-syntax`:  hence, this PR also includes some minor changes to `crucible-syntax` to make it a bit more useful as a library, and also to fix a small issue with its entry block construction.


crucible-concurrency/sv-benchmarks/c/pthread-lit/foo
```diff
@@ -0,0 +1,7 @@
+Build profile: -w ghc-8.10.4 -O1
+In order, the following will be built (use -v for more details):
+ - crucibles-0.1.0.0 (lib) (file src/Explore.hs changed)
+ - crucibles-0.1.0.0 (exe:cruces) (dependency rebuilt)
+Preprocessing library for crucibles-0.1.0.0..
+Building library for crucibles-0.1.0.0..
+[ 9 of 11] Compiling Explore          ( src/Explore.hs, /Users/abakst/git/crucible/dist-newstyle/build/x86_64-osx/ghc-8.10.4/crucibles-0.1.0.0/build/Explore.o, /Users/abakst/git/crucible/dist-newstyle/build/x86_64-osx/ghc-8.10.4/crucibles-0.1.0.0/build/Explore.dyn_o )
```
spernsteiner, line 7: Committed by accident?

crux-mir/lib/libcore/crucible/concurrency.rs
```diff
@@ -0,0 +1,62 @@
...
+pub fn join<T>(thid : u64) -> T {
+    panic!("crucible::concurrency::join should never be executed!")
+}
+
+fn thread_exit<T>(f:T) {}
+
+/**
+Funny game we're playing here:
+
+In the caller of "spawn", we'll actually effectively overide the behavior of
```
RyanGlScott, line 30:
```suggestion
In the caller of "spawn", we'll actually effectively override the behavior of
```

crucible-concurrency/test/Main.hs
```diff
@@ -0,0 +1,71 @@
...
+          (test_action input outFile) -- action whose result is tested
+       | input <- sort inputs
+       , let outFile = replaceExtension input ".out"
+       , let goodFile = replaceExtension input ".out.good"
+       ]
+
+testSimulator :: FilePath -> FilePath -> IO ()
+testSimulator inFile outFile =
+  do contents <- T.readFile inFile
+     debugh   <- openFile "/dev/null" WriteMode
```
RyanGlScott, line 60: Is `debugh` actually used anywhere?

crucible-concurrency/DesignNotes.md
```diff
@@ -0,0 +1,185 @@
+# Crucibles/Cruces: concurrency-enabled symbolic simulation
```
RyanGlScott, line 1: Is there a place that explains the meaning behind the name "Crucibles"? I was sort of expecting it to more closely resemble the name of the package, `crucible-concurrency`.

crucible-concurrency/sv-benchmarks/c/pthread-lit/foo
abakst: Eek thanks, yes.

crucible-concurrency/DesignNotes.md
abakst: I guess it's the plural of crucible (multiple threads, etc...). `crucible-concurrency` seems like a mouthful for text, but you're right it could at least be explained. Now that you point it out, the cabal file should probably be named crucible-concurrency, etc.

crucible-concurrency/test/Main.hs
abakst: Nope! Thanks!

crucible-concurrency/DesignNotes.md
```diff
@@ -0,0 +1,187 @@
...
+recognizing certain _hooks_ in the Crucible code to be explored. These _hooks_
+can be thought of as functions denoting that a global access is about to occur,
+that a thread is about to be spawned, that a lock is going to be acquired, etc.
+
+When Crucibles encounters one of these hooks, it decides whether or not the
+current thread should be preempted by consulting a scheduling algorithm. When
+execution terminates, the same scheduling algorithm decides if it should try to
+find a new thread interleaving to explore, or if should terminate.
+
+## Using crucibles-concurrency in a new language
```
RyanGlScott, line 25: `crucibles-concurrency` or `crucible-concurrency`?

crucible-concurrency/DesignNotes.md
```diff
@@ -0,0 +1,187 @@
...
+The main thing a client language needs to provide in order to get support from
+Crucibles is to implement a set of primitive matchers. These matchers inspect a
+Crucible Call statement, and can return a Crucibles-defined primitive value. The
+primitives that Crucibles understands are enumerated by the `ExplorePrimitive`
+type in `Crucibles.Primitives`.
+
+An example implementation for `crucible-syntax` is defined by
+`crucibleSyntaxPrimitives` in the same module.
+
+## Modifying crucibles-concurrency
```
RyanGlScott, line 36: `crucibles-concurrency` or `crucible-concurrency`?

RyanGlScott (commented):
> I guess it's the plural of crucible (multiple threads, etc...).

Gotcha. I suppose my confusion was rooted in that I kept trying to interpret "Crucibles" as "multiple instances of Crucible" rather than as a separate name altogether. For whatever reason, I didn't seem to have the same issue with Crux versus Cruces—I guess the use of Spanish/Latin adds its own distinct flair :)

In any case, I appreciate the footnote you've added to the README.

crucible-concurrency/src/Cruces/CrucesMain.hs
```diff
@@ -0,0 +1,167 @@
...
+      undefined
+  where
+    isFn x (ACFG _ _ g) = handleName (cfgHandle g) == x
+
+run :: (Crux.Logs) => (Crux.CruxOptions, CrucesOptions) -> IO ()
+run (cruxOpts, opts) =
+  do let ?dpor = not (noDpor opts)
+     let ?bound = maxPreemptions opts
+     let [fn] = Crux.inputFiles cruxOpts
+     -- Some ng <- newIONonceGenerator
```
atomb, line 111: Would this line be better removed?

crucible-concurrency/src/Cruces/CrucesMain.hs
```diff
@@ -0,0 +1,167 @@
...
+     ha      <- newHandleAllocator
+     theInput <- readFile fn
+     case MP.parse (skipWhitespace *> many (sexp atom) <* eof) fn (pack theInput) of
+       Left err ->
+         do putStrLn $ errorBundlePretty err
+            exitFailure
+       Right v ->
+         withIONonceGenerator $ \nonceGen ->
+           -- withYicesOnlineBackend FloatRealRepr nonceGen ProduceUnsatCores noFeatures $ \sym -> do
```
atomb, line 120: Ditto. I imagine it inherits support for all of the standard Crux provers now. Is that right?

crucible-concurrency/src/Cruces/ExploreCrux.hs
```diff
@@ -0,0 +1,165 @@
...
+  Crux.SimulatorCallback $ \sym symOnline -> do
+    (mainHdl, prims, globs, fns) <- mkSym sym
+    let simCtx = initSimContext sym emptyIntrinsicTypes ha outh fns emptyExtensionImpl emptyExploration
+
+    let st0  = InitialState simCtx emptyGlobals defaultAbortHandler C.UnitRepr $
+                   runOverrideSim C.UnitRepr (exploreOvr symOnline cruxOpts (regValue <$> callFnVal mainHdl emptyRegMap))
+
+    feats <- case symOnline of
+               Just Crux.SomeOnlineSolver ->
+                 do -- _check <- pathSatisfiabilityFeature sym (considerSatisfiability sym)
```
atomb, line 69: Another no-longer-needed line?

crucible-concurrency/src/Crucibles/CruciblesMain.hs
```diff
@@ -0,0 +1,193 @@
...
+      undefined
+  where
+    isFn x (ACFG _ _ g) = handleName (cfgHandle g) == x
+
+run :: (Crux.Logs) => (Crux.CruxOptions, CruciblesOptions) -> IO ()
+run (cruxOpts, opts) =
+  do let ?dpor = not (noDpor opts)
+     let ?bound = maxPreemptions opts
+     let [fn] = Crux.inputFiles cruxOpts
+     -- Some ng <- newIONonceGenerator
```
atomb, line 112: Ditto.

crucible-concurrency/src/Crucibles/CruciblesMain.hs
```diff
@@ -0,0 +1,193 @@
...
+     ha      <- newHandleAllocator
+     theInput <- readFile fn
+     case MP.parse (skipWhitespace *> many (sexp atom) <* eof) fn (pack theInput) of
+       Left err ->
+         do putStrLn $ errorBundlePretty err
+            exitFailure
+       Right v ->
+         withIONonceGenerator $ \nonceGen ->
+           -- withYicesOnlineBackend FloatRealRepr nonceGen ProduceUnsatCores noFeatures $ \sym -> do
```
atomb, line 121: Ditto.

crucible-concurrency/src/Crucibles/CruciblesMain.hs
```diff
@@ -0,0 +1,193 @@
...
+                   -- (num, mfailed) <- exploreExecutions @DPOR sym ha debugH checkSat interpret mainHdl gvs fnBindings
+                   -- hPutStrLn outh ("== Explored " ++ show num ++ " executions ==")
+                   -- case mfailed of
+                   --   Just failed ->
+                   --     do hPutStrLn outh "UNSAFE:"
+                   --        forM_ failed $ \((p,e):_) ->
+                   --          do hPrint outh (ppSimError e)
+                   --             hPutStrLn outh ("assert " ++ show (printSymExpr p) ++ " failed.")
+                   --   Nothing ->
+                   --      hPutStrLn outh "Proved all goals."
```
atomb, line 170: This is probably obsolete at this point?

crucible-concurrency/src/Crucibles/DPOR.hs
```diff
@@ -0,0 +1,434 @@
...
+                 pending.at eventBeforeIID %=
+                   (pure . (Set.singleton pend <>) . fromMaybe mempty)
+               _ -> return ()
+
+    -- search e =
+    --   case e ^. eventInfo of
+    --     Read{} -> True
+    --     Write{} -> True
+    --     ThreadSpawn -> e ^. eventThread == thId
+    --     _ -> False
```
atomb, line 391: Ditto.

crucible-concurrency/src/Crucibles/DPOR.hs
```diff
@@ -0,0 +1,434 @@
...
+--   SchedAlgM DPOR EventID
+-- walkPrevID stop eid
+--   | eid == 0 = return eid
+--   | otherwise =
+--     do e <- view (exec.event eid)
+--        if stop e then
+--          return eid
+--         else
+--          do prev <- view (exec.prevEvent.at eid.to (fromMaybe (error "no parent event in walkPrevID")))
+--             walkPrevID stop prev
```
atomb, line 424: Ditto.

atomb (approved):
This is really an impressive bit of work! I appreciate the design notes and the number of tests and benchmarks you included. I didn't deeply review most of the Haskell code, since a lot of it implements algorithms I'm not very familiar with, but it looks good from what I can tell (though I did notice a few bits of commented-out code that could probably be removed).

crux-mir/lib/libstd/sys_common/poison.rs
```diff
@@ -20,7 +20,11 @@ pub struct Flag {
 
 impl Flag {
     pub const fn new() -> Flag {
-        Flag { failed: AtomicBool::new(false) }
+        #[cfg(feature = "opaque-poison-atomics")]
+        { Flag { failed: AtomicBool::new_unmodeled(false) } }
+
+        #[cfg(not(feature = "opaque-poison-atomics"))]
+        { Flag { failed: AtomicBool::new_unmodeled(false) } }
```
spernsteiner, line 27: This has the same code in both `cfg` cases - is that correct?

crux-mir/translate_libs.sh
```diff
@@ -28,6 +28,46 @@ translate_2015() {
...
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
+
+
+if [[ $SHOULD_MODEL_ATOMICS -eq "1" ]]; then
```
spernsteiner, line 64: `[[` is a bash builtin - either switch to `[` or change the shebang line from `/bin/sh` to `/bin/bash`

crux-mir/lib/libcore/crucible/concurrency.rs
```diff
@@ -0,0 +1,62 @@
+//! Crucible concurrency intrinsics
+
+
+// Signal a possible preemption point. `ro` is true if this
+// preemption point MUST be read only. It is always safe to
+// pass `false`.
+#[unstable(feature = "crucible_intrinsics", issue = "none")]
+pub fn sched_yield<T>(ro: bool, x: T) {}
```
spernsteiner, line 8: Looking at the implementations, it seems `sched_yield`, `mutex_lock`, and `mutex_unlock` all require `x` to be represented as `MirReferenceRepr`.  Is it possible to change the signatures here to take `x: *const T` to reflect that?  I think all references and pointers are coercible to `*const T` so it hopefully won't require changes elsewhere in the code.  If it would be a pain to change then please add a comment here explaining the requirements on `T`.

crux-mir/lib/libstd/sys_common/poison.rs
abakst: No, good catch. I caught another location in liballoc/sync.rs as well.

crux-mir/lib/libstd/sys/crux/mutex.rs
```diff
@@ -22,14 +24,16 @@ impl Mutex {
     pub unsafe fn lock(&self) {
         // TODO: Currently, we never run `drop` code, so locks are never released.  Once we support
         // drops, we can enable this assertion to check for (invalid) reentrant locking.
-        //assert!(!self.locked.get());
-        self.locked.set(true);
+        // assert!(!self.locked.get());
+        concurrency::mutex_lock(self);
+        self.locked.set(true); //TODO: Why does this yield an empty mux tree in crux?
```
spernsteiner, line 29: Is this TODO still relevant?  If so, under what conditions does it give the "empty mux tree" error?  (I assume it's not always doing it since otherwise nothing would work.)

crux-mir/src/Mir/Concurrency.hs
```diff
@@ -0,0 +1,157 @@
...
+  case t of
+    MirReferenceMux tree ->
+       let refs = fst <$> viewFancyMuxTree tree
+           extr r =
+             case r of
+               MirReference root p ->
+                 Text.pack (show root ++ mirPathName p)
+               -- N.B.: I don't think we should ever hit this:
+               MirReference_Integer t rv ->
+                 error "Unexpected MirReference_Integer in mirRefName"
```
spernsteiner, line 138: Looking at how this gets called, I agree that under the current implementation this case will never be hit.  But that relies on the fact that `std::sys::crux::Mutex` has nonzero size, which IMO is a bit unintuitive.  Would it be easy enough to just add support for this case?  `rv` is just a symbolic `usize`, same as in the `Index_RefPath` case.

crux-mir/src/Mir/Concurrency.hs
```diff
@@ -0,0 +1,157 @@
...
+       in (extr <$> refs)
+
+mirPathName :: C.IsSymInterface sym => MirReferencePath sym tpr t -> String
+mirPathName p =
+  case p of
+    Empty_RefPath -> ""
+    Any_RefPath _ p -> mirPathName p
+    Field_RefPath _ p idx -> mirPathName p ++ "." ++ show (indexVal idx)
+    Variant_RefPath _ p idx -> mirPathName p ++ "." ++ show (indexVal idx)
+    Index_RefPath _ p idx -> mirPathName p ++ "[" ++ show (W4.printSymExpr idx) ++ "]"
```
spernsteiner, line 148: Are you going to run into trouble here if you ever get two `idx`s that are syntactically distinct but semantically equal?  Imagine having symbolic integers `i` and `j` where the path condition implies that `i == j`.  `&arr[i]` and `&arr[j]` point to the same thing, but will produce different strings here.

crux-mir/lib/libstd/sys/crux/mutex.rs
abakst: Agreed: I think I confused myself when rebasing and thought that this was actually a pre-existing comment somehow, referring to something that I wasn't observing. While this PR _is_ experimental, it's not so experimental that it's crashing on every run.

spernsteiner:
I only looked at the crux-mir parts, but overall those parts look good.  My only real worry is the `Index_RefPath` case I commented on above.

crux-mir/src/Mir/Concurrency.hs
abakst: Hm I think the connection between the size of `Mutex` and this constructor is a bit subtle. However, I suppose that since this denotes an integer interpreted as a pointer, the integer value itself should work fine as a name (i.e., address). 

If the pointer can never be dereferenced as indicated by the comment on the constructor, then I guess there will be no observable effects related to the resource. Is that right? Or is just that reads and writes will have to live in `unsafe` code? Is there a requirement that the result of the cast never aliases a value that would be denoted by a `MirReference`?

spernsteiner: On second thought, you should probably continue to forbid this, but maybe with a more informative comment (either here or on the definition of `Mutex`).

You could indeed use the integer/address as the name.  But if `Mutex` ever became zero-sized, you would run into the separate issue that multiple `Mutex`es that are conceptually distinct might have the same address.  You can see this in action with this code:
```Rust
    let v = vec![(), (), ()];
    println!("{:p} {:p} {:p}", &v[0], &v[1], &v[2]);
```
which prints `0x1 0x1 0x1`.  Since `()` is zero-sized (so dereferencing `&()` never accesses any memory), `Vec::index` is allowed to cast any non-null, well-aligned integer into a `&()` reference.  It happens to always choose the smallest legal value, which for `()` is `1`.  If `Mutex` was zero-sized, it would do the same thing there, so you could end up with a vector of mutexes where `ms[0]` and `ms[1]` actually have the same address (`MirReference_Integer 1`) internally.

abakst: Thanks, I'll note that

crux-mir/src/Mir/Concurrency.hs
abakst: Yes! That's an issue -- the right resolution for now is to simply treat all elements as potentially interfering.

PR #736, opened 2021-05-12
SV-COMP: Adapt parsing to Task-Definition Format Version 2.0

RyanGlScott:
The `.yml` files included in https://github.com/sosy-lab/sv-benchmarks have undergone some changes since SV-COMP 2020. In particular, they now adhere to `format_version: 2.0`, as described [here](https://github.com/sosy-lab/sv-benchmarks/tree/99d37c5b4072891803b9e5c154127c912477f705#task-definitions). This means that the benchmark sets' descriptions and architectures are no longer described in `.cfg` files, but instead they are included with each `.yml` task file as `options`.

This changes the parsing logic in `Crux.SVCOMP` accordingly.

### DIFF
--- a/crux/src/Crux/SVCOMP.hs
+++ b/crux/src/Crux/SVCOMP.hs
@@ -11,7 +11,6 @@ module Crux.SVCOMP where
 
 import           Config.Schema
 import           Control.Applicative
-import           Control.Monad
 import qualified Data.Attoparsec.Text as Atto
 import           Data.Map (Map)
 import qualified Data.Map as Map
@@ -91,8 +90,6 @@ data ComputedVerdict
 data BenchmarkSet =
   BenchmarkSet
   { benchmarkName :: Text
-  , benchmarkDescription :: Text
-  , benchmarkArchWidth   :: Maybe Int
   , benchmarkTasks :: [VerificationTask]
   }
  deriving (Show,Eq,Ord)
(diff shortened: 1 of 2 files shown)


crux/src/Crux/SVCOMP.hs
```diff
@@ -186,7 +167,7 @@ loadVerificationTask fp =
  decodeVT :: Yaml.Value -> IO VerificationTask
  decodeVT (Yaml.Object o) =
   do case HM.lookup "format_version" o of
-       Just (Yaml.String "1.0") -> return ()
+       Just (Yaml.String "2.0") -> return ()
        _ -> fail $ unwords ["Expected verification task version 1.0 while parsing", show fp]
```
robdockins, line 171: should update this error message

crux/src/Crux/SVCOMP.hs
```diff
@@ -280,4 +275,4 @@ deduplicateTasks :: [BenchmarkSet] -> TaskMap
 deduplicateTasks = Map.unionsWith (++) . map f
   where
   f bs = Map.fromList
-           [ ((t, benchmarkArchWidth bs), [bs]) | t <- benchmarkTasks bs ]
+           [ (t, [bs]) | t <- benchmarkTasks bs ]
```
robdockins, line 278: Should update the comment to this function as well; target architecture is not longer a separate component of the task map.

robdockins (approved):
Looks good, modulo a couple of string and comments that need updating.

crux/src/Crux/SVCOMP.hs
RyanGlScott: Good catch!

crux/src/Crux/SVCOMP.hs
RyanGlScott: Right you are. Fixed.

PR #738, opened 2021-05-12
crux: add output-for-ide option

Ptival:
For vscode-crux-llvm, we need a way of communicating relevant information to
the client process.  Most of Crux's current output is directed at a human user.
This adds the possibility of having machine-readable output instead.

Open questions (could be solved now or later):
- Do we want to still output the human-relevant bits in `outputForIDE` mode?
- Should we have a more robust scheme for indicating the beginning and end of each output?

Open question to resolve before merging:
- Is the current output flushing greedy enough for the IDE? (I'll figure this out, but the PR can be reviewed either way)

### DIFF
--- a/crux-llvm/svcomp/Main.hs
+++ b/crux-llvm/svcomp/Main.hs
@@ -310,6 +310,7 @@ evaluateSingleTask writeHdl jsonHdl cruxOpts llvmOpts bsRoot num task inpBCFile 
       let cruxOpts' = cruxOpts { outDir = taskRoot, inputFiles = inputs }
 
       let ?outputConfig = OutputConfig False writeHdl writeHdl False
+      let ?outputForIDE = False
 
       mres <- try $
                do res <- Crux.runSimulator cruxOpts' (simulateLLVMFile inpBCFile llvmOpts)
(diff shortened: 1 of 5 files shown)


Ptival:
Before someone asks, the `outputForIDE` flag cannot be part of the `OutputOptions`, because these are defined unconditionally by executables prior to even looking at the configuration and arguments to the executable.

That's why I added a separate boolean to the `Logs` constraint.

travitch:
Would it be more desirable to just turn all of our reporting into structured messages (i.e., ADT values)? We could then just stream those out of the verifier loop and do the rendering outside of the verifier. It could look like just writing ADT values out over a channel from the verifier core.

For command line usage, spin up a thread to consume messages over that channel and pretty print them for consumption.

For IDE usage, convert them to JSON (or something else) and write them out over a socket.

We've been moving a few libraries to that model using Kevin's lumberjack library https://hackage.haskell.org/package/lumberjack

Ptival:
I think that'd be reasonable, as long as we get alignment on it (I assume this has consequences on all `crux-*` projects), I'm happy trying to make such changes.

Paging @atomb since this came up as part of an effort on the CASE front.

atomb:
I think @travitch makes a good point, and we've already been talking about moving more code over to use `lumberjack`. If you think you have the capacity to make changes along those lines, @Ptival, I think it's a great idea.

Ptival:
Superseded by #755.

PR #746, opened 2021-05-21
crux-llvm: Search in more places for the default lib-dir

### LINKED ISSUE #731: Make crux-llvm header and library bitcode files easier to find when using releases
Currently, `crux-llvm` expects `crucible.h`, any C++ library bitcode files, and runtime support files for executable counterexamples to exist in a `c-src` subdirectory of the library directory. The `libdir` configuration variable can control where to look for these files, but its default value is only meaningful when the executables we build are located somewhere that Cabal understands. When we distribute binary tarballs, this isn't true, so users need to specify a value for `libdir` manually.

RyanGlScott:
Previously, `crux-llvm` was relying entirely on `getDataDir` to find the location of the default `lib-dir`. However, this proves fragile enough that we should consult other locations besides it. (See the discussion in #731.) The most obvious location to search is in a directory relative to the `crux-llvm` binary itself, which serves the common use case of invoking a `crux-llvm` bindist.

Fixes #731.

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Config.hs
+++ b/crux-llvm/src/Crux/LLVM/Config.hs
@@ -4,9 +4,13 @@
 
 module Crux.LLVM.Config where
 
+import           Control.Applicative ( Alternative(..) )
 import           Control.Exception ( Exception, displayException, throwIO )
+import           Control.Monad ( guard )
 import           Control.Monad.State ( liftIO, MonadIO )
-import           System.FilePath ( (</>) )
+import           System.Directory ( doesDirectoryExist )
+import           System.Environment ( getExecutablePath )
+import           System.FilePath ( (</>), joinPath, normalise, splitPath, takeDirectory )
 
 import qualified Data.LLVM.BitCode as LLVM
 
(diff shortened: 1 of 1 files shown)


RyanGlScott:
I'm not quite sure how to add a regression test for this, but I did confirm that I can create a bindist with this patch and have it work as expected when invoking `crux-llvm`:

```
$ RUNNER_OS=Linux .github/ci.sh bundle_crux_llvm_files
<elided>

$ ./crux-llvm-2021-05-21-Linux-x86_64/bin/crux-llvm test.c
[Crux] Simulating function "main"
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 1
[Crux]   Disproved: 0
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Valid.
```

RyanGlScott:
I had a conversation with @kquick about the best way to design this. In particular, he wondered why we didn't turn `libDir` into a list of paths to search through, rather than having it be a single path like it is now. I originally considered this approach, but I decided against it since `libDir` is used to find the paths to files that must be passed wholesale to `clang`, such as [`libcxx*.bc`](https://github.com/GaloisInc/crucible/blob/d332aa6b2fd5fd12446f7c38f4a14e00e3c68ca9/crux-llvm/c-src/libcxx-7.1.0.bc), [`print-model.c`](https://github.com/GaloisInc/crucible/blob/79428012dae4e3c922f75c6f41f418c8aacca665/crux-llvm/src/Crux/LLVM/Compile.hs#L273), and [`concrete-backend.c`](https://github.com/GaloisInc/crucible/blob/79428012dae4e3c922f75c6f41f418c8aacca665/crux-llvm/src/Crux/LLVM/Compile.hs#L280). As a result, we can't simply pass a bunch of paths to `clang` separated with something like `-I`/`-L` and let `clang` do the rest.

We could envision having `crux-llvm` take the list of paths and search through them on its own, passing the final results to `clang`, but the benefit would be marginal. One of the main benefits of the search path approach would be that a user could, say, override the default path to `libcxx*.bc` with their own version while keeping the other defaults the same. With the way `crux-llvm` is written currently, however, all of `libcxx*.bc`, `print-model.c`, `concrete-backend.c`, etc. must live in the same directory. In other words, if you override one file, you must override them all. This makes the scope considerably larger than what we'd like, so for now, we'll go with the current approach of having `libDir` be a single directory.

kquick (approved):

PR #749, opened 2021-05-28
crux-mir translation and library changes for Cryptol support

spernsteiner:
This PR contains the `crux-mir` parts of GaloisInc/saw-script#1313, specifically:

* Add the `cryptol!` macro to `lib/crucible`, with placeholders for `cryptol::load` and `cryptol::override_` (to be overridden in `crux-mir-comp`).

* Add placeholder for `munge`, to be overridden in `crux-mir-comp`.

* Add support for `#[repr(transparent)]`.

* Add custom translation for unchecked arithmetic and `rotate_left`/`right` intrinsics.

### DIFF
--- a/crux-mir/src/Mir/Overrides.hs
+++ b/crux-mir/src/Mir/Overrides.hs
@@ -13,13 +13,14 @@
 
 module Mir.Overrides (bindFn, getString) where
 
-import Control.Lens ((^.), (%=), (.=), use)
+import Control.Lens ((^.), (^?), (%=), (.=), use, ix, _Wrapped)
 import Control.Monad
 import Control.Monad.IO.Class
 import Control.Monad.State (get)
 import Control.Monad.Trans (lift)
 import Control.Monad.Trans.Maybe (MaybeT(..))
 
+import Data.BitVector.Sized (mkBV)
 import qualified Data.BitVector.Sized as BV
 import qualified Data.ByteString as BS
 import qualified Data.Char as Char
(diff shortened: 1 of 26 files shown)


crux-mir/src/Mir/Overrides.hs
```diff
@@ -84,9 +85,11 @@ import Crux.Types (Model(..), Vars(..), Vals(..), Entry(..))
 
 import Mir.DefId
 import Mir.FancyMuxTree
-import Mir.Generator (CollectionState)
+import Mir.Generator (CollectionState, collection, handleMap, MirHandle(..))
 import Mir.Intrinsics
-import Data.BitVector.Sized (mkBV)
+import qualified Mir.Mir as M
+
+import Debug.Trace
```
atomb, line 92: Is this import still necessary?

crux-mir/src/Mir/Overrides.hs
```diff
@@ -376,17 +411,38 @@ bindFn ::
...
+  , Empty :> MirSliceRepr (BVRepr w) :> (asBaseType -> AsBaseType btpr) <- cfgArgTypes cfg
+  , Just Refl <- testEquality w (knownNat @8)
+  , UnitRepr <- cfgReturnType cfg
+  = bindFnHandle (cfgHandle cfg) $ UseOverride $
+    mkOverride' "crucible_override_" UnitRepr $ do
+        RegMap (Empty :> RegEntry _ strRef :> RegEntry _ expr) <- getOverrideArgs
+        str <- getString strRef >>= \x -> case x of
+            Just x -> return x
+            Nothing -> fail $ "dump_what4: desc string must be concrete"
+        traceM $ Text.unpack str ++ " = " ++ show (printSymExpr expr)
```
atomb, line 443: Ah, this is the use of `Debug.Trace`. I believe `OverrideSim` wraps `IO`, so you can probably use something like `liftIO $ putStrLn ...` here.

atomb (approved):
This looks nice!

PR #751, opened 2021-06-01
Use `Paths_*` for version numbers in `--version` output

### LINKED ISSUE #750: `crux-llvm --version` prints an out-of-date version number
If you ask `crux-llvm` what its version is, you'll get:

```
$ cabal exec crux-llvm -- --version
crux version: 0.1, crux-llvm version: 0.1
```

But this isn't the current version of `crux-llvm` at all!

https://github.com/GaloisInc/crucible/blob/ad0c5919d5bc2c260db3f10711e36e5477cc6568/crux-llvm/crux-llvm.cabal#L3

The culprit is this line of code, which hardcodes the version number:

https://github.com/GaloisInc/crucible/blob/79428012dae4e3c922f75c6f41f418c8aacca665/crux-llvm/src/CruxLLVMMain.hs#L32

We should instead use the `Version` from `Paths_crux_llvm` to ensure that it stays up-to-date with version changes.

RyanGlScott:
This avoids needing to hardcode the version numbers used in `.cabal` files as strings. Fixes #750.

### DIFF
--- a/crucible-concurrency/crucible-concurrency.cabal
+++ b/crucible-concurrency/crucible-concurrency.cabal
@@ -53,6 +53,8 @@ library
 
 executable cruces
   main-is:             crucibles/Main.hs
+  other-modules:       Paths_crucible_concurrency
+  autogen-modules:     Paths_crucible_concurrency
   -- other-extensions:
   build-depends:       base >=4.7 && < 5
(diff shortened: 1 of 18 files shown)


robdockins (approved):
Looks good to me

PR #752, opened 2021-06-02
What4 now allows idempotent config setting (as of 2e43136)

kquick:


RyanGlScott:
This seems reasonable. The `crux-mir` CI fails on this PR, and although I doubt it's the fault of your patch, it might be worth taking a look at.

kquick:
Yes, that mir failure looks like a test timeout (ran for 3h30min) and it looks unrelated because the other GHC mir tests passed.

RyanGlScott (approved):

PR #759, opened 2021-06-08
Sequences

### LINKED ISSUE #587: Incorrect counterexample in `crux-llvm`
Given `test.c`:
```c
#include <crucible.h>
#define N 2

int main()
{
    int i;
    for (i = 1; i <= N; i++) {
      int i0 = i;
      for (int j = 1; j <= i0; j++) {
        int b = crucible_int32_t("b");
        switch (b) {
	  case 1: break;
	  case 2: i++; break;
	  default:
	    return i;
	}
      }
    }
    crucible_assert(i <= 4, __FILE__, __LINE__);

    return 0;
}
```
Running `crux-llvm test.c` produces a counter example file whose contents include:
```c
size_t const crucible_values_number_int32_t = 4;
const char* crucible_names_int32_t[] = { "b","b","b","b" };
int32_t const crucible_values_int32_t[] = { 0x1,0x2,0x0,0x2 };
```
However this is not a valid counter-example, as executing the program with this input results in returning from `main` before hitting the assertion.  Instead, the counterexample should have the value `0x2` as the third entry:
```c
int32_t const crucible_values_int32_t[] = { 0x1,0x2,0x2,0x2 };
```

robdockins:
Add a new `Sequence` type to crucible.

This will be used to implement the event log needed to close #587.  It may also prove helpful when we eventualy refactor the LLVM memory model.

### DIFF
--- a/crucible-syntax/src/Lang/Crucible/Syntax/Atoms.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/Atoms.hs
@@ -50,7 +50,7 @@ data Keyword = Defun | DefBlock | DefGlobal
              | Just_ | Nothing_ | FromJust
              | Inj | Proj
              | AnyT | UnitT | BoolT | NatT | IntegerT | RealT | ComplexRealT | CharT | StringT
-             | BitvectorT | VectorT | FPT | FunT | MaybeT | VariantT | RefT
+             | BitvectorT | VectorT | FPT | FunT | MaybeT | VariantT | StructT | RefT
              | Half_ | Float_ | Double_ | Quad_ | X86_80_ | DoubleDouble_
              | Unicode_ | Char8_ | Char16_
              | The
@@ -64,6 +64,7 @@ data Keyword = Defun | DefBlock | DefGlobal
              | ToAny | FromAny
              | VectorLit_ | VectorReplicate_ | VectorIsEmpty_ | VectorSize_
              | VectorGetEntry_ | VectorSetEntry_ | VectorCons_
+             | MkStruct_ | GetField_ | SetField_
              | Deref | Ref | EmptyRef
              | Jump_ | Return_ | Branch_ | MaybeBranch_ | TailCall_ | Error_ | Output_ | Case
              | Print_ | PrintLn_
@@ -133,6 +134,7 @@ keywords =
   , ("->", FunT)
   , ("Maybe", MaybeT)
   , ("Variant", VariantT)
+  , ("Struct", StructT)
 
     -- string sorts
   , ("Unicode", Unicode_)
@@ -181,6 +183,11 @@ keywords =
     -- Variants
   , ("inj", Inj)
   , ("proj", Proj)
+
+    -- Structs
+  , ("struct", MkStruct_)
+  , ("get-field", GetField_)
+  , ("set-field", SetField_)
 
     -- Maybe
   , ("just" , Just_)
--- a/crucible-syntax/src/Lang/Crucible/Syntax/Atoms.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/Atoms.hs
@@ -19,7 +19,6 @@ import Control.Applicative
 import Data.Char
 import Data.Functor
 import Data.Ratio
-import Data.Semigroup ( (<>) )
 import Data.Text (Text)
 import qualified Data.Text as T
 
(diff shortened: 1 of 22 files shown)


crucible/src/Lang/Crucible/Simulator/SymSequence.hs
```diff
@@ -0,0 +1,302 @@
...
+import qualified Data.Parameterized.Map as MapF
+
+import           Lang.Crucible.Types
+import           What4.Interface
+import           What4.Partial
+
+------------------------------------------------------------------------
+-- SymSequence
+
+data SymSequence sym a where
```
RyanGlScott, line 36: It might be worth giving this a `Show`/`ShowF`/`Pretty` instance to make it easier to print these for debugging purposes in the future.

crucible-syntax/test-data/simulator-tests/seq-test1.cbl
```diff
@@ -0,0 +1,20 @@
...
+    (assert! (equal? 0 (seq-length n)) "nil length test")
+
+    (let s1 (seq-cons 5 n))
+
+    (assert! (not (seq-nil? s1)) "cons test")
+    (assert! (equal? 1 (seq-length s1)) "cons length test")
+
+    (let v (from-just (seq-head s1) "head s1"))
+    (let t (from-just (seq-tail s1) "tail s1"))
+    (let u (from-just (seq-uncons s1) "uncons s1"))
```
RyanGlScott, line 14: I don't think there's any `assert!`/`print` statement involving this `u` value. I think it would be worthwhile to add one for consistency with the other operations featured in this test case.

RyanGlScott (commented):
Hooray! This looks good to me from a quick glance (aside from some very minor suggestions).

crucible/src/Lang/Crucible/Simulator/RegValue.hs
```diff
@@ -143,6 +148,7 @@ instance Show (FnVal sym a r) where
 -- | Version of 'MuxFn' specialized to 'RegValue'
 type ValMuxFn sym tp = MuxFn (Pred sym) (RegValue sym tp)
 
+
```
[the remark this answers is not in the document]
robdockins: nope

crucible/src/Lang/Crucible/Simulator/SymSequence.hs
robdockins: agreed

crucible-syntax/test-data/simulator-tests/seq-test1.cbl
robdockins: Absolutely.  Turns out, I have to implement syntax for structs first, which is why I stopped here for now. :-)

crucible-syntax/src/Lang/Crucible/Syntax/Atoms.hs
```diff
@@ -50,7 +49,7 @@ data Keyword = Defun | DefBlock | DefGlobal
              | Just_ | Nothing_ | FromJust
              | Inj | Proj
              | AnyT | UnitT | BoolT | NatT | IntegerT | RealT | ComplexRealT | CharT | StringT
-             | BitvectorT | VectorT | FPT | FunT | MaybeT | VariantT | RefT
+             | BitvectorT | VectorT | SequenceT | FPT | FunT | MaybeT | VariantT | StructT | RefT
```
RyanGlScott, line 52: The `crucible-syntax` README ought to be updated in light of these new types being introduced. (At the very least, the "Types" section.)

crucible/src/Lang/Crucible/Simulator/SymSequence.hs
```diff
@@ -0,0 +1,507 @@
...
+    loop SymSequenceNil = pure []
+    loop (SymSequenceCons _ v tl) = (:) <$> eval v <*> loop tl
+    loop (SymSequenceAppend _ xs ys) = (++) <$> loop xs <*> loop ys
+    loop (SymSequenceMerge _ p xs ys) =
+      do b <- conc p
+         if b then loop xs else loop ys
+
+-- | Given a pretty printer for elements,
+--   print a symbolic sequence.
+prettySymSequence :: IsExpr (SymExpr sym) =>
```
RyanGlScott, line 402: Would it be more convenient to also have a `Pretty` instance with `pretty = prettySymSequence pretty`?

RyanGlScott (approved):

PR #761, opened 2021-06-08
Remove redundant fmap in HasModel instance for Model

RyanGlScott:
`fmap id` can simply be replaced with `id`.


robdockins:
Made obsolete by #769

PR #762, opened 2021-06-09
Data type for log messages across crux projects

Ptival:
This pull request builds on top of #755 and will be rebased as that one evolves into its final form.

This draft pull request serves as a way to check for CI integration ahead of the final pull request, as well as gathering feedback on the current design.

For feedback, you should only look at the very last commit, authored by me, rather than the whole set of changes which contains all the changes from #755.

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Log.hs
+++ b/crux-llvm/src/Crux/LLVM/Log.hs
@@ -1,0 +1,61 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module Crux.LLVM.Log
+  ( CruxLLVMLogMessage (..),
+    SupportsCruxLLVMLogMessage,
+    cruxLLVMLogMessageToSayWhat,
+    sayCruxLLVM,
+  )
+where
+
+import Crux.Log (SayLevel (..), SayWhat (..), cruxLogTag)
+import qualified Crux.Log as Log
+import Data.Text as Text (Text, pack, unwords)
+
+data CruxLLVMLogMessage
+  = Breakpoint Text
+  | ClangInvocation [Text]
+  | Executable Text
+  | FailedToBuildCounterexampleExecutable
+  | SimulatingFunction Text
+  | UsingPointerWidthForFile Integer Text
+
+type SupportsCruxLLVMLogMessage msgs =
+  (?injectCruxLLVMLogMessage :: CruxLLVMLogMessage -> msgs)
+
+sayCruxLLVM ::
+  Log.Logs msgs =>
+  SupportsCruxLLVMLogMessage msgs =>
+  CruxLLVMLogMessage ->
+  IO ()
+sayCruxLLVM msg =
+  let ?injectMessage = ?injectCruxLLVMLogMessage
+   in Log.say msg
+
+clangLogTag :: Text
+clangLogTag = "CLANG"
+
+cruxLLVMLogMessageToSayWhat :: CruxLLVMLogMessage -> SayWhat
+cruxLLVMLogMessageToSayWhat (ClangInvocation params) =
+  SayWhat Noisily clangLogTag (Text.unwords params)
+cruxLLVMLogMessageToSayWhat (Breakpoint line) =
+  SayWhat Simply cruxLogTag ("*** break on line: " <> line)
+cruxLLVMLogMessageToSayWhat (Executable exe) =
+  SayWhat Simply cruxLogTag ("*** debug executable: " <> exe)
+cruxLLVMLogMessageToSayWhat FailedToBuildCounterexampleExecutable =
+  SayWhat Fail cruxLogTag "Failed to build counterexample executable"
+cruxLLVMLogMessageToSayWhat (SimulatingFunction function) =
+  SayWhat Simply cruxLogTag ("Simulating function " <> function)
+cruxLLVMLogMessageToSayWhat (UsingPointerWidthForFile width file) =
+  SayWhat
+    Simply
+    cruxLogTag
+    ( Text.unwords
+        [ "Using pointer width:",
+          pack (show width),
+          "for file",
+          file
+        ]
+    )
--- a/crux-llvm/svcomp/SVComp/Log.hs
+++ b/crux-llvm/svcomp/SVComp/Log.hs
@@ -1,0 +1,123 @@
+{-# LANGUAGE ConstraintKinds #-}
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE OverloadedStrings #-}
+
+module SVComp.Log
+  ( EvaluationProcessFailureReason (..),
+    SupportsSVCompLogMessage,
+    SVCompLogMessage (..),
+    SVCompSkipReason (..),
+    saySVComp,
+    svCompLogMessageToSayWhat,
+  )
+where
+
+import Crux (SayLevel (..), SayWhat (..))
+import qualified Crux.Log as Log
+import Crux.SVCOMP (ComputedVerdict (..))
+import qualified Data.Text as T
+
+data EvaluationProcessFailureReason
+  = ExitedWithFailureCode T.Text
+  | StoppedBySignal T.Text
+  | TerminatedBySignal T.Text
+  | UnknownStatus
+
+data SVCompSkipReason
+  = DueToBlacklist
+  | NoInputFiles
+
+svCompSkipReasonSuffix :: SVCompSkipReason -> T.Text
+svCompSkipReasonSuffix DueToBlacklist = " due to blacklist"
+svCompSkipReasonSuffix NoInputFiles = " because no input files are present"
+
+data SVCompLogMessage
+  = DoubleFaultWhileTryingToRemove T.Text
+  | ElapsedWallClockTime T.Text
+  | Evaluating
+      T.Text
+      -- ^ task root
+      T.Text
+      -- ^ source file
+  | EvaluationProcessFailed EvaluationProcessFailureReason
+  | FailedToCompileTask T.Text
+  | SimulatorThrewException
+  | VerdictExpecting
+      Bool
+      -- ^ the expected verdict (True for success, False for failure)
+      ComputedVerdict
+      -- ^ the actual verdict
+  | NoVerdict [T.Text]
+  | Skipping SVCompSkipReason T.Text
+
+type SupportsSVCompLogMessage msgs =
+  (?injectSVCompLogMessage :: SVCompLogMessage -> msgs)
+
+saySVComp ::
+  Log.Logs msgs =>
+  SupportsSVCompLogMessage msgs =>
+  SVCompLogMessage ->
+  IO ()
+saySVComp msg =
+  let ?injectMessage = ?injectSVCompLogMessage
+   in Log.say msg
+
+
+svCompTag :: T.Text
+svCompTag = "SVCOMP"
+
+svCompFail :: T.Text -> SayWhat
+svCompFail = SayWhat Fail svCompTag
+
+svCompOK :: T.Text -> SayWhat
+svCompOK = SayWhat OK svCompTag
+
+svCompWarn :: T.Text -> SayWhat
+svCompWarn = SayWhat Warn svCompTag
+
+svCompLogMessageToSayWhat :: SVCompLogMessage -> SayWhat
+svCompLogMessageToSayWhat (DoubleFaultWhileTryingToRemove path) =
+  svCompFail ("Double fault! While trying to remove: " <> path)
+svCompLogMessageToSayWhat (ElapsedWallClockTime time) =
+  svCompOK ("Elapsed wall-clock time: " <> time)
+svCompLogMessageToSayWhat (Evaluating taskRoot sourceFile) =
+  svCompOK
+    ( T.unlines
+        [ "Evaluating:",
+          "  " <> taskRoot,
+          "  " <> sourceFile
+        ]
+    )
+svCompLogMessageToSayWhat (EvaluationProcessFailed (ExitedWithFailureCode code)) =
+  svCompFail ("Evaluation process exited with failure code " <> code)
+svCompLogMessageToSayWhat (EvaluationProcessFailed (TerminatedBySignal signal)) =
+  svCompWarn ("Evaluation process terminated by signal " <> signal)
+svCompLogMessageToSayWhat (EvaluationProcessFailed (StoppedBySignal signal)) =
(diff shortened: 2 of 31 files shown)


crux-llvm/svcomp/SVComp/Log.hs
```diff
@@ -0,0 +1,105 @@
...
+  | ElapsedWallClockTime T.Text
+  | Evaluating
+      T.Text
+      -- ^ task root
+      T.Text
+      -- ^ source file
+  | EvaluationProcessFailed EvaluationProcessFailureReason
+  | FailedToCompileTask T.Text
+  | SimulatorThrewException
+  | VerdictExpecting Bool ComputedVerdict
```
RyanGlScott, line 40: Add a comment explaining what this `Bool` means.

crux/src/Crux/Log.hs
```diff
@@ -25,23 +38,154 @@ module Crux.Log (
...
+cruxSimply = SayWhat Simply cruxLogTag
+
+cruxWarn :: T.Text -> SayWhat
+cruxWarn = SayWhat Warn cruxLogTag
+
+
+cruxMessageToSayWhat :: SayWhatHandler ext => CruxLogMessage ext -> IO SayWhat
+
+cruxMessageToSayWhat AttemptingProvingVCs =
+  pure (cruxSimply "Attempting to prove verification conditions.")
```
kquick, line 115: Very minor note: I prefer to only use `pure` in `Applicative`, but here it's `IO` which is a `Monad` so I would have used `return`.

crux/src/Crux/Log.hs
```diff
@@ -25,23 +38,154 @@ module Crux.Log (
...
+-- | Main function used to log/output a general text message of some kind
+say :: Logs => CruxLogMessage () -> IO ()
+say =
+  let ?extToSayWhat = \ () -> pure SayNothing in
+  LJ.writeLog (?outputConfig ^. logWhat) <=< cruxMessageToSayWhat
+
 -- | Main function used to log/output a general text message of some kind
-say :: Logs => SayLevel -> Text -> Text -> IO ()
-say lvl frm = LJ.writeLog (?outputConfig ^. logWhat) . SayWhat lvl frm
+sayExt :: Logs => SayWhatHandler ext => ext -> IO ()
```
kquick, line 187: Could we use something other than "ext" to represent the front-end?  Using `ext` conflates with the crucible `ext` and they are fairly different.  The closest analogy would probably be the crucible `personality` parameter, but I have a hard time coming up with a function name to use here (and related) that is a short form for personality.   You could just use something like `says` or `sayCruxMsg`... or anything else that seems reasonable.

crux-llvm/src/Crux/LLVM/Log.hs
```diff
@@ -0,0 +1,46 @@
...
+  )
+where
+
+import Crux.Log (SayLevel (..), SayWhat (..), cruxLogTag)
+import Data.Text as Text (Text, pack, unwords)
+
+data CruxLLVMLogMessage
+  = ClangInvocation [Text]
+  | DebugBreakpoint Text
+  | DebugExecutable Text
```
kquick, line 15: Similar to my comment below, maybe we should remove the `Debug...` prefix from these messages since that imposes a severity.  I realize we'd probably need to replace it with _something_ since just `Breakpoint` and `Executable` are too short and non-unique.

crux-llvm/src/Crux/LLVM/Log.hs
```diff
@@ -0,0 +1,46 @@
...
+  | FailedToBuildCounterexampleExecutable
+  | SimulatingFunction Text
+  | UsingPointerWidthForFile Integer Text
+
+clangLogTag :: Text
+clangLogTag = "CLANG"
+
+cruxLLVMLogMessageToSayWhat :: CruxLLVMLogMessage -> IO SayWhat
```
kquick, line 23: See note below: if we remove `IO` from these, they can just be a contravariant functor that the crux LLVM front-end contramaps onto the base logger.

crux/src/Crux/Log.hs
```diff
@@ -25,23 +38,154 @@ module Crux.Log (
...
 import           Lang.Crucible.Backend ( AssumptionReason )
 import qualified Lang.Crucible.Simulator.SimError as CSE
 
 import           Crux.Types
+import           Crux.Version ( version )
 
 ----------------------------------------------------------------------
 -- Various functions used by the main code to generate logging output
 
+newtype LogDoc = LogDoc (Doc ())
```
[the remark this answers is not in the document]
Ptival: We can but it would be an orphan instance, and could potentially cause problems if someone else were to write such an instance?

crux/src/Crux/Log.hs
```diff
@@ -10,9 +13,19 @@ module Crux.Log (
...
+  , CruxLogMessage(..)
   , SayWhat(..)
+  , SayWhatHandler
   , SayLevel(..)
+  , cruxLogTag
+  , cruxMessageToSayWhat
+  , cruxNoisily
+  , cruxOK
+  , cruxSimply
+  , cruxWarn
```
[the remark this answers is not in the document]
Ptival: Agreed.

crux/src/Crux/Log.hs
```diff
@@ -25,23 +38,154 @@ module Crux.Log (
...
+  | SimulationComplete
+  | SimulationTimedOut
+  | StartedGoal Integer
+  | TotalPathsExplored Word64
+  | Version T.Text Version
+  | WarnDisablingBranchCoverageRequiresPathSatisfiability
+  | WarnDisablingProfilingIncompatibleWithPathSplitting
+  | WarnPathsUnexplored Int
+  | WarnSkippingUnsatCoresBecauseMCSatEnabled
+  | WarnUnsupportedTimeoutFor String -- ^ name of the backend
```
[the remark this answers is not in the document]
Ptival: Agreed.

crux/src/Crux/Log.hs
```diff
@@ -25,23 +38,154 @@ module Crux.Log (
...
+  deriving ( Generic, ToJSON )
+
+
+type SayWhatHandler ext =
+  ( ?extToSayWhat :: ext -> IO SayWhat
```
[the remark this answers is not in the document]
Ptival: I like this, will attempt.

crux/src/Crux/Log.hs
Ptival: Ah, I tend to use `return` inside actually-effectful `do`-blocks, and `pure` when there really was nothing effectful going on.

crux/src/Crux/Log.hs
Ptival: Yeah we can definitely change this. I don't love `says`, because it reminds me of `get`/`gets`, `put`/`puts`, but does not have the same semantics.

Will think of something.

crux-llvm/src/Crux/LLVM/Log.hs
Ptival: Yeah, and I do feel like the names here tend to be very general and are best used qualified. For instance, `CruxLogMessage` has constructors like `Goals`, `Help`, `Version` that are bound to collide with other constants, and I think it's better in general to always qualify them to avoid any confusion.

crux/src/Crux/Log.hs
kquick: OK.  Like I said, it was just a very minor observation, so I'm fine with your reasoning.

crux/src/Crux/Log.hs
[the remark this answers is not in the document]
kquick: Fair point, we should probably stick with `LogDoc`.

crux-llvm/svcomp/SVComp/Log.hs
```diff
@@ -0,0 +1,105 @@
...
+        )
+    )
+svCompMessageToSayWhat (EvaluationProcessFailed (ExitedWithFailureCode code)) =
+  pure (svCompFail ("Evaluation process exited with failure code " <> code))
+svCompMessageToSayWhat (EvaluationProcessFailed (TerminatedBySignal signal)) =
+  pure (svCompWarn ("Evaluation process terminated by signal " <> signal))
+svCompMessageToSayWhat (EvaluationProcessFailed (StoppedBySignal signal)) =
+  pure (svCompWarn ("Evaluation process stopped by signal " <> signal))
+svCompMessageToSayWhat (EvaluationProcessFailed UnknownStatus) =
+  pure (svCompFail ("Could not retrieve evaluation process status"))
```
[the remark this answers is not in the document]
Ptival: Thanks, fixed!

crux/src/Crux/Log.hs
Ptival: New commit shows a proposed change:
https://github.com/GaloisInc/crucible/blob/3c333baf99c624b0810d5ee72294f22c4952cbd3/crux/src/Crux/Log.hs#L99-L105

`say` is now polymorphic over the message data type, but has a constraint for an implicit parameter turning that message into a `SayWhat`.

Ptival:
Additional thoughts and concerns:

1. We are also considering moving from `Text` to `Doc` for `SayWhat`, so that layout decisions can be delayed to log handlers rather than being made at log call sites. Considering doing this in a separate PR, but feedback welcome here too.

2. One possible point of contention is the handling of extensibility of logging.

    I had initially considered having it all in a single parameterized data type `CruxLogMessage ext` with an extension constructor `ExtendedLogMessage :: ext -> CruxLogMessage`, and a type class guaranteeing the existence of a log classifier `ext -> SayWhat`. I did not like the typeclass approach as it enforced the uniqueness of such a classifier, so instead decided to use implicit parameters. I also found the single data type a bit awkward, and went instead with just having entirely separate data types.

    So with the current design, you can call `say` with whatever datatype `t`, as long as you have a `?handleMessage :: t -> SayWhat` in context. This is flexible enough to also allow mixing and matching multiple data types, as long as you shadow the `?handleMessage` name with the appropriate handler, e.g.:

    ```haskell
    foo ::
      (?handleCruxLogMessage :: Log.Handler CruxLogMessage) =>
      (?handleMirLogMessage :: Log.Handler MirLogMessage) =>
    ...
    foo = do
      let ?handleMessage = ?handleMirLogMessage -- you can optionally nest a 'do' here to make a local scope
      doSomeMirStuff
      let ?handleMessage = ?handleCruxLogMessage
      doSomeCruxStuff
    ```

    It's slightly verbose but quite flexible, if you have better suggestions, let's discuss it! :-)

langston-barrett (approved):
The parts that touch `uc-crux-llvm` LGTM.

Ptival:
Reconverting to draft because there is one more structural change we would like to make before merging, sorry for the false start!

robdockins:
I'm not completely set on this idea, but I think I'd prefer to see a combinator used for setting the `logHandler`.  It doesn't make any real semantic difference, but I think it would help readers of the code to understand the scope of the log handlers.  I'm thinking something like:

```
withLogHandler :: Handler a -> (LogsHandling a => b) -> b
withLogHandler h x = let ?handleMessage = h in x
```

Ptival:
@robdockins The API you looked at is now outdated, but I believe your comment stays relevant.

In the current iteration of the design, the end user responsibility is to provide the following:

https://github.com/GaloisInc/crucible/blob/2985cf4dacbe855a85b96fae654840560d2fb3b3/uc-crux-llvm/test/Test.hs#L140-L148

which is a bit ugly but flexible. I tried to avoid type classes because I don't want logging set in stone and want to allow local overrides.

robdockins:
Another design choice to consider would be to use a `Prism` for the injection function.  That would allow you to both inject a value and also to project values from the overall log message sum.  I don't know if that would be useful here, but it might be.

uc-crux-llvm/test/Test.hs
```diff
@@ -110,19 +112,47 @@ _exampleStruct =
...
   do
+    -- log structure is "Crux + (CruxLLVM + UCCruxLLVMTest)"
+    let ?injectCruxLogMessage = Left
+    let ?injectCruxLLVMLogMessage = Right . Left
+    let ?injectUCCruxLLVMTestLogMessage = Right . Right
+    let logMessageToSayWhat =
+          either Crux.cruxLogMessageToSayWhat
+            (either
+              Log.cruxLLVMLogMessageToSayWhat
+              ucCruxLLVMTestLogMessageToSayWhat)
```
kquick, line 148: I'm still concerned that this performs the conversion at the caller side, rather than the callee side, which will mean that if we use a different `LogAction` for the IDE that it will only be receiving `SayWhat` messages and have lost a lot of the structural info that would be useful.

My suggestion would be to create the following here:

    data UCCruxLLVMTestLogMsg = UCCruxLLVMLog UCCruxLLVMTestLogMessage
                              | CruxLLVMLog CruxLLVMLogMessage

Then parameterize `OutputConfig` by `msg` (where here in uc-crux-llvm Test.hs, `msg` is `UCCruxLLVMLog`) and require the initialization point (like here in Test.hs) to provide a `_logFrontEnd :: LogAction IO msg` (where your default can be a contravariant functor that converts to `SayWhat` that is applied to the `SayWhat` log action, but the initialization could change that to be an IDE-related thing.  

In Crux.Log you can then have:


    class  ToLogMsg o m where
        toLogMsg :: o -> m

    sayMsg :: (ToLogMsg o msg, Logs msg) => o -> IO ()
    sayMsg = LJ.writeLog (_logFrontEnd ?outputConfig) . toLogMsg

Using this approach, the caller uses `sayMsg` which defers any conversion to the `_logFrontEnd` log action, and that allows the initialization code to setup the type of log action that is appropriate for its configuration (default to user, JSON to IDE, etc.).  

Each front end then provides a `ToLogMsg` instance that helps convert other Crux module messages into the `msg` type.  For example, this uc-crux-llvm Test.hs might have:

    instance ToLogMsg UCCruxLLVMTestLogMessage UCCruxLLVMTestLogMsg
        where toLogMsg = UCCruxLLVMLog

     instance ToLogMsg CruxLLVMLogMessage UCCruxLLVMTestLogMsg
        where toLogMsg = CruxLLVMLog

which allows crux-llvm to log it's own messages as long as there is a ToLogMsg class constraint for its messages.

I'm sure that there's a hole or two in this approach, but the main advantage is allowing the initialization point to determine how the logged messages are converted and emitted (or not) rather than having to spread that information to all the call sites and only providing the actual log emitter with a generic form.

robdockins:
I think this is an improvement over the previous version.

kquick (approved):

robdockins (approved):

Ptival:
Rebased all previous commits as one. I checked the diff against the non-rebased version and they are the same. Once all tests are green, I will merge!

PR #764, opened 2021-06-14
Update crucible-go for crux lumberjack changes.

kquick:


Ptival (approved):
Assuming the previous behavior corresponded to `True`, this looks good.

PR #766, opened 2021-06-17
Enable manual trigger of crux-llvm binary tarballs

atomb:


.github/workflows/crux-llvm-build.yml
```diff
@@ -251,7 +251,7 @@ jobs:
           VERSION: ${{ needs.outputs.outputs.crux-llvm-version }}
 
       - uses: actions/upload-artifact@v2
-        if: startsWith(github.ref, 'refs/heads/release-')
+        if: startsWith(github.ref, 'refs/heads/release-') || github.event == 'workflow_dispatch'
```
ldettwy, line 254:
```suggestion
```

This logic isn't really doing anything for us except adding complexity. Unless there's something special about the artifact in question, it's best to always upload it as a result.

atomb: My thought was that we just wanted artifacts for releases and manual triggers, and not for every additional commit to an active PR (since there may be many of those). I'm not really opposed to just removing the condition if you don't think the large number of tarballs created will lead to any trouble.

Issue #763, opened 2021-06-10
Assumptions/Assertions inaccessible to execution features


robdockins:
The bookeeping for making logical assumptions or assertions about programs under simulation is currently handled by the `Lang.Crucible.Backend` module, which exposes the `IsBoolSolver` API.  This API is designed to cooperate closely with the steps of the symbolic simulator, and maintains stacks of assumptions that closely correspond to the control-flow paths followed by simulation.

Because this bookkeeping is done entirely internally, it isn't possible for an `ExecutionFeature` to monitor or take action when assumptions or assertions are made.  I think it would be better design to rework this part of the API so that these events are exposed as part of the execution of programs and reorganize, e.g. the `OnlineBackend` as an execution feature instead.  The motivation for this is the desire for Crux to keep track of events that occur along the execution of a program, and attach those events to proof goals.  Currently, there isn't an easy way to do this.

RyanGlScott:
@robdockins, was this issue addressed by #945, or did that solve a different problem?

robdockins:
That solved a different, but related, problem.

PR #767, opened 2021-06-21
Refactor `IsSymInterface` (attempt 1)

robdockins:
The goal of this refactoring is to address #763. To do this, we need to ensure that we only add assertions or assumptions to the logical program state via program states that can be observed by execution features.  Ultimately, we may remove the `IsBoolSolver` interface altogether.

To do this, we need to refactor the handling of errors, assertions, and assumptions so that they are always done in a context that can properly step the simulator abstract machine.

### DIFF
--- a/crucible-concurrency/crucibles/Main.hs
+++ b/crucible-concurrency/crucibles/Main.hs
@@ -9,9 +9,9 @@
 {-# LANGUAGE ImplicitParams #-}
 module Main where
 
-import qualified Data.Text.IO as T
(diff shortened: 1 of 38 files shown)


robdockins:
Ultimately, I think this refactoring is worth doing.  However, after working on it for a bit, I don't think it is the path of least resistance to solving #587, so I'm going to put it down for a while.

robdockins:
This is now quite out of date, and I don't know that we will go in this direction anyway.

PR #768, opened 2021-06-21
uc-crux-llvm: Check programs for "crash equivalence"

langston-barrett:
This commit adds another feature from the original paper
"Under-Constrained Symbolic Execution: Correctness Checking for Real
Code" by Ramos and Engler, namely the ability to check two versions of
the same function for "crash equivalence" or "crash ordering", where one
version is "crash less than" the other just if the all the undefined
behaviors present in the former are also present in the latter.

The idea is that this can be used e.g. to test patches: patches should
generally only *remove* instances of undefined behavior, not add them.

This can also be used as a generic, automated "rough equivalence" check.

There's also some refactoring & plumbing through of source locations, it's probably easiest to review the three commits separately.

### DIFF
--- a/uc-crux-llvm/README.md
+++ b/uc-crux-llvm/README.md
@@ -1,8 +1,12 @@
 # UC-Crux-LLVM
 
 UC-Crux-LLVM is a tool for under-constrained symbolic execution of C programs.
-It can be used to find undefined behavior and failing assertions, or for simple
-functions, to formally verify the absence of such behaviors.
+It can be used to:
+
+- find undefined behavior and failing assertions,
+- verify the absence of undefined behavior and failing assertions,
+- deduce sufficient function preconditions to avoid undefined behavior, and
+- check two versions of a program for crash-equivalence.
 
 **UC-Crux-LLVM is still in development.**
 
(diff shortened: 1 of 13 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Equivalence.hs
```diff
@@ -0,0 +1,261 @@
...
+-- This ignores:
+--
+-- * timeouts,
+-- * errors that are missing annotations, and
+-- * executions that encountered unimplemented features of UC-Crux-LLVM.
+--
+-- The behavior of this function is documented in the README, updates should be
+-- reflected there.
+--
+-- The composition of this function with 'crashDiffIsEmpty' is a partial order
```
travitch, line 140: Did this mean `crashDiffNonEmpty`? `crashDiffIsEmpty` returns `Bool`

travitch (approved):
Seems pretty reasonable to me. I just had two questions on the documentation

uc-crux-llvm/src/UCCrux/LLVM/Equivalence.hs
langston-barrett: Yeah, this is a bit vague - "composition" doesn't fully make sense here. I was thinking that a (total) function `[Located (Explanation ...)] -> [Located (Explanation ...)] -> Bool` is semantically comparable to a subset of the cross product `([Located (Explanation ...)], [Located (Explanation ...)])`, i.e. a binary relation (and in particular, a partial order) on `[Located (Explanation ...)]`. I can add more of that explanation inline here.

uc-crux-llvm/README.md
```diff
@@ -202,6 +206,22 @@ Uncertain results:
...
+
+UC-Crux-LLVM can check two different versions of the same program (or two
+implementations of the same interface) for *crash-equivalence*, meaning
+the two implementations are considered the same unless UC-Crux-LLVM can find a
+bug in one but not the other.
+
+The argument to the `--check-equivalence` flag is a second program to check for
+equivalence. Use `--entry-points` to check specific functions, or `--explore` to
+check all functions from both programs that share a name. If the
+`--strict-crash-equivalence` isn't passed to UC-Crux-LLVM, it actually checks
```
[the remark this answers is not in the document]
langston-barrett: Thanks for pointing out the awkward phrasing, I fixed this in the latest commit.

langston-barrett:
Hmm, I'm seeing
```
Run defsetup()  { echo import Distribution.Simple; echo main = defaultMain; }
warning: Nix search path entry 'channel:nixos-20.09' cannot be downloaded, ignoring
error: file 'nixpkgs' was not found in the Nix search path (add it using $NIX_PATH or -I), at (string):1:13
Error: Process completed with exit code 1.
```

langston-barrett:
I think the CI failures here are unrelated.

PR #770, opened 2021-07-07
crucible-llvm: Add llvm debug statements

glguy:


crucible-llvm/src/Lang/Crucible/LLVM/Extension/Syntax.hs
```diff
@@ -227,6 +227,43 @@ data LLVMStmt (f :: CrucibleType -> Type) :: CrucibleType -> Type where
      !(f (LLVMPointerType wptr)) {- Second pointer -} ->
      LLVMStmt f (BVType wptr)
 
+  -- | Debug information
+  LLVM_Debug ::
+    !(LLVM_Dbg f c)              {- Debug variant -} ->
+    LLVMStmt f UnitType
+
+-- | Debug statement variants - these have no semantic meaning
+data LLVM_Dbg f c where
```
travitch, line 236: I wasn't in the loop on the initial design - what is the motivation for having the `c` type parameter exposed? It is existentially quantified away in the `LLVM_Debug` statement, so there is no type-level connection there. All of the uses in this `LLVM_Dbg` type either fix that type or have an associated repr, so you can definitely already recover those types by pattern matching.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
```diff
@@ -1820,13 +1820,32 @@ callFunctionWithCont :: forall s arch ret a.
...
+          ptr <- case asScalar val of
+                   Scalar _ PtrRepr ptr -> pure ptr
+                   _ -> fail "dbg.addr: expected pointer"
+          extensionStmt (LLVM_Debug (LLVM_Dbg_Addr ptr lv di)) >> k
+
+     | L.ValSymbol "llvm.dbg.value" <- fn =
+       do (val, lv, di) <- dbgArgs args
+          case asScalar val of
+            Scalar _ repr v1 -> extensionStmt (LLVM_Debug (LLVM_Dbg_Value repr v1 lv di)) >> k
+            NotScalar -> fail "dbg.value: expected scalar"
```
travitch, line 1842: In these `fail` cases, is it possible to report more about what was actually received? It might speed up debugging when the LLVM folks give us some surprises

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
```diff
@@ -1847,6 +1866,30 @@ callFunctionWithCont instr tailCall_ fnTy fn args assign_f k
...
 
+-- | Match the arguments used by @dbg.addr@, @dbg.declare@, and @dbg.value@.
+dbgArgs :: 
+  [L.Typed L.Value] {- ^ debug call arguments -} ->
+  LLVMGenerator s arch ret (LLVMExpr s arch, L.DILocalVariable, L.DIExpression)
+dbgArgs args =
+  do (valArg, lvArg, diArg) <-
+       case args of
+         [valArg, lvArg, diArg] -> pure (valArg, lvArg, diArg)
+         _ -> fail "dbg expected 3 arguments"
```
travitch, line 1877: Similar request here - is it possible to include extra contextual information about what we got that was unexpected?

crucible-llvm/src/Lang/Crucible/LLVM/Extension/Syntax.hs
glguy: I did that because it made it fit into the -FC world of typeclasses more neatly. I'd be happy to replace this with something better. I originally didn't have that parameter and then seemed like I didn't have good support for fixing up the TraversableFC instance for LLVM statements without it.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
glguy: I'll look to see if there's some code that succinctly describes what LLVM AST we got .

travitch: Just showing the offending `ScalarView` (it doesn't have a Show instance right now, but looks like it could trivially have one) would probably do the trick.

crucible-llvm/src/Lang/Crucible/LLVM/Extension/Syntax.hs
travitch: Ah I was afraid of that. Fair enough. If it becomes an issue we can figure out what to do with it later

travitch (approved):

robdockins (approved):

PR #771, opened 2021-07-07
crux: add command-line arguments for websockets communication

Ptival:
Opening this draft pull request to discuss the design of such changes.
I'm not super happy with the current design, and would like to discuss some alternatives.

The current constraints are:
- `mkOutputConfig` is responsible for generating the `LogAction`s,
- `mkOutputConfig` is a pure function,
- `mkOutputConfig` returns the `OutputConfig`,
- `WebSockets.runClient` effective type is `(Connection -> IO a) -> IO a`, and is "bracketed", in the sense that the connection is only kept open while the callback `IO` action is executing.

These constraints make it quite awkward to set up logging to output to websockets: I must already have a handle to the `Connection` by the time I call `mkOutCfg`, so that it can use said connection when setting up the `LogAction`s, but the entire program continuation must be passed to `WebSockets.runClient` for that connection to stay alive for the duration of execution, resulting in the following:

https://github.com/GaloisInc/crucible/blob/e4d33e93d92bd8dc242faa694b503b6f737ab0a7/crux/src/Crux.hs#L185-L188

This is serviceable, but I'm considering whether we would benefit from an inversion of control in `mkOutCfg`: if instead of being `mkOutCfg :: ... -> OutputConfig` function, it was `withOutCfg :: ... -> (Logs msgs => IO a) -> IO a`, I think we would be able to call `WebSockets.runClient` from within it rather than prior to it, and guarantee that the continuation executes while the connection remains open.

I think the best people to ponder this are likely @kquick and @robdockins , so let me know what you think if/when you have some time to consider it.

Not particularly urgent, it will look about the same from the outside regardless of this internal design decision, so I can keep working without knowing what solution we end up choosing.

### DIFF
--- a/crux-llvm/crux-llvm.cabal
+++ b/crux-llvm/crux-llvm.cabal
@@ -86,6 +86,7 @@ library
     Paths_crux_llvm
 
   build-depends:
+    aeson,
     bv-sized,
     config-schema,
     data-binary-ieee754,
@@ -117,6 +118,31 @@ executable crux-llvm
     build-depends: unix
 
   other-modules: RealMain
+
+
+executable crux-llvm-for-ide
+  import: bldflags
+
+  hs-source-dirs: for-ide
+
+  build-depends:
+    aeson,
+    ansi-terminal,
+    crux-llvm,
+    lumberjack,
+    websockets >= 0.12
+
+  main-is: Main.hs
+
+  if os(windows)
+    hs-source-dirs: for-ide/windows
+  else
+    hs-source-dirs: for-ide/unix
+    build-depends: unix
+
+  other-modules:
+    Paths_crux_llvm
+    RealMain
 
 
 executable crux-llvm-svcomp
--- a/crux-llvm/for-ide/Main.hs
+++ b/crux-llvm/for-ide/Main.hs
@@ -1,0 +1,94 @@
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE TemplateHaskell #-}
+
+module Main (main) where
+
+import Control.Lens (makeLenses, set, view)
+import Crux (OutputConfig)
+import qualified Crux
+import Crux.Config.Common (CruxOptions)
+import Crux.LLVM.Config (LLVMOptions, llvmCruxConfig)
+import CruxLLVMMain
+  ( CruxLLVMLogging,
+    mainWithOptions,
+    withCruxLLVMLogging,
+  )
+import qualified Data.Aeson as JSON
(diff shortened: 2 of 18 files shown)


kquick:
I think your situation is not too dissimilar to the need for the out/err handles passed to remain valid.  

Assuming you need to re-use the same websocket connection and cannot make one on-demand inside the logger, you'll need to have some way to ensure that the connection stays valid for the duration.  You could do an inversion of control like you noted, or you could try embedding the IDE comm setup in the `mkOutCfg` passed, altering the output config returned from the default.

```haskell
mainIDEStartup = mainWithOutputConfig $
                 \mbCruxOpts -> 
                   let host = fromMaybe "localhost" ideHost mbCruxOpts
                       port = fromMaybe 80 idePort mbCruxOpts
                   in withOptionalIDECommunication host port $ \conn ->
                       let ocfg = defaultOutputconfig mbCruxOpts
                       in ocfg { _logMsg = optionallyOutputForIDE conn (_logMsg ocfg) } 
```

This doesn't necessarily solve your lifetime management constraints, but it keeps your general impact low.  The `withOptionalIDECommunication logic could further be sunk into the above if it made sense and simplified things; I probably mangled the logic with the `fromMaybe` handling and the net might be simpler than what I showed above.

You could even have a different executable `.hs` entrypoint  with the above in it which would keep the various websocket considerations outside of the core code.

[As a minor note, if you used `Maybe WS.Connection -> Maybe CruxOptions -> ...` instead of `Maybe CruxOptions -> Maybe WS.Connection -> ...` I think it might make your set of changes a bit smaller.]

Just some high level thoughts without being closely involved in all the considerations and constraints. Let me know if this is useful; happy to chat about approaches directly as well.

kquick (changes requested):
I made a general comment below.  Also note that using  separate executable might keep the ideHost/idePort arguments/parameters separate, which would ensure normal `crux-llvm` and `crux-mir` users didn't get confused by them.

Ptival:
How about this?

I turned the IDE-aware crux-llvm into a separate executable. At the price of a tiny bit of redundancy (to handle the Unix/Windows executable setup), it makes it so that the IDE-related options only show up for that executable.

I still have to fix these:
https://github.com/GaloisInc/crucible/blob/cb09b47cbf472ca479e5ca03e1f80c75de52593e/crux-llvm/for-ide/Main.hs#L46-L47
but I think I see how to do it.

kquick:
Yes, I like the direction this approach is heading.

robdockins (commented):
I agree with Kevin that a separate executable seems like the best way to handle this.

kquick (approved):

PR #772, opened 2021-07-13
crux: add --no-colors option

Ptival:
Something I thought could be useful, and it seem @RyanGlScott also thought so?

### DIFF
--- a/crux/src/Crux.hs
+++ b/crux/src/Crux.hs
@@ -195,27 +195,33 @@ showVersion ::
 showVersion nm ver = sayCrux (Log.Version nm ver)
 
 
--- | Create an OutputConfig for Crux, based on an indication of
--- whether colored output should be used, the normal and error output
--- handles, and the parsed CruxOptions.
+-- | Create an OutputConfig for Crux, based on an indication of whether colored
+-- output should be used, the normal and error output handles, and the parsed
+-- CruxOptions.
 --
--- If no CruxOptions are available (i.e. Nothing, as used in the
--- preliminary portions of loadOptions), then a default output stance
--- is applied; the default stance is described along with the
--- individual option below.
+-- If no CruxOptions are available (i.e. Nothing, as used in the preliminary
+-- portions of loadOptions), then a default output stance is applied; the
+-- default stance is described along with the individual option below.
 --
 -- The following CruxOptions affect the generated OutputConfig:
 --
+--  * noColorsErr    (default stance: False when the error handle supports
+--    colors, as reported by System.Console.ANSI.hSupportsANSIColor)
+--  * noColorsOut    (default stance: False when the output handle supports
+--    colors, as reported by System.Console.ANSI.hSupportsANSIColor)
 --  * printFailures  (default stance: True)
 --  * quietMode      (default stance: False)
 --  * simVerbose     (default stance: False)
 mkOutputConfig ::
   JSON.ToJSON msgs =>
-  Bool -> Handle -> Handle ->
+  (Handle, Bool) -> (Handle, Bool) ->
   (msgs -> SayWhat) -> Maybe CruxOptions ->
   OutputConfig msgs
-mkOutputConfig withColors outHandle errHandle logMessageToSayWhat opts =
-  let lgWhat = let la = LJ.LogAction $ logToStd withColors outHandle errHandle
+mkOutputConfig (outHandle, outWantsColor) (errHandle, errWantsColor) logMessageToSayWhat opts =
+  let consensusBetween wants maybeRefuses = wants && not (maybe False maybeRefuses opts)
+      outSpec = (outHandle, consensusBetween outWantsColor noColorsOut)
+      errSpec@(_, errShouldColor) = (errHandle, consensusBetween errWantsColor noColorsErr)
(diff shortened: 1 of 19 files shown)


crux/src/Crux.hs
```diff
@@ -213,7 +214,8 @@ mkOutputConfig ::
   (msgs -> SayWhat) -> Maybe CruxOptions ->
   OutputConfig msgs
 mkOutputConfig withColors outHandle errHandle logMessageToSayWhat opts =
-  let lgWhat = let la = LJ.LogAction $ logToStd withColors outHandle errHandle
+  let withColorsConsensus = withColors && not (maybe False noColors opts)
+      lgWhat = let la = LJ.LogAction $ logToStd withColorsConsensus outHandle errHandle
```
kquick, line 218: What if we dropped the `withColors` argument here? Callers can use the opts to set an override if  they want it.

For the default, we already have System.Console.ANSI available, so we could use `hSupportsANSIColor` to get the default value, at the cost of changing `mkOutputConfig` from a pure function to an IO monad operation.  This will have a little bit of impact, but the removal of the extra argument and the possibility that this will automatically default to doing the right thing could be pretty helpful.  What do you think?

Ptival: I actually tried that initially, but then I wasn't sure where to force the value in places such as:

https://github.com/GaloisInc/crucible/blob/ede38ab268f7eed2264b7276b06dfa3e6c6670e5/uc-crux-llvm/src/UCCrux/LLVM/Main.hs#L90-L93

but I guess what you'd do is more or less:
```haskell
mainWithOutputTo h =
  mainWithOutputConfig $
    set noColors True
    >>> Crux.mkOutputConfig False h h ucCruxLLVMLoggingToSayWhat
```
?

Basically I just want to make sure that we allow:
- a sensible default (e.g. what you suggest of detecting ANSI support),
- a way for a tool override the default for its main use case,
- a way for users to refuse colors.

One thing that I'm not sure about is whether the consensus should be colored output unless anyone disagrees, or whether, say, command-line options preference should prevail over tool's default preference.

We might want to rephrase the option/doc depending on the semantics we land upon.

Ptival: It's actually slightly less nice, because `mkOutputConfig` receives a `Maybe CruxOptions`, and so in the `Nothing` case, we can't set `noColors` to `True` short of creating a whole default `CruxOptions`...

kquick: I was thinking of the `set noColors true` technique for the forcing locations, but based on your latest observation, maybe it's better to keep the `Bool` but update `defaultOutputConfig` in Crux to use the `hSupportsANSIColor` in the IO monad.  That still allows for the "normal" config, but could be less invasive.  Does that seem reasonable?

To your semantics question, I'd assume that the `noColors` set to`True` would override any default coloring, regardless of origin (cfgfile, env-var, or cmd-line).

robdockins: Is `hSupportsANSIColor` portable to Windows? That would be my only concern.  Otherwise, it seems nice to query it when making a `defaultOutputConfig`.

kquick: The documentation isn't specific on this point, but I imagine this is a variation of the `hSupportsANSI`, which is reported to work for Windows.  These functions _are_ GHC specific, but I think we're already pretty far down that particular path.

Ptival: Updated. Currently the option is set if `stdout` supports ANSI. Should it instead be the conjunction of support on `stdout` and `stderr`, or does this seem better?

crux/src/Crux.hs
```diff
@@ -205,6 +205,7 @@ showVersion nm ver = sayCrux (Log.Version nm ver)
 --
 -- The following CruxOptions affect the generated OutputConfig:
 --
+--  * noColors       (default stance: False)
```
kquick, line 208: This default stance note needs to change now that we are using `hSupportsANSIColor`

crux/src/Crux.hs
kquick: Good point.  I suspect it would be best to keep those independently indicated down to the point where the output is configured and then configure each output according to the corresponding channel's color capabilities.  Otherwise one or the other will be suboptimal if the compatibilities don't match.

Ptival: Alright, so I rebased over the current master branch, with options to independently disable colors for output, errors, or both.

crux/src/Crux.hs
Ptival: Fixed!

kquick:
Looks great, thanks for driving this all the way through!

Ptival:
So, one caveat is that the help/version still use colors when `--no-colors` is passed.

The reason for that is that these two are called "early", before all options are parsed. I've looked into moving the color options to be early. It's doable, but if we want to avoid duplicating `mkOutputConfig`, then we kind of have to turn its `Maybe CruxOptions` input into something like `Maybe (Either ColorOptions CruxOptions)`.

It cascades into a lot of changes, so I think I will leave this for another pull request where we can discuss the design decisions of those changes.

kquick:
That's unfortunate, but I agree that's a minor issue and can be left for another time.

PR #773, opened 2021-07-13
crux: log prover milestones

Ptival:
(Not sure who the best reviewer is, so I'm pinging a few of you!)

For vscode-crux-llvm, we would like to have a notification when Crux knows what
goals it is about to start solving, and whenever each goal is started and
ended.

The goals are too verbose for displaying to the user, so they are not.  We
already have a progress bar system for the goals starting/ending, so these
messages are also not displayed to the user.

Because I did not want to build on the assumption that a goal ending was
necessarily the last started goal, I also made changes so that the ending of a
goal repeats the goal number.  This should be more future-proof in case we ever
process goals in parallel.  It can also help IDEs be a little more stateless.

Also, my apologies for inadvertently introducing 'StartedGoal' in a previous
commit where it was unused.

### DIFF
--- a/crux/src/Crux.hs
+++ b/crux/src/Crux.hs
@@ -648,6 +648,10 @@ doSimWithResults cruxOpts simCallback sym execFeatures profInfo monline goalProv
       let ctx = execResultContext res
       inFrame profInfo "<Prove Goals>" $ do
         todo <- getProofObligations sym
+        let showGoal = \ g -> ( T.pack (show (view labeledPred g))
+                              , T.pack (show (view labeledPredMsg g))
+                              )
+        sayCrux $ Log.Goals (showGoal . proofGoal <$> maybe [] goalsToList todo)
         when (isJust todo) $ sayCrux Log.AttemptingProvingVCs
         (nms, proved) <- goalProver cruxOpts ctx explainFailure todo
         mgt <- provedGoalsTree sym proved
--- a/crux/src/Crux.hs
+++ b/crux/src/Crux.hs
@@ -648,10 +648,7 @@ doSimWithResults cruxOpts simCallback sym execFeatures profInfo monline goalProv
       let ctx = execResultContext res
       inFrame profInfo "<Prove Goals>" $ do
         todo <- getProofObligations sym
-        let showGoal = \ g -> ( T.pack (show (view labeledPred g))
-                              , T.pack (show (view labeledPredMsg g))
-                              )
-        sayCrux $ Log.Goals (showGoal . proofGoal <$> maybe [] goalsToList todo)
+        sayCrux $ Log.ProofObligations (LogProofObligation <$> maybe [] goalsToList todo)
         when (isJust todo) $ sayCrux Log.AttemptingProvingVCs
         (nms, proved) <- goalProver cruxOpts ctx explainFailure todo
         mgt <- provedGoalsTree sym proved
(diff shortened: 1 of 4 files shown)


crux/src/Crux.hs
```diff
@@ -648,6 +648,10 @@ doSimWithResults cruxOpts simCallback sym execFeatures profInfo monline goalProv
       let ctx = execResultContext res
       inFrame profInfo "<Prove Goals>" $ do
         todo <- getProofObligations sym
+        let showGoal = \ g -> ( T.pack (show (view labeledPred g))
+                              , T.pack (show (view labeledPredMsg g))
+                              )
```
kquick, line 653: Would it be better to just log the goal object itself here and then let the logger decide which portions it would like to show?

crux/src/Crux/Goal.hs
```diff
@@ -188,7 +184,7 @@ proveGoalsOffline adapters opts ctx explainFailure (Just gs0) = do
     failfast = proofGoalsFailFast opts
 
     go :: SupportsCruxLogMessage msgs
-       => (Integer -> IO (), IO ())
+       => (Integer -> IO (), Integer -> IO ())
```
kquick, line 187: This type is repeated in a couple of places; would a `type` alias simplify the declarations, or is it still so few places that there's no real benefit?

crux/src/Crux/Log.hs
```diff
@@ -177,8 +185,8 @@ cruxLogMessageToSayWhat SimulationTimedOut =
 cruxLogMessageToSayWhat SkippingUnsatCoresBecauseMCSatEnabled =
   cruxWarn "Warning: skipping unsat cores because MC-SAT is enabled."
 
-cruxLogMessageToSayWhat (StartedGoal goalNumber) =
-  cruxOK (T.pack ("Started goal " ++ show goalNumber))
```
kquick, line 181: This message already existed; is this an intentional behavioral change for existing uses?

kquick (commented):
In general I like this and it's along the lines of the type of changes I would have expected to need for good IDE interaction.

crux/src/Crux/Log.hs
Ptival: I accidentally introduced this message in 970ed02d76ca75e95ba8bdfee490fffeebba917e.

I don't think it corresponded to anything pre-existing, and so this message was actually never printed except in my personal logging extension attempts. I had meant to stash this away, but missed it when shuffling things around.

crux/src/Crux/Goal.hs
Ptival: Actually this made me come up with fairly nice type synonyms. Will update!

crux/src/Crux.hs
Ptival: Reasonable question. I had two concerns:

1. JSON serializability, but since we had talked about putting `Doc`s in the log data types, I think it no longer makes sense to try and have the `...LogMessage` data types have derivable JSON instances (as I originally did), and so storing more complex data in the log messages would no longer be a concern accordingly.

2. Type indices, those goals are at the very least indexed by `sym`, so we would have to choose whether to existentially wrap it in the `Goals` constructor, or reflect it in the type of `CruxLogMessage`. With the former, we no longer get deriving for the `Generic` instance, so we must write all `ToJSON` cases by hand without `genericToEncoding`. With the latter, I'm not sure whether it creates problems, it's just a little more noise but maybe it's all fine?

I'm attempting this change locally, will report on the outcome.

kquick: OK.  The second concern seems fairly significant, so I wouldn't go to extraordinary lengths to make this work.  In general, my thoughts are that whenever reasonable, we can defer the output representation to the logger instead of the point of log, but it will be nice to keep these relatively simple as well rather than incurring  a significant technical burden just for logging/output.

Ptival: I have attempted the existential wrapping, and it's not too bad: d91b9978b83feb6e9ba8b97b5419d4080e7d0ec7

We need to have `sym` be partially instantiated to `ExprBuilder t st fs` so as to have the `Show` instance without adding constraints. I'm happy with this change.

kquick:
This looks really good to me, thanks @Ptival !

robdockins (approved):

PR #775, opened 2021-07-19
Replace use of `which` with `findExecutable`

### LINKED ISSUE #774: `crux-llvm`: Use `findExecutable` instead of `/usr/bin/which`
When attempting to compile a C file, `crux-llvm` will locate `clang` by shelling out to `/usr/bin/which`:

https://github.com/GaloisInc/crucible/blob/ede38ab268f7eed2264b7276b06dfa3e6c6670e5/crux-llvm/src/Crux/LLVM/Compile.hs#L64

This technically works, but it's a bit delicate, since one could hypothetically invoke `crux-llvm` on a machine where `which` doesn't live in `/usr/bin` (or perhaps doesn't live anywhere at all). Moreover, it's not necessary to resort to an external executable for this task, as the `process` library exposes a [`findExecutable`](https://hackage.haskell.org/package/directory-1.3.6.2/docs/System-Directory.html#v:findExecutable) function for this very purpose. This ticket tracks replacing this use of `which` with `findExecutable` (or a related function from `process`).

RyanGlScott:
Fixes #774.

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Compile.hs
+++ b/crux-llvm/src/Crux/LLVM/Compile.hs
@@ -61,7 +61,12 @@ anyCPPFiles = any isCPlusPlus
 getClang :: IO FilePath
 getClang = attempt (map findExecutable clangs)
   where
-  clangs = [ "clang", "clang-4.0", "clang-3.6", "clang-3.8", "clang-7", "clang-8" ]
+  clangs = "clang"
+         : [ "clang-" ++ ver
+           | ver <- ["3.6", "3.7", "3.8", "3.9", "4.0", "5.0", "6.0"]
+                    -- Starting with LLVM 7, the version numbers used in binaries only display the major version. See
+                    -- https://releases.llvm.org/7.0.0/tools/clang/docs/ReleaseNotes.html#non-comprehensive-list-of-changes-in-this-release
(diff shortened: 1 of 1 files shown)


crux-llvm/src/Crux/LLVM/Compile.hs
```diff
@@ -59,22 +59,21 @@ anyCPPFiles = any isCPlusPlus
 -- (NOTE: do not look for environment var "CLANG". That is assumed
 --  to be tried already.)
 getClang :: IO FilePath
-getClang = attempt (map inPath clangs)
+getClang = attempt (map findExecutable clangs)
   where
-  inPath x = head . lines <$> readProcess "/usr/bin/which" [x] ""
-  clangs   = [ "clang", "clang-4.0", "clang-3.6", "clang-3.8", "clang-7", "clang-8" ]
+  clangs = [ "clang", "clang-4.0", "clang-3.6", "clang-3.8", "clang-7", "clang-8" ]
```
robdockins, line 64: May as well expand this list while we're at it.

robdockins (approved):
Let's add `clang-X` strings for more recent clangs, but otherwise looks good to me.

kquick (approved):
LGTM!

PR #776, opened 2021-07-19
When debug metadata doesn't match ignore it

glguy:
There are a lot of old formats for debug metadata. This change makes using the old formats non-fatal.


glguy:
I left in the descriptive text even though it isn't currently printed anyway to help someone when they are upgrading crucible to handle new metadata formats.

robdockins (approved):
I'd like to see some comments explaining which version(s) of LLVM you are targeting with the `dbgArgs` function, and mentioning that we are deliberately ignoring ones that don't look like those.  That will help guide whoever looks at this next.

Otherwise, seems OK to me.

PR #782, opened 2021-07-21
Add debug-intrinsics option to control translating llvm.dbg statements

### LINKED ISSUE #778: `crux-llvm` crashes (Input block type ... does not match expected) during SSA conversion
Consider this program, slightly minimized from SV-COMP's [`soft_float_1-2a.c.cil.c`](https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/bitvector/soft_float_1-2a.c.cil.c):

```c
unsigned int addflt(unsigned int a , unsigned int b )
{
  unsigned int res ;
  unsigned int ma ;
  unsigned int mb ;
  unsigned int delta ;
  int ea ;
  int eb ;
  unsigned int tmp ;
  unsigned int __retres10 ;

  {
  if (a < b) {
    tmp = a;
    a = b;
    b = tmp;
  } else {

  }
  if (! b) {
    __retres10 = a;
    goto return_label;
  } else {

  }
  {
  ma = a & ((1U << 24U) - 1U);
  ea = (int )(a >> 24U) - 128;
  ma = ma | (1U << 24U);
  mb = b & ((1U << 24U) - 1U);
  eb = (int )(b >> 24U) - 128;
  mb = mb | (1U << 24U);
  delta = ea - eb;
  mb = mb >> delta;
  }
  if (! mb) {
    __retres10 = a;
    goto return_label;
  } else {

  }
  ma = ma + mb;
  if (ma & (1U << 25U)) {
    if (ea == 127) {
      __retres10 = 4294967295U;
      goto return_label;
    } else {

    }
    ea = ea + 1;
    ma = ma >> 1U;
  } else {

  }
  {
  ma = ma & ((1U << 24U) - 1U);
  res = ma | ((unsigned int )(ea + 128) << 24U);
  }
  __retres10 = res;
  return_label: /* CIL Label */
  return (__retres10);
}
}
int main(void)
{
  if (addflt(2,2)) {
    return 0;
  } else {
    return 1;
  }
}
```

Before commit 690d989a3c1760f1e110520f7701d7684183ee00, `crux-llvm` would verify this program. After commit 690d989a3c1760f1e110520f7701d7684183ee00, however, `crux-llvm` crashes:

```
$ cabal run exe:crux-llvm -- ../../../Crux/test.c 
Up to date
Input block type [IntrinsicRepr LLVM_pointer [BVRepr 32], IntrinsicRepr LLVM_pointer [BVRepr 32], IntrinsicRepr LLVM_pointer [BVRepr 32]] does not match expected [IntrinsicRepr LLVM_pointer [BVRepr 32], IntrinsicRepr LLVM_pointer [BVRepr 32]]:
while SSA converting function addflt
  error, called at src/Lang/Crucible/CFG/SSAConversion.hs:976:24 in crucible-0.5-inplace:Lang.Crucible.CFG.SSAConversion
```

RyanGlScott:
Unfortunately, Clang can sometimes generate ill-scoped `llvm.dbg` statements (see #778), so this patch adds a `debug-intrinsics` option that controls whether `llvm.dbg` statements should be translated at all. The default value for `crux-llvm` is `no`, as `crux-llvm` does not yet make use of `llvm.dbg` information. We may wish to revisit this choice in the future should the upstream LLVM bug (https://bugs.llvm.org/show_bug.cgi?id=51155) be fixed.

Fixes #778.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
@@ -83,6 +83,7 @@ module Lang.Crucible.LLVM.Translation
   , translateModule
 
   , module Lang.Crucible.LLVM.Translation.Constant
+  , module Lang.Crucible.LLVM.Translation.Options
   , module Lang.Crucible.LLVM.Translation.Types
   ) where
 
@@ -116,6 +117,7 @@ import           Lang.Crucible.LLVM.Translation.Aliases
 import           Lang.Crucible.LLVM.Translation.Constant
 import           Lang.Crucible.LLVM.Translation.Expr
 import           Lang.Crucible.LLVM.Translation.Monad
+import           Lang.Crucible.LLVM.Translation.Options
 import           Lang.Crucible.LLVM.Translation.Instruction
 import           Lang.Crucible.LLVM.Translation.Types
 import           Lang.Crucible.LLVM.TypeContext
@@ -187,7 +189,7 @@ buildRegTypeMap m0 bb = foldM stmt m0 (L.bbStmts bb)
 
 
 -- | Generate crucible code for each LLVM statement in turn.
-generateStmts :: (?laxArith :: Bool, ?debugIntrinsics :: Bool)
+generateStmts :: (?transOpts :: TranslationOptions)
         => TypeRepr ret
         -> L.BlockLabel
         -> [L.Stmt]
@@ -308,7 +310,7 @@ findFile _ = Nothing
 -- | Lookup the block info for the given LLVM block and then define a new crucible block
 --   by translating the given LLVM statements.
 defineLLVMBlock
-        :: (?laxArith :: Bool, ?debugIntrinsics :: Bool)
+        :: (?transOpts :: TranslationOptions)
         => TypeRepr ret
         -> Map L.BlockLabel (LLVMBlockInfo s)
         -> L.BasicBlock
@@ -326,7 +328,7 @@ defineLLVMBlock _ _ _ = fail "LLVM basic block has no label!"
 --
 --   This step introduces a new dummy entry point that simply jumps to the LLVM entry
 --   point.  It is inconvenient to avoid doing this when using the Generator interface.
-genDefn :: (?laxArith :: Bool, ?debugIntrinsics :: Bool)
+genDefn :: (?transOpts :: TranslationOptions)
         => L.Define
         -> TypeRepr ret
         -> LLVMGenerator s arch ret (Expr ext s ret)
@@ -357,8 +359,7 @@ genDefn defn retType =
 --
 -- | Translate a single LLVM function definition into a crucible CFG.
 transDefine :: forall arch wptr.
-               ( HasPtrWidth wptr, wptr ~ ArchWidth arch
-               , ?laxArith :: Bool, ?optLoopMerge :: Bool, ?debugIntrinsics :: Bool )
+               (HasPtrWidth wptr, wptr ~ ArchWidth arch, ?transOpts :: TranslationOptions)
             => HandleAllocator
             -> LLVMContext arch
             -> L.Define
@@ -377,7 +378,7 @@ transDefine halloc ctx d = do
                   f = genDefn d retType
     sng <- newIONonceGenerator
     (SomeCFG g, []) <- defineFunctionOpt InternalPos sng h def $ \ng cfg ->
-      if ?optLoopMerge then earlyMergeLoops ng cfg else return cfg
+      if optLoopMerge ?transOpts then earlyMergeLoops ng cfg else return cfg
     case toSSA g of
       C.SomeCFG g_ssa -> return (symb, (decl, C.AnyCFG g_ssa))
 
@@ -387,7 +388,7 @@ transDefine halloc ctx d = do
 -- | Translate a module into Crucible control-flow graphs.
 -- Note: We may want to add a map from symbols to existing function handles
 -- if we want to support dynamic loading.
-translateModule :: (?laxArith :: Bool, ?optLoopMerge :: Bool, ?debugIntrinsics :: Bool)
+translateModule :: (?transOpts :: TranslationOptions)
                 => HandleAllocator -- ^ Generator for nonces.
                 -> GlobalVar Mem   -- ^ Memory model to associate with this context
                 -> L.Module        -- ^ Module to translate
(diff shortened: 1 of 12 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
```diff
@@ -187,7 +187,7 @@ buildRegTypeMap m0 bb = foldM stmt m0 (L.bbStmts bb)
 
 
 -- | Generate crucible code for each LLVM statement in turn.
-generateStmts :: (?laxArith :: Bool)
+generateStmts :: (?laxArith :: Bool, ?debugIntrinsics :: Bool)
```
RyanGlScott, line 190: Since the option controlling lax arithmetic was already being passed around as an implicit parameter, I adopted the same convention for `debugIntrinsics`. Let me know if you'd prefer something else.

crux-llvm/src/Crux/LLVM/Config.hs
```diff
@@ -161,6 +162,9 @@ llvmCruxConfig = do
          loopMerge <- Crux.section "opt-loop-merge" Crux.yesOrNoSpec False
                         "Insert merge blocks in loops with early exits (i.e. breaks or returns). This may improve simulation performance."
 
+         debugIntrinsics <- Crux.section "debug-intrinsics" Crux.yesOrNoSpec False
```
RyanGlScott, line 165: I'm unclear on what determines if something deserves to be both a Crux configuration file option and a command-line option. For now, I went with making `debug-intrinsics` solely a configuration file option, but I can also make it a command-line option if you prefer.

crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
robdockins: I think this is fine.  If we end up with more options like this, we should probably bundle them into a record or something.

crux-llvm/src/Crux/LLVM/Config.hs
robdockins: I don't think we have a principled rule, just a gut feeling about which things are common enough to be worth putting in command line options.  I think this is find as a config-file only option.

robdockins (changes requested):
Looks OK to me, with the caveat that we'll need to update downstream clients to provide the new option, including in the test suites.

RyanGlScott:
OK. Should I interpret your request for changes as a call to try building downstream clients first before merging this PR?

robdockins:
> OK. Should I interpret your request for changes as a call to try building downstream clients first before merging this PR?

No, just to make sure the tests in this repo pass.  Probably worth opening a ticket about this for SAW, though to get it on the radar.

RyanGlScott:
I've opened GaloisInc/saw-script#1388. My most recent push addresses the `crucible-llvm` test suite failures.

robdockins (approved):
Looks OK to me, assuming CI passes.

RyanGlScott:
In an attempt to tame the zoo of LLVM translation–related functions, I've introduced a `TranslationOptions` record data type, which is directly inspired by the existing `MemOptions` type.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Options.hs
```diff
@@ -0,0 +1,55 @@
...
+--   LLVM translation.
+data TranslationOptions
+  = TranslationOptions
+    { debugIntrinsics :: !Bool
+      -- ^ Should we translate @llvm.dbg@ intrinsic statements? The upside to
+      --   translating them is that they encode debugging information about a
+      --   program that can be useful for external clients. The downside is
+      --   that they can bloat the size of translated programs, despite being
+      --   no-ops during simulation.
+    , laxArith :: !Bool
```
robdockins, line 26: I think you've reversed the sense of the `laxArith` option in these comments.  When `False`, we apply strict arithmetic checking; when `True` we omit those checks.

I'm open to the idea that we should reverse the entire sense of this option ad call it something else; but the easiest thing to do would be to update these comments.

robdockins:
I like the changes with the translation options, modulo my remark above about the `laxArithmetic` documentation.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Options.hs
RyanGlScott: Oops, that's a a silly oversight on my part. I've changed "Should we produce proof obligations" to "Should we omit proof obligations".

RyanGlScott:
@glguy, are you OK with pursuing this direction of travel (as part of the approach suggested in https://github.com/GaloisInc/crucible/issues/778#issuecomment-885903648)? I want to make sure that we address any remaining concerns you have before going forward with this.

glguy:
I think it's fine as a stop-gap solution. I'd rather see debug statements that move ahead of their definitions simply removed than all debug information discarded. (Even better the more complete analysis pass Rob has been imaging). I think it's wrong to disable debug translation by default and that it makes sense to turn it off for the test cases that need it turned off. We should be testing as much of the code as possible.

AFAIK debugging intrinsics are only generated when the code is compiled with -g. If we're only going to strip those intrinsics out, why are we compiling with them turned on? (Am I simply wrong about that situation?)

robdockins:
The debug intrinsics are only emitted with `-g`; but if you turn it off, you also lose source locations, which are super useful.

RyanGlScott:
To be clear, I'd also like to have a pass during translation which turns use-before-def `llvm.dbg.value` statements into something that Crucible can use. The main reasons I'm pursuing this approach are that:

1. I need _some_ way to make programs like the one above work.
2. I don't have a sense for how much work it would take to implement the aforementioned translation pass, and I don't want to hold things up on that. I'd advocate for keeping #778 open as a reminder to implement this pass (or open a new issue for it, if you'd prefer).
3. Even in a world where we do have such a translation pass, it's conceivable that an expert user may wish to strip out `llvm.dbg` statements anyway for optimization purposes. I tend to agree that, ideally, including `llvm.dbg` statements would be the default behavior, but we need to make sure that they work robustly before we can make it the default.
4. As @robdockins mentions, I don't think telling users not to compile with `-g` is a viable option, given that LLVM debug metadata is used for a variety of different purposes.

robdockins:
Let's merge this and open a new ticket about doing def/use analysis during LLVM->Crucible translation so we can detect and discard (or eventually fix up) `dbg.value` statements if they get ahead of their defs.

RyanGlScott:
I've opened #798 to track the remaining work.

PR #786, opened 2021-07-22
`crucible-llvm`: Add override for `exit`

### LINKED ISSUE #784: Add override for `exit`
The [`Problem01_label00`](https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/eca-rers2012/Problem01_label00.c) SV-COMP benchmark makes use of `exit`, which Crucible currently cannot handle due to a lack of an appropriate override. We should add one.

RyanGlScott:
Fixes #784.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
@@ -487,6 +487,18 @@ callAssert mvar sym (Empty :> _pfn :> _pfile :> _pline :> ptxt ) =
         let err = AssertFailureSimError "Call to assert()" (UTF8.toString txt)
         liftIO $ addFailedAssertion sym err
 
+callExit :: (IsSymInterface sym)
+         => sym
+         -> RegEntry sym (BVType 32)
+         -> OverrideSim p sym ext r args ret (RegValue sym UnitType)
+callExit sym ec = liftIO $
+  do cond <- bvEq sym (regValue ec) =<< bvLit sym knownNat (BV.zero knownNat)
+     -- If the argument is non-zero, throw an assertion failure. Otherwise,
+     -- simply stop the current thread of execution.
+     assert sym cond "Call to exit() with non-zero argument"
+     loc <- getCurrentProgramLoc sym
+     abortExecBecause $ EarlyExit loc
+
 callPrintf
   :: (IsSymInterface sym, HasPtrWidth wptr, HasLLVMAnn sym)
   => sym
@@ -669,6 +681,16 @@ llvmAbortOverride =
      do let err = AssertFailureSimError "Call to abort" ""
         liftIO $ assert sym (falsePred sym) err
   )
+
+llvmExitOverride
+  :: forall sym p
+   . (IsSymInterface sym)
+  => LLVMOverride p sym
+         (EmptyCtx ::> BVType 32)
+         UnitType
+llvmExitOverride =
+  [llvmOvr| void @exit( i32 ) |]
+  (\_ sym args -> Ctx.uncurryAssignment (callExit sym) args)
 
 llvmGetenvOverride
   :: (IsSymInterface sym, HasPtrWidth wptr)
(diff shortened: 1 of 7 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -487,6 +487,20 @@ callAssert mvar sym (Empty :> _pfn :> _pfile :> _pline :> ptxt ) =
         let err = AssertFailureSimError "Call to assert()" (UTF8.toString txt)
         liftIO $ addFailedAssertion sym err
 
+callExit :: (IsSymInterface sym)
+         => sym
+         -> RegEntry sym (BVType 32)
+         -> OverrideSim p sym ext r args ret (RegValue sym UnitType)
+callExit sym ec = liftIO $
```
robdockins, line 494: This is an odd pattern.  I think I'd rather see an assertion that the code is zero, followed by an `abortExecBecause`.

crux-llvm/test-data/golden/T784-fail.good
```diff
@@ -0,0 +1,8 @@
+[Crux] Found counterexample for verification goal
+[Crux]   test-data/golden/T784-fail.c:4:3: error: in main
+[Crux]   exit
+[Crux] 
+[Crux] Found counterexample for verification goal
```
robdockins, line 5: I don't think we should be seeing this unreachable code error.

robdockins: Humm... that also defies my expectations.  I wonder if this is because "assume after assert" is not the default anymore.

robdockins: Yeah, a quick test confirms that this is related to the `assertThenAssume` option.

We should probably make concretely false assertions abort regardless of that option.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
RyanGlScott: Ah, good suggestion. As an added bonus, this no longer exhibits the weird unreachable code error seen in https://github.com/GaloisInc/crucible/pull/786#discussion_r675112827.

crux-llvm/test-data/golden/T784-fail.good
RyanGlScott: Right, it looks like `abort()` is doing things quite differently from how the override for `exit()` now works. This PR is technically only about `abort()`, so would you be OK if I opened a separate issue about your suggested `assertThenAssume` changes?

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -487,6 +487,20 @@ callAssert mvar sym (Empty :> _pfn :> _pfile :> _pline :> ptxt ) =
...
+  do cond <- bvEq sym (regValue ec) =<< bvLit sym knownNat (BV.zero knownNat)
+     -- First, check that the argument is zero. If it is zero, we wish to stop
+     -- execution, without failing.
+     assert sym cond "Call to exit()"
+     -- If we get to this point, the argument is non-zero, so fail.
```
robdockins, line 499: With `assert` you state the condition that must be true, and a message to print when it is _not_, so here we'd wany to say "called ext() with nonzero condition".

The abort below is for _all_ cases, 0 and nonzero to kill this thread of execution, so it shouldn't be an `AssertionFailure`.  Maybe we should add a new abort reason for this case, as `InfeasibleBranch` isn't really right either.

crux-llvm/test-data/golden/T784-fail.good
RyanGlScott: I've opened https://github.com/GaloisInc/crucible/issues/787 for this.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
RyanGlScott: Thanks for the tips! I've added an `EarlyExit` constructor to `AbortExecReason`. Let me know if you think it deserves to be more descriptive than that.

robdockins: Looks good.

robdockins (approved):

PR #788, opened 2021-07-22
Symbolic I/O

travitch:
This PR implements a set of overrides supporting symbolic I/O operations. The core is implemented as a set of overrides in the new crucible-symio package. These core overrides are language neutral, and should be suitable for nearly any frontend that needs to support I/O. The overrides are instantiated for the LLVM frontend in crucible-llvm.


crucible-symio/src/What4/CachedArray.hs
```diff
@@ -0,0 +1,837 @@
...
+      -- | an optional override for the abstract domain of the index
+    , symIdxAbs :: Maybe (AbsIndex ctx)
+    }
+
+deriving instance (W4.IsSymExprBuilder sym => Eq (SymIndex sym ctx))
+deriving instance (W4.IsSymExprBuilder sym => Ord (SymIndex sym ctx))
+
+-- | An offset is an index into the last element of the array index
+-- A value range is always representable as a base + offset
+data SymOffset sym ctx where
```
RyanGlScott, line 346:
```suggestion
newtype SymOffset sym ctx where
```

crucible-symio/src/What4/CachedArray.hs
```diff
@@ -0,0 +1,837 @@
...
+  SymOffset :: W4.SymExpr sym (CtxFirst ctx) -> SymOffset sym ctx
+
+data FirstIndex ctx where
```
RyanGlScott, line 349:
```suggestion
newtype FirstIndex ctx where
```

crucible-symio/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Revision history for crucible-vfs
```
RyanGlScott, line 1:
```suggestion
# Revision history for crucible-symio
```

crucible-symio/LICENSE
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2020 Galois Inc.
```
RyanGlScott, line 1: Perhaps mention 2021 as well here?

crucible-symio/src/What4/CachedArray.hs
```diff
@@ -0,0 +1,837 @@
...
+data CachedArray sym (ctx :: Ctx.Ctx W4.BaseType) (tp :: W4.BaseType) where
+  CachedArray ::
+    {
+      arrMap :: IM.IntervalsMap AbsIntervalEnd ctx (PMuxTree sym (ArrayEntry sym ctx tp))
+    , arrConstraints :: forall a. (NonEmptyCtx ctx => a) -> a
+    , arrTypeRepr :: W4.BaseTypeRepr tp
+    , _arrNonce :: ArrayNonce
+    } -> CachedArray sym ctx tp
+
+instance TestEquality (W4.SymExpr sym) => Eq (CachedArray sym idx tp) where
```
RyanGlScott, line 297: Is the `TestEquality (W4.SymExpr sym)` constraint actually needed here?

crucible-symio/src/What4/CachedArray.hs
```diff
@@ -0,0 +1,837 @@
...
+
+
+incNonceEntry ::
+  ArrayEntry sym ctx tp ->
+  IO (ArrayEntry sym ctx tp)
+incNonceEntry (ArrayEntry vals _) = do
+  nonce <- freshArrayNonce
+  return $ ArrayEntry vals nonce
+
+instance TestEquality (W4.SymExpr sym) => Eq (ArrayEntry sym ctx tp) where
```
RyanGlScott, line 325: Is the `TestEquality (W4.SymExpr sym)` constraint actually needed here?

crucible-symio/src/What4/CachedArray.hs
```diff
@@ -0,0 +1,837 @@
...
+  e1 == e2 = entryNonce e1 == entryNonce e2
+
+instance OrdF (W4.SymExpr sym) => Ord (ArrayEntry sym ctx tp) where
```
RyanGlScott, line 328: Is the `OrdF (W4.SymExpr sym)` constraint actually needed here?

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -0,0 +1,463 @@
...
+    , fDescMap :: Map.Map Natural (W4P.PartExpr (W4.Pred sym) (SymIO.FileHandle sym ptrW))
+    } -> FDescMap sym ptrW
+
+-- | An existential wrapper around an 'OverrideSim' action
+--
+-- This enables us to make explicit that all of the type variables are free and
+-- can be instantiated at any types since this override action has to be
+-- returned from the 'initialLLVMFileSystem' function.
+data SomeOverrideSim sym a where
+  SomeOverrideSim :: (forall p ext rtp args ret . () -> OverrideSim p sym ext rtp args ret a) -> SomeOverrideSim sym a
```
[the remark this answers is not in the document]
travitch: I couldn't convince GHC to let me say what I wanted without it at the time, but it seems like I added enough type signatures elsewhere/changed enough other things that it isn't needed

crucible-symio/src/Lang/Crucible/SymIO/Loader.hs
```diff
@@ -0,0 +1,167 @@
...
+instance X.Exception FileSystemLoadError
+
+-- | The specification for the symbolic contents of a file in the symbolic
+-- filesystem
+--
+-- There will be multiple specifications including:
+--
+--   * Complete symbolic file specifications (including concrete regions)
+--   * Symbolic overlays on otherwise concrete files
+data SymbolicFileContents =
```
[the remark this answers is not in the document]
travitch: My inclination is to leave this as data because I'm planning on adding more fields in the near term (e.g., specifying constraints over ranges of symbolic values)

travitch:
Aside from the usual review feedback, I'd appreciate commentary on:
* How is the interface for specifying initial filesystem contents?
* Do the semantics for standard I/O handling in crux-llvm make sense?
* Should crucible-symio know anything about standard I/O? My inclination is that it shouldn't (and that should be handled by the code instantiating it), but I'd rather defer those changes to avoid having this large PR diverge further.

crucible-symio/src/Data/Parameterized/IntervalsMap.hs
```diff
@@ -0,0 +1,325 @@
...
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE FunctionalDependencies #-}
+
+-- must come after TypeFamilies, see also https://gitlab.haskell.org/ghc/ghc/issues/18006
+{-# LANGUAGE NoMonoLocalBinds #-}
+
+module Data.Parameterized.IntervalsMap
```
robdockins, line 33: This seems like we should sink it down into `parameterized-utils`

crux-llvm/test-data/golden/stdio/write-stdout.good
```diff
@@ -0,0 +1 @@
+[Crux] Overall status: Valid.
```
robdockins, line 1: Do we not expect to see the actual output here?

robdockins (commented):
A couple of general comments, just looking at the crux-llvm integration, so far (I haven't dived into the main changes yet).

1) I don't know it belongs in this PR necessarily, but we should open a ticket about implementing `errno`.
2) We should clearly indicate in the documentation that the symbolic I/O interface is very experimental.  I think we'll probably need to adjust how filesystem contents, etc, are specified as we actually use this stuff going forward, so let's be upfront about the fact that this is likely to change.
3) I think conservative defaults for the SymIO package itself are good, and that hooking up stdin etc. is also good for crux-llvm in particular.
4) We should have some tests that exercise symbolic files.

crux-llvm/test-data/golden/stdio/write-stdout.good
travitch: ha ha I was wondering about that too. It turns out that the test suite only captures output from the handles passed to `Crux.mkOutputConfig`. And those both go to a special other handle in the test suite.

So we could hook the output up to those handles instead of actual stdout/stderr. I wasn't sure if that was a good idea or not, given that there seems to be other processing on those handles. I'm happy to switch it over if we think that is reasonable.

travitch:
> A couple of general comments, just looking at the crux-llvm integration, so far (I haven't dived into the main changes yet).
> 
>     1. I don't know it belongs in this PR necessarily, but we should open a ticket about implementing `errno`.

Yes, this and tracking the working directory are high on my list

>     2. We should clearly indicate in the documentation that the symbolic I/O interface is very experimental.  I think we'll probably need to adjust how filesystem contents, etc, are specified as we actually use this stuff going forward, so let's be upfront about the fact that this is likely to change.

Good idea - totally agree

>     3. I think conservative defaults for the SymIO package itself are good, and that hooking up stdin etc. is also good for crux-llvm in particular.
> 
>     4. We should have some tests that exercise symbolic files.

crux-llvm/test-data/golden/stdio/write-stdout.good
robdockins: Yeah, we should use the `mkOutputConfig` handles.  That way we can capture `stdout` into IDE windows, etc, as we go forward.

travitch: Okay, fixed. I think there is only one handle there, unfortunately, which means stdout and stderr are combined for now.

robdockins: I think that's fine.  If we enhance the output configuration to have separate handles, we can change this then.

travitch:
I'm also hoping to either squash all of this or force push a rebase that turns it into a small number of commits

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -524,3 +540,21 @@ are here, but they are arranged to not collide with any valid names for actual
 files in the symbolic filesystem.
 
 -}
+
+{- Note [File Descriptor Sequence]
```
robdockins, line 544: The documentation for `dup` and such indicates that the order in which fds are allocated is pretty deterministic.  Real code can (although usually doesn't) depend on that behavior.  I think this design is fine for most purposes, but we should probably start an "errata" list or something.

Other designs we could consider:
1) Accurately following UNIX behavior (allocating the smallest unused fd)
2) Protect the abstraction further by allocating fresh non-dereferencable pointers instead of actually using small integers.

2) is even less accurate than what we are doing now, but is more aggressive about maintaining the abstraction.

travitch: I was thinking about the implications of just making the file descriptor source be symbolic so that it could take part in branching and merging. I'm not sure what the implications would be...

As for 2, I think it is noted somewhere but we would have to build an escape hatch to support stdin/stdout/stderr. That isn't necessarily a problem, but just another consideration.

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -0,0 +1,560 @@
...
+    -- out; we don't have any guards against overflow on that right now
+    , fDescMap :: Map.Map Natural (W4P.PartExpr (W4.Pred sym) (SymIO.FileHandle sym ptrW))
+    } -> FDescMap sym ptrW
+
+-- | A wrapper around a RankN 'OverrideSim' action
+--
+-- This enables us to make explicit that all of the type variables are free and
+-- can be instantiated at any types since this override action has to be
+-- returned from the 'initialLLVMFileSystem' function.
+newtype SomeOverrideSim sym a where
```
robdockins, line 133: This might be a useful utility definition, and better placed elsewhere...

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -0,0 +1,560 @@
...
+  $ SymIO.symIOIntrinsicTypes
+
+-- | Resolve a symbolic file descriptor to a known allocated file handle.
+-- The partial result is undefined if the descriptor is not found in the
+-- file handle table.
+getHandle
+  :: forall sym ptrW
+   . IsSymInterface sym
+  => sym
+  -> W4.SymNat sym
```
robdockins, line 267: Would it make more sense to model FDs as a muxtree instead?

travitch: In this particular case I'm wondering why it is a Nat instead of a Bitvector. Did you mean for this function on the internals, or in general? In general, I'm a bit worried about the `FD_SET` family of macros for e.g., `select` and `poll` that might be assuming bitvector operations are possible (it is a bit hard to tell).

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -0,0 +1,560 @@
...
+      , Just numBytes <- BVS.asUnsigned <$> W4.asBV (regValue size) -> do
+          -- We have a concrete size of the write and an IO Handle to write
+          -- to. Write each byte that is concrete, or replace it with a '?'
+          sym <- getSymInterface
+          F.forM_ [0..numBytes - 1] $ \idx -> do
+            idxBV <- liftIO $ W4.bvLit sym ptrw (BVS.mkBV ptrw idx)
+            byteVal <- liftIO $ SymIO.evalChunk chunk idxBV
+            case W4.asBV byteVal of
+              Just (BVS.asUnsigned -> concByte) -> liftIO $ BS.hPut hdl (BS.pack [fromIntegral concByte])
+              Nothing -> liftIO $ BS.hPut hdl (BS.pack [63])
```
robdockins, line 467: Is 63 the ASCII code for '?' ... let's comment that.

travitch: Yeah. I was being too lazy. Let me just import Data.ByteString.Char8 and do the right thing here. Thanks for the reminder

travitch: Okay, fixed in f9cfe51

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
robdockins: Not sure.  It looks like `fdset` are generally implemented as a bitset array, with size expected to be no more than `FD_SETSIZE`.

As far as I can tell, the only call side for `getHandle` has immediately previously coerced a bitvector into a natural.  Better not to do this if possible, that translation really sucks for symbolic values.

travitch: Yup, let me see if I can change this one to just be a bitvector at the very least

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
robdockins: Well, it couldn't just be the FD source.  You'd actually need to track which FDs have been closed and such, because they get recycled.

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -0,0 +1,560 @@
...
+openFile
+  :: (IsSymInterface sym, HasLLVMAnn sym, HasPtrWidth wptr)
+  => LLVMFileSystem wptr
+  -> LLVMOverride p sym
+           (EmptyCtx ::> LLVMPointerType wptr
+                     ::> BVType 32)
+           (BVType 32)
+openFile fsVars =
+  [llvmOvr| i32 @open( i8*, i32 ) |]
+  -- TODO add mode support by making this a varargs function
```
robdockins, line 388: I don't remember seeing documentation about which flags and such are supported.  Probably worth writing that down.

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
travitch: Okay, I switched it to be a SymBV instead of a SymNat

crucible-symio/src/What4/CachedArray.hs
```diff
@@ -0,0 +1,837 @@
...
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE PatternSynonyms #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+-- must come after TypeFamilies, see also https://gitlab.haskell.org/ghc/ghc/issues/18006
+{-# LANGUAGE NoMonoLocalBinds #-}
+
+module What4.CachedArray
```
robdockins, line 33: It would be nice to have some overall sense of what is going on in this module; it's not obvious just from the exposed API.

I guess it eventually wants to find it's way down into What4... is it ready for that?

travitch: @danmatichuk Can you comment on what the CachedArray is trying to accomplish and whether or not it might be more generic than this memory model? I mostly left that part alone since I'm honestly not quite sure. It seemed like a deferred calculation of the contents of array fragments, but that is as far as I got...

crucible-symio/src/Lang/Crucible/SymIO/Types.hs
```diff
@@ -0,0 +1,208 @@
...
+
+-- | The crucible type of file handles.
+type FileHandleType w = ReferenceType (MaybeType (FilePointerType w))
+
+-- |  A file handle is a mutable file pointer that increments every time it is read.
+type FileHandle sym w = RegValue sym (FileHandleType w)
+
+-- | A 'File' represents a file in the filesystem independent
+-- of any open handles to it
+data File sym w = File (NatRepr w) (SymInteger sym)
```
robdockins, line 138: These correspond basically to `inode`s right?

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
travitch: I added documentation on that to the README for crux-llvm. Do you want the errata duplicated in this module?

crucible-symio/src/Lang/Crucible/SymIO/Types.hs
travitch: Yes, I believe that is a good characterization

crucible-symio/src/What4/CachedArray.hs
danmatichuk: It's a general storage structure for What4 that is intended to support arrays which have data at a mix of both concrete and symbolic indexes. Logically it's just a multi-dimensional array, but supports reading and writing out "chunks" to the underlying store (i.e where a chunk spans continuous locations in the first index of the array). When indexing into an N-dimensional array, we take the abstract domain of N expressions and use it as an index into an N-level nested interval map. 

For the filesystem, this allows us to easily define a file that is fully concrete initially (i.e. chunk with a domain spanning the file contents), while efficiently supporting writes on top of this file at (potentially) symbolic locations.

There is probably an argument for merging this idea with the existing ConcreteArray in What4.

crucible-symio/src/Lang/Crucible/SymIO.hs
```diff
@@ -0,0 +1,790 @@
...
+  FileHandle sym wptr ->
+  W4.SymBV sym wptr ->
+  C.OverrideSim p sym arch r args ret (DataChunk sym wptr, W4.SymBV sym wptr)
+readChunk' fsVar fhdl sz = readChunk fsVar fhdl sz $ \case
+  Left FileHandleClosed -> do
+    sym <- C.getSymInterface
+    liftIO $ addFailedAssertion sym $ AssertFailureSimError "Failed to read array due to closed file handle." "readArray': Closed file handle"
+  Right arr -> return arr
+
+-- | Returns a predicate indicating whether or not the file handle
```
robdockins, line 439: Is this comment correct? It mentions EOF, but the function seems to be checking if the handle is open.

travitch: Also going to refer this one to @danmatichuk. I'm also curious, reading around there, if we might be conflating EOF and closed. EOF can change if more data comes in, while being closed cannot.

crucible-symio/src/Lang/Crucible/SymIO/Types.hs
```diff
@@ -0,0 +1,208 @@
...
+     return $ FilePointer b off
+
+
+type DataChunk sym w = CA.ArrayChunk sym (BaseBVType w) (BaseBVType 8)
+
+type SizedDataChunkType w = SymbolicStructType (EmptyCtx ::> BaseArrayType (EmptyCtx ::> BaseBVType w) (BaseBVType 8) ::> BaseBVType w)
+type SizedDataChunk sym w = SymStruct sym (EmptyCtx ::> BaseArrayType (EmptyCtx ::> BaseBVType w) (BaseBVType 8) ::> BaseBVType w)
+
+
+pattern FileHandleRepr :: () => (1 <= w, ty ~ FileHandleType w) => NatRepr w -> TypeRepr ty
```
robdockins, line 198: I think this is worth a comment.  Especially, it's not immediately clear what is meant if the reference contains `Nothing`. I gather this is the state of a closed handle, but I'm not 100% sure.

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
robdockins: No, better to have it in one place, I think.  Maybe reference it here? I'm indifferent.

crucible-symio/src/What4/CachedArray.hs
robdockins: OK, I think those would be helpful comments to add to the module documentation.  I can see this being a useful addition to What4, but I don't know that we should hold up this PR for that.

crucible-symio/src/Lang/Crucible/SymIO.hs
robdockins: Or if you seek, etc.  We should definitely keep track of EOF separately from closed.

danmatichuk: Right, this comment is wrong (at some point they were conflated, but that was fixed). This is checking that the handle is still open. If a handle reaches EOF it can still be read from, it will just return that zero bytes have been read.

travitch: I'll update the comment

crucible-symio/src/Lang/Crucible/SymIO/Types.hs
danmatichuk: This is correct, a `FileHandle` that points to `Nothing` simply represents a closed handle.

crucible-symio/src/What4/CachedArray.hs
travitch: Okay, I added this documentation and a bit more that I think is probably correct

crucible-symio/src/Lang/Crucible/SymIO/Types.hs
travitch: Added a bit of detail (and some additional comments on some of the related types)

robdockins:
I think all my immediate concerns are addressed, except that:

1) I don't remember seeing any tests exercising symbolic files or symbolic overlays
2) We should open tickets about the various code migration tasks mentioned above

travitch:
Thanks - let me think about a good test for the symbolic code. I'll also make the relevant tickets

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
travitch: Opened #801

crucible-symio/src/Data/Parameterized/IntervalsMap.hs
travitch: Created issue #802

travitch:
> I think all my immediate concerns are addressed, except that:
> 
>     1. I don't remember seeing any tests exercising symbolic files or symbolic overlays

Okay, test added

>     2. We should open tickets about the various code migration tasks mentioned above

I added those tickets, as well as a task list (#803) for the next steps on crucible-symio

atomb:
I think the other reviewers have already covered all the ground I could have. It looks like very valuable new functionality, with good documentation and testing (especially after the most recent commits)!

robdockins (approved):

Issue #783, opened 2021-07-22
Add more toggles for memory model-related options


RyanGlScott:
SV-COMP has various benchmark programs that are expected to verify despite technically having memory model issues. For instance, [`s3_clnt.blast.01.i.cil-1.c`](https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/openssl/s3_clnt.blast.01.i.cil-1.c) is expected to verify successfully in the `unreach-call` category, but Crux will currently reject it thusly:

```
$ crux-llvm --config config.crux --target=i386-unknown-linux-elf sv-benchmarks/c/openssl/s3_clnt.blast.01.i.cil-1.c 
[Crux] Using pointer width: 32 for file crux-build/crux~s3_clnt.blast.01.i.cil-1.bc
[Crux] Simulating function main
[Crux] Attempting to prove verification conditions.
[Crux] Counterexample found, skipping remaining goals
[Crux] Found counterexample for verification goal
[Crux]   sv-benchmarks/c/openssl/s3_clnt.blast.01.i.cil-1.c:1120:19: error: in ssl3_connect
[Crux]   Error during memory load
[Crux]     No previous write to this location was found
[Crux]       Attempting load at type: i32
[Crux]       Via pointer: (20, 0x1c:[32])
[Crux]     Performing overall load at type: i32
[Crux]       Via pointer: (20, 0x1c:[32])
...
```

In order to make Crux handle these sorts of SV-COMP programs, we'll need to add a toggle to suppress this particular class of error. We already have various memory-model–related toggles, however. These include `lax-pointer-ordering` and `lax-constant-equality`. Although we could just tack on more one-off toggles like these, it might make sense to consolidate all of these options somehow, similarly to how `ub-sanitizers` consolidates toggles related to undefined behavior.

robdockins:
In my ideal vision for this, we have a tree-based taxonomy of errors.  Then users can specify sets of errors to check for by specifying boolean combinations of sets of errors as named by the taxonomy.

For example, we might have a category of `memory-error` that includes `no-previous-write`, and `unaligned-pointer`, `invalid-cast`, etc. etc.

robdockins:
[...]

PR #794, opened 2021-08-03
`crucible-llvm`: Add `readUnwrittenMemory` option, use it in `readMem'`

RyanGlScott:
This is an attempt at relaxing Crucible's requirement that all reads must come from previously written memory, as described in #783. The main changes are:

* A new `freshLLVMVal` function, which generates a fresh `LLVMVal` constant based on the `StorageType` supplied as an argument.
* A new `readUnwrittenMemory` option in `MemOptions`, which relaxes the aforementioned requirement.
* The `readMem'` function, which is responsible for throwing the `No previous write to this location was found` error, now takes a `MemOptions` as an implicit argument (which requires quite a bit of plumbing to accomplish). When `readUnwrittenMemory` is set to `True`, instead of failing, `readMem'` will return a value computed by `freshLLVMVal`.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs
@@ -672,8 +672,8 @@ appendArray sym errCtx (NoErr _ v1) (NoErr _ v2) =
 
 -- | Make a partial LLVM array value.
 --
--- It returns 'Unassigned' if any of the elements of the vector are
--- 'Unassigned'. Otherwise, the 'AssertionTree' on the returned value
+-- It returns 'Err' if any of the elements of the vector are
+-- 'Err'. Otherwise, the 'Pred' on the returned 'NoErr' value
 -- is the 'And' of all the assertions on the values.
 mkArray :: forall sym. (IsExprBuilder sym, IsSymInterface sym) =>
   sym ->
@@ -695,8 +695,8 @@ mkArray sym tp vec =
 
 -- | Make a partial LLVM struct value.
 --
--- It returns 'Unassigned' if any of the struct fields are 'Unassigned'.
--- Otherwise, the 'AssertionTree' on the returned value is the 'And' of all the
+-- It returns 'Err' if any of the struct fields are 'Err'.
+-- Otherwise, the 'Pred' on the returned 'NoErr' value is the 'And' of all the
 -- assertions on the values.
 mkStruct :: forall sym. IsExprBuilder sym =>
   sym ->
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
@@ -34,11 +34,12 @@ module Lang.Crucible.LLVM.MemModel.Value
   , explodeStringValue
 
   , llvmValStorableType
+  , freshLLVMVal
   , isZero
   , testEqual
   ) where
 
-import           Control.Lens (view, over, _2)
+import           Control.Lens (view, over, _2, (^.))
 import           Control.Monad (foldM, join)
 import           Data.ByteString (ByteString)
 import qualified Data.ByteString as BS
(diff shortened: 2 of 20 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
```diff
@@ -120,6 +121,23 @@ llvmValStorableType v =
...
 
+freshLLVMVal :: IsSymInterface sym =>
+                sym -> StorageType -> IO (LLVMVal sym)
+freshLLVMVal sym tp =
+  case storageTypeF tp of
+    Bitvector bytes ->
+      case mkNatRepr (bytesToBits bytes) of
+        Some repr ->
+          case isPosNat repr of
+            Just LeqProof -> LLVMValInt <$> freshNat sym emptySymbol
```
robdockins, line 132: I think the block number should just be `0` in this case.  This corresponds to treating the freshly-created value as a "raw" bitvector, which I think is what we want.

robdockins:
You'll also want to modify the `NoErr` case of `assertSafe`.  In the "read-before-write" case, if the partial condition `p` is anything other than `true`, we'll need to construct `if p then v else freshVal` as the return value.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
robdockins: This means that anything we read from uninitialized memory will not be a dereferencable pointer, which seems fine.

RyanGlScott:
> In the "read-before-write" case, if the partial condition `p` is anything other than `true`, we'll need to construct `if p then v else freshVal` as the return value.

Oh, how would one construct an `ite` as an `LLVMVal`? I'm aware of `muxLLVMVal`, but that works over `PartLLVMVal`s, not `LLVMVal`s.

robdockins:
> > In the "read-before-write" case, if the partial condition `p` is anything other than `true`, we'll need to construct `if p then v else freshVal` as the return value.
> 
> Oh, how would one construct an `ite` as an `LLVMVal`? I'm aware of `muxLLVMVal`, but that works over `PartLLVMVal`s, not `LLVMVal`s.

Humm, yeah, that's a good point.  I think you'll have to push it down into the leaves of the `LLVMVal`.

RyanGlScott:
> Humm, yeah, that's a good point. I think you'll have to push it down into the leaves of the `LLVMVal`.

I see. I think `muxLLVMVal.muxval` is already implementing this pushing logic? I'll try factoring that out and seeing if that works.

robdockins:
Yeah, but what you're doing here is simpler.  You know at each leaf that what you're muxing with is a fresh value of the right type, so you can just create it inline, rather than creating a fresh value and then muxing them together.  Then you don't have to worry about what to do when you have mismatches, etc.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs
```diff
@@ -318,21 +319,25 @@ totalLLVMVal :: (IsExprBuilder sym)
 totalLLVMVal sym = NoErr (truePred sym)
 
 -- | Take a partial value and assert its safety
-assertSafe :: (IsSymInterface sym)
+assertSafe :: (IsSymInterface sym, ?memOpts :: MemOptions)
```
RyanGlScott, line 322: One thing that just dawned on me: I'm modifying all uses of `assertSafe`, but really, I think we only care about the particular call site in `readMem'`, which is responsible for checking reads of unwritten memory. Perhaps I should create a separate version of `assertSafe` for this specific need?

robdockins: I think that would be a sensible thing to do, yeah.  At a minimum, this new operation should have a different name.

RyanGlScott: This idea ends up being trickier to implement than I imagined. `readMem'` itself isn't actually the function that invokes `assertSafe`. Rather, it returns an `Err` value that is eventually passed to `doLoad`, which invokes `assertSafe` on it.

One idea I had was to make a specialized version of `assertSafe` that is specific to `doLoad`, but even that's too broad. There are other functions besides `readMem'` that produce `Err` values that make it to `doLoad`, such as `readMemInvalidate` (which `Err`s upon reading invalidated memory), `readMemStore` (which `Err`s upon reading invalid memory), and `readMem` (which can potentially have a failing `NoErr` due to bad alignment or an unreadable region).

I suppose if we wanted to distinguish between these cases, we'd have to check what `MemoryErrorReason` was used to annotate the `Pred` attached to the `PartLLVMVal`. Is it feasible to do so?

robdockins: I don't think looking at the `MemoryErrorReason` is a good idea.

robdockins: Maybe the right play is to go back to our original idea.  You can use the same `freshLLVMVal` you just wrote in the base case of `readMem'` whenever `readUnwrittenMemory` is enabled.

RyanGlScott: Ah, of course. That should have been obvious to me in hindsight, but better late than never.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
RyanGlScott: Good suggestion.

RyanGlScott:
Now that I've resolved the latest review comments, I have good news and bad news.

The good news is that `read-unwritten-memory` works much better than in the previous, hacky version. See the `T783` test case for evidence of this.

The bad news is that even with `read-unwritten-memory`, `crux-llvm` still can't run on `s3_clnt.blast.01.i.cil-1.c` in the `unreach-call` category without claiming it's false. Even ignoring the `NoSatisfyingWrite` assertion, it will separately trip up an `UnreadableRegion` and `ReadBadAlignment` alignment assertion in `readMem`. I suppose this means we need even more toggles.

This PR has grown larger than I'd expected it would, so I'm somewhat inclined to work towards merging it in its current state and add finer-grained toggling later. Does that sound reasonable?

robdockins (approved):
I agree, let's merge this before diving into the other memory errors that need knobs.

PR #796, opened 2021-08-03
Float printf

robdockins:
Add support for floating-point conversion codes to `crux-llvm`.

This basically just coerces the floating-point value to a real value, and prints it as a rational, if it is concrete.

We still do no have full support for the "%a" conversions (which print bit-exact representations of the float), and some of the formatting does not exactly match a proper printf implementation, but it isn't clear how much we care about those kinds of details.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Printf.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Printf.hs
@@ -158,6 +158,16 @@ insertThousands sep = reverse . go . reverse
   go (a:b:c:xs@(_:_)) = a:b:c:sep:go xs
   go xs = xs
 
+
+addLeadingZeros ::
+  Maybe Int -> -- precision
+  String ->
+  String
+addLeadingZeros Nothing digits = digits
+addLeadingZeros (Just p) digits =
+   let n = max 0 (p - length digits) in
+   replicate n '0' ++ digits
+
 formatSignedDec
   :: Integer -- value to format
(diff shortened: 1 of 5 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Printf.hs
```diff
@@ -170,19 +170,19 @@ formatSignedDec i minwidth prec flags = do
                | Set.member PrintfPosSpace flags -> " "
                | otherwise -> ""
   let digits = N.showInt (abs i) []
-  let precdigits = if prec > 0 then
-                      let n = max 0 (prec - length digits) in
-                      replicate n '0' ++ digits
-                   else
-                      digits
+  let precdigits = case prec of
```
RyanGlScott, line 173: The code which is used to compute `precdigits` is repeated four different times in this module, which suggests that it should be factored out into its own function. (There are other similarities among the various `format*` functions, but it's not clear if one can factor them out as cleanly.)

RyanGlScott (approved):
This looks sensible to me. It might be worth mentioning the shortcomings of the current approach in [this document](https://github.com/GaloisInc/crucible/blob/ea42ba91f79d6616ad284565526fe0cb732cf49f/crucible-llvm/doc/limitations.md). In fact, I think the `floatprint.c` test case exhibits some of the ways in which `crucible-llvm`'s approach differs from a proper `printf` implementation, so one can point to that as an example.

PR #799, opened 2021-08-05
insertvalue translation fixes

robdockins:


crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
```diff
@@ -346,14 +346,18 @@ extractValue (ZeroExpr (StructType si)) is =
...
-           extractValue (BaseExpr tpr (getStruct idx x)) is
 extractValue (StructExpr vs) (i:is)
    | fromIntegral i < Seq.length vs = extractValue (snd $ Seq.index vs $ fromIntegral i) is
 extractValue (VecExpr _ vs) (i:is)
    | fromIntegral i < Seq.length vs = extractValue (Seq.index vs $ fromIntegral i) is
+extractValue (BaseExpr (StructRepr ctx) x) (i:is)
+   | Just (Some idx) <- Ctx.intIndex (fromIntegral i) (Ctx.size ctx) = do
+           let tpr = ctx Ctx.! idx
+           extractValue (BaseExpr tpr (getStruct idx x)) is
+extractValue (BaseExpr (VectorRepr elTp) x) (i:is)
```
RyanGlScott, line 357: Crucible newbie question: does `VectorRepr` encompass both LLVM arrays and vectors? I ask since the [docs for `extractvalue`/`insertvalue`](https://llvm.org/docs/LangRef.html#extractvalue-instruction) indicate that the type must be an aggregate type—that is, either a struct or array type, but not a vector type. (It seems unlikely that we'd be given a program where the type of `extractvalue`/`insertvalue` is a non-aggregate type, but you never know.)

RyanGlScott (approved):

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
robdockins: LLVM indeed explicitly excludes vector types from being "aggregate".  Despite this, the memory layout for vectors and arrays is basically the same, provided you assume the vector elements are packed as tightly as possible (no padding), so we use crucible `Vector` to model both.

For vectors the `insertelement` and `extractelement` instructions are used instead.  I think the distinction is largely a waste of time, especially since vectors can participate in `getelementpointer` despite the documentation indicating otherwise.

PR #800, opened 2021-08-06
crucible-wasm: Only define ?recordLLVMAnnotation once

### LINKED ISSUE #791: `crucible-wasm`: Factor out seven separate occurrences of `?recordLLVMAnnotation = \_ _ -> return ()`
Currently, `Lang.Crucible.Wasm.Memory` has seven different occurrences of a `let ?recordLLVMAnnotation = \_ _ -> return ()` binding. We should instead factor these out into `HasLLVMAnn` constraints and bind `?recordLLVMAnnotation` a single time at the beginning of simulation, like we do in `crux-llvm` and `crucible-go`.

I originally noticed this when adding another implicit parameter to `Lang.Crucible.Wasm.Memory`, but this task could be done independently of that.

RyanGlScott:
Fixes #791.

### DIFF
--- a/crucible-wasm/src/Lang/Crucible/Wasm.hs
+++ b/crucible-wasm/src/Lang/Crucible/Wasm.hs
@@ -49,6 +49,7 @@ import Lang.Crucible.FunctionHandle
 import Lang.Crucible.Simulator
 import Lang.Crucible.Types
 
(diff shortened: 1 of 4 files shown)


robdockins (approved):
Thanks for doing this.

PR #805, opened 2021-08-11
Guard whether overrides fail on abnormalExitBehavior

### LINKED ISSUE #785: Allow overrides for early-termination functions (e.g., `abort`) to stop execution without erroring
Currently, Crucible will exit with an error whenever it overrides a function which exits early with a non-zero exit code, such as `abort`. This poses a challenge for SV-COMP's `unreach-call` category, which only considers functions that exit due to using the `reach_error()` function (and no others) as failing. In order to support this SV-COMP use case, we will need to add an option which changes the overrides to stop executing a path without throwing a non-zero exit code. The following hack, for instance, shows how this can be done for `abort`:

```diff
diff --git a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
index e4a4f72f..c3ae8729 100644
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
@@ -39,7 +39,7 @@ import           Data.Parameterized.Context ( pattern (:>), pattern Empty )
 import qualified Data.Parameterized.Context as Ctx
 
 import           What4.Interface
-import           What4.ProgramLoc (plSourceLoc)
+import           What4.ProgramLoc (plSourceLoc, initializationLoc)
 
 import           Lang.Crucible.Backend
 import           Lang.Crucible.CFG.Common
@@ -665,9 +665,9 @@ llvmAbortOverride
   => LLVMOverride p sym EmptyCtx UnitType
 llvmAbortOverride =
   [llvmOvr| void @abort() |]
-  (\_ sym _args ->
+  (\_ _sym _args ->
      do let err = AssertFailureSimError "Call to abort" ""
-        liftIO $ assert sym (falsePred sym) err
+        liftIO $ abortExecBecause $ AssertionFailure $ SimError initializationLoc err
   )
 
 llvmGetenvOverride
```

We should generalize this hack in a more principled way that can be toggled by the user.

RyanGlScott:
This commit adds a new `IntrinsicsOptions` data type which allows fine tuning the behavior of LLVM overrides. Currently, there is only one option, `abnormalExitBehavior`, which controls the behavior of functions which abnormally end execution (e.g., `abort()`). There are currently three options: `AlwaysFail` (the default), `OnlyAssertFail` (which fails if the function in question is `__assert_fail()`, and does not fail otherwise), and `NeverFail` (which is like `OnlyAssertFail` without the special case for `__assert_fail()`). The special case for `__assert_fail()` is motivated by SV-COMP's `unreach-call` category, where the `reach_error` function (implemented in terms of `__assert_fail()`) indicates failure, a status no other function (not even `abort()` etc.) enjoy.

The last commit, with the same title as this PR, contains the interesting part of the pull request. The remaining commits are mostly plumbing.

Fixes #785.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Options.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Options.hs
@@ -1,0 +1,48 @@
+------------------------------------------------------------------------
+-- |
+-- Module           : Lang.Crucible.LLVM.Intrinsics.Options
+-- Description      : Definition of options that affect LLVM overrides
+-- Copyright        : (c) Galois, Inc 2021
+-- License          : BSD3
+-- Maintainer       : Rob Dockins <rdockins@galois.com>
+-- Stability        : provisional
+------------------------------------------------------------------------
+module Lang.Crucible.LLVM.Intrinsics.Options
+  ( IntrinsicsOptions(..)
+  , AbnormalExitBehavior(..)
+  , defaultIntrinsicsOptions
+  ) where
+
+-- | Should Crucible fail when simulating a function which triggers an abnormal
+-- exit, such as @abort()@?
+data AbnormalExitBehavior
+  = AlwaysFail
+    -- ^ Functions which trigger an abnormal exit will always cause Crucible
+    --   to fail.
+  | OnlyAssertFail
+    -- ^ The @__assert_fail()@ function will cause Crucible to fail, while
+    --   other functions which triggern an abnormal exit will not cause
+    --   failures. This option is primarily useful for SV-COMP.
+  | NeverFail
+    -- ^ Functions which trigger an abnormal exit will never cause Crucible
+    --   to fail. This option is primarily useful for SV-COMP.
+  deriving Eq
+
+-- | This datatype encodes a variety of tweakable settings that to LLVM
+--   overrides.
+newtype IntrinsicsOptions
+  = IntrinsicsOptions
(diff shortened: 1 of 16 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Options.hs
```diff
@@ -0,0 +1,47 @@
...
+    -- ^ Functions which trigger an abnormal exit will never cause Crucible
+    --   to fail. This option is primarily useful for SV-COMP.
+  deriving Eq
+
+-- | This datatype encodes a variety of tweakable settings that to LLVM
+--   overrides.
+newtype IntrinsicsOptions
+  = IntrinsicsOptions
+    { abnormalExitBehavior :: AbnormalExitBehavior
+      -- ^ TODO RGS
```
robdockins, line 36: Let's fill out this TODO

robdockins (approved):

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Options.hs
RyanGlScott: Thanks for catching that. I should really just write Haddocks as I go, since I never seem to remember to revisit these TODOs :)

PR #808, opened 2021-08-18
Broaden scope of read-unwritten-memory to lax-loads-and-stores

RyanGlScott:
This renames the `read-unwritten-memory` Crux option to `lax-loads-and-stores`, reflecting its new purpose of being a catch-all option for relaxing various validity checks related to loads and stores. This is meant for SV-COMP mode, which has a different view of fatal errors than Crucible does.

This addresses some of the goals of #783. It does not address the goal of designing a generic taxonomy of errors.


crucible-llvm/src/Lang/Crucible/LLVM/Utils.hs
```diff
@@ -0,0 +1,15 @@
...
+-- Copyright        : (c) Galois, Inc 2021
+-- License          : BSD3
+-- Maintainer       : Rob Dockins <rdockins@galois.com>
+-- Stability        : provisional
+------------------------------------------------------------------------
+module Lang.Crucible.LLVM.Utils (applyUnless) where
+
+-- | If the first argument is 'False', apply the second argument to the third.
+-- Otherwise, simply return the third argument.
+applyUnless :: Applicative f => Bool -> (a -> f a) -> a -> f a
```
RyanGlScott, line 14: I wasn't sure of a good place to put this, so I invented a new module for my needs. Let me know if there is a better home for this.

RyanGlScott (commented):
Some notes:

* There are many more `MemoryErrorReason`s/`UndefinedBehavior`s that one could include under the purview of `lax-loads-and-stores`, but for now, I'm limiting myself to those necessary to get SV-COMP to work. In particular, these changes were sufficient to make [`s3_clnt.blast.01.i.cil-1.c`](https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/openssl/s3_clnt.blast.01.i.cil-1.c) no longer incorrectly fail.
* Unfortunately, `s3_clnt.blast.01.i.cil-1.c` times out now. An `Unknown` result is better than a spurious failure, of course, but this means that `s3_clnt.blast.01.i.cil-1.c` isn't really suitable as a test case. I've added a more minimal test case as `T783b`, renaming `T783` to `T783a` in the process.
* I've marked this as a draft for the time being, as I'd like to try it out on more `unreach-call` programs to see if there are other things that should be included in `lax-loads-and-stores`.

RyanGlScott:
Having run this patch on a large number of `unreach-call` SV-COMP benchmarks, I'm happy with the current state of this PR. As a result, I'll mark this as ready for review.

crucible-llvm/src/Lang/Crucible/LLVM/Utils.hs
robdockins: *shrug* Seems OK to me.  I haven't seen this exact function before, so I don't know if there's some better place to grab it from.

robdockins (approved):

PR #813, opened 2021-08-20
`crucible-llvm`: Support translating `fneg` instructions

RyanGlScott:
Recent versions of LLVM have an `fneg` instruction. This proves relatively straightforward to support in `crucible-llvm`, as `crucible` already has a `FloatNeg` constructor for `App`.

This bumps the `llvm-pretty` and `llvm-pretty-bc-parser` submodules to include elliottt/llvm-pretty#80 and GaloisInc/llvm-pretty-bc-parser#159, respectively.

### DIFF
--- a/crux-llvm/test-data/golden/fneg.c
+++ b/crux-llvm/test-data/golden/fneg.c
@@ -1,0 +1,10 @@
+#include <crucible.h>
+
+double negate(double x) {
+  return -x;
+}
+
+int main() {
+  double x = crucible_double("x");
+  return (negate(x) > 0);
+}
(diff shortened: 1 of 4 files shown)


crux-llvm/test-data/golden/fneg.c
```diff
@@ -0,0 +1,10 @@
+#include <crucible.h>
+
+double negate(double x) {
+  return -x;
+}
+
+int main() {
+  double x = crucible_double("x");
+  return (x > 0);
```
robdockins, line 9: This test case doesn't actually call `negate`.  It probably should.

robdockins (approved):
Nice to see this is pretty straightforward.  I'm happy with this modulo the comment re. the test case once everything lands upstream.

crux-llvm/test-data/golden/fneg.c
RyanGlScott: Well that's embarrassing. Fixed.

PR #814, opened 2021-08-22
Implement memcpy and memset with arrayCopy and arraySet

andreistefanescu:
Depends on https://github.com/GaloisInc/what4/pull/152

### DIFF
--- a/crucible-saw/src/Lang/Crucible/Backend/SAWCore.hs
+++ b/crucible-saw/src/Lang/Crucible/Backend/SAWCore.hs
@@ -1007,6 +1007,25 @@ evaluateExpr sym sc cache = f []
                SAWExpr <$> SC.scArrayUpdate sc sc_idx_type sc_elm_type sc_arr sc_idx sc_elm
           | otherwise -> unimplemented "multidimensional UpdateArray"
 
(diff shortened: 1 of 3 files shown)


crucible-saw/src/Lang/Crucible/Backend/SAWCore.hs
```diff
@@ -1007,6 +1007,25 @@ evaluateExpr sym sc cache = f []
                SAWExpr <$> SC.scArrayUpdate sc sc_idx_type sc_elm_type sc_arr sc_idx sc_elm
           | otherwise -> unimplemented "multidimensional UpdateArray"
 
+        B.CopyArray w a_repr dest_arr dest_idx src_arr src_idx len ->
```
robdockins, line 1010: Why are we not handling the "equal array range" constructor? Did I miss something?

robdockins: On second thought, nevermind.  This is basically dead code now.  I should probably go ahead and delete it.

robdockins (approved):
Looks OK to me, once we land the What4 changes and update the submodules.

Issue #351, opened 2019-11-21
Create SV-COMP competition entry


robdockins:
`crux` is nearly, but not quite, set up to be an SV-COMP competitor.  I think we could feasibly enter in the following SV-COMP categories: 1) ReachSafety, 2) MemSafety, 4) NoOverflows, 6) SoftwareSystems 8) C-Falsification, and maybe 9) Java-Overall (leaving out concurrency and termination).

To properly enter the competition, we'd need to do some additional work to parse the contest benchmark metadata files, produce counterexample witnesses in the expected format, and generally make sure we can run in the contest environment.

In addition, we should add the ability to check the `valid-memtrack` property, which states that memory should be freed before becoming inaccessible.

robdockins:
CF #352

RyanGlScott:
> In addition, we should add the ability to check the `valid-memtrack` property, which states that memory should be freed before becoming inaccessible.

I opened a separate issue (#790) for this part.

RyanGlScott:
We have successfully whipped `crux-llvm` into competition shape, as chronicled [here](https://galois.com/blog/2022/01/crux-in-sv-comp-2022-a-retrospective/). I think this issue has served its purpose, so I'll close it in favor of the various other issues under the "SV-Comp" GitHub label.

PR #815, opened 2021-08-24
Overhaul crux-llvm-svcomp to fit in with SV-COMP's benchmark harness

RyanGlScott:
This is the current snapshot of my work to rewrite `crux-llvm-svcomp` to be capable of participating in SV-COMP (see #351). It is not complete yet, but there is enough scaffolding in place that I wanted to upstream it to avoid bitrot. This is split up into two commits, one of which is focused exclusively on the code in `crux-llvm-svcomp` itself, and the other of which focuses on auxiliary scripting used to produce `crux-llvm-svcomp` bindists.

### DIFF
--- a/crux/src/Crux/SVCOMP.hs
+++ b/crux/src/Crux/SVCOMP.hs
@@ -6,7 +6,9 @@
 {-# LANGUAGE ApplicativeDo #-}
 {-# LANGUAGE DeriveAnyClass #-}
 {-# LANGUAGE DeriveGeneric #-}
+{-# LANGUAGE KindSignatures #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE RecordWildCards #-}
 module Crux.SVCOMP where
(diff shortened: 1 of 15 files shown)


crux/src/Crux/SVCOMP.hs
```diff
@@ -31,44 +35,94 @@ import Crux.Config
...
+    _       -> Left "Invalid architecture"
+
+-- | We cannot verify an SV-COMP program without knowing the architecture and
+-- the specification. Unfortunately, Crux's command-line argument parser makes
+-- it somewhat awkward to require certain arguments (without defaults). As a
+-- hacky workaround, we parse the command-line arguments for the
+-- architecture/specification as if they were optional and later check here to
+-- see if they were actually provided, throwing an error if they were not
+-- provided.
+processSVCOMPOptions :: SVCOMPOptions Maybe -> IO (SVCOMPOptions Identity)
```
RyanGlScott, line 74: This is an extremely gross hack I came up with to make certain command-line arguments required. I hate it. If there is a better way to do this in Crux, please let me know.

crux/src/Crux/SVCOMP/Witness.hs
```diff
@@ -0,0 +1,329 @@
...
+--   specified in https://github.com/sosy-lab/sv-witnesses.
+
+{-# Language NamedFieldPuns #-}
+module Crux.SVCOMP.Witness where
+
+import           Data.ByteString (ByteString)
+import qualified Data.ByteString.Char8 as B (unpack)
+import qualified Data.List as L (intercalate)
+import           Data.Maybe (catMaybes)
+import           Data.Time.Format.ISO8601 (iso8601Show)
```
RyanGlScott, line 11: Note to self: this module was introduced in `time-1.9`, and unfortunately, GHC 8.6 is bundled with an earlier version of `time`, causing this to fail to build on GHC 8.6. I'll need to see if I can backport `iso8601Show` somehow.

Issue #816, opened 2021-08-25
Better error message


varosi-chaosgroup:
> [Crux] Function should have no arguments

If this message show information about which function does not have arguments - it'll be much more informative and it will help to find the problem.

I have used galoisinc/crux-llvm:0.4 Docker image.

RyanGlScott:
Indeed, this could potentially be dealt with at the same time as a fix for #812.

Issue #790, opened 2021-07-28
SV-COMP: Check the `valid-memtrack` and `valid-memcleanup` properties


RyanGlScott:
SV-COMP has two similar, but still distinct, properties called `valid-memtrack` and `valid-memcleanup`, neither of which Crux checks for at the moment. Here is how they are described in the [SV-COMP 2021 rules](https://sv-comp.sosy-lab.org/2021/rules.php):

> * `G valid-memtrack`
> 
>   All allocated memory is tracked, i.e., pointed to or deallocated (counterexample: memory leak). More precisely: There exists no finite execution of the program on which the program lost track of some previously allocated memory. (Comparison to Valgrind: This property is violated if Valgrind reports 'definitely lost'.)
> * `G valid-memcleanup`
> 
>   All allocated memory is deallocated before the program terminates. In addition to valid-memtrack: There exists no finite execution of the program on which the program terminates but still points to allocated memory. (Comparison to Valgrind: This property can be violated even if Valgrind reports 'still reachable'.)

What is the difference between the two? My understanding of the description above, as well as reading [these](https://groups.google.com/g/sv-comp/c/ap9f2tUPVow) SV-COMP [mailing](https://groups.google.com/g/sv-comp/c/Slug4p2DACM) list posts, is that the following program would violate both properties:

```c
#include <stdlib.h>

int *f() {
  int *x = malloc(sizeof(int));
  int *y = malloc(sizeof(int));
  x = y;
}

int main() {
  int *x = f();
  free(x);
  return 0;
}
```

On the other hand, this program would violate `valid-memcleanup` but not `valid-memtrack`:

```c
#include <stdlib.h>

int main() {
  int *x = malloc(sizeof(int));
  return 0;
}
```

The difference depends on whether one considers programs which have live variables that still point to un-freed memory upon exit to constitute a memory leak or not. `valid-memtrack` answers "no" to this question, whereas `valid-memcleanup` answers "yes" to this question.

Checking `valid-memtrack` is a prerequisite to competing in the `MemSafety` category of SV-COMP, while checking `valid-memcleanup` is a prerequisite to competing in the `MemCleanup` category.

robdockins:
Conceptually, I think `valid-memcleanup` would be relatively easy.  We just need to ensure at each program exit point that all (heap) allocations have been freed.  I think the only slightly tricky thing is determining what, exactly, is meant by "program exit point."  We might need to slightly rearrange how we handle the `EarlyExit` abort.

[...]

PR #826, opened 2021-08-27
`crux-llvm`: Add `supply-main-arguments` option

### LINKED ISSUE #812: `crucible-llvm`: Support `main` functions with arguments
`crucible-llvm` currently supports `main` functions with this type signature:

```c
int main() { ... }
```

But not this type signature:

```c
int main(int argc, char **argv) { ... }
```

If you feed the latter program to `crux-llvm`, it will choke thusly:

```
$ crux-llvm test.c                                                                             
Function should have no arguments
```

However, it would be useful to support `main` functions of the latter form as well. As one data point, various programs in the SV-COMP benchmark suite, such as [`floats-esbmc-regression/digits_for.c`](https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/floats-esbmc-regression/digits_for.c), use `main` with this type signature. What's more, they never use `argc` nor `argv`, so `crux-llvm` would likely work for these programs if it just passed dummy arguments to `main`!

Some possible ways forward would be:

1. Allow users to configure what arguments to pass to `main`, perhaps through the command line or a configuration file.
2. Teach `crucible-llvm` to pass `argc=1` and `argv=[]` to `main` when it has the latter type signature. There is a question of whether this should be the default behavior, or if this should require users to opt in.

RyanGlScott:
When enabled, `supply-main-arguments` will allow simulating `int main(int argc, char *argv[])` functions by supplying `argc=0` and `argv={}` arguments. This is a very specialized option, but one that ends up being very useful for SV-COMP's purposes, as there are many SV-COMP benchmark programs that define `main` in this way.

Fixes #812.

While I was in town, I spruced up the error message that is given when an entrypoint function has arguments, thereby fixing #816.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -74,6 +74,7 @@ module Lang.Crucible.LLVM.MemModel
   , doInvalidate
   , doCalloc
   , doFree
+  , doAlloca
   , doLoad
   , doStore
   , doArrayStore
@@ -404,14 +405,8 @@ evalStmt sym = eval
 
   eval (LLVM_Alloca _w mvar (regValue -> sz) alignment loc) =
      do mem <- getMem mvar
-        blkNum <- liftIO $ nextBlock (memImplBlockSource mem)
-        blk <- liftIO $ natLit sym blkNum
-        z <- liftIO $ bvLit sym PtrWidth (BV.zero PtrWidth)
-
-        let heap' = G.allocMem G.StackAlloc blkNum (Just sz) alignment G.Mutable loc (memImplHeap mem)
-        let ptr = LLVMPointer blk z
-
-        setMem mvar mem{ memImplHeap = heap' }
+        (ptr, mem') <- liftIO $ doAlloca sym mem sz alignment loc
+        setMem mvar mem'
         return ptr
 
   eval (LLVM_Load mvar (regValue -> ptr) tpr valType alignment) =
@@ -542,6 +537,25 @@ ptrMessage msg ptr ty =
           , "  address " ++ show (G.ppPtr ptr)
           , "  at type " ++ show (G.ppType ty)
           ]
+
+-- | Allocate memory on the stack frame of the currently executing function.
+doAlloca ::
+  (IsSymInterface sym, HasPtrWidth wptr) =>
+  sym ->
+  MemImpl sym ->
+  SymBV sym wptr {- ^ allocation size -} ->
+  Alignment      {- ^ pointer alignment -} ->
+  String         {- ^ source location for use in error messages -} ->
+  IO (LLVMPtr sym wptr, MemImpl sym)
+doAlloca sym mem sz alignment loc = do
+  blkNum <- liftIO $ nextBlock (memImplBlockSource mem)
+  blk <- liftIO $ natLit sym blkNum
+  z <- liftIO $ bvLit sym PtrWidth (BV.zero PtrWidth)
+
+  let heap' = G.allocMem G.StackAlloc blkNum (Just sz) alignment G.Mutable loc (memImplHeap mem)
+  let ptr   = LLVMPointer blk z
+  let mem'  = mem{ memImplHeap = heap' }
+  pure (ptr, mem')
 
 -- | Load a 'RegValue' from memory. Both the 'StorageType' and 'TypeRepr'
 -- arguments should be computed from a single 'MemType' using
--- a/crux-llvm/src/Crux/LLVM/Config.hs
+++ b/crux-llvm/src/Crux/LLVM/Config.hs
@@ -31,7 +31,7 @@ data CError =
     ClangError Int String String
   | LLVMParseError LLVM.Error
   | MissingFun String
-  | BadFun
+  | BadFun String
   | EnvError String
   | NoFiles
     deriving Show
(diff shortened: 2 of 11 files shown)


crux-llvm/src/Crux/LLVM/Config.hs
```diff
@@ -82,6 +89,7 @@ data LLVMOptions = LLVMOptions
   , noCompile :: Bool
   , optLevel :: Int
   , symFSRoot :: Maybe FilePath
+  , supplyMainArguments :: Bool
```
travitch, line 92: Do we think that we might want more expressive arguments here in the future? If we made this an ADT, we would have an easy path in the future for just adding a new constructor:
```
data MainArguments = NoArguments -- ^ main ()
                                   | EmptyArguments  -- ^ main (int argc, char* argv[]) where argc = 0
                                   | SomeArguments (NonEmpty String) -- ^ main(int argc, char* argv[]) where argc > 0
```

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -543,6 +538,25 @@ ptrMessage msg ptr ty =
           , "  at type " ++ show (G.ppType ty)
           ]
 
+-- | Allocate memory on the stack frame of the currently executing function.
```
travitch, line 541: I like giving this a name and exposing it

crux-llvm/src/Crux/LLVM/Simulate.hs
```diff
@@ -228,18 +239,64 @@ prepLLVMModule llvmOpts halloc sym bcFile memVar = do
...
+          sym       = ctx^.ctxSymInterface
+          dl        = llvmDataLayout (trans^.transContext.llvmTypeCtx)
+          tp        = ArrayType 0 (PtrType (MemType i8))
+          tp_sz     = memTypeSize  dl tp
+          alignment = memTypeAlign dl tp
+          loc       = "crux-llvm main(argc, argv) simulation"
+
+      argc <- liftIO $ LLVMPointer <$> natLit sym 0 <*> bvLit sym w (BV.zero w)
+      sz   <- liftIO $ bvLit sym PtrWidth $ bytesToBV PtrWidth tp_sz
+      (argv, mem') <- liftIO $ doAlloca sym mem sz alignment loc
```
travitch, line 294: Could you comment on why `doAlloca` over just `doCalloc`?

travitch (approved):

crux-llvm/src/Crux/LLVM/Simulate.hs
RyanGlScott: My goal was to simulate a function that looks roughly like this:

```hs
int crucible_main() {
  int argc = 0;
  char *argv[] = {};
  return main(argc, argv);
}
```

With this in mind, `doAlloca` seemed like a more natural fit, since the LLVM bitcode to create `argv` will use the `alloca` instruction anyway. There's no reason I couldn't use `doCalloc`, I suppose, but then the memory would be on the heap instead of on the stack. That could potentially have ramifications if we ever enter the [`valid-memcleanup`/`valid-memsafety`](https://github.com/GaloisInc/crucible/issues/790) categories (which care a lot about the lifetime of heap-allocated data), so `doAlloca` seemed like the simpler approach.

Having typed all of this out, I should turn this into a comment in the source code as well.

crux-llvm/src/Crux/LLVM/Config.hs
RyanGlScott: You're right—I could foresee someone wanting `SomeArguments` in the future. I can easily turn the `yes`/`no` options for `supply-main-arguments` to `empty`/`none` to keep the door open for extending it in the future. Does `config-schema` allow parsing `empty`, `none`, and something like `["args", "go", "here"]` all in the same option? My understanding of `config-schema`'s [`ValueSpec`](https://hackage.haskell.org/package/config-schema-1.2.2.0/docs/Config-Schema-Spec.html#t:ValueSpec) data type is that it permits this, but I haven't tried implementing something like that yet.

crux-llvm/src/Crux/LLVM/Simulate.hs
travitch: I think that is reasonable. `argv` *does* actually live on the stack in most binaries (on Linux, in the aux vector), though in a different spot, so it is reasonable. I was just curious what level of fidelity you were looking for.

crux-llvm/src/Crux/LLVM/Config.hs
travitch: Ah, I'm actually not sure what config-schema supports. I don't see anything quite like that in the exposed API, but that doesn't mean one could definitely not write something like that. No need to do this right now, given that we don't know when we'll need that capability.

Given the structure of `config-schema`, we might instead need to specify command line arguments by supplying multiple `--argument` parameters and just saying that specifying that is mutually-exclusive with `supply-main-arguments`.

crux-llvm/src/Crux/LLVM/Simulate.hs
RyanGlScott: I've turned https://github.com/GaloisInc/crucible/pull/826/files#r698825652 into a comment above `checkMainWithArguments`.

crux-llvm/src/Crux/LLVM/Config.hs
RyanGlScott: I prototyped what this might look like with the following code:

```hs
supplyMainArgumentsSpec :: ValueSpec SupplyMainArguments
supplyMainArgumentsSpec =
  (NoArguments <$ atomSpec "none") <!>
  (EmptyArguments <$ atomSpec "empty") <!>
  (SomeArguments <$> Crux.listSpec Crux.stringSpec)
```

To my delight, this works exactly like I expected it would. In light of this, I'm quite happy with this design, since it should be straightforward to add `SomeArguments` later if we desire it.

PR #834, opened 2021-09-02
Update correct MemImpl when supply-main-arguments is enabled

### LINKED ISSUE #833: `supply-main-arguments: empty` causes inscrutable `No implementation or override found for pointer` error
This simple program:

```c
int a;

int getService() {
  a = 9;
  return 0;
}

int main(int argc, char *argv[]) {
  return getService();
}
```

Will fail when `supply-main-arguments: empty` is used in a config file:

```
$ cabal run exe:crux-llvm -- --config test2.config test2.c
Resolving dependencies...
Up to date
[Crux] Using pointer width: 64 for file crux-build/crux~test2.bc
[Crux] *** debug executable: results/test2/debug-9
[Crux] *** break on line: 9
[Crux]   test2.c:9:10: error: in main
[Crux]   Details:
[Crux]     No implementation or override found for pointer: "getService"
[Crux]     No implementation or override found for pointer: "getService"
[Crux]     Attempting to load callable function with type: i32()
[Crux]       Via pointer: Global symbol "getService" (2, 0x0:[64])
[Crux]     In memory state:
[Crux]       Stack frame main
[Crux]       Base memory
[Crux]         Allocations:
[Crux]           StackAlloc 5 0x0:[64] Mutable 8-byte-aligned crux-llvm main(argc, argv) simulation
[Crux]           GlobalAlloc 4 0x4:[64] Mutable 4-byte-aligned [global variable  ] a
[Crux]           GlobalAlloc 3 0x0:[64] Immutable 1-byte-aligned [defined function ] main
[Crux]           GlobalAlloc 2 0x0:[64] Immutable 1-byte-aligned [defined function ] getService
[Crux]           GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [external function] llvm.dbg.value
[Crux]         Writes:
[Crux]             4 |->   *(4, 0x0:[64]) := 0
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

On the other hand, the same program when `int main() { ... }` is used instead.

RyanGlScott:
Previously, `checkMainWithArguments` was incorrectly using a `MemImpl` from a `PreppedLLVM` rather than looking up the `MemImpl` in the `GlobalVar Mem`. Doing so led to #833. Luckily, this is easily fixed.

Fixes #833.

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Simulate.hs
+++ b/crux-llvm/src/Crux/LLVM/Simulate.hs
@@ -21,7 +21,7 @@ import qualified Data.List as List
 import Data.Maybe ( fromMaybe )
 import qualified Data.Parameterized.Map as MapF
 import qualified Data.Traversable as T
-import Control.Lens ((&), (%~), (%=), (^.), view)
+import Control.Lens ((&), (%~), (%=), (^.), use, view)
 import Control.Monad.State(liftIO)
 import Data.Text as Text (Text, pack)
 import GHC.Exts ( proxy# )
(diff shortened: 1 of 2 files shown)


travitch (approved):
Whoops, I totally flubbed that previous review then.

RyanGlScott:
Hah, no worries. Luckily, SV-COMP quickly alerted me to my mistake (and then some).

PR #837, opened 2021-09-03
uc-crux-llvm: Update README demo with current output

langston-barrett:


RyanGlScott (approved):

langston-barrett:
CI failed, but this is just a one-file change to a README, so I'm pretty sure its not related. I made this PR from my fork, and I think the issue is that the build is missing some secrets or some such.

Issue #335, opened 2019-11-06
`crux-llvm` test suite is fragile


robdockins:
The crux-llvm test suite relies on the local `clang` install to compile C files into bitcode.  This can cause the test suite to fail if, e.g., clang is too new and produces bitcode we can't parse.

The crux output also, unfortunately, leaks some environmental details (e.g. absolute paths of the compiled files), which mean they will fail on other machines for spurious reasons when this information is captured in the "golden" output files.

We should probably work directly with pre-compiled .bc files or find some other method to make the test suite more robust. @kquick, ideas?

langston-barrett:
In general, I prefer pre-compiled `.ll` files to `.bc`. We have checked in `.bc` files on the CHESS program and regretted it later: when it became necessary to recompile these files with another compiler version or set of compiler flags, this often caused the tests to fail in ways that were harder to review with the normal `git` toolset, because the files in question weren't text.

The problem with `.ll` files is that they aren't immediately consumable by Crux, you would still need `llvm-as` installed. However, `llvm-as` is _much_ more deterministic across versions than `clang`.

robdockins:
That's a good point about the determinism of `llvm-as` vs `clang`.  For better/worse OSX also doesn't supply `llvm-as`, so we don't have to deal with Apple's upgrade cycle.

RyanGlScott:
I ran into a situation recent in #841 where using `llvm-as` instead of `clang` to compile `.ll` files in the test suite would have been preferable. For context, #841 implements overrides for `htonl` and similar functions, which behave differently on big-endian and little-endian targets. When `clang` compiles an `.ll` file with `-emit-llvm -c` on a little-endian machine (such as my own), the resulting `.bc` file will always assume a little-endian architecture, even if the input `.ll` file specifies big-endianness in its data layout. As a result, it's effectively impossible to test big-endian–only code paths in `crux-llvm` with the way the test suite is set up right now.

Using `llvm-as` instead of `clang` would solve this problem, as `llvm-as` respects the endianness specified in an `.ll` file's data layout. With that approach, we could check in two different `.ll` files—one assuming a big-endian data layout and another assuming little-endian—and cover all possible code paths in `htonl`'s override.

langston-barrett:
Some kinds of tests can also be written in the S-expression Crucible-LLVM syntax, see discussion on #523.

langston-barrett:
See https://github.com/GaloisInc/crucible/issues/1379 for internal implementation details sneaking into the output.

PR #841, opened 2021-09-07
Add overrides for `hton{s,l}` and `ntoh{s,l}`

### LINKED ISSUE #840: `crucible-llvm`: Add overrides for `htons` and friends
`crux-llvm` fails to verify the [`aws-c-common/aws_byte_cursor_read_be16_harness.i`](https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/aws-c-common/aws_byte_cursor_read_be16_harness.i) SV-COMP benchmark program due to lacking an override for [`htons`](https://linux.die.net/man/3/htons):

```
$ cabal run exe:crux-llvm -- --config crux-llvm/svcomp/config-files/unreach-call.config test.c
Up to date
[Crux] Counterexample found, skipping remaining goals
[Crux]   test.c:4052:12: error: in aws_ntoh16
[Crux]   Details:
[Crux]     No implementation or override found for pointer: "ntohs"
[Crux]     No implementation or override found for pointer: "ntohs"
[Crux]     Attempting to load callable function with type: i16(i16)
[Crux]       Via pointer: Global symbol "ntohs" (29, 0x0:[64])
[Crux]     In memory state:
[Crux]       Stack frame aws_ntoh16
[Crux]       Stack frame aws_byte_cursor_read_common_harness
...
```

Other SV-COMP benchmark programs fail due to lacking overrides for `ntohs`, `htonl`, and `ntohl`. These would be relatively straightforward to add support for. We already have overrides for the `llvm.bswap` intrinsics, so implementing support for `htons` and friends would be a variation on an existing theme.

RyanGlScott:
These are identical to the overrides for `llvm.bswap.*`, except that they only swap bytes if being run on a little-endian machine. I have added a test case which ensures that when these functions are run on a little-endian architecture, their behavior corresponds to that of `bswap_16` and `bswap_32` from `byteswap.h`.

Fixes #840.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
@@ -252,6 +252,10 @@ declare_overrides =
   , basic_llvm_override Libc.llvmPutCharOverride
   , basic_llvm_override Libc.llvmExitOverride
   , basic_llvm_override Libc.llvmGetenvOverride
+  , basic_llvm_override Libc.llvmHtonlOverride
+  , basic_llvm_override Libc.llvmHtonsOverride
+  , basic_llvm_override Libc.llvmNtohlOverride
+  , basic_llvm_override Libc.llvmNtohsOverride
 
   , basic_llvm_override Libc.cxa_atexitOverride
   , basic_llvm_override Libc.posixMemalignOverride
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
@@ -550,10 +550,7 @@ llvmBSwapOverride widthRepr =
         -- From the LLVM docs:
         -- declare i16 @llvm.bswap.i16(i16 <id>)
         [llvmOvr| #width8 $nm( #width8 ) |]
-        (\_ sym args -> liftIO $
-            let vec :: SymBV sym (width * 8)
-                vec = regValue (args^._1)
-            in bvSwap sym widthRepr vec)
+        (\_ sym args -> Ctx.uncurryAssignment (Libc.callBSwap sym widthRepr) args)
     }}}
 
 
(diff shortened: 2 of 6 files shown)


robdockins:
If it's not too difficult to put together, I'd also like to see a test that `ntohl` and friends are no-ops on big-endian architectures.

RyanGlScott:
The tricky part is figuring out how to make Clang emit LLVM bitcode that assumes big endianness. I would have thought `clang -mbig-endian` would do the trick, but that doesn't appear to make a difference based on my limited testing. The alternative is to manually edit the little-endian `.ll` file to turn it into something big-endian. That would be tricky, but it has the advantage of possibly working.

RyanGlScott:
Unfortunately, my suggestion of manually editing the `.ll` file doesn't quite work either. This is because `crux-llvm`'s test suite driver passes `.ll` files directly to `crux-llvm`, which in turn passes them directly to `clang -c -emit-llvm`. When `clang` produces a `.bc` file from this, it seems to ignore whether the `.ll` file specifies `E` (indicating big-endianness) in its data layout, since if you use `llvm-dis` on the resulting `.bc` file, it will always give you something with `e` (indicating little-endianness) in its data layout. Urgh.

The only way I could see making this work is we changed `crux-llvm`'s test suite slightly. We'd either need to add:

1. A mode that first calls `llvm-as` on an `.ll` file, and then passing the resulting `.bc` file to `crux-llvm` with the `--no-compile` flag enabled, or:
2. Check in the `.bc` file directly and teach the `crux-llvm` test suite to recognize `.bc` files.

RyanGlScott:
Checking in `.bc` files directly (i.e., option 2) is probably unwise, given the discussion in #335. Speaking of which, #335 suggests option (1) in https://github.com/GaloisInc/crucible/pull/841#issuecomment-914581800 as a way to make the `crux-llvm` test suite less fragile when using different versions of LLVM. Unfortunately, I don't think implementing a fix for #335 is a small lift. I propose not addressing it in this PR, instead posting a follow-up comment in #335 to remind us to revisit this after #335 has been fixed.

robdockins (approved):

Issue #187, opened 2019-04-09
Missing floating point operations


robdockins:
There are a number of llvm intrinsics and `math.h` functions for floating point operations that are exercised by the `sv-benchmark` suite that are not currently implemented.

The following ones should be relatively easy to hook up, because they already appear to have What4 representations and corresponding solver support (although we should check to ensure that the semantics of each operation match exactly):

- [ ] `nan` (although the tag semantics will be lost)
- [ ] `remainder`
- [ ] `llvm.fabs.*`
- [ ] `llvm.maxnum.*`
- [ ] `llvm.minnum.*`
- [ ] `llvm.round.*`
- [ ] `llvm.sqrt.*`

The following will probably require additional work.  For most, it's not obvious if they can easily be reduced to simpler operations that respect all the corner cases of their semantics.  In other cases, the semantics probably cannot be exactly represented at all (because e.g., they distinguish between different NaNs):

- [ ] `fdim`
- [ ] `fesetround` (modifes the rounding mode)
- [ ] `modff`
- [ ] `llvm.cos.*`
- [ ] `llvm.sin.*`
- [ ] `llvm.copysign.*`

There are a whole additional collection of llvm intrinsics and operations from `math.h` we could imagine implementing in addition to the ones listed here, of course.  Let's edit this ticket as we encounter examples of these in the wild.  Many of them will pose nontrivial design questions as we decide what to do about each of these operations in the various interpretations we have for them.

RyanGlScott:
I recently tested `crux-llvm` out on a handful of floating-point SV-COMP benchmarks. Here is a tally of missing overrides that I encountered:

* [ ] 6 fail due to not overriding `fesetround`
* [ ] 5 fail due to not overriding `__fpclassifyf`
* [x] 5 fail due to not overriding `__isnan` (https://github.com/GaloisInc/crucible/pull/927)
* [x] 5 fail due to not overriding `__isinf` (https://github.com/GaloisInc/crucible/pull/1110)
* [ ] 4 fail due to not overriding `__signbit`
* [x] 2 fail due to not overriding `__isinff` (https://github.com/GaloisInc/crucible/pull/1110)
* [ ] 1 fails due to not overriding `__fpclassify`
* [x] 1 fails due to not overriding `__isnanf` (https://github.com/GaloisInc/crucible/pull/927)
* [ ] 1 fails due to not overriding `fdim`
* [ ] 1 fails due to not overriding `modff`
[...]

PR #851, opened 2021-09-14
Add overrides for `llvm.copysign.{f32,f64}`

RyanGlScott:
LLVM 12 is much more eager to simplify certain expressions to calls to `llvm.copysign.f*`, as illustrated in https://github.com/GaloisInc/crucible/issues/187#issuecomment-917047901. It proves relatively straightforward to just add overrides for the `llvm.copysign.f{32,64}` instrinsics, so this patch accomplishes just that. The one caveat is that we cannot accurately represent the semantics of `llvm.copysign.f*` with respect to NaN arguments, since What4 does not possess a way to distinguish the sign bits of NaN values. I decided not to let perfect be the enemy of good, however, so I simply made a note of this limitation in the `crucible-llvm/doc/limitations.md` file.

Addresses one bullet point in #187.

### DIFF
--- a/crux-llvm/test-data/golden/copysign.c
+++ b/crux-llvm/test-data/golden/copysign.c
@@ -1,0 +1,33 @@
+#include <crucible.h>
+#include <float.h>
+#include <math.h>
+#include <stdbool.h>
+
+// A reasonable specification for copysign, assuming that x isn't NAN or
+// INFINITY.
+float copysign_spec(float x, float y) {
+  float magnitude = fabs(x);
+  if (signbit(y)) {
+    return -magnitude;
+  } else {
+    return magnitude;
+  }
+}
+
+int main() {
+  // The general cases
+  float x = crucible_float("x");
+  float y = crucible_float("y");
+  assuming(!isnan(x));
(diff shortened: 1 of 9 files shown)


crux-llvm/test-data/golden/copysign.c
```diff
@@ -0,0 +1,45 @@
...
+float copysign_spec(float x, float y) {
+  float magnitude = fabs(x);
+  if (signbit(y)) {
+    return -magnitude;
+  } else {
+    return magnitude;
+  }
+}
+
+bool float_approx_equal(float x, float y) {
```
robdockins, line 17: For my curiosity, is this actually necessary here? Flipping the sign bit is one of the few things you can do with floating-point where you'll actually get the exactly equalities you expect.

robdockins (approved):

crux-llvm/test-data/golden/copysign.c
RyanGlScott: I used `FLT_EPSILON` due to my general paranoia about using `==` with floating-point values, but I think you're right in that I'm being excessively paranoid here. I've removed this in favor of plain old `==`.

Issue #847, opened 2021-09-09
`crucible-llvm`: Add overrides for `llvm.abs.i*`


RyanGlScott:
LLVM 12 [adds an overloaded `llvm.abs.i*` intrinsic](https://reviews.llvm.org/D81829). It's fairly easy to produce one of these with Clang 12, as the following code demonstrates:

```c
int my_abs(int x) {
  int a = x;
  if (a < 0) a = -a;
  return a;
}
```
```
$ ~/Software/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-20.04/bin/clang -O1 test.c -S -emit-llvm
$ cat test.ll
; ModuleID = 'test.c'
source_filename = "test.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

; Function Attrs: norecurse nounwind readnone uwtable willreturn
define dso_local i32 @my_abs(i32 %0) local_unnamed_addr #0 {
  %2 = call i32 @llvm.abs.i32(i32 %0, i1 true)
  ret i32 %2
}

; Function Attrs: nofree nosync nounwind readnone speculatable willreturn
declare i32 @llvm.abs.i32(i32, i1 immarg) #1

attributes #0 = { norecurse nounwind readnone uwtable willreturn "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" "unsafe-fp-math"="false" "use-soft-float"="false" }
attributes #1 = { nofree nosync nounwind readnone speculatable willreturn }

!llvm.module.flags = !{!0}
!llvm.ident = !{!1}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{!"clang version 12.0.0 (https://github.com/llvm/llvm-project/ b978a93635b584db380274d7c8963c73989944a1)"}
```

While we're at it, we should add an override for `abs` as well. Most of the time, calls to `abs` will get inlined away, but this might not be the case if compiled with `-O0`.

Issue #845, opened 2021-09-08
`crucible-llvm`: Add override for `llvm.ubsantrap`


RyanGlScott:
LLVM 12 [adds](https://reviews.llvm.org/D89959#change-9A7rHNe4Wrq3) an `llvm.ubsantrap(i8)` intrinsic, and `clang` now tends to generate `llvm.ubsantrap` instead of `llvm.trap` when `-fsanitize=... -fsanitize-trap=...` is used. Here is one example of how to generate bitcode using `llvm.ubsantrap`:

```c
#include <stdint.h>
#include <stdio.h>

int main() {
  int32_t x = 2147483647 + 1;
  printf("%d\n", x);
  return 0;
}
```
```
$ ~/Software/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-20.04/bin/clang -O1 -S -emit-llvm test.c -fsanitize=signed-integer-overflow -fsanitize-trap=signed-integer-overflow
test.c:5:26: warning: overflow in expression; result is -2147483648 with type 'int' [-Winteger-overflow]
  int32_t x = 2147483647 + 1;
                         ^
1 warning generated.

$ cat test.ll
; ModuleID = 'test.c'
source_filename = "test.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

; Function Attrs: noreturn nounwind uwtable
define dso_local i32 @main() local_unnamed_addr #0 {
  call void @llvm.ubsantrap(i8 0) #2, !nosanitize !2
  unreachable, !nosanitize !2
}

; Function Attrs: cold noreturn nounwind
declare void @llvm.ubsantrap(i8 immarg) #1

attributes #0 = { noreturn nounwind uwtable "disable-tail-calls"="false" "frame-pointer"="none" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" "unsafe-fp-math"="false" "use-soft-float"="false" }
attributes #1 = { cold noreturn nounwind }
[...]

PR #855, opened 2021-09-16
Rudimentary LLVM 12 support

RyanGlScott:
This PR contains a series of commits which allow the `crucible-llvm`, `crux-llvm`, and `uc-crux-llvm` test suites to pass with LLVM 12. In summary, we now:

* Include more recent `llvm-pretty` and `llvm-pretty-bc-parser` commits which allow parsing bitcode that is specific to LLVM 12,
* Allow having `crux-llvm` test cases that are meant to be run with a minimum version of Clang in mind,
* Crudely translate `freeze` instructions in a way that just barely gets the job done (and test this in `crux-llvm` with Clang 12 or later), and
* Fix an oddity in a `uc-crux-llvm` test case that causes LLVM 12 to optimize an entire function away to `ret undef`, which doesn't seem like what was intended.

Refer to the individual commits in the commit history for more specifics.

There are additional overrides that we will need to implement to be able to handle certain "in-the-wild" programs with LLVM 12 (see #845 and #847), but I will handle those in separate PRs.

### DIFF
--- a/uc-crux-llvm/test/programs/linked_list_sum.c
+++ b/uc-crux-llvm/test/programs/linked_list_sum.c
@@ -9,5 +9,6 @@ int linked_list_sum(struct list *l) {
     total += next->head;
     next = next->tail;
   }
+  return total;
 }
 
--- a/crucible-llvm/doc/limitations.md
+++ b/crucible-llvm/doc/limitations.md
@@ -39,3 +39,22 @@ The implementations of some floating-point operations are imprecise with
 respect to NaN values. For example, `crucible-llvm`'s implementation of the
 `copysign` function will always return a positive, "quiet" NaN value if its
 first argument is a NaN, regardless of the sign of the second argument.
+
+
+`freeze` instruction limitations
+================================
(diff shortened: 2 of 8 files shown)


uc-crux-llvm/test/programs/linked_list_sum.c
```diff
@@ -9,5 +9,6 @@ int linked_list_sum(struct list *l) {
     total += next->head;
     next = next->tail;
   }
+  return total;
```
RyanGlScott, line 12: As the commit message of 2725066f8b25211c3e92e9b29946c55c91d9e1cc explains, I needed to make this change in order to prevent Clang 12 from optimizing the entirety of the `linked_list_sum` function to just `ret undef`. I think this is an OK thing to do, but please double-check me @langston-barrett.

langston-barrett: Yep, that looks like it was just a mistake on my part when creating the test. Thanks for the fix!

langston-barrett (approved):

PR #856, opened 2021-09-17
Add overrides for `llvm.ubsantrap`, `llvm.abs.*`, and friends

### LINKED ISSUE #845: `crucible-llvm`: Add override for `llvm.ubsantrap`

RyanGlScott:
This adds support for some overrides that were introduced in LLVM 12, `llvm.ubsantrap` and `llvm.abs.*`. Refer to the individual commits in this PR for specifics. While I was in town, this also cleans up `crucible-llvm`'s treatment of overrides involving the `long` type slightly.

Fixes #845 and #847.


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -782,6 +824,84 @@ callBSwap ::
...
+  RegEntry sym (BVType w) ->
+  RegEntry sym (BVType 1) ->
+  OverrideSim p sym ext r args ret (RegValue sym (BVType w))
+callLLVMAbs sym widthRepr src (regValue -> isIntMinPoison) = do
+  shouldCheckIntMin <- liftIO $
+    -- Per https://releases.llvm.org/12.0.0/docs/LangRef.html#id451, the second
+    -- argument must be a constant.
+    case asBV isIntMinPoison of
+      Just bv -> pure (bv /= BV.zero (knownNat @1))
+      Nothing -> addFailedAssertion sym "Call to llvm.abs.* with non-constant second argument"
```
langston-barrett, line 902: This might be a case where we would want to throw [`MalformedLLVMModule`](https://github.com/GaloisInc/crucible/blob/master/crucible-llvm/src/Lang/Crucible/LLVM/MalformedLLVMModule.hs) instead of adding an assertion - It's not really a safety condition violation, but more of an issue that the module isn't semantically meaningful if this isn't a constant.

It may also be worth considering adding a check to LLVM module translation that the argument is syntactically constant, in order to fail earlier/more eagerly.

crux-llvm/test-data/golden/golden/T847-fail1.ll
```diff
@@ -0,0 +1,42 @@
...
+attributes #1 = { nounwind readnone willreturn "disable-tail-calls"="false" "frame-pointer"="all" "less-precise-fpmad"="false" "no-infs-fp-math"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="true" "stack-protector-buffer-size"="8" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "tune-cpu"="generic" "unsafe-fp-math"="false" "use-soft-float"="false" }
+attributes #2 = { nounwind readnone willreturn }
+
+!llvm.module.flags = !{!0}
+!llvm.ident = !{!1}
+!llvm.commandline = !{!2}
+
+!0 = !{i32 1, !"wchar_size", i32 4}
+!1 = !{!"clang version 12.0.0 (https://github.com/llvm/llvm-project/ b978a93635b584db380274d7c8963c73989944a1)"}
+!2 = !{!"/home/rscott/Software/clang+llvm-12.0.0-x86_64-linux-gnu-ubuntu-20.04/bin/clang-12 -fno-discard-value-names -frecord-command-line -S -D CRUCIBLE -emit-llvm -I ../../../c-src/includes -O0 T847-fail1.c -o T847-fail1.ll"}
```
langston-barrett, line 42: I wonder if there's a way to compile this so that it doesn't include paths on your machine... No biggie, though.

langston-barrett (approved):
Thanks for all the comments and tests!

crux-llvm/test-data/golden/golden/T847-fail1.ll
RyanGlScott: Yes, this is an annoyingly non-portable aspect of the `-frecord-command-line` flag. Some possible mitigations are:

1. Manually edit the bitcode to strip out non-portable artifacts like this one. While this would work, it would be an extra step that someone would have to remember to do in case this `.ll` file needed to be regenerated.
2. Don't use `-frecord-command-line` at all. I'd really prefer using it, though, as it makes it possible for others to know exactly how to recreate the `.ll` file. 
3. Figure out whatever tools like Nix do to avoid this problem. (Perhaps you might know this?)

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
RyanGlScott: Good suggestion about `MalformedLLVMModule`. I'll change this code to throw that instead.

I don't have a strong opinion about whether we check for the second argument being constant here or during module translation. Do you know if there are examples of other parts of the translation code where we check for constants in function arguments?

RyanGlScott: I've pushed a change such that we now error out with `malformedLLVMModule "Call to llvm.abs.* with non-constant second argument" [printSymExpr isIntMinPoison]` instead.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -782,6 +825,86 @@ callBSwap ::
...
+  if shouldCheckIntMin
+     then do isNotIntMinUB <- annotateUB sym ub isNotIntMin
+             let err = AssertFailureSimError "Undefined behavior encountered" $
+                       show $ UB.explain ub
+             assert sym isNotIntMinUB err
+             pure srcAbs
+     else -- Per https://releases.llvm.org/12.0.0/docs/LangRef.html#id452,
+          -- return INT_MIN if the argument is INT_MIN and we are not returning
+          -- poison.
+          bvIte sym isNotIntMin srcAbs src
```
robdockins, line 867: I think we can simplify this.  If `src = INT_MIN` then `negate src = INT_MIN` also, so I think you can just return `srcAbs`.

robdockins (approved):
I think we can slightly simplify the term we produce for abs, but otherwise looks good.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
RyanGlScott: Good catch. I've applied your suggestion.

PR #859, opened 2021-09-21
uc-crux-llvm: Refactor into more modules for code reuse

langston-barrett:
I'm planning to reuse many of these functions in another PR, so this PR collects the refactoring-type changes that precede the functional changes.


langston-barrett:
Commit https://github.com/GaloisInc/crucible/pull/859/commits/df9701cf6f8742bcf3a9ef3c72da88abc2ef5afb requires a new release of parameterized-utils after https://github.com/GaloisInc/parameterized-utils/pull/121 is merged.

uc-crux-llvm/src/UCCrux/LLVM/FullType/StorageType.hs
```diff
@@ -0,0 +1,36 @@
...
+toStorageType :: HasPtrWidth wptr => FullTypeRepr m ft -> StorageType
+toStorageType fullTypeRepr =
+  let memType = toMemType fullTypeRepr
+  in case LLVMMem.toStorableType memType of
+       Nothing ->
+         panic
+           "Impossible: 'toMemType' returned a metadata type?"
+           [ "Either that, or the failure condition of"
+           , "'toStorableType' has changed."
+           ]
```
langston-barrett, line 35: For the sake of review, I'll note that this isn't a *new* call to `panic` per se - This function previously returned a `Maybe`, and then wherever that `Maybe` was used, there were calls to `panic`.

uc-crux-llvm/src/UCCrux/LLVM/Setup.hs
```diff
@@ -261,34 +215,48 @@ generate sym mts ftRepr selector (ConstrainedShape shape) =
...
+          <$> (SymValue <$> malloc sym ftRepr selector (fromIntegral n))
           <*> pure (Shape.ShapeAllocated n)
       (Shape.ShapePtr _constraints (Shape.ShapeInitialized vec), FTPtrRepr ptPtdTo) ->
         do
           let num = Seq.length vec
-          ptr <- malloc sym ftRepr selector num
-          let ftPtdTo = asFullType mts ptPtdTo
-          size <- sizeBv proxy sym ftPtdTo 1
+          Some numRepr <-
+            case NatRepr.mkNatRepr (fromIntegral num) of
```
kquick, line 224: It is slightly safer to use `toEnum` here since it fixes both source and target types.  The `fromIntegral` can have silent wraparound effects because it's polymorphic in the target type (same concern for all the uses of `fromIntegral`in this code).  That doesn't apply here ATM, but just sharing from my bag of general Haskell practices.

kquick (approved):

uc-crux-llvm/src/UCCrux/LLVM/Setup.hs
langston-barrett: See #876

PR #860, opened 2021-09-22
SV-COMP: various witness-related tweaks

RyanGlScott:
This contains various changes that I needed to get `crux-llvm-svcomp`'s witness files accepted by the various witness validators used in SV-COMP, including:

* An option to name witness automaton files as `witness.graphml`, as that works best with `benchexec`.
* A hack to make sure that witness files never print decimals in `creationtime`s, as the witness linter will reject these.
* Some notes on how to factor witness validation into scoring.

Refer to each individual commit message for further details.

### DIFF
--- a/crux-llvm/svcomp/README.md
+++ b/crux-llvm/svcomp/README.md
@@ -108,4 +108,153 @@ $ PYTHONPATH=../benchexec python3 -m benchexec.benchexec -r SV-COMP21_unreach-ca
 This will start a run of the `SV-COMP21_unreach-call` benchmark definition as
 defined in `../benchmark-defs/crux.xml`. (See the "Preparing the necessary
 files for SV-COMP benchmark runs" section.) To execute a different benchmark
-definition, change the argument to `-r`.
+definition, change the argument to `-r`. When the run is complete, it will
+place various files under the `results/` directory, including:
+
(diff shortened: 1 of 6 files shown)


crux-llvm/svcomp/README.md
```diff
@@ -108,4 +108,148 @@ $ PYTHONPATH=../benchexec python3 -m benchexec.benchexec -r SV-COMP21_unreach-ca
...
+   script that I use to automate this process. It will output something like
+   this:
+
+   ```bash
+   $ ./CountScore.hs /home/rscott/bench-defs/crux-llvm-svcomp-2021-09-15-Linux-x86_64/results/crux.2021-09-16_02-25-33.results.SV-COMP21_no-overflow.xml.bz2.merged.xml.bz2
+   Statistics:                     631 Files
+     correct:                      249
+       correct true:               61
+       correct false:              188
+     correct-unconfirmed:          249
```
robdockins, line 241: Looks like this is just the same count as the `correct` category.  I'd expect to se `28` here.

robdockins (approved):
Looks good to me, modulo what I think is a minor bug in the `CountScore` script.

crux-llvm/svcomp/README.md
RyanGlScott: Oops, that's an embarrassing mistake. Fixed.

PR #863, opened 2021-09-23
uc-crux-llvm: Use a fixed configuration in test suite

langston-barrett:
Uses #832 to stop grabbing the configuration from the environment and instead fix its values. This is substantially cleaner and more flexible, if a bit more verbose.


kquick (approved):
The implementation looks good.  I can't really comment in depth on the semantic implications of the configuration change, but this seems reasonable.

I'll note that I had a small concern that this left us exposed to a lack of test coverage on the normal configuration process, but in discussions with Langston it was identified that the `crux-llvm` tests (at least) are still utilizing the standard configuration, so there's ultimately better coverage rather than a loss of coverage between the two.

langston-barrett:
Thanks for the review, @kquick!

PR #865, opened 2021-09-24
uc-crux-llvm: Minor test suite fixes

langston-barrett:
The bound on 'time' is necessary for function 'secondsToNominalDiffTime'.

Also, remove a superfluous print statement that was introduced during
development.

### DIFF
--- a/uc-crux-llvm/test/Test.hs
+++ b/uc-crux-llvm/test/Test.hs
@@ -63,7 +63,6 @@ import qualified Data.Text as Text
 import qualified Data.Map as Map
 import           Data.Maybe (fromMaybe, isNothing)
 import qualified Data.Set as Set
-import qualified Data.Time as Time
 import           GHC.Generics (Generic)
 import           System.FilePath ((</>))
 import           System.IO (IOMode(WriteMode), withFile)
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Be warned that imposing `time >= 1.9.1` will cut off support for GHC 8.6, as 8.6 is bundled with `time-1.8.0.2`. This is a problem I ran into separately in #815, which also uses an API that was introduced in `time-1.9`—see [this code](https://github.com/GaloisInc/crucible/blob/726662c904ff1e0c506e576280aac6c0303007e8/crux/src/Crux/SVCOMP/Witness.hs#L22-L601) that I added to support older versions of `time`.

I wonder if it would be worth factoring out this code in a separate `.Compat` module and backport an implementation of `secondsToNominalDiffTime` as well. Luckily, the definition of `secondsToNominalDiffTime` is pretty simple, so this should be a smaller lift.

langston-barrett:
> Be warned that imposing time >= 1.9.1 will cut off support for GHC 8.6, as 8.6 is bundled with time-1.8.0.2. 

Ah, thanks very much for that pointer @RyanGlScott! I should probably just do the same thing as `parseNominalDiffTime` in Crux.

kquick (approved):

PR #866, opened 2021-09-24
CI: Don't try to upload artifacts from forks

langston-barrett:
Builds from forks will fail on these steps, see https://github.com/GaloisInc/crucible/runs/3589806066


RyanGlScott (approved):
I like the cut of this PR's jib, especially if it makes the CI more robust.

langston-barrett:
I don't think this worked the way I expected... The problem this was intended to fix persisted in https://github.com/GaloisInc/crucible/pull/869/checks?check_run_id=3735228237.

PR #868, opened 2021-09-27
uc-crux-llvm: Postprocess Crux options in test suite

langston-barrett:
This has the effect of (1) validating the options and logging any warnings,
which may be helpful in keeping the test suite up to date with any shifting
requirements on valid Crux configurations, and (2) creating the build directory,
which seems to be causing CI failures on Ubuntu 20.04 (but confusingly, not on
18.04).

### DIFF
--- a/uc-crux-llvm/test/Test.hs
+++ b/uc-crux-llvm/test/Test.hs
@@ -86,7 +86,7 @@ import           Lang.Crucible.LLVM.Translation as CLLVM (defaultTranslationOpti
 import qualified Lang.Crucible.LLVM.MemModel as CLLVM
 
 import qualified Crux
-import qualified Crux.Config.Common as Crux (PathStrategy(SplitAndExploreDepthFirst))
+import qualified Crux.Config.Common as Crux (PathStrategy(SplitAndExploreDepthFirst), postprocessOptions)
 import qualified Crux.Log as Log
 
 import           Crux.LLVM.Compile (genBitCode)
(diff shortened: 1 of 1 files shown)


langston-barrett:
Hmm, this does not seem to fix the problem: https://github.com/GaloisInc/crucible/pull/868/checks?check_run_id=3720807498

I can't reproduce that issue locally, but I'll try to think of some other mitigation.

uc-crux-llvm/test/Test.hs
```diff
@@ -248,29 +250,32 @@ withOptions llvmModule file k =
...
-        , CruxLLVM.optLevel = 1
-        , CruxLLVM.symFSRoot = Nothing
-        , CruxLLVM.supplyMainArguments = CruxLLVM.NoArguments
-        }
+      do clang <- fromMaybe "clang" <$> lookupEnv "CLANG"
+         llvmLink <- fromMaybe "llvm-link" <$> lookupEnv "LLVM_LINK"
+         return $
+           CruxLLVM.LLVMOptions
+             { CruxLLVM.clangBin = clang
+             , CruxLLVM.linkBin = llvmLink
```
langston-barrett, line 260: This whole block was indented but only these two options actually changed.

kquick (approved):
The CI upload step failures look unrelated, but I'll leave that to you to determine.

langston-barrett:
> The CI upload step failures look unrelated, but I'll leave that to you to determine.

Those should be taken care of by #866. Thanks!

PR #869, opened 2021-09-27
uc-crux-llvm: Split the test suite into more modules

langston-barrett:
This will make the code in Utils easier to reuse in new tests in new modules.


kquick (approved):
I'm assuming there are no changes to the logic, just a re-arrangement into different files.  If that's not the case, I'm happy to re-review any logic change.

langston-barrett:
> I'm assuming there are no changes to the logic, just a re-arrangement into different files.

That's right!

PR #870, opened 2021-09-27
Set OS_TAG to fix crux-mir build

atomb:


atomb:
It's even simpler than that: it fixes an error from the packaging script complaining that `OS_TAG` isn't defined. :)

PR #886, opened 2021-10-13
Dbg def use

robdockins:


crux-llvm/test-data/golden/shrink.c.orig
```diff
@@ -0,0 +1,27 @@
+unsigned int addflt(unsigned int a , unsigned int b )
```
RyanGlScott, line 1: There is already a test case for this [here](https://github.com/GaloisInc/crucible/blob/796d1f595e7cbcd421e68050fa5cc864327b7aab/crux-llvm/test-data/golden/T778.c). Perhaps they should be consolidated?

robdockins: Maybe... but I want to use the `.ll` here to really be sure that we retain the problematic def/use inversion across future compiler changes. What do you think?

crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
```diff
@@ -386,19 +418,21 @@ transDefine halloc ctx d = do
...
+-- Return the translated module and a list of warning messages
+-- generated during translation.
 -- Note: We may want to add a map from symbols to existing function handles
 -- if we want to support dynamic loading.
 translateModule :: (?transOpts :: TranslationOptions)
                 => HandleAllocator -- ^ Generator for nonces.
                 -> GlobalVar Mem   -- ^ Memory model to associate with this context
                 -> L.Module        -- ^ Module to translate
-                -> IO (Some ModuleTranslation)
+                -> IO (Some ModuleTranslation, [(L.Symbol, Position, Text.Text)])
```
RyanGlScott, line 429: There's a lot of tuples in this return type. I wonder if it would be worth wrapping the `(L.Symbol, Position, Text.Text)` part in a newtype.

crux-llvm/src/Crux/LLVM/Log.hs
```diff
@@ -25,6 +25,7 @@ data CruxLLVMLogMessage
   | FailedToBuildCounterexampleExecutable
   | SimulatingFunction Text
   | UsingPointerWidthForFile Integer Text
+  | TranslationWarning Text
```
RyanGlScott, line 28: Rather than stuff the entire warning message here as a bare `Text`, it would probably be more in the spirit of structured message logging to store the list of warning triples (or the newtypes on top of them here), and then migrate the pretty-printing code from `sayTranslationWarning` to `cruxLLVMLogMessageToSayWhat`.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
```diff
@@ -1879,45 +1883,53 @@ callFunction instr _tailCall fnTy _fn _args _assign_f =
...
-     | L.ValSymbol "llvm.dbg.declare" <- fn =
-       do mbArgs <- dbgArgs args
+     | L.ValSymbol "llvm.dbg.declare" <- fn
+     , debugIntrinsics ?transOpts =
+       do mbArgs <- dbgArgs defSet args
           case mbArgs of
             Right (asScalar -> Scalar _ PtrRepr ptr, lv, di) ->
               extensionStmt (LLVM_Debug (LLVM_Dbg_Declare ptr lv di)) >> k
+            Left msg -> addWarning (Text.pack msg) >> k
             _ -> k
```
RyanGlScott, line 1905: I know this technically isn't your fault, but having these catch-call `_ -> k` cases is a bit worrisome, as it's possible that we could fall through silently if given unusual debug information (which, as we've seen, is totally within the realm of possibility). Perhaps we should give a warning here at least?

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+updateUseSet lab bi bim = if newuse == block_use_set bi then Nothing else Just bi{ block_use_set = newuse }
+  where
+  newuse = loop (block_body bi)
+
+  loop [] = mempty
+  loop (L.Result nm i _md:ss) = Set.union (instrUse lab i bim) (Set.delete nm (loop ss))
+  loop (L.Effect i _md:ss) = Set.union (instrUse lab i bim) (loop ss)
+
+instrUse :: L.BlockLabel -> L.Instr -> Map L.BlockLabel (LLVMBlockInfo s) -> Set L.Ident
+instrUse from i bim = Set.unions $ case i of
```
RyanGlScott, line 185: Curiously, this function only calls `Set.unions` once at the top level, whereas `useVal` invokes `Set.unions` recursively. Is there a good reason not to pick a single convention for both functions?

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+     ]
+
+-- | Compute use sets for each basic block. Thess sets list all the
+-- virtual registers that must be assigned before jumping to a
+-- block.
+--
+-- This is essentially a backward fixpoint computation based on the
+-- identifiers used in the block statements.  We iterate the use/def
+-- transfer function until no more changes are made.  Because it is a
+-- backward analysis, we (heuristicly) examine the blocks in
```
RyanGlScott, line 143:
```suggestion
-- backward analysis, we (heuristically) examine the blocks in
```

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+-- descending order, and re-add blocks to the workset based on
+-- predecessor maps.
+--
+-- This fixpoint computation terminates for the usual reasons: the transfer
+-- function is monotonic (use sets only increase) and has no infinite
+-- chains (in the worst case, all the finitely-many identifers in the
```
RyanGlScott, line 149:
```suggestion
-- chains (in the worst case, all the finitely-many identifiers in the
```

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+
+      -- | Map from labels to assignments that must be made before
+      -- jumping.  If this is the block info for label l',
+      -- and the map has [(i1,v1),(i2,v2)] in the phi_map for block l,
+      -- then basic block l is required to assign i1 = v1 and i2 = v2
+      -- before jumping to block l'.
+    , block_phi_map :: !(Map L.BlockLabel (Seq (L.Ident, L.Type, L.Value)))
+    }
+
+-- | Construct the block info map for the given function definition.  This collections
```
RyanGlScott, line 80:
```suggestion
-- | Construct the block info map for the given function definition.  This collects
```

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+import qualified Data.Set as Set
+
+import qualified Text.LLVM.AST as L
+
+import           Lang.Crucible.CFG.Generator
+import           Lang.Crucible.Panic ( panic )
+
+type LLVMBlockInfoMap s = Map L.BlockLabel (LLVMBlockInfo s)
+
+-- | Information about an LLVM basic block computed before be we begin the
```
RyanGlScott, line 44:
```suggestion
-- | Information about an LLVM basic block computed before we begin the
```

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+--   translation proper.
+data LLVMBlockInfo s
+  = LLVMBlockInfo
+    {
+      -- | The crucible block label corresponding to this LLVM block
+      block_label :: Label s
+
+      -- | The computed "use" set for this block.  This is the set
```
RyanGlScott, line 52:
```suggestion
      -- | The computed \"use\" set for this block.  This is the set
```

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+      -- of identifiers that must be assigned prior to jumping to this
+      -- block. They are either used directly in this block or used
+      -- by a successor of this block.  Note! "metadata" nodes
```
RyanGlScott, line 55:
```suggestion
      -- by a successor of this block.  Note! \"metadata\" nodes
```

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -0,0 +1,280 @@
...
+      -- do not contribute to the use set.  Note! values referenced
```
RyanGlScott, line 56: It would be good to explain somewhere _why_ metadata notes do not contribute to the use set, citing the relevant LLVM ticket.

uc-crux-llvm/src/UCCrux/LLVM/Main.hs
```diff
@@ -219,7 +219,7 @@ translateLLVMModule ::
   IO SomeModuleContext'
 translateLLVMModule llOpts halloc memVar moduleFilePath llvmMod =
   do
-    Some trans <-
+    (Some trans, _warns) <- -- TODO? should we do something with these warnings?
```
RyanGlScott, line 222: What do you think about this, @langston-barrett? Should `uc-crux-llvm` emit warnings discovered during translation?

RyanGlScott (commented):
This is exciting! I have not read every line thoroughly, but if the example from #778 works, I'm happy.

crucible-llvm/src/Lang/Crucible/LLVM/Translation.hs
robdockins: Yeah, that's probably a good idea.

robdockins:
I've verified locally that the program from #778 now works as expected.  When `debug-intrinsics` are enabled, we get the expected warning about def/use violation in debug metadata and the program is accepted.

PR #892, opened 2021-10-19
Add a hook for interpreting simulation results to Crux

langston-barrett:
This was inspired by a need in UC-Crux. In that tool, I need to register some overrides that will put some `Pred sym` away in an `IORef` when they're executed. I'd like to be able to consider whether or not those predicates are satisfiable after execution ends.

The general idea is to allow for an "interpreting results" step that has access to the symbolic backend (`ExprBuilder`). If this feature seems helpful and ergonomic to others, I'm happy to update all the tools in this repo at least to be compatible with it (at least).


langston-barrett:
> Concretely, you mean change that return value from `SimulatorHooks` to `IO SimulatorHooks`?

Yes, exactly.

PR #898, opened 2021-10-28
uc-crux-llvm: Overrides that check inferred function contracts

langston-barrett:
'Check' overrides take the result of contract inference on a function f
(namely, 'Constraints') and create an override that has the same
signature as f, and when called will

- Create a bunch of predicates that represent whether or not the
  constraints were violated at this call
- Stash them away in an IORef
- Call the original implementation of f

The hope is that this will be useful for refining overly pessimistic
(sufficient but unnecessary) preconditions in inferred contracts, or for
indicating the presence of bugs (when the contract really *should*
hold).

This is not currently integrated with other functionality (i.e., it's
not reachable from the CLI), but is instead a minimal prototype.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
@@ -77,11 +77,11 @@ data LLVMOverride p sym args ret =
   , llvmOverride_args    :: CtxRepr args -- ^ A representation of the argument types
   , llvmOverride_ret     :: TypeRepr ret -- ^ A representation of the return type
   , llvmOverride_def ::
-       forall ext rtp args' ret'.
+       forall rtp args' ret'.
          GlobalVar Mem ->
          sym ->
          Ctx.Assignment (RegEntry sym) args ->
-         OverrideSim p sym ext rtp args' ret' (RegValue sym ret)
+         OverrideSim p sym LLVM rtp args' ret' (RegValue sym ret)
     -- ^ The implementation of the intrinsic in the simulator monad
     -- (@OverrideSim@).
   }
--- a/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
@@ -62,6 +62,7 @@ import qualified Prettyprinter.Render.Text as PP
 import qualified Text.LLVM.AST as L
 
 import           Data.Parameterized.Classes (IndexF)
+import           Data.Parameterized.Ctx (Ctx)
 import qualified Data.Parameterized.Context as Ctx
 import qualified Data.Parameterized.Fin as Fin
 import           Data.Parameterized.TraversableFC.WithIndex (FoldableFCWithIndex, ifoldrMFC)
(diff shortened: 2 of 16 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
```diff
@@ -0,0 +1,358 @@
...
+  Monad g =>
+  (forall x. IndexF (t f z) x -> f x -> g m) ->
+  t f z ->
+  g m
+ifoldMapMFC f = ifoldrMFC (\i x acc -> fmap (<> acc) (f i x)) mempty
+
+-- | Create a predicate that checks that a Crucible(-LLVM) value conforms to the
+-- 'ConstrainedShape'.
+--
+-- TODO: Add selector for provenance
```
travitch, line 157: Could we add a note here about why the return value is a `Seq (Pred sym)` instead of just a `Pred sym`?

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
```diff
@@ -115,9 +126,96 @@ newtype CreateOverrideFn arch =
         IsSymInterface sym =>
         HasLLVMAnn sym =>
         sym ->
-        IO (PolymorphicLLVMOverride (Crux.Crux sym) sym arch)
+        IO (PolymorphicLLVMOverride arch (Crux.Crux sym) sym)
+    }
+
+newtype SymCreateOverrideFn sym arch =
```
travitch, line 132: Could we get a note on what this is for?

travitch (approved):

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -77,11 +77,11 @@ data LLVMOverride p sym args ret =
   , llvmOverride_args    :: CtxRepr args -- ^ A representation of the argument types
   , llvmOverride_ret     :: TypeRepr ret -- ^ A representation of the return type
   , llvmOverride_def ::
-       forall ext rtp args' ret'.
+       forall rtp args' ret'.
          GlobalVar Mem ->
          sym ->
          Ctx.Assignment (RegEntry sym) args ->
-         OverrideSim p sym ext rtp args' ret' (RegValue sym ret)
+         OverrideSim p sym LLVM rtp args' ret' (RegValue sym ret)
```
langston-barrett, line 84: Oh @robdockins, I wanted to check in with you on this - I changed this because I need to create overrides that call CFGs from the LLVM module. Does this seem OK to you?

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
langston-barrett: Yeah, good call - In brief, it's because eventually I'll add more "provenance" information, e.g. what `Constraint` is this `Pred` from, and *where* in the value does that constraint apply (i.e., a `Cursor`).

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
langston-barrett: Done!

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
langston-barrett: I added a new commit that just resolves the TODO and takes care of this along the way.

PR #906, opened 2021-11-01
`crucible-llvm`: Control granularity of reading uninitialized memory

### LINKED ISSUE #844: SV-COMP: `unreach-call` failure in `reducercommutativity/max.i`
`crux-llvm-svcomp` incorrectly concludes that `reach_error` is reachable in the [`reducercommutativity/max.i`](https://github.com/sosy-lab/sv-benchmarks/blob/99d37c5b4072891803b9e5c154127c912477f705/c/reducercommutativity/max.i) SV-COMP benchmark program:

```
$ cabal run exe:crux-llvm-svcomp -- --config crux-llvm/svcomp/config-files/unreach-call.config --svcomp-arch 32bit --svcomp-spec /home/rscott2/bench-defs/sv-benchmarks/c/properties/unreach-call.prp /home/rscott2/bench-defs/sv-benchmarks/c/reducercommutativity/max.i
Resolving dependencies...
Up to date
[Crux] Using pointer width: 32 for file crux-build/crux~max.bc
[Crux] Counterexample found, skipping remaining goals
[Crux]   rscott2/bench-defs/sv-benchmarks/c/reducercommutativity/max.i:12:83: error: in reach_error
[Crux]   Call to assert()
[Crux]   Details:
[Crux]     void reach_error()
[Crux] Goal status:
[Crux]   Total: 56
[Crux]   Proved: 54
[Crux]   Disproved: 1
[Crux]   Incomplete: 1
[Crux]   Unknown: 0
[SVCOMP] Verification result: FALSIFIED (unreach-call)
```

There are several other programs of a similar caliber that also give the same error, but `max.i` is probably the most self-contained example. Crux claims that `N = 0x20` is a counterexample, which is untrue. I think the use of `iteration-bound: 50` in `unreach-call.config` might be partially responsible for this, but I'm not sure.

RyanGlScott:
Previously, `laxLoadsAndStores` would make it so that all reads from uninitialized memory would return a new, fresh value. This isn't always what you want, however, as this implies that consecutive reads from the same, uninititalized variable would return different results. This patch adds another knob in the form of the `indeterminateLoadBehavior` field of `MemOptions`, which has two settings:

* `unstable-symbolic`: The previous behavior, in which each read from uninitialized memory will return a fresh symbolic value each time.
* `stable-symbolic` (now the default): After memory is allocated (be it through `alloca`, `malloc`, `calloc`, etc.), it is written with a fresh symbolic value. This makes it so that reading from the same section of uninitialized memory multiple times will always return the same symbolic value.

Note that `indeterminateLoadBehavior` only takes effect when `laxLoadsAndStores` is enabled.

Fixes #844.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -706,7 +706,9 @@ doMallocSize sz sym allocType mut loc mem alignment = do
   let heap' = G.allocMem allocType blkNum sz alignment mut loc (memImplHeap mem)
   let ptr   = LLVMPointer blk z
   let mem'  = mem{ memImplHeap = heap' }
-  mem'' <- if laxLoadsAndStores ?memOpts && mut == G.Mutable
+  mem'' <- if laxLoadsAndStores ?memOpts
+                && mut == G.Mutable
+                && allocType == G.HeapAlloc
                 && indeterminateLoadBehavior ?memOpts == StableSymbolic
            then do bytes <- freshConstant sym emptySymbol
                               (BaseArrayRepr (Ctx.singleton $ BaseBVRepr ?ptrWidth)
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -719,6 +719,8 @@ doMallocSize sz sym allocType mut loc mem alignment = do
            else pure mem'
   return (ptr, mem'')
 
+
+
 bindLLVMFunPtr ::
   (IsSymInterface sym, HasPtrWidth wptr) =>
   sym ->
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -44,6 +44,7 @@ module Lang.Crucible.LLVM.MemModel
   , BlockSource(..)
   , nextBlock
   , MemOptions(..)
+  , IndeterminateLoadBehavior(..)
   , defaultMemOptions
   , laxPointerMemOptions
 
@@ -547,7 +548,8 @@ ptrMessage msg ptr ty =
 
 -- | Allocate memory on the stack frame of the currently executing function.
 doAlloca ::
-  (IsSymInterface sym, HasPtrWidth wptr) =>
+  ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+  , ?memOpts :: MemOptions ) =>
   sym ->
   MemImpl sym ->
   SymBV sym wptr {- ^ allocation size -} ->
@@ -562,7 +564,14 @@ doAlloca sym mem sz alignment loc = do
   let heap' = G.allocMem G.StackAlloc blkNum (Just sz) alignment G.Mutable loc (memImplHeap mem)
   let ptr   = LLVMPointer blk z
   let mem'  = mem{ memImplHeap = heap' }
-  pure (ptr, mem')
+  mem'' <- if laxLoadsAndStores ?memOpts
+                && indeterminateLoadBehavior ?memOpts == StableSymbolic
+           then do bytes <- freshConstant sym emptySymbol
+                              (BaseArrayRepr (Ctx.singleton $ BaseBVRepr ?ptrWidth)
+                                             (BaseBVRepr (knownNat @8)))
+                   doArrayStore sym mem' ptr alignment bytes sz
+           else pure mem'
+  pure (ptr, mem'')
 
 -- | Load a 'RegValue' from memory. Both the 'StorageType' and 'TypeRepr'
 -- arguments should be computed from a single 'MemType' using
@@ -630,7 +639,8 @@ sextendBVTo sym w w' x
 --
 -- Precondition: the multiplication /size * number/ does not overflow.
 doCalloc ::
-  (IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym) =>
+  ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+  , ?memOpts :: MemOptions ) =>
   sym ->
   MemImpl sym ->
   SymBV sym wptr {- ^ size   -} ->
@@ -653,7 +663,8 @@ doCalloc sym mem sz num alignment = do
 
 -- | Allocate a memory region.
 doMalloc
-  :: (IsSymInterface sym, HasPtrWidth wptr)
+  :: ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+     , ?memOpts :: MemOptions )
   => sym
   -> G.AllocType {- ^ stack, heap, or global -}
   -> G.Mutability {- ^ whether region is read-only -}
@@ -666,7 +677,8 @@ doMalloc sym allocType mut loc mem sz alignment = doMallocSize (Just sz) sym all
 
 -- | Allocate a memory region of unbounded size.
 doMallocUnbounded
-  :: (IsSymInterface sym, HasPtrWidth wptr)
+  :: ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+     , ?memOpts :: MemOptions )
   => sym
   -> G.AllocType {- ^ stack, heap, or global -}
   -> G.Mutability {- ^ whether region is read-only -}
@@ -677,7 +689,8 @@ doMallocUnbounded
 doMallocUnbounded = doMallocSize Nothing
 
 doMallocSize
-  :: (IsSymInterface sym, HasPtrWidth wptr)
+  :: ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+     , ?memOpts :: MemOptions )
   => Maybe (SymBV sym wptr) {- ^ allocation size -}
   -> sym
   -> G.AllocType {- ^ stack, heap, or global -}
@@ -692,7 +705,17 @@ doMallocSize sz sym allocType mut loc mem alignment = do
   z      <- bvLit sym PtrWidth (BV.zero PtrWidth)
   let heap' = G.allocMem allocType blkNum sz alignment mut loc (memImplHeap mem)
   let ptr   = LLVMPointer blk z
-  return (ptr, mem{ memImplHeap = heap' })
+  let mem'  = mem{ memImplHeap = heap' }
+  mem'' <- if laxLoadsAndStores ?memOpts && mut == G.Mutable
+                && indeterminateLoadBehavior ?memOpts == StableSymbolic
+           then do bytes <- freshConstant sym emptySymbol
+                              (BaseArrayRepr (Ctx.singleton $ BaseBVRepr ?ptrWidth)
+                                             (BaseBVRepr (knownNat @8)))
+                   case sz of
+                     Just sz' -> doArrayStore sym mem' ptr alignment bytes sz'
+                     Nothing  -> doArrayStoreUnbounded sym mem' ptr alignment bytes
+           else pure mem'
+  return (ptr, mem'')
 
 bindLLVMFunPtr ::
   (IsSymInterface sym, HasPtrWidth wptr) =>
@@ -1300,7 +1323,8 @@ storeConstRaw sym mem ptr valType alignment val = do
 
 -- | Allocate a memory region on the heap, with no source location info.
 mallocRaw
-  :: (IsSymInterface sym, HasPtrWidth wptr)
+  :: ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+     , ?memOpts :: MemOptions )
   => sym
   -> MemImpl sym
   -> SymBV sym wptr {- ^ size in bytes -}
@@ -1311,7 +1335,8 @@ mallocRaw sym mem sz alignment =
 
 -- | Allocate a read-only memory region on the heap, with no source location info.
 mallocConstRaw
-  :: (IsSymInterface sym, HasPtrWidth wptr)
+  :: ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+     , ?memOpts :: MemOptions )
   => sym
   -> MemImpl sym
   -> SymBV sym wptr
@@ -1683,14 +1708,16 @@ registerGlobal (MemImpl blockSource gMap sMap hMap mem) symbols ptr =
 
 -- | Allocate memory for each global, and register all the resulting
 -- pointers in the global map.
-allocGlobals :: (IsSymInterface sym, HasPtrWidth wptr)
+allocGlobals :: ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+                , ?memOpts :: MemOptions )
              => sym
              -> [(L.Global, [L.Symbol], Bytes, Alignment)]
              -> MemImpl sym
              -> IO (MemImpl sym)
 allocGlobals sym gs mem = foldM (allocGlobal sym) mem gs
 
-allocGlobal :: (IsSymInterface sym, HasPtrWidth wptr)
+allocGlobal :: ( IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym
+               , ?memOpts :: MemOptions )
             => sym
             -> MemImpl sym
             -> (L.Global, [L.Symbol], Bytes, Alignment)
--- a/crucible-llvm/CHANGELOG.md
+++ b/crucible-llvm/CHANGELOG.md
@@ -1,0 +1,15 @@
+# next
+* A new `indeterminateLoadBehavior` flag in `MemOptions` now controls now
+  reading from uninitialized memory works when `laxLoadsAndStores` is enabled.
+  If `StableSymbolic` is chosen, then allocating memory will also initialize it
+  with a fresh symbolic value so that subsequent reads will always return that
+  same value. If `UnstableSymbolic` is chosen, then each read from a section of
+  uninitialized memory will return a distinct symbolic value each time.
+
+  As a result of this change, `?memOpts :: MemOptions` constraints have been
+  added to the following functions:
+  * `Lang.Crucible.LLVM.Globals`:
+    `initializeAllMemory` and `initializeMemoryConstGlobals`
+  * `Lang.Crucible.LLVM.MemModel`:
+    `doAlloca`, `doCalloc`, `doMalloc`, `doMallocUnbounded`, `mallocRaw`,
+    `mallocConstRaw`, `allocGlobals`, and `allocGlobal`
--- a/crucible-llvm/CHANGELOG.md
+++ b/crucible-llvm/CHANGELOG.md
@@ -9,7 +9,8 @@
   As a result of this change, `?memOpts :: MemOptions` constraints have been
   added to the following functions:
   * `Lang.Crucible.LLVM.Globals`:
-    `initializeAllMemory` and `initializeMemoryConstGlobals`
+    `initializeAllMemory`, `initializeMemoryConstGlobals`, `populateGlobals`,
+    `populateAllGlobals`, and `populateConstGlobals`
   * `Lang.Crucible.LLVM.MemModel`:
     `doAlloca`, `doCalloc`, `doMalloc`, `doMallocUnbounded`, `mallocRaw`,
     `mallocConstRaw`, `allocGlobals`, and `allocGlobal`
--- a/crucible-llvm/crucible-llvm.cabal
+++ b/crucible-llvm/crucible-llvm.cabal
@@ -12,6 +12,7 @@ Synopsis:      Support for translating and executing LLVM code in Crucible
 Description:
    Library providing LLVM-specific extensions to the crucible core library
    for Crucible-based simulation and verification of LLVM-compiled applications.
+extra-source-files: CHANGELOG.md, README.md
 
 common bldflags
   ghc-options: -Wall
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
@@ -25,6 +25,7 @@
 {-# LANGUAGE PatternSynonyms       #-}
 {-# LANGUAGE RankNTypes            #-}
 {-# LANGUAGE ScopedTypeVariables   #-}
+{-# LANGUAGE TypeApplications      #-}
 {-# LANGUAGE TypeOperators         #-}
 
 module Lang.Crucible.LLVM.Globals
@@ -50,6 +51,7 @@ import qualified Data.Set as Set
 import           Data.String
 import           Control.Monad.State (StateT, runStateT, get, put)
 import           Data.Maybe (fromMaybe)
+import qualified Data.Parameterized.Context as Ctx
 
 import qualified Text.LLVM.AST as L
 import qualified Text.LLVM.PP as LPP
@@ -253,6 +255,7 @@ allocLLVMFunPtr sym llvm_ctx mem decl =
 --   filtered list transitively reference.
 populateGlobals ::
   ( ?lc :: TypeContext
+  , ?memOpts :: MemOptions
   , 16 <= wptr
   , HasPtrWidth wptr
   , HasLLVMAnn sym
@@ -267,12 +270,13 @@ populateGlobals select sym gimap mem0 = foldM f mem0 (Map.elems gimap)
   f mem (gl, _) | not (select gl)    = return mem
   f _   (_,  Left msg)               = fail msg
   f mem (gl, Right (mty, Just cval)) = populateGlobal sym gl mty cval gimap mem
-  f mem (_ , Right (_, Nothing))     = return mem
+  f mem (gl, Right (mty, Nothing))   = populateExternalGlobal sym gl mty mem
 
 
 -- | Populate all the globals mentioned in the given @GlobalInitializerMap@.
 populateAllGlobals ::
   ( ?lc :: TypeContext
+  , ?memOpts :: MemOptions
   , 16 <= wptr
   , HasPtrWidth wptr
   , HasLLVMAnn sym
@@ -288,6 +292,7 @@ populateAllGlobals = populateGlobals (const True)
 --   given @GlobalInitializerMap@ (and any they transitively refer to).
 populateConstGlobals ::
   ( ?lc :: TypeContext
+  , ?memOpts :: MemOptions
   , 16 <= wptr
   , HasPtrWidth wptr
   , HasLLVMAnn sym
@@ -298,6 +303,40 @@ populateConstGlobals ::
   IO (MemImpl sym)
 populateConstGlobals = populateGlobals f
   where f = L.gaConstant . L.globalAttrs
+
+
+-- | Ordinarily external globals do not receive initalizing writes.  However,
+--   when 'lax-loads-and-stores` is enabled in the `stable-symbolic` mode, we
+--   populate external global variables with fresh bytes.
+populateExternalGlobal :: forall sym wptr.
+  ( ?lc :: TypeContext
+  , 16 <= wptr
+  , HasPtrWidth wptr
+  , IsSymInterface sym
+  , HasLLVMAnn sym
+  , HasCallStack
+  , ?memOpts :: MemOptions
+  ) =>
+  sym ->
+  L.Global {- ^ The global to populate -} ->
+  MemType {- ^ Type of the global -} ->
+  MemImpl sym ->
+  IO (MemImpl sym)
+populateExternalGlobal sym gl memty mem
+  | laxLoadsAndStores ?memOpts
+  , indeterminateLoadBehavior ?memOpts == StableSymbolic
+
+  =  do bytes <- freshConstant sym emptySymbol
+                    (BaseArrayRepr (Ctx.singleton $ BaseBVRepr ?ptrWidth)
+                        (BaseBVRepr (knownNat @8)))
+        let dl = llvmDataLayout ?lc
+        let sz = memTypeSize dl memty
+        let tyAlign = memTypeAlign dl memty
+        sz' <- bvLit sym PtrWidth (bytesToBV PtrWidth sz)
+        ptr <- doResolveGlobal sym mem (L.globalSym gl)
+        doArrayConstStore sym mem ptr tyAlign bytes sz'
+
+  | otherwise = return mem
 
 
 -- | Write the value of the given LLVMConst into the given global variable.
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
@@ -148,7 +148,8 @@ makeGlobalMap ctx m = foldl' addAliases globalMap (Map.toList (llvmGlobalAliases
 -- allocates space for global variables, but does not set their
 -- initial values.
 initializeAllMemory
-   :: (IsSymInterface sym, HasPtrWidth wptr)
+   :: ( IsSymInterface sym, HasPtrWidth wptr, HasLLVMAnn sym
+      , ?memOpts :: MemOptions )
    => sym
    -> LLVMContext arch
    -> L.Module
@@ -156,7 +157,8 @@ initializeAllMemory
 initializeAllMemory = initializeMemory (const True)
 
 initializeMemoryConstGlobals
-   :: (IsSymInterface sym, HasPtrWidth wptr)
+   :: ( IsSymInterface sym, HasPtrWidth wptr, HasLLVMAnn sym
+      , ?memOpts :: MemOptions )
    => sym
    -> LLVMContext arch
    -> L.Module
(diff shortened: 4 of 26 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -555,7 +556,13 @@ doAlloca sym mem sz alignment loc = do
   let heap' = G.allocMem G.StackAlloc blkNum (Just sz) alignment G.Mutable loc (memImplHeap mem)
   let ptr   = LLVMPointer blk z
   let mem'  = mem{ memImplHeap = heap' }
-  pure (ptr, mem')
+  mem'' <- if laxLoadsAndStores ?memOpts
```
RyanGlScott, line 559: The key changes are here, in `doAlloca`...

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -685,7 +696,16 @@ doMallocSize sz sym allocType mut loc mem alignment = do
   z      <- bvLit sym PtrWidth (BV.zero PtrWidth)
   let heap' = G.allocMem allocType blkNum sz alignment mut loc (memImplHeap mem)
   let ptr   = LLVMPointer blk z
-  return (ptr, mem{ memImplHeap = heap' })
+  let mem'  = mem{ memImplHeap = heap' }
+  mem'' <- if laxLoadsAndStores ?memOpts && mut == G.Mutable
```
RyanGlScott, line 700: ...here, in `doMallocSize`...

crux-llvm/test-data/golden/T844.c
```diff
@@ -0,0 +1,10 @@
+#include <stdlib.h>
```
RyanGlScott, line 1: ...and here, which adds a test case.

langston-barrett:
I wonder if it's worth supporting an enumeration of possible semantics of reading from uninitialized memory? i.e.
```haskell
data LaxUninitializedReads = FreshValueEveryTime | OneValue
```
where `FreshValueEveryTime` is the previous behavior and `OneValue` is the behavior in this PR.

It also might be worth splitting this out from `lax-loads-and-stores`, which disables some less-than-completely specified set of checks.

robdockins:
This specific implementation of stable nondeterministic values is about what I'd expect to see, so that looks good to me.

As @langston-barrett alludes to, I think the UI aspects of this are a bit tricker, and I'm not sure I have a well-formed opinion about what they should be.  Perhaps it does make sense to be a little more systematic about this, and break out specifically the fact that there are different possible interpretations of the "indeterminate" values from the C spec:

1) It is UB to read from uninitialized memory (the default behavior without flags)
2) You get "unstable" nondeterministic values (the current behavior with `lax-loads-and-stores`)
3) You get "stable" nondeterministic values (the behavior in this PR)
4) You get "poison" values (a possible future implementation when we get poison values figured out)

RyanGlScott:
Splitting out the uninitialized memory initialization component from the rest of `lax-loads-and-stores` would be tricky since the rest of `lax-loads-and-stores`' behavior is specifically designed to make sure that filling in uninitialized memory with symbolic values doesn't crash later. Besides the code in `doAlloca` and `doMalloc` that I added in this PR, there are three other places where `laxLoadsAndStores` currently has an effect. Two of these places are in `readMem`:

https://github.com/GaloisInc/crucible/blob/c12434ddaffe798208805cc51ca7a639cc7719d1/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs#L722-L727

And the third place is in `doMemcpy`:

https://github.com/GaloisInc/crucible/blob/bd78bbfc5bb495a30c1755d50f702c21c4266051/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L950-L951

What would happen if we removed these three uses of `laxLoadsAndStores`, but kept filling in uninitialized memory with symbolic values à la this PR? I tried this to see what would happens:

<details>

```diff
diff --git a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
index 641a6fc2..1a8f08d1 100644
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -967,7 +967,7 @@ doMemcpy sym w mem mustBeDisjoint dest src len = do
 
   let mop = MemCopyOp (gsym_dest, dest) (gsym_src, src) len (memImplHeap mem)
 
-  p1' <- applyUnless (laxLoadsAndStores ?memOpts)
+  p1' <- -- applyUnless (laxLoadsAndStores ?memOpts)
                      (Partial.annotateME sym mop UnreadableRegion) p1
   p2' <- Partial.annotateME sym mop UnwritableRegion p2
 
diff --git a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
index 9e111214..5deced6f 100644
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
@@ -719,11 +719,11 @@ readMem sym w gsym l tp alignment m = do
     -- Otherwise, fall back to the less-optimized read case
     _ -> readMem' sym w (memEndianForm m) gsym l m tp alignment (memWrites m)
 
-  part_val' <- applyUnless (laxLoadsAndStores ?memOpts)
+  part_val' <- -- applyUnless (laxLoadsAndStores ?memOpts)
                            (Partial.attachSideCondition sym p2 (UB.ReadBadAlignment (RV l) alignment))
                            part_val
-  applyUnless (laxLoadsAndStores ?memOpts)
-              (Partial.attachMemoryError sym p1 mop UnreadableRegion)
+  -- applyUnless (laxLoadsAndStores ?memOpts)
+  (Partial.attachMemoryError sym p1 mop UnreadableRegion)
               part_val'
 
 data CacheEntry sym w =
```
</details>

If I then run the `crux-llvm` test suite, it uncovers a test failure in `T783c`:

<details>

```
    T783c.c
      z3
        loop
          pre-clang11
            z3 4.8.7
              T783c
                T783c crux run:                          OK (0.37s)
                T783c crux output:                       FAIL
                  test/Test.hs:234:
                  MISMATCH for expected (test-data/golden/T783c.good)
                             and actual (test-data/golden/T783c.z3.out) output:
                      F-expect>|[Crux] Overall status: Valid.
                      F-actual>|[Crux] Found counterexample for verification goal
                      F-actual>|[Crux]   test-data/golden/T783c.c:18:22: error: in HpCdrProcessLastSession
                      F-actual>|[Crux]   Error during memory load
                      F-actual>|[Crux]     The region wasn't allocated or wasn't large enough
                      F-actual>|[Crux]     Performing overall load at type: i8
                      F-actual>|[Crux]       Via pointer: (5, 0xb2:[64])
                      F-actual>|[Crux]     In memory state:
                      F-actual>|[Crux]       Branch frame
                      F-actual>|[Crux]         Writes:
                      F-actual>|[Crux]           Indexed chunk:
                      F-actual>|[Crux]             5 |->
                      F-actual>|[Crux]               *(5, 0x0:[64]) := arrayMap [BV 0::[64]] (select c@1:a 0x2:[64]) c@1:a
                      F-actual>|[Crux]       Stack frame HpCdrProcessLastSession
                      F-actual>|[Crux]         No writes or allocations
                      F-actual>|[Crux]       Stack frame main
                      F-actual>|[Crux]         Allocations:
                      F-actual>|[Crux]           HeapAlloc 5 0x66:[64] Mutable 16-byte-aligned <malloc> test-data/golden/T783c.c:27:27
                      F-actual>|[Crux]         Writes:
                      F-actual>|[Crux]           Indexed chunk:
                      F-actual>|[Crux]             5 |->   *(5, 0x0:[64]) := c@1:a
                      F-actual>|[Crux]       Base memory
                      F-actual>|[Crux]         Allocations:
                      F-actual>|[Crux]           GlobalAlloc 4 0x0:[64] Immutable 1-byte-aligned [defined function ] main
                      F-actual>|[Crux]           GlobalAlloc 3 0x0:[64] Immutable 1-byte-aligned [defined function ] HpCdrProcessLastSession
                      F-actual>|[Crux]           GlobalAlloc 2 0x0:[64] Immutable 1-byte-aligned [external function] malloc
                      F-actual>|[Crux]           GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [external function] llvm.dbg.value
                      F-actual>|[Crux] Found counterexample for verification goal
                      F-actual>|[Crux]   test-data/golden/T783c.c:19:25: error: in HpCdrProcessLastSession
                      F-actual>|[Crux]   Error during memory load
                      F-actual>|[Crux]     The region wasn't allocated or wasn't large enough
                      F-actual>|[Crux]     Performing overall load at type: i8
                      F-actual>|[Crux]       Via pointer: (5, 0xb6:[64])
                      F-actual>|[Crux]     In memory state:
                      F-actual>|[Crux]       Branch frame
                      F-actual>|[Crux]         Writes:
                      F-actual>|[Crux]           Indexed chunk:
                      F-actual>|[Crux]             5 |->   *(5, 0x0:[64]) := let -- test-data/golden/T783c.c:17:23
                      F-actual>|[Crux]                                           v29 = select c@1:a 0x2:[64]
                      F-actual>|[Crux]                                           -- test-data/golden/T783c.c:18:22
                      F-actual>|[Crux]                                           v43 = bvSum (bvZext 64 (select c@1:a 0x0:[64])) 0x1:[64]
                      F-actual>|[Crux]                                           -- test-data/golden/T783c.c:18:22
                      F-actual>|[Crux]                                           v48 = select (arrayMap [BV 0::[64]] v29 c@1:a) v43
                      F-actual>|[Crux]                      in arrayMap [BV 0::[64]] v29 [BV 1::[64]] v48 c@1:a
                      F-actual>|[Crux]                     *(5, 0x0:[64]) := arrayMap [BV 0::[64]] (select c@1:a 0x2:[64]) c@1:a
                      F-actual>|[Crux]       Stack frame HpCdrProcessLastSession
                      F-actual>|[Crux]         No writes or allocations
                      F-actual>|[Crux]       Stack frame main
                      F-actual>|[Crux]         Allocations:
                      F-actual>|[Crux]           HeapAlloc 5 0x66:[64] Mutable 16-byte-aligned <malloc> test-data/golden/T783c.c:27:27
                      F-actual>|[Crux]         Writes:
                      F-actual>|[Crux]           Indexed chunk:
                      F-actual>|[Crux]             5 |->   *(5, 0x0:[64]) := c@1:a
                      F-actual>|[Crux]       Base memory
                      F-actual>|[Crux]         Allocations:
                      F-actual>|[Crux]           GlobalAlloc 4 0x0:[64] Immutable 1-byte-aligned [defined function ] main
                      F-actual>|[Crux]           GlobalAlloc 3 0x0:[64] Immutable 1-byte-aligned [defined function ] HpCdrProcessLastSession
                      F-actual>|[Crux]           GlobalAlloc 2 0x0:[64] Immutable 1-byte-aligned [external function] malloc
                      F-actual>|[Crux]           GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [external function] llvm.dbg.value
                      F-actual>|[Crux] Overall status: Invalid.
```
</details>

langston-barrett:
> Splitting out the uninitialized memory initialization component from the rest of lax-loads-and-stores would be tricky since the rest of lax-loads-and-stores' behavior is specifically designed to make sure that filling in uninitialized memory with symbolic values doesn't crash later.

Thanks for looking into this, @RyanGlScott! Since it wasn't immediately obvious to either you or me how those other uses of `laxLoadsAndStores` related to uninitialized memory specifically, we should consider adding comments. Optimally these would explain *how* those uses are related to the semantics of uninitialized memory, but I would certainly settle for *that* they're related.

RyanGlScott:
At present, the only use sites of `laxLoadsAndStores` are in service of making reading from uninitialized memory work. That being said, we [intended `laxLoadsAndStores` to be a catch-all option](https://github.com/GaloisInc/crucible/issues/783#issuecomment-896251687) for different classes of memory load/store errors, so this could conceivably change in the future. I'll make sure to add comments next to each use site of `laxLoadsAndStores` to state their purpose.

If we can't disentangle the different use sites of `laxLoadsAndStores`, then there is still a question of how we should design the UI to allow picking between the different behaviors in https://github.com/GaloisInc/crucible/pull/906#issuecomment-956634140. I'm open to suggestions here, as I'm not sure what the best design would be.

RyanGlScott:
Here's one design possibility: by default, make `lax-loads-and-stores` imply one of option 2 or 3 from https://github.com/GaloisInc/crucible/pull/906#issuecomment-956634140 by default. We would also add a separate `load-indeterminate-behavior` option that has as possible values `unstable-symbolic` (i.e., option 2) and `stable-symbolic` (i.e., option 3), which only apply if `lax-loads-and-stores` is enabled. We could conceivably add another value corresponding to option 4 (`poison`, perhaps) in the future.

Alternatively, we could make another breaking change and change `lax-loads-and-stores` itself to have the `unstable-symbolic`/`stable-symbolic` enumeration as possible values instead of `yes`/`no`. This seems a bit strange to me, however, as `lax-loads-and-stores` does more than just control the behavior of indeterminate reads.

langston-barrett:
> Here's one design possibility: by default, make lax-loads-and-stores imply one of option 2 or 3 from #906 (comment) by default. We would also add a separate load-indeterminate-behavior option that has as possible values unstable-symbolic (i.e., option 2) and stable-symbolic (i.e., option 3), which only apply if lax-loads-and-stores is enabled. We could conceivably add another value corresponding to option 4 (poison, perhaps) in the future.

I like this idea a lot!

RyanGlScott:
I've pushed a new version of this PR which incorporates the suggestion in https://github.com/GaloisInc/crucible/pull/906#issuecomment-967315967. I've added a handful of documentation about this in `MemOptions`, so that would be worth taking another look at.

I'll fire off an SV-COMP run with this PR to see if there are any expected side effects of this patch.

robdockins (approved):

RyanGlScott:
I did a quick smoke-test of the `no-overflow` category in SV-COMP, and unfortunately, this patch introduces some regressions. There are 24 programs that previously returned the correct answer using the `unstable-symbolic` approach that now report the incorrect answer when using the `stable-symbolic` approach. One such program is [`uthash-2.0.2/uthash_BER_test1-1.i`](https://github.com/sosy-lab/sv-benchmarks/blob/2e1723fde6aa65a250dcb677efa45edaa4b6b631/c/uthash-2.0.2/uthash_BER_test1-1.i). With this configuration file:

```
-- For no-overflow, check the following ub-sanitizers
ub-sanitizers: [ "signed-integer-overflow", "shift",  "integer-divide-by-zero" ]
supply-main-arguments: empty

-- Don't produce counterexamples upon abnormal exits
abnormal-exit-behavior: never-fail

-- Generic SV-COMP options
clang-opts: "-fbracket-depth=1024"
solver: "z3"
timeout: 30
floating-point: "ieee"
goal-timeout: 30
make-executables: no
skip-report: yes
iteration-bound: 50
recursion-bound: 100
lax-pointer-ordering: yes
lax-constant-equality: yes
lax-arithmetic: yes
proof-goals-fail-fast: yes
hash-consing: yes
lazy-compile: no
unsat-cores: no
path-sat: yes
lax-loads-and-stores: yes
indeterminate-load-behavior: unstable-symbolic
```

The verification succeeds:

```
$ cabal run exe:crux-llvm -- --target=i386-unknown-linux-elf --config crux-llvm/svcomp/config-files/no-overflow.config /home/rscott2/bench-defs/sv-benchmarks/c/uthash-2.0.2/uthash_BER_test1-1.i
Up to date
[Crux] Using pointer width: 32 for file crux-build/crux~uthash_BER_test1-1.bc
[Crux] Overall status: Valid.
```

If I change `unstable-symbolic` to `stable-symbolic` in the configuration file, however, the verification fails:

```
$ cabal run exe:crux-llvm -- --target=i386-unknown-linux-elf --config crux-llvm/svcomp/config-files/no-overflow.config /home/rscott2/bench-defs/sv-benchmarks/c/uthash-2.0.2/uthash_BER_test1-1.i
Up to date
[Crux] Using pointer width: 32 for file crux-build/crux~uthash_BER_test1-1.bc
[Crux] Counterexample found, skipping remaining goals
[Crux]   rscott2/bench-defs/sv-benchmarks/c/uthash-2.0.2/uthash_BER_test1-1.i:708:24: error: in test_int_int
[Crux]   llvm.trap() called
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

The remaining 23 programs are also from the `uthash` family, so I suspect they have similar causes. I'll try to produce a more minimal example.

RyanGlScott:
Through the magic of CReduce, here is a tinier example:

```c
int a = 0;

int main() {
  a++;
  return 0;
}
```

With this config file:

```
ub-sanitizers: [ "signed-integer-overflow" ]
lax-loads-and-stores: yes
indeterminate-load-behavior: stable-symbolic
```

One gets:

```
$ cabal run exe:crux-llvm -- --config wat.config wat.c
Up to date
[Crux] Using pointer width: 64 for file crux-build/crux~wat.bc
[Crux] *** debug executable: results/wat/debug-4
[Crux] *** break on line: 4
[Crux]   wat.c:4:4: error: in main
[Crux]   llvm.ubsantrap() called
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

This almost suggests a bug in the implementation of `stable-symbolic`, as surely there shouldn't be any uninitialized memory in this program?

robdockins:
It sure looks like the global variable is getting initialized with symbolic bytes instead of zero initialized.

robdockins:
Indeed, we can use `crucible_dump_memory` to look at the memory state at the top of "main", and we see the following:

```
Stack frame main
  No writes or allocations
Base memory
  Allocations:
    GlobalAlloc 5 0x4:[64] Mutable 4-byte-aligned [global variable  ] a
    GlobalAlloc 4 0x0:[64] Immutable 1-byte-aligned [defined function ] main
    GlobalAlloc 3 0x0:[64] Immutable 1-byte-aligned [external function] llvm.trap
    GlobalAlloc 2 0x0:[64] Immutable 1-byte-aligned [external function] llvm.sadd.with.overflow.i32
    GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [external function] crucible_dump_memory
  Writes:
    Indexed chunk:
      5 |->   *(5, 0x0:[64]) := c@1:a
              *(5, 0x0:[64]) := c@1:a
```

There are two (?) writes to the global variable "a" with a symbolic value.  With "unstable-symbolc" selected, or with "lax-loads-and-stores" disabled, we see the expected zero initializer instead.

robdockins:
The problem is here, I think, where we use `doMalloc` to allocate space for global variables.

https://github.com/GaloisInc/crucible/blob/4f81a72d3f0aaabcfa67e7bd69bcec00d6e40b3c/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1698-L1705

This, I think, now violates the expected invariant that global variables are not already written to when "populateGobals" is called and messes up the initialization phase. I'm still not entirely sure why we see the exact results we are getting, however.

robdockins:
Indeed, altering `doMalloc` so that it only enters the `stable-symbolic` case if `allocType = G.HeapAlloc` causes this bug do disappear.

However, I'd still like to understand why it doesn't work correctly as-is.  What I expect to see is the global variable first written with symbolic bytes, and then the zero initializer being written on top, and that makes me suspect another bug is lurking here somewhere.

robdockins:
The problem appears to be in the SMT array case of `writeMemWithAllocationCheck`.  Disabling this case causes the behavior I expect (zero initializer is written after symbolic bytes).

robdockins:
I am now pretty confident the underlying problem is here:

https://github.com/GaloisInc/crucible/blob/4f81a72d3f0aaabcfa67e7bd69bcec00d6e40b3c/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs#L1266-L1272

The fall through case there should probably panic instead of doing nothing, and we need a separate case for `LLVMZero`.

RyanGlScott:
Well spotted! I've kicked off another `no-overflow` run with these latest changes. Fingers crossed that this nabs all the issues seen above!

RyanGlScott:
I have good news and bad news.

The good news is that I believe the latest round of commits fixed the bug observed in https://github.com/GaloisInc/crucible/pull/906#issuecomment-970390307, as that particular test case in `no-overflow` now passes once again.

The bad news is that there are still some lingering regressions. However, they're not the fault of this patch, but rather due to the a somewhat brittle hack that `crux-llvm-svcomp` uses to treat lacking a function override as `UNKNOWN` rather than `FALSIFIED`. Namely, it [uses a regex to search for a particular `crucible-llvm` error message](https://github.com/GaloisInc/crucible/blob/4f81a72d3f0aaabcfa67e7bd69bcec00d6e40b3c/crux-llvm/svcomp/def-files/crux.py#L40). Unfortunately, this hack broke when the error message changed in commit 16cc1c23bc81d54131440acb0701e7a41793e2ce.

I've kicked off another `no-overflow` run which uses a modified regex that should catch the new error message output. Third try's the charm?

RyanGlScott:
The latest results are in. There are only two regressions this time, one in [`busybox-1.22.0/head-2.i`](https://github.com/sosy-lab/sv-benchmarks/blob/2e1723fde6aa65a250dcb677efa45edaa4b6b631/c/busybox-1.22.0/head-2.i) and the other in [`busybox-1.22.0/yes-2.i`](https://github.com/sosy-lab/sv-benchmarks/blob/2e1723fde6aa65a250dcb677efa45edaa4b6b631/c/busybox-1.22.0/yes-2.i). Both failures are of a similar nature. Here is a minimized example of the issue:

```c
extern int a;

int main() {
  return a;
}
```

With `unstable-symbolic`, verification succeeds. With `stable-symbolic`, we instead get:

```
$ cabal run exe:crux-llvm -- --config crux-llvm/svcomp/crux-llvm-svcomp-2021-11-16-Linux-x86_64/stable.config wat.c
Up to date
[Crux] Using pointer width: 64 for file crux-build/crux~wat.bc
[Crux] Counterexample found, skipping remaining goals
[Crux]   wat.c:4:10: error: in main
[Crux]   Error during memory load
[Crux]     No previous write to this location was found
[Crux]       Attempting load at type: i32
[Crux]     Performing overall load at type: i32
[Crux]       Via pointer: Global symbol "a" (2, 0x0:[64])
[Crux]     In memory state:
[Crux]       Stack frame main
[Crux]       Base memory
[Crux]         Allocations:
[Crux]           GlobalAlloc 2 0x4:[64] Mutable 4-byte-aligned [global variable  ] a
[Crux]           GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [defined function ] main
[Crux]     in context:
[Crux]       main
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

robdockins:
Interesting... from a purely SV-COMP perspective, I guess we should treat external globals just like heap allocations (i.e., fill them with symbolic bytes)?  It doesn't seem right to assume they will be zero-initialized.

From a more general perspective... it seems pretty weird to assume anything about external globals really, but I guess this would just be something to document about how lax-loads-and-stores works.

RyanGlScott:
I wonder if `unstable-symbolic` is the thing that's buggy? In this program, for instance:

```c
#include <stdlib.h>

int a;

int main() {
  if (a && !a) {
    abort();
  }
  return 0;
}
```

Verification succeeds with `stable-symbolic` and `unstable-symbolic`. Intuitively, I would have expected the latter to fail.

robdockins:
What is the expected result for the benchmark examples?

RyanGlScott:
Both `head-2.i` and `yes-2.i` are expected to verify. That being said, the particular `extern` globals that they're using are `optarg` and `stdout`, which conceivably could have definitions somewhere.

robdockins:
> I wonder if `unstable-symbolic` is the thing that's buggy? In this program, for instance:
> 
> ```c
> #include <stdlib.h>
> 
> int a;
> 
> int main() {
>   if (a && !a) {
>     abort();
>   }
>   return 0;
> }
> ```
> 
> Verification succeeds with `stable-symbolic` and `unstable-symbolic`. Intuitively, I would have expected the latter to fail.

I'm not surprised by this.  Global variables are zero-initialized.

RyanGlScott:
> I'm not surprised by this. Global variables are zero-initialized.

My mistake!

As far what the behavior of `extern` should be, I'm somewhat perturbed that `unstable-symbolic` seems to initialize `extern` variables without corresponding definitions. I'd rather that it be an error to use such a variable, regardless of the `indeterminate-load-behavior`. This will mean that that we'll technically regress on some programs, but I'd argue that we should have never succeeded on those programs in the first place. If nothing else, we can add another regex hack to `crux-llvm-svcomp` that checks to see if an undeclared `extern` was used, much like we currently check for unimplemented overrides.

robdockins:
I think it's entirely in the spirit of "lax loads" to assume that external globals are backed by some properly-allocated memory.  The "unstable" code path just replaces any load that would otherwise fail with fresh bytes, so that behavior makes sense.  We'd have to go out of our way to make the external global case work for stable loads, though, as they currently aren't allocated at all. I think that's probably worth doing.

RyanGlScott:
OK. I suppose I was imagining that we'd want to treat `extern` variables and `extern` functions identically, but further investigation reveals that LLVM treats them quite differently. In that case, I'm not opposed to special-casing `extern` variables in the `stable-symbolic` case.

RyanGlScott:
If my understanding of how the code in `crucible-llvm` works is correct, the thing we'd need to change is [`makeGlobalMap`](https://github.com/GaloisInc/crucible/blob/79428012dae4e3c922f75c6f41f418c8aacca665/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs#L138-L141)? Somewhat brutally, that would require plumbing `?memOpts :: MemoryOpts` constraints to `translateModule`, which is `makeGlobalMap`'s call site. Urk.

robdockins:
I think we can do it all in `populateGlobals`. I have a patch incoming.

RyanGlScott:
Thanks, @robdockins! That was enough to fix `yes-2.i`, but not `head-2.i`. The latter program looks like this (minimized):

```c
extern char *optarg;

int main() {
  if (*optarg) {
    return 0;
  } else {
    return 1;
  }
}
```

This verifies with `unstable-symbolic`, but fails with `stable-symbolic`:

```
$ cabal run exe:crux-llvm -- --config crux-llvm/svcomp/crux-llvm-svcomp-2021-11-16-Linux-x86_64/stable.config wat.c
Up to date
[Crux] Using pointer width: 64 for file crux-build/crux~wat.bc
[Crux] Counterexample found, skipping remaining goals
[Crux]   wat.c:4:7: error: in main
[Crux]   Error during memory load
[Crux]     No previous write to this location was found
[Crux]       Attempting load at type: i8
[Crux]     Performing overall load at type: i8
[Crux]       Via pointer: 0x0:[64]
[Crux]     In memory state:
[Crux]       Stack frame main
[Crux]       Base memory
[Crux]         Allocations:
[Crux]           GlobalAlloc 2 0x8:[64] Mutable 8-byte-aligned [global variable  ] optarg
[Crux]           GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [defined function ] main
[Crux]         Writes:
[Crux]             2 |->   *(2, 0x0:[64]) := c@1:a
[Crux]     in context:
[Crux]       main
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

robdockins:
Looking briefly at the original head-2.c benchmark, I don't see how that can possibly be memory safe.  It seems to be assuming that any nonzero pointer is dereferenceable; which just doesn't make sense to me.

RyanGlScott:
Out of curiosity, what series of steps leads `head-2.c` to dereferencing an unwritten-to pointer? From a quick glance, it almost looks like `optarg` would have been initialized by `getopt`, but my eyes glazed over when I tried to step through the execution of that program in my head.

robdockins:
Humm, I should have looked closer I guess.  I'm seeing that this program returns UNKNOWN with reasonable iteration bounds, and thinks more or less forever otherwise, even without lax loads. I'm confused.

RyanGlScott:
Yes. To clarify, I'm defining "regression" as any benchmark that flipped from a correct verdict or `UNKOWN` to an incorrect verdict. `head-2.c` is an example of flipping from `UNKNOWN` to an incorrect verdict.

https://github.com/GaloisInc/crucible/pull/906#issuecomment-971896984 isn't really a faithful minimization of `head-2.c` in that sense. Nevertheless, it does show an example where `stable-symbolic` and `unstable-symbolic` return wildly different answers. Is this desirable?

robdockins:
This particular benchmark is calling directly into a bunch of file I/O stuff that we don't really have hooked up, so I'm not sure we should really consider this a regression.  Without lax loads turned on, we end up bumping into the fact that `stdin` isn't initialized.

As to the larger question, "unstable symbolic" basically makes "every pointer dereferenceable" as I put it above, whereas "stable symbolic" is more conservative (and more reasonable, frankly). Maybe the unstable case should at least check that the pointer to load through is at least allocated (I guess it isn't now?)

RyanGlScott:
Since the SV-COMP freeze date was pushed back to November 22, I'll wait a little longer to gather some `unreach-call` data with this PR. In the meantime, I've submitted sosy-lab/benchexec#787 to update the version of `crux.py` that it uses, which will be needed regardless.

RyanGlScott:
> As to the larger question, "unstable symbolic" basically makes "every pointer dereferenceable" as I put it above, whereas "stable symbolic" is more conservative (and more reasonable, frankly). Maybe the unstable case should at least check that the pointer to load through is at least allocated (I guess it isn't now?)

Alright, I'm OK with this difference (although I should document it, since I found that surprising).

RyanGlScott:
The results from running against the `unreach-call` SV-COMP benchmark set are in. Here are what the results were before this PR:

```
Statistics:             7753 Files                                                                                                                                                                                               
  correct:              2030                                                                                                                                                                                                     
    correct true:       933                                                                                                                                                                                                      
    correct false:      1097                                                                                                                                                                                                     
  incorrect:            1596                                                                                                                                                                                                     
    incorrect true:     0                                                                                                                                                                                                        
    incorrect false:    1596                                                                                                                                                                                                     
  unknown:              4127                                                                                                                                                                                                     
Score:                  -22573 (max: 13248)
```

And after this PR:

```
Statistics:           7753 Files
  correct:            2094
    correct true:      990
    correct false:    1104
  incorrect:          1554
    incorrect true:      2
    incorrect false:  1552
  unknown:            4105
  Score:            -21812 (max: 13248)
```

Our score improved overall, which is nice. There are some concerning things, however:

* Six programs changed from a correct verdict to an `UNKNOWN` verdict:
  * `reducercommutativity/rangesum40.yml`
  * `loops/eureka_01-1.yml`
  * `nla-digbench-scaling/dijkstra-u_valuebound20.yml`
  * `nla-digbench-scaling/lcm1_valuebound10.yml`
  * `seq-mthreaded/pals_floodmax.4.1.ufo.UNBOUNDED.pals.yml`
  * `aws-c-common/aws_add_size_checked_harness.yml`
* Three programs changed from a correct verdict to an incorrect verdict:
  * `ldv-linux-3.4-simple/32_1_cilled_ok_nondet_linux-3.4-32_1-drivers--platform--x86--intel_menlow.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.yml`
  * `ldv-linux-3.4-simple/43_1a_cilled_ok_nondet_linux-43_1a-drivers--net--arcnet--arc-rimi.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.yml`
  * `ldv-linux-3.4-simple/43_1a_cilled_ok_nondet_linux-43_1a-drivers--net--arcnet--com90io.ko-ldv_main0_sequence_infinite_withcheck_stateful.cil.out.yml`
* One program changed from a correct verdict to simply erroring out before it ever reports a verdict:
  * `loops/matrix-2.yml`
  The error in particular is `fd:4: hFlush: resource vanished (Broken pipe)`.
* There are now two programs that report incorrect true verdicts:
  * `ldv-regression/test28-1.yml`
  * `list-ext3-properties/sll_of_sll_nondet_append-2.yml`
  
  `ldv-regression/test28-1.yml` is a red herring, as that is the result of LLVM 12 more aggressively optimizing away a program that uses uninitialized memory. Moreover, that issue has been [fixed upstream](https://github.com/sosy-lab/sv-benchmarks/commit/b46a294e0963263b9f0811b5bf4b0081221ae767), so I don't think we'd see that on an actual SV-COMP competition run. I don't know what is causing the other program to report an incorrect true verdict, however.

RyanGlScott:
I talked about this with @robdockins, and we concluded that the benefits from this patch outweigh the drawbacks (in this case, the small handful of regressions observed in https://github.com/GaloisInc/crucible/pull/906#issuecomment-975636496). I've opened #926 as a reminder to investigate these more thoroughly. Some of these might not be fixable—for instance, it might the case that these programs only verified with `unstable-symbolic` due to a lucky fluke—but it would at least be worthwhile to characterize the underlying issues.

Once CI looks like it will pass, I'll land this.

PR #913, opened 2021-11-05
uc-crux-llvm: Check inferred function contracts

langston-barrett:
Still WIP. Needs:

- [x] Plumbing to expose through CLI
- [x] Test cases
- [x] Probably formatting fixes in the pretty-printed messages

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Constraints.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Constraints.hs
@@ -20,6 +20,7 @@ Stability    : provisional
 module UCCrux.LLVM.Constraints
   ( -- * Types of constraints
     ShapeConstraint (..),
+    ppShapeConstraint,
     Constraint (..),
     ppConstraint,
     RelationalConstraint (..),
@@ -116,6 +117,18 @@ data ShapeConstraint (m :: Type) (atTy :: FullType m) where
 
 deriving instance Eq (ShapeConstraint m atTy)
 
+ppShapeConstraint :: ShapeConstraint m atTy -> Doc ann
+ppShapeConstraint =
+  \case
+    Allocated sz ->
+      PP.pretty "points to allocated space for"
+      PP.<+> PP.viaShow sz
+      PP.<+> PP.pretty "elements"
+    Initialized sz ->
+      PP.pretty "points to"
+      PP.<+> PP.viaShow sz
+      PP.<+> PP.pretty "initialized elements"
+
 -- | A 'Constraint' is in general, something that can be \"applied\" to a
 -- symbolic value to produce a predicate. In practice, these represent
 -- \"missing\" function preconditions that are deduced by
@@ -134,7 +147,7 @@ instance Eq (Constraint m atTy) where
       (Aligned n1, Aligned n2) -> n1 == n2
       (BVCmp op1 _ bv1, BVCmp op2 _ bv2) -> op1 == op2 && bv1 == bv2
 
-ppConstraint :: Constraint m ft -> Doc Void
+ppConstraint :: Constraint m ft -> Doc ann
 ppConstraint =
   \case
     Aligned alignment ->
@@ -150,10 +163,10 @@ ppConstraint =
     BVCmp L.Isgt w bv -> PP.pretty "is (signed) greater than " <> signed w bv
     BVCmp L.Isge w bv -> PP.pretty "is (signed) greater than or equal to " <> signed w bv
   where
-    signed :: forall w. (1 <= w) => NatRepr w -> BV w -> PP.Doc Void
+    signed :: forall ann w. (1 <= w) => NatRepr w -> BV w -> PP.Doc ann
     signed w bv =
       PP.viaShow (BV.asSigned w bv) PP.<+> PP.parens (PP.pretty (BV.ppHex w bv))
-    unsigned :: forall w. (1 <= w) => NatRepr w -> BV w -> PP.Doc Void
+    unsigned :: forall ann w. (1 <= w) => NatRepr w -> BV w -> PP.Doc ann
     unsigned w bv =
       PP.viaShow (BV.asUnsigned bv) PP.<+> PP.parens (PP.pretty (BV.ppHex w bv))
 
--- a/uc-crux-llvm/src/UCCrux/LLVM/Cursor.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Cursor.hs
@@ -36,13 +36,14 @@ module UCCrux.LLVM.Cursor
     selectorCursor,
     Where (..),
     selectWhere,
+    ppWhere,
+    ppSelector
   )
 where
 
 {- ORMOLU_DISABLE -}
 import           Control.Lens (Lens, lens, (^.))
 import           Data.Semigroupoid (Semigroupoid(o))
-import           Data.Void (Void)
 import           Data.Type.Equality
 import           Prettyprinter (Doc)
 import qualified Prettyprinter as PP
@@ -211,12 +212,12 @@ ppCursor ::
   -- | Top level, e.g. the name of a variable
   String ->
   Cursor m inTy atTy ->
-  Doc Void
+  Doc ann
 ppCursor top =
   \case
     Here _ft -> PP.pretty top
     Dereference 0 (Field _fieldTypes idx cursor) ->
-      ppCursor top cursor <> PP.pretty "->" <> PP.pretty (show idx)
+      ppCursor top cursor <> PP.pretty "->" <> PP.viaShow idx
     Dereference 0 what -> PP.pretty "*" <> ppCursor top what
     Dereference idx what -> ppCursor top what <> PP.pretty ("[" ++ show idx ++ "]")
     Index idx _len cursor -> ppCursor top cursor <> PP.pretty ("[" ++ show idx ++ "]")
@@ -253,6 +254,20 @@ selectWhere =
     SelectReturn fSymb _ ->
       let L.Symbol f = getFuncSymbol fSymb
        in ReturnValue f
+
+ppWhere :: Where -> PP.Doc ann
+ppWhere =
+  \case
+    Arg n -> PP.pretty "in argument #" <> PP.viaShow n
+    Global g -> PP.pretty "in global" PP.<+> PP.pretty g
+    ReturnValue f ->
+      PP.pretty "in return value of skipped function" PP.<+> PP.pretty f
+
+ppSelector :: Selector m argTypes inTy atTy -> PP.Doc ann
+ppSelector selector =
+  ppWhere (selectWhere selector) PP.<+>
+    PP.pretty "at" PP.<+>
+    ppCursor "<top>" (selector ^. selectorCursor)
 
 -- | For documentation of the type parameters, see the comment on 'Cursor'.
 --
--- a/uc-crux-llvm/src/UCCrux/LLVM/Main.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Main.hs
@@ -86,6 +86,7 @@ import qualified UCCrux.LLVM.Logging as Log
 import qualified UCCrux.LLVM.Main.Config.FromEnv as Config.FromEnv
 import           UCCrux.LLVM.Main.Config.Type (TopLevelConfig)
 import qualified UCCrux.LLVM.Main.Config.Type as Config
+import           UCCrux.LLVM.Module (defnSymbolToString)
 import           UCCrux.LLVM.Run.EntryPoints (makeEntryPointsOrThrow)
 import           UCCrux.LLVM.Run.Explore (explore)
 import           UCCrux.LLVM.Run.Result (BugfindingResult(..), SomeBugfindingResult(..))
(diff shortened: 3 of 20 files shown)


langston-barrett:
Here's what it looks like right now.

```c
#include <limits.h>

int f(int x) __attribute__((noinline))  {
  return INT_MAX - 2 + x;
}

int g(int x) __attribute__((noinline))  {
  return f(0); // no problem
}

int h(int x) __attribute__((noinline))  {
  return f(3); // overflow!
}
```
Executing from `g` and checking the contract for `f` (safe):
```
$ cabal v2-run exe:uc-crux-llvm -- --entry-points f --check-from g tmp.c
[UC-Crux-LLVM] Results for f
[UC-Crux-LLVM] Function is safe if deduced preconditions are met:
[UC-Crux-LLVM] Arguments:
[UC-Crux-LLVM]   An integer: is (signed) less than 2 (0x2:[32])
[UC-Crux-LLVM] Globals:
[UC-Crux-LLVM] Return values of skipped functions:
When executing from g,
encountered these calls to functions which had inferred contracts:
When calling f
In g at tmp.c:8:10
With arguments:
- Argument #0 x: literal integer: unsigned value = 0, signed value =  0, width = 32
Checked constraints:
Constraint: is (signed) less than 2 (0x2:[32])
Applied at: in argument #0 at <top>
Result: concretely true
```
Executing from `h` and checking the contract for `f` (violation):
```
$ cabal v2-run exe:uc-crux-llvm -- --entry-points f --check-from g tmp.c
[UC-Crux-LLVM] Results for f
[UC-Crux-LLVM] Function is safe if deduced preconditions are met:
[UC-Crux-LLVM] Arguments:
[UC-Crux-LLVM]   An integer: is (signed) less than 2 (0x2:[32])
[UC-Crux-LLVM] Globals:
[UC-Crux-LLVM] Return values of skipped functions:
When executing from h,
encountered these calls to functions which had inferred contracts:
When calling f
In h at tmp.c:12:10
With arguments:
- Argument #0 x: literal integer: unsigned value = 3, signed value =  3, width = 32
Checked constraints:
Constraint: is (signed) less than 2 (0x2:[32])
Applied at: in argument #0 at <top>
Result: concretely false
```

travitch:
Very cool! It is almost unsettling to see an idea just work

uc-crux-llvm/src/UCCrux/LLVM/Run/Result.hs
```diff
@@ -77,19 +79,26 @@ ppFunctionSummaryTag =
...
 data FunctionSummary m (argTypes :: Ctx (FullType m))
   = Unclear (NonEmpty (Located Uncertainty))
   | FoundBugs (NonEmpty (Located TruePositive))
   | SafeWithPreconditions DidHitBounds Unsoundness (Constraints m argTypes)
   | SafeUpToBounds Unsoundness
   | AlwaysSafe Unsoundness
 
-data SomeBugfindingResult
-  = forall m arch argTypes.
+data SomeBugfindingResult m arch
```
travitch, line 92: This could use a description of its contents - in particular, why is there a sequence of results?

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
```diff
@@ -138,6 +144,20 @@ data SomeCheckedConstraint m sym (argTypes :: Ctx (FullType m)) =
   forall inTy atTy.
     SomeCheckedConstraint (CheckedConstraint m sym argTypes inTy atTy)
 
+data SomeCheckedConstraint' m =
+  forall sym argTypes inTy atTy.
+    SomeCheckedConstraint' (CheckedConstraint m sym argTypes inTy atTy)
+
+-- NOTE(lb): The explicit kind signature here is necessary for GHC 8.8/8.6
+-- compatibility.
+data CheckedCall m sym arch (argTypes :: Ctx (FullType m)) =
```
travitch, line 153: What does this type represent? Is this recording all of the constraints checked at a given call site (in a given context)?

travitch: Also, are these created inside the generated checking override? Do we strip the override from the call stack?

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
```diff
@@ -0,0 +1,406 @@
...
+import           UCCrux.LLVM.Run.EntryPoints (EntryPoints, getEntryPoints)
+import qualified UCCrux.LLVM.Run.Simulate as Sim
+import qualified UCCrux.LLVM.Run.Loop as Loop
+import           UCCrux.LLVM.Run.Result (SomeBugfindingResult)
+import qualified UCCrux.LLVM.Run.Result as Result
+import           UCCrux.LLVM.Setup (SymValue)
+import           UCCrux.LLVM.Shape (Shape)
+{- ORMOLU_ENABLE -}
+
+newtype SomeCheckedCalls m sym arch =
```
travitch, line 95: This type looks complicated enough to warrant an explanation. What is it for? What should a caller think of that inner continuation as?

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
```diff
@@ -0,0 +1,406 @@
...
+              ovFunCtx <-
+                case makeFunctionContext modCtx func argFTys argCTys of
+                  Left err ->
+                    panic
+                      "checkInferredContracts"
+                      [Text.unpack (ppFunctionContextError err)]
+                  Right funCtx' -> return funCtx'
+              let ?memOpts = CruxLLVM.memOpts llOpts
+              case testEquality argFTys types of
+                Nothing -> panic "checkInferredContracts" []
```
travitch, line 206: Is the error here indicating an unexpected type mismatch?

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
```diff
@@ -0,0 +1,406 @@
...
+            Sim.SimulatorHooks
+              { Sim.createOverrideHooks = map fst ovs
+              , Sim.resultHook =
+                \sym mem args cruxResult ucResult ->
+                  return $
+                    CheckResult $
+                      \k -> k sym mem args cruxResult ucResult (map snd ovs)
+              })
+  where
+    overrides ::
```
travitch, line 183: I'd appreciate a comment here. Is this just creating one checking override for each passed-in contract?

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
```diff
@@ -318,7 +340,7 @@ createCheckOverride ::
   AppContext ->
   ModuleContext m arch ->
   -- | Predicates checked during simulation
-  IORef (Map CheckOverrideName [(Stack sym, Seq (SomeCheckedConstraint m sym argTypes))]) ->
+  IORef [CheckedCall m sym arch argTypes] ->
```
travitch, line 343: How should I think about this `IORef`? Does the order that entries are added matter? Who is ultimately consulting it?

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
```diff
@@ -0,0 +1,406 @@
...
+
+           funCtx <-
+             case makeFunctionContext modCtx entry argFTys (Crucible.cfgArgTypes cfg) of
+               Left err ->
+                 panic
+                   "checkInferredContracts"
+                   [Text.unpack (ppFunctionContextError err)]
+               Right funCtxF -> return funCtxF
+           result <-
+            checkInferredContracts
```
travitch, line 258: Does it matter that there are multiple `IORef`s tracking checked functions under here? Is it necessary that there are multiple?

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
```diff
@@ -0,0 +1,406 @@
...
+              halloc
+              cruxOpts
+              llOpts
+              (emptyConstraints argFTys)
+              cfg
+              contracts
+           return (entry, SomeCheckResult funCtx result)
+
+-- | Infer preconditions for a group of functions, then for the ones that are
+-- safe-with-preconditions, check their inferred preconditions by seeing if they
```
travitch, line 271: Safe with preconditions, but also where no execution limits were hit

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
langston-barrett: Ah whoops, I meant to go back and add more details here. Yeah, that's exactly right. I'm considering adding another kind of phantom type variable `f` to give type-level names to functions like the current `m` gives type-level names to LLVM modules. Then we could avoid panics/possible problems like this. I'll continue to think about it.

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
langston-barrett: When you say "there are multiple `IORef`s", do you mean the `IORef`s that are passed to `createCheckOverride`?

If so, then I think it's fine in the sense that `checkInferredContracts` returns a way to read each of them (obliquely, through a `SomeCheckedCalls`).

> Is it necessary that there are multiple?

I think it makes the interface to `createCheckOverride` nicer if the results can depend on the (type-level parameters that indicate) types of the arguments to the function. Since these functions might all have different, non-statically-known types, we have to then have one `IORef` per check override.

uc-crux-llvm/src/UCCrux/LLVM/Constraints.hs
```diff
@@ -116,6 +117,18 @@ data ShapeConstraint (m :: Type) (atTy :: FullType m) where
 
 deriving instance Eq (ShapeConstraint m atTy)
 
+ppShapeConstraint :: ShapeConstraint m atTy -> Doc ann
+ppShapeConstraint =
+  \case
+    Allocated sz ->
+      PP.pretty "points to allocated space for"
+      PP.<+> PP.pretty (show sz)
```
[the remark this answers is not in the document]
langston-barrett: Agreed, and I added a custom HLint rule so I'll be consistent about this in the future.

uc-crux-llvm/src/UCCrux/LLVM/Cursor.hs
```diff
@@ -254,6 +255,20 @@ selectWhere =
       let L.Symbol f = getFuncSymbol fSymb
        in ReturnValue f
 
+ppWhere :: Where -> PP.Doc ann
+ppWhere =
+  \case
+    Arg n -> PP.pretty "in argument #" <> PP.pretty (show n)
```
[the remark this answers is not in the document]
langston-barrett: Agreed, and I added a custom HLint rule so I'll be consistent about this in the future.

uc-crux-llvm/src/UCCrux/LLVM/Main.hs
```diff
@@ -156,41 +163,44 @@ mainWithConfigs appCtx cruxOpts topConf =
...
-           void $
-             flip Map.traverseWithKey results $
-               \func (SomeBugfindingResult result _) ->
-                 Log.sayUCCruxLLVM
-                   ( Log.Results
-                       (Text.pack func)
-                       (Result.printFunctionSummary (summary result))
-                   )
+           let printResult results =
+                 void $
```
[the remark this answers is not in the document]
langston-barrett: Nice catch!

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
langston-barrett: Added comments.

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
langston-barrett: You're absolutely right thanks. Added comments.

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
langston-barrett: Yep, exactly. Comment added.

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Check.hs
langston-barrett: Added a comment.

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
langston-barrett: You're right - fixed!

uc-crux-llvm/src/UCCrux/LLVM/Run/Result.hs
langston-barrett: Comment added.

uc-crux-llvm/src/UCCrux/LLVM/Run/Check.hs
travitch: Okay, that makes sense to me. Yes - I was thinking of that set of IORefs

travitch (approved):

PR #914, opened 2021-11-06
crucible-llvm: Annotate errors with call stacks

langston-barrett:
Still WIP, since:

- [x] There are probably other projects in this repo that will need changes to typecheck
- [x] The Crux-LLVM golden tests are probably failing
- [x] Most importantly: There's a spot where I could use help on how to proceed (see inline comment)

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Eval.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Eval.hs
@@ -45,7 +45,7 @@ llvmExtensionEval sym _iTypes _logFn eval e =
   case e of
     X86Expr ex -> X86.eval sym eval ex
 
-    LLVM_SideConditions _tp conds val ->
+    LLVM_SideConditions _ _tp conds val ->
       do conds' <- traverse (traverseF (\x -> RV @sym <$> eval x)) (NE.toList conds)
          forM_ conds' (assertSideCondition sym)
          eval val
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Eval.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Eval.hs
@@ -4,8 +4,10 @@
 {-# LANGUAGE TypeApplications #-}
 module Lang.Crucible.LLVM.Eval
   ( llvmExtensionEval
+  , callStackFromMemVar
   ) where
 
+import           Control.Lens ((^.), view)
 import           Control.Monad (forM_)
 import qualified Data.List.NonEmpty as NE
 import           Data.Parameterized.TraversableF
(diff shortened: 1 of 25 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Eval.hs
```diff
@@ -30,18 +32,20 @@ assertSideCondition ::
   LLVMSideCondition (RegValue' sym) ->
   IO ()
 assertSideCondition sym (LLVMSideCondition (RV p) ub) =
-  do p' <- annotateUB sym ub p
+  do p' <- annotateUB sym noCallStack ub p
```
langston-barrett, line 35: In order to get a callstack from the LLVM memory here, it looks like

- Expression extension evaluators will need (read-only) access to the `CrucibleState`. Not sure if this is acceptable (e.g., is there an `IORef` in there that would be problematic if it were modified by evaluation of an expression?).
- I'll need to find some way to feed in the `mvar` (memory global variable) into `llvmExtensionEval`. For statements, this is embedded in their syntax. Not sure if the same thing is appropriate or possible here.

robdockins: Extensions already essentially have unlimited access to mutate the crucible state in the statement evaluator, I don't see a problem threading it in here.

I'm guessing you'll have access to the `mvar` as necessary when generating side conditions, so that's probably the way to go, yeah.

robdockins:
Glancing through the code, I don't see where the callstacks are generated from.  Are they just compute from the sequence of pushed frames in the memory?

robdockins:
This PR highlights something that I think we'll need to address going forward as we refactor the memory model.  Right now, the memory is tightly coupled with a program's control flow via the `StackFrame` and `BranchFrame` constructors.  If we ever want to take concurrency seriously, we'll need to decouple these concepts, as control flow is an essentially thread-local concept.

langston-barrett:
The only remaining CI failure seems unrelated (it's during installation of Nix).

RyanGlScott:
> it's during installation of Nix

Sigh. It looks like NixOS/nix#3605 is biting again.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
```diff
@@ -573,7 +575,10 @@ llvmAbsOverride ::
 llvmAbsOverride w =
   let nm = L.Symbol ("llvm.abs.i" ++ show (natValue w)) in
     [llvmOvr| #w $nm( #w, i1 ) |]
-    (\_memOpts sym args -> Ctx.uncurryAssignment (Libc.callLLVMAbs sym w) args)
+    (\mvar sym args ->
+     do callStack <-
+          getCallStack . view (to memImplHeap . memState) <$> readGlobal mvar
```
robdockins, line 580: callStackFromMemVar'  ?

robdockins (approved):

PR #915, opened 2021-11-06
crucible-llvm: Simplify 'MemoryErrorReason'

langston-barrett:
There was only one constructor of this type that depended on type
parameters 'sym' and 'w', and that was a pointer that is duplicated in
the 'MemoryOp' that accompanies each 'MemoryErrorReason'.

Also, create a new error datatype instead of using a pretty-printed doc.

In addition to making 'MemoryError' more normalized, 'MemoryErrorReason'
is now friendlier to programmatic use. In particular, it's orderable.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Errors/MemoryError.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Errors/MemoryError.hs
@@ -25,18 +25,19 @@ module Lang.Crucible.LLVM.Errors.MemoryError
 , ppMemoryOp
 , MemoryErrorReason(..)
 , ppMemoryErrorReason
+, FuncLookupError(..)
+, ppFuncLookupError
 
 , concMemoryError
 , concMemoryOp
-, concMemoryErrorReason
 ) where
 
 import           Prelude hiding (pred)
 
 import           Data.Text (Text)
-import           Data.Void
 import qualified Text.LLVM.PP as L
 import qualified Text.LLVM.AST as L
+import           Type.Reflection (SomeTypeRep(SomeTypeRep))
 import           Prettyprinter
 
 import           What4.Interface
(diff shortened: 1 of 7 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Errors/MemoryError.hs
```diff
@@ -64,23 +64,43 @@ data MemoryOp sym w
...
   | ApplyViewFail ValueView
   | Invalid StorageType
   | Invalidated Text
-  | NoSatisfyingWrite StorageType (LLVMPtr sym w)
+  | NoSatisfyingWrite StorageType
   | UnwritableRegion
   | UnreadableRegion
-  | BadFunctionPointer (Doc Void)
+  | BadFunctionPointer FuncLookupError
   | OverlappingRegions
```
langston-barrett, line 82: Whoops, I meant to make this `deriving (Eq, Ord)` as well.

crux-llvm/test-data/golden/golden/invoke-test.good
```diff
@@ -2,9 +2,9 @@
 [Crux]   internal: error: in _ZN3std2rt10lang_start17h4f32aa1279b9079fE
 [Crux]   Failed to load function handle
 [Crux]   Details:
-[Crux]     No implementation or override found for pointer: "_ZN3std2rt19lang_start_internal17h3dc68cf5532522d7E"
+[Crux]     No implementation or override found for pointer
 [Crux]     The given pointer could not be resolved to a callable function
-[Crux]     No implementation or override found for pointer: "_ZN3std2rt19lang_start_internal17h3dc68cf5532522d7E"
+[Crux]     No implementation or override found for pointer
 [Crux]     Attempting to load callable function with type: i64({ }*, [3 x i64]*, i64, i8**)
 [Crux]       Via pointer: Global symbol "_ZN3std2rt19lang_start_internal17h3dc68cf5532522d7E" (1, 0x0:[64])
```
langston-barrett, line 9: Note that the function name that is removed from this message appears here at line 9. It was just duplicated between the `MemoryOp` and the `MemoryErrorReason`.

robdockins (approved):

PR #916, opened 2021-11-08
uc-crux-llvm: Add source position traces to debug log

langston-barrett:
Makes it easier to comprehend where the simulator went. These data are already being collected by Crux, so why not make use of them?

### DIFF
--- a/crux/src/Crux/Report.hs
+++ b/crux/src/Crux/Report.hs
@@ -91,7 +91,7 @@ provedGoalLocs = concatMap Fold.toList . provedGoalTraces
 -- | Return a list of all of the traces referenced in a set of proved goals.
 --
 -- This returns a sequence-of-sequences because a single 'ProvedGoals' can
--- involve many 'Branch', which mirror the branching structure of the program
+-- involve many 'Branch'es, which mirror the branching structure of the program
 -- execution that led to each individual 'ProvedGoal' or 'NotProvedGoal'.
 provedGoalTraces :: ProvedGoals -> Seq (Seq ProgramLoc)
 provedGoalTraces =
--- a/crux/src/Crux/Report.hs
+++ b/crux/src/Crux/Report.hs
@@ -12,6 +12,8 @@ import qualified Data.Foldable as Fold
 import Data.Functor.Const
 import Data.List (partition)
 import Data.Maybe (fromMaybe)
+import qualified Data.Sequence as Seq
+import           Data.Sequence (Seq)
 import qualified Data.Set as Set
 import qualified Data.Text as Text
 import Control.Exception (catch, SomeException(..))
@@ -84,10 +86,21 @@ maybeGenerateSource opts files =
 -- | Return a list of all program locations referenced in a set of
 -- proved goals.
 provedGoalLocs :: ProvedGoals -> [ProgramLoc]
-provedGoalLocs (Branch goals1 goals2) = provedGoalLocs goals1 ++
-                                        provedGoalLocs goals2
-provedGoalLocs (ProvedGoal _ err locs _) = locs ++ [simErrorLoc err]
-provedGoalLocs (NotProvedGoal _ err _ locs _) = locs ++ [simErrorLoc err]
+provedGoalLocs = concatMap Fold.toList . provedGoalTraces
+
+-- | Return a list of all of the traces referenced in a set of proved goals.
+--
+-- This returns a sequence-of-sequences because a single 'ProvedGoals' can
+-- involve many 'Branch', which mirror the branching structure of the program
+-- execution that led to each individual 'ProvedGoal' or 'NotProvedGoal'.
+provedGoalTraces :: ProvedGoals -> Seq (Seq ProgramLoc)
+provedGoalTraces =
+  \case
+    Branch pgs1 pgs2 -> provedGoalTraces pgs1 <> provedGoalTraces pgs2
+    ProvedGoal _ err locs _ ->
+      Seq.singleton (Seq.fromList locs Seq.|> simErrorLoc err)
+    NotProvedGoal _ err _ locs _ ->
+      Seq.singleton (Seq.fromList locs Seq.|> simErrorLoc err)
 
 -- | Return a list of all files referenced in a set of proved goals.
 provedGoalFiles :: ProvedGoals -> Set.Set FilePath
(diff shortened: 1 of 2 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
```diff
@@ -536,7 +536,17 @@ mkCallbacks appCtx modCtx funCtx halloc callbacks constraints cfg llvmOpts =
...
       IO r
     mkResultHook sym skipOverrideRef unsoundOverrideRef explRef memRef argShapeRef cruxResult resHook =
-      do unsoundness' <-
+      do for_ (traces cruxResult) $
+           \trace ->
+             do liftIO $ (appCtx ^. log) Hi "Trace:"
+                for_ trace $
+                  \loc ->
+                    let prettyLoc = ppProgramLoc loc
+                    in if prettyLoc == "internal"
```
RyanGlScott, line 545: This relies on the implementation of `Position`'s `Pretty` output, right? Wouldn't it be less fragile to match against the `InternalPos` constructor instead?

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
```diff
@@ -555,6 +565,16 @@ mkCallbacks appCtx modCtx funCtx halloc callbacks constraints cfg llvmOpts =
          argShapes <- fromMaybe rd <$> IORef.readIORef argShapeRef
          resHook sym mem argShapes cruxResult ucCruxResult
 
+    traces :: Crux.CruxSimulationResult -> Set [What4.ProgramLoc]
+    traces =
+      let go =
+            \case
+              Crux.Branch pgs1 pgs2 -> go pgs1 ++ go pgs2
```
RyanGlScott, line 572: This loop is very close to `crux`'s [`provedGoalLocs`](https://github.com/GaloisInc/crucible/blob/a6669a5fe657edecc61e294dd706e04ebb071776/crux/src/Crux/Report.hs#L84-L90) function. Is it possible to just use that instead? Note that `provedGoalLocs` returns more `ProgramLoc`s than this code does, as `provedGoalLocs` also includes `simErrorLoc`s—I'm not sure if that is a deal-breaker for `uc-crux-llvm`.

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
langston-barrett: Yep, this is sheer laziness. Will fix.

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
langston-barrett: Ah, thanks for pointing out that function!

langston-barrett: I generalized the function in Crux, let me know what you think.

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
```diff
@@ -536,7 +537,16 @@ mkCallbacks appCtx modCtx funCtx halloc callbacks constraints cfg llvmOpts =
...
     mkResultHook sym skipOverrideRef unsoundOverrideRef explRef memRef argShapeRef cruxResult resHook =
-      do unsoundness' <-
+      do for_ (traces cruxResult) $
+           \trace ->
+             do liftIO $ (appCtx ^. log) Hi "Trace:"
+                for_ trace $
+                  \loc ->
+                    if What4.InternalPos == What4.plSourceLoc loc
+                    then pure ()
+                    else liftIO $ (appCtx ^. log) Hi ("  " <> ppProgramLoc loc)
```
RyanGlScott, line 547: This can be simplified slightly with `Control.Monad.unless`:

```suggestion
                    unless (What4.InternalPos == What4.plSourceLoc loc) $
                      liftIO $ (appCtx ^. log) Hi ("  " <> ppProgramLoc loc)
```

crux/src/Crux/Report.hs
```diff
@@ -84,10 +86,17 @@ maybeGenerateSource opts files =
...
 -- proved goals.
 provedGoalLocs :: ProvedGoals -> [ProgramLoc]
-provedGoalLocs (Branch goals1 goals2) = provedGoalLocs goals1 ++
-                                        provedGoalLocs goals2
-provedGoalLocs (ProvedGoal _ err locs _) = locs ++ [simErrorLoc err]
-provedGoalLocs (NotProvedGoal _ err _ locs _) = locs ++ [simErrorLoc err]
+provedGoalLocs = concatMap Fold.toList . provedGoalTraces
+
+-- | Return a list of all of the traces referenced in a set of proved goals.
+provedGoalTraces :: ProvedGoals -> Seq (Seq ProgramLoc)
```
RyanGlScott, line 92: Why do you use nested `Seq`s here? Perhaps you're trying to preserve some structure here, but if so, the structure should be documented.

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
langston-barrett: Thanks!

crux/src/Crux/Report.hs
langston-barrett: Documented, let me know if that makes things more clear.

crux/src/Crux/Report.hs
```diff
@@ -84,10 +86,21 @@ maybeGenerateSource opts files =
...
-provedGoalLocs (Branch goals1 goals2) = provedGoalLocs goals1 ++
-                                        provedGoalLocs goals2
-provedGoalLocs (ProvedGoal _ err locs _) = locs ++ [simErrorLoc err]
-provedGoalLocs (NotProvedGoal _ err _ locs _) = locs ++ [simErrorLoc err]
+provedGoalLocs = concatMap Fold.toList . provedGoalTraces
+
+-- | Return a list of all of the traces referenced in a set of proved goals.
+--
+-- This returns a sequence-of-sequences because a single 'ProvedGoals' can
+-- involve many 'Branch', which mirror the branching structure of the program
```
RyanGlScott, line 94:
```suggestion
-- involve many 'Branch'es, which mirror the branching structure of the program
```

crux/src/Crux/Report.hs
RyanGlScott: Thanks, that's helpful.

RyanGlScott (approved):

crux/src/Crux/Report.hs
langston-barrett: Ah, for some reason I always thought Haddock didn't like other text surrounding the quotes. TIL! Thanks for the review.

PR #921, opened 2021-11-11
uc-crux-llvm: Introduce a newtype for pre-simulation memory

### LINKED ISSUE #912: uc-crux-llvm: Newtype for pre-simulation, populated LLVM memory
UC-Crux-LLVM passes around a lot of `MemImpl`s, often for the sake of displaying data to a user, e.g., printing values with `ppLLVMValWithGlobals`. Many of these uses are of the *initial* program memory, i.e. memory as it was laid out just before starting simulation. A `newtype` would help make this more clear.

langston-barrett:
This concept was used throughout the codebase, so it's nice to have types act as documentation.

Fixes #912.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Newtypes/PreSimulationMem.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Newtypes/PreSimulationMem.hs
@@ -1,0 +1,23 @@
+{-
+Module           : UCCrux.LLVM.Newtypes.PreSimulationMem
+Copyright        : (c) Galois, Inc 2020
+License          : BSD3
+Maintainer       : Langston Barrett <langston@galois.com>
+Stability        : provisional
+-}
+
+module UCCrux.LLVM.Newtypes.PreSimulationMem
+  ( PreSimulationMem
+  , makePreSimulationMem
+  , getPreSimulationMem
+  ) where
+
+import           Lang.Crucible.LLVM.MemModel (MemImpl)
+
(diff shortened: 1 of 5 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Newtypes/PreSimulationMem.hs
```diff
@@ -0,0 +1,23 @@
...
+module UCCrux.LLVM.Newtypes.PreSimulationMem
+  ( PreSimulationMem
+  , makePreSimulationMem
+  , getPreSimulationMem
+  ) where
+
+import           Lang.Crucible.LLVM.MemModel (MemImpl)
+
+-- | LLVM memory just before symbolic execution. Populated with globals, defined
+-- functions, and arguments to the entry point.
```
travitch, line 18: I like newtypes, but what is the motivation? Is it that we want to inspect pre-states sometimes and want to make sure that we don't mix up memimpls?

travitch (approved):

uc-crux-llvm/src/UCCrux/LLVM/Newtypes/PreSimulationMem.hs
langston-barrett: Yeah, pretty much! It's to distinguish this kind of use from other `MemImpl`s, hopefully making it easier to understand what's going on, and specifically, what sorts of reads one would expect to work on this particular `MemImpl`.

PR #923, opened 2021-11-17
Add crucible-syntax support for arbitrary syntax extensions

bboston7:
This change adds a `ParserHooks` datatype that library users can
leverage to define parsing rules for additional types and syntax
extensions.  It also provides a `defaultParserHooks` instance that
existing crucible-syntax client code can use to preserve the current
behavior of the library.

### DIFF
--- a/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
@@ -29,9 +29,16 @@ module Lang.Crucible.Syntax.Concrete
     ExprErr(..)
   -- * Parsing and Results
   , ACFG(..)
+  , ParserHooks(..)
+  , defaultParserHooks
   , top
   , cfgs
   , prog
+  -- * Low level parsing operations
+  , SyntaxState(..)
+  , atomName
+  , freshAtom
+  , operands
   -- * Rules for pretty-printing language syntax
   , printExpr
   )
@@ -73,6 +80,7 @@ import qualified Data.Text as T
 import qualified Data.Vector as V
 import Numeric.Natural
 
+import qualified Lang.Crucible.CFG.Extension as LCCE
 import Lang.Crucible.Syntax.ExprParse hiding (SyntaxError)
 import qualified Lang.Crucible.Syntax.ExprParse as SP
 
@@ -112,20 +120,20 @@ printExpr = toText (PrintRules rules)
         rules (Kw Registers) = Just (Special 0)
         rules _ = Nothing
 
-data E s t where
-  EAtom  :: !(Atom s t) -> E s t
-  EReg   :: !Position -> !(Reg s t) -> E s t
-  EGlob  :: !Position -> !(GlobalVar t) -> E s t
-  EDeref :: !Position -> !(E s (ReferenceType t)) -> E s t
-  EApp   :: !(App () (E s) t) -> E s t
+data E s ext t where
+  EAtom  :: !(Atom s t) -> E s ext t
+  EReg   :: !Position -> !(Reg s t) -> E s ext t
+  EGlob  :: !Position -> !(GlobalVar t) -> E s ext t
+  EDeref :: !Position -> !(E s ext (ReferenceType t)) -> E s ext t
+  EApp   :: !(App ext (E s ext) t) -> E s ext t
 
-data SomeExpr :: * -> * where
-  SomeE :: TypeRepr t -> E s t -> SomeExpr s
-  SomeOverloaded :: AST s -> Keyword -> [SomeExpr s] -> SomeExpr s
-  SomeIntLiteral :: AST s -> Integer -> SomeExpr s
+data SomeExpr ext s where
+  SomeE :: TypeRepr t -> E s ext t -> SomeExpr ext s
+  SomeOverloaded :: AST s -> Keyword -> [SomeExpr ext s] -> SomeExpr ext s
+  SomeIntLiteral :: AST s -> Integer -> SomeExpr ext s
 
-data SomeBVExpr :: * -> * where
-  SomeBVExpr :: (1 <= w) => NatRepr w -> E s (BVType w) -> SomeBVExpr s
+data SomeBVExpr ext s where
+  SomeBVExpr :: (1 <= w) => NatRepr w -> E s ext (BVType w) -> SomeBVExpr ext s
 
 data ExprErr s where
   TrivialErr :: Position -> ExprErr s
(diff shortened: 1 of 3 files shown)


robdockins:
I wonder if this might look a little nicer to use implicit parameters to pass the parser hooks around instead of doing it explicitly?

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -1749,49 +1842,53 @@ termStmt' retTy =
...
+  ACFG :: forall (s :: *) (init :: Ctx CrucibleType) (ret :: CrucibleType) ext .
+          ( LCCE.IsSyntaxExtension ext ) =>
           CtxRepr init -> TypeRepr ret ->
-          CFG () s init ret ->
-          ACFG
+          CFG ext s init ret ->
+          ACFG ext
 
-deriving instance Show ACFG
+deriving instance Show (ACFG ())
```
travitch, line 1881: Very minor - is it possible to make this cover the generic case? Probably not a huge deal if not

travitch (commented):
Looks straightforward to me - I like it!

I'll leave the call on implicit parameters up to Rob, but will register my appreciation for making the hooks explicit. It isn't that noisy, and it makes it clear where data flows

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -112,20 +120,20 @@ printExpr = toText (PrintRules rules)
         rules (Kw Registers) = Just (Special 0)
         rules _ = Nothing
 
-data E s t where
-  EAtom  :: !(Atom s t) -> E s t
-  EReg   :: !Position -> !(Reg s t) -> E s t
-  EGlob  :: !Position -> !(GlobalVar t) -> E s t
-  EDeref :: !Position -> !(E s (ReferenceType t)) -> E s t
-  EApp   :: !(App () (E s) t) -> E s t
+data E s ext t where
```
RyanGlScott, line 123: Why does `ext` appear after `s` here? That seems to go against the `ext s` convention used in other data types (`SomeExpr`, `SomeBVExpr`, `Stmt`, `Block`, `CFG`, etc.).

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -1749,49 +1842,53 @@ termStmt' retTy =
...
-                 depCons (located (reading (check argTy))) $
+                 depCons (located (reading (check argTy hooks))) $
                    \(Posd loc arg) ->
                      emptyList *>
                        (Output lbl <$> eval loc arg)
 
 
 
-data Rand s t = Rand (AST s) (E s t)
+data Rand s ext t = Rand (AST s) (E s ext t)
```
RyanGlScott, line 1866: Similarly, `Rand ext s t` seems like it would fit in with the existing conventions better than `Rand s ext t`.

crucible-syntax/src/Lang/Crucible/Syntax/Prog.hs
```diff
@@ -62,7 +62,7 @@ doParseCheck fn theInput pprint outh =
          do when pprint $
               forM_ v $
                 \e -> T.hPutStrLn outh (printExpr e) >> hPutStrLn outh ""
-            cs <- top ng ha [] $ cfgs v
+            cs <- top ng ha [] $ (cfgs defaultParserHooks) v
```
RyanGlScott, line 65: These parentheses are redundant:
```suggestion
            cs <- top ng ha [] $ cfgs defaultParserHooks v
```

crucible-syntax/src/Lang/Crucible/Syntax/Prog.hs
```diff
@@ -95,7 +95,7 @@ simulateProgram fn theInput outh profh opts setup =
             extendConfig opts (getConfiguration sym)
             ovrs <- setup @() @_ @() sym ha
             let hdls = [ (SomeHandle h, p) | (FnBinding h _,p) <- ovrs ]
-            parseResult <- top ng ha hdls $ cfgs v
+            parseResult <- top ng ha hdls $ (cfgs defaultParserHooks) v
```
RyanGlScott, line 98:
```suggestion
            parseResult <- top ng ha hdls $ cfgs defaultParserHooks v
```

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -244,18 +281,20 @@ repUntilLast sp = describe "zero or more followed by one" $ repUntilLast' sp
           (cons p emptyList <&> \(x, ()) -> ([], x)) <|>
           (cons p (repUntilLast' p) <&> \(x, (xs, lst)) -> (x:xs, lst))
 
-_isBaseType :: MonadSyntax Atomic m => m (Some BaseTypeRepr)
-_isBaseType =
+_isBaseType :: MonadSyntax Atomic m => ParserHooks ext -> m (Some BaseTypeRepr)
+_isBaseType hooks =
   describe "base type" $
-  do Some tp <- isType
+  do Some tp <- (isType hooks)
```
RyanGlScott, line 287:
```suggestion
  do Some tp <- isType hooks
```

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -244,18 +281,20 @@ repUntilLast sp = describe "zero or more followed by one" $ repUntilLast' sp
...
 
-_isFloatingType :: MonadSyntax Atomic m => m (Some FloatInfoRepr)
-_isFloatingType =
+_isFloatingType :: MonadSyntax Atomic m
+                => ParserHooks ext
+                -> m (Some FloatInfoRepr)
+_isFloatingType hooks =
   describe "floating-point type" $
-  do Some tp <- isType
+  do Some tp <- (isType hooks)
```
RyanGlScott, line 297:
```suggestion
  do Some tp <- isType hooks
```

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
bboston7: Yep.  Done

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
bboston7: Good point.  I fixed this for `E` and `Rand`.

bboston7:
@robdockins The last commit I just pushed up makes the parser hooks an implicit parameter

PR #925, opened 2021-11-22
Assorted SV-COMP fixes

RyanGlScott:
This adds a variety of tweaks and fixes that I have incorporated based on SV-COMP reviews for Crux.


robdockins (approved):
I'm not set up to test the results of the driver script changes, but I think that addresses the concerns of the reviewers.

PR #930, opened 2021-11-29
uc-crux-llvm: Improve tracking of unclassified errors

langston-barrett:
Introduce a new type 'Bug' that keeps both more and less information than the old type 'Unclassified'. It tracks the program location of the issue, but leaves out any symbolic terms or copies of the program memory.

The new type can be ordered and compared for equality, which will facilitate programatic reasoning about unclassified errors.


langston-barrett:
Whoops, I forgot CI doesn't like PRs from forks. Will reopen.

PR #931, opened 2021-11-29
uc-crux-llvm: Improve tracking of unclassified errors

langston-barrett:
Introduce a new type 'Bug' that keeps both more and less information than the old type 'Unclassified'. It tracks the program location and callstack of the issue, but leaves out any symbolic terms or copies of the program memory.

The new type can be ordered and compared for equality, which will facilitate programatic reasoning about unclassified errors.


uc-crux-llvm/src/UCCrux/LLVM/Bug.hs
```diff
@@ -0,0 +1,96 @@
...
+
+import           UCCrux.LLVM.Bug.UndefinedBehaviorTag (UndefinedBehaviorTag, getUndefinedBehaviorTag, makeUndefinedBehaviorTag)
+import           UCCrux.LLVM.PP (ppProgramLoc)
+{- ORMOLU_ENABLE -}
+
+-- | This is different from 'Lang.Crucible.LLVM.Errors.BadBehavior' in that
+-- it stores less data.
+data BugBehavior
+  = BBUndefinedBehaviorTag !UndefinedBehaviorTag
+  | BBMemoryErrorReason MemoryErrorReason
```
travitch, line 48: If the other field is strict, is there a reason to not make this one strict?

uc-crux-llvm/src/UCCrux/LLVM/Bug.hs
```diff
@@ -0,0 +1,96 @@
...
+    BBUndefinedBehaviorTag ub -> UB.explain (getUndefinedBehaviorTag ub)
+    BBMemoryErrorReason mer -> MemErrors.ppMemoryErrorReason mer
+
+instance PP.Pretty BugBehavior where
+  pretty = ppBugBehavior
+
+-- | A possible bug: What it is, and where it can occur.
+data Bug =
+  Bug
+    { bugBehavior :: BugBehavior
```
travitch, line 63: Same strictness question here

uc-crux-llvm/src/UCCrux/LLVM/Bug/UndefinedBehaviorTag.hs
```diff
@@ -0,0 +1,630 @@
...
+makeUndefinedBehaviorTag = UndefinedBehaviorTag . fmapF (const (Const ()))
+
+-- | This instance is a coarsening of that for 'UndefinedBehavior'. In
+-- particular, there may be instances of 'UndefinedBehavior' that do not compare
+-- equal, but their projections under 'makeUndefinedBehaviorTag' do compare
+-- equal.
+--
+-- The less coarse this becomes, the more effortful it gets to write, but also
+-- the better it will be for deduplicating possible bugs/errors.
+instance Eq UndefinedBehaviorTag where
```
travitch, line 45: Can this just call `testEqualityC`?

uc-crux-llvm/src/UCCrux/LLVM/Bug/UndefinedBehaviorTag.hs
```diff
@@ -0,0 +1,630 @@
...
+      (UB.SRemByZero (Const ()) (Const ()), UB.SRemByZero (Const ()) (Const ())) -> True
+      (UB.SDivOverflow (Const ()) (Const ()), UB.SDivOverflow (Const ()) (Const ())) -> True
+      (UB.SRemOverflow (Const ()) (Const ()), UB.SRemOverflow (Const ()) (Const ())) -> True
+      (UB.AbsIntMin (Const ()), UB.AbsIntMin (Const ())) -> True
+      -- NB: We ignore what kind of poison it was.
+      (UB.PoisonValueCreated _p1, UB.PoisonValueCreated _p2) -> True
+      _ -> False
+
+-- | See comment on 'Eq' instance.
+instance Ord UndefinedBehaviorTag where
```
travitch, line 77: Likewise, can we reuse `compareC` here?

uc-crux-llvm/src/UCCrux/LLVM/Bug/UndefinedBehaviorTag.hs
langston-barrett: That would be nice, but I tried it and couldn't make it work. Specializing the type of `testEqualityC`...
```haskell
(forall x y. f x -> f y -> Maybe (x :~: y)) -> t f -> t f -> Bool

(forall x y. f x -> f y -> Maybe (x :~: y)) -> 
  UndefinedBehavior f -> 
  UndefinedBehavior f -> 
  Bool

(forall x y. Const () x -> Const () y -> Maybe (x :~: y)) -> 
  UndefinedBehavior (Const ()) -> 
  UndefinedBehavior (Const ()) -> 
  Bool
```
There's the rub: There's no safe way to make a function that could produce `x :~: y` out of two `Const ()`s.

We could consider adding enough type information to `UndefinedBehavior` (i.e., adding a bunch of `NatRepr`s) so that we could make `UndefinedBehaviorTag ~ UndefinedBehavior TypeRepr`, which would be easier to handle.

travitch: Ah, I just had another hacky thought. You can define:
```
data NoValues (tp :: CrucibleType) where
  NoValues :: NoValues BoolType

equalWithoutValue :: NoValues tp1 -> NoValues tp2 -> Maybe (tp1 :~: tp2)
equalWithoutValue NoValues NoValues = Just Refl
```

langston-barrett: Those are both good ideas, but I don't think either can work. Addressing the second one first:
```haskell
data OneValue (tp :: CrucibleType) where
  OneValue :: OneValue BoolType

equalWithoutValue :: OneValue tp1 -> OneValue tp2 -> Maybe (tp1 :~: tp2)
equalWithoutValue OneValue OneValue = Just Refl

newtype UndefinedBehaviorTag =
  UndefinedBehaviorTag { getUndefinedBehaviorTag :: UndefinedBehavior OneValue }

makeUndefinedBehaviorTag :: UndefinedBehavior e -> UndefinedBehaviorTag
makeUndefinedBehaviorTag = UndefinedBehaviorTag . fmapF (const OneValue)
```
yields
```
src/UCCrux/LLVM/Bug/UndefinedBehaviorTag.hs:46:58-71: error:
    • Couldn't match type ‘x’ with ‘BoolType’
      ‘x’ is a rigid type variable bound by
        a type expected by the context:
          forall (x :: CrucibleType). e x -> OneValue x
        at src/UCCrux/LLVM/Bug/UndefinedBehaviorTag.hs:46:57-72
      Expected type: e x -> OneValue x
        Actual type: e x -> OneValue BoolType
    • In the first argument of ‘fmapF’, namely ‘(const OneValue)’
      In the second argument of ‘(.)’, namely ‘fmapF (const OneValue)’
      In the expression: UndefinedBehaviorTag . fmapF (const OneValue)
   |
46 | makeUndefinedBehaviorTag = UndefinedBehaviorTag . fmapF (const OneValue)
   |
```
which makes sense - there are constructors of `UndefinedBehavior` like
```haskell
  FreeBadOffset ::
    (1 <= w) =>
    e (LLVMPointerType w) ->
    UndefinedBehavior e
```
but we have nothing to put in that field - there aren't any matching `OneValue` constructors that can have that type parameter (`LLVMPointerType w`).

And for the first one, it's also impossible to write `makeUndefinedBehaviorTag` - there just isn't enough information contained in an `UndefinedBehavior` to reconstruct a `TypeRepr` (which you can also see from the `FreeBadOffset` constructor above - it has a constraint on `w` but no `NatRepr w`). That's why I suggested:

> We could consider adding enough type information to UndefinedBehavior (i.e., adding a bunch of NatReprs) so that we could make UndefinedBehaviorTag ~ UndefinedBehavior TypeRepr, which would be easier to handle.

langston-barrett: I've pushed a version that uses `unsafeCoerce`.

robdockins: Hold on now... not happy with a solution that ends with "use unsafeCoerce" unless we really have no options.

robdockins: It looks like you can arrange to have an `IsExpr` contraint in the right places, so you should be able to get a `BaseTypeRepr` in the places where you need them.

travitch: That right place might just be in `makeUndefinedBehaviorTag` - you could then use that constraint when mapping over terms to case on a type repr and produce a type-correct surrogate value

travitch: And that surrogate value could just be the type repr

langston-barrett: Nice suggestion, @robdockins! That works like a charm. I did have to add
```haskell
llvmPointerType :: IsExpr (SymExpr sym) => LLVMPtr sym w -> TypeRepr (LLVMPointerType w)
llvmPointerType ptr =
  case exprType (llvmPointerOffset ptr) of
    BaseBVRepr w -> LLVMPointerRepr w
```
to `Lang.Crucible.LLVM.MemModel.Pointer` to make it work, but that seems OK to me?

robdockins (approved):

robdockins:
Looks like GHC 8.10 is being picky about the GADT extension, but this otherwise looks good to me.

PR #939, opened 2021-12-09
uc-crux-llvm: Improve error handling in cases of a malformed LLVM module

langston-barrett:
Previously, UC-Crux `panic`ed in a few places that could potentially indicate malformed LLVM modules. This change makes it more clear what might be Clang/rustc/whatever bugs.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
@@ -121,20 +121,16 @@ ppFunctionContextError =
     BadMemType _ -> "Bad MemType"
 
 throwFunctionContextError :: FunctionContextError -> a
-throwFunctionContextError =
-  \case
-    err@(BadMemType {}) ->
-      malformedLLVMModule
-        "loopOnFunction"
-        (PP.pretty (ppFunctionContextError err))
-        []
-    err@(BadLift {}) ->
-      malformedLLVMModule
-        "loopOnFunction"
-        (PP.pretty (ppFunctionContextError err))
-        []
-    err@(BadLiftArgs {}) ->
-      panic "loopOnFunction" [Text.unpack (ppFunctionContextError err)]
+throwFunctionContextError err =
+  case err of
+    BadMemType {} -> isMalformed
+    BadLift {} -> isMalformed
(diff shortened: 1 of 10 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
```diff
@@ -117,24 +120,51 @@ ppFunctionContextError =
...
   ArchOk arch =>
   ModuleContext m arch ->
   DefnSymbol m ->
   Ctx.Assignment (FullTypeRepr m) fullTypes ->
   Ctx.Assignment CrucibleTypes.TypeRepr (MapToCrucibleType arch fullTypes) ->
-  Either FunctionContextError (FunctionContext m arch fullTypes)
+  FunctionContext m arch fullTypes
 makeFunctionContext modCtx defnSymb argFullTypes argTypes =
+  case tryMakeFunctionContext modCtx defnSymb argFullTypes argTypes of
+    Left err -> throwFunctionContextError err
```
travitch, line 156: I suggest putting `HasCallStack` constraints on `throwFunctionContextError` and all of the functions that call it, so that you get call stacks for free on panics.

uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
```diff
@@ -117,24 +120,51 @@ ppFunctionContextError =
...
 
+throwFunctionContextError :: FunctionContextError -> a
+throwFunctionContextError err =
+  case err of
+    BadMemType {} -> isMalformed
+    BadLift {} -> isMalformed
+    BadLiftArgs {} -> doPanic
+  where
+    nm = "throwFunctionContextError"
+    doPanic = panic nm [Text.unpack (ppFunctionContextError err)]
```
travitch, line 131: If you do add `HasCallStack` constraints, these intermediate bindings could be problematic (unless they get their own `HasCallStack` constraints

uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
langston-barrett: Okay, let me read a little bit about that and I'll add those constriants! Thanks for the suggestion.

PR #945, opened 2022-01-05
Separate backend

robdockins:
This is a large, but mostly mechanical change. The idea is to separate the control-flow-specific aspects of the "symbolic backend" from the purely expression-constructing aspects. This will allow more principled use of multiple symbolic execution contexts from separate threads of execution, and is also probably a worthwhile conceptual distinction to draw.

The main idea is that a Crucible "backend" now contains a What4 "expression builder" rather than the two aspects being directly bundled into a single object. This allows multiple backends to share the same expression builder without stomping on each other's assumption contexts, and makes it easier to, e.g., configure path satisfiability solvers without committing to that choice elsewhere.

There is still some "context isolation" work that should go into the underlying What4 library, but that will come in a separate PR.


crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -378,58 +379,61 @@ ppLoc :: ProgramLoc -> PP.Doc ann
...
-  , IsSymExprBuilder sym
+  ( IsSymExprBuilder sym
   , IsInterpretedFloatSymExprBuilder sym
   )
 
 -- | This class provides operations that interact with the symbolic simulator.
 --   It allows for logical assumptions/assertions to be added to the current
 --   path condition, and allows queries to be asked about branch conditions.
-class IsBoolSolver sym where
+class IsBoolSolver sym bak | bak -> sym where
```
RyanGlScott, line 389: I would find it helpful to have the Haddocks clarify what exactly the `bak` type variable represents, as well as the relationship between `bak` and `sym`.

RyanGlScott (commented):
I haven't made an effort to review this in depth yet, partly because this is still marked as a draft. If you'd like a more thorough review, let me know which commits are the most interesting to look at—I get the feeling a lot of these changes are mostly mechanical, which makes the diff harder to analyze.

Also, would it be possible to give a high-level summary of the more noticeable changes in the `crucible` changelog? It's likely not possible to anticipate every kind of change that will be required, but I get the feeling there are a lot of common patterns in what kind of code has to be adapted.

robdockins:
Yeah, this PR isn't 100% baked yet. I want to be sure I can push these changes all the way through SAWScript before I'm sure it's ready.

robdockins:
I have enough downstream PRs now in place and passing tests that I think this is ready to be looked at.  @spernsteiner, I'm mostly interested in your perspective on the crux-mir changes, as I have less of an understanding of how that code works.

@RyanGlScott, I think the most interesting commits to examine are probably:
https://github.com/GaloisInc/crucible/pull/945/commits/84cad0a4ab0b259962ab649a68191e81a62bc443
https://github.com/GaloisInc/crucible/pull/945/commits/9c6b4fa54e2e077808af3fad16a389562eae0503
https://github.com/GaloisInc/crucible/pull/945/commits/619c4c725084dd031de82fd172f03df73ba218dd
https://github.com/GaloisInc/crucible/pull/945/commits/fcd54ce453174706532597bd0eb7e2e3183ba02f

The rest are pretty much just the same kinds of changes repeated in a bunch of places.

crucible/src/Lang/Crucible/Simulator/RegMap.hs
```diff
@@ -141,17 +144,26 @@ muxAny s itefns p (AnyValue tpx x) (AnyValue tpy y)
   | Just Refl <- testEquality tpx tpy =
        AnyValue tpx <$> muxRegForType s itefns tpx p x y
   | otherwise =
-    addFailedAssertion s $
```
RyanGlScott, line 144: What is the rationale for using `throwIO` over `addFailedAssertion` here (and in other places in this commit)?

crux/src/Crux.hs
```diff
@@ -301,29 +313,30 @@ defaultOutputConfig logMessagesToSayWhat = do
...
     "default" -> case selectedSolvers of
-                   [oneSolver] -> CCS.withDefaultFloatRepr proxy oneSolver k
-                   _           -> k FloatUninterpretedRepr
+                   [oneSolver] -> CCS.withDefaultFloatRepr proxy proxy' oneSolver k
+                   _           -> k WEB.FloatUninterpretedRepr
     fm -> fail ("Unknown floating point mode: " ++ fm ++ "; expected one of [real|ieee|uninterpreted|default]")
 
-floatReprString :: FloatModeRepr fm -> String
-floatReprString floatRepr =
+_floatReprString :: WEB.FloatModeRepr fm -> String
```
RyanGlScott, line 334: Is this function still needed? If not, can it be removed?

crucible/src/Lang/Crucible/Utils/MuxTree.hs
```diff
@@ -78,6 +83,78 @@ muxTreeCmpOp sym f xt yt = orOneOf sym folded =<< sequence zs
   xs = viewMuxTree xt
   ys = viewMuxTree yt
 
+
+-- | Compute an equality predicate on mux trees.
+--
+--   NOTE! This assumes the equality relation
+--   defined by `Eq` is the semantic equality
+--   relation on `a`.
```
RyanGlScott, line 91:
```suggestion
--   relation on @a@.
```

crucible/src/Lang/Crucible/Utils/MuxTree.hs
```diff
@@ -78,6 +83,78 @@ muxTreeCmpOp sym f xt yt = orOneOf sym folded =<< sequence zs
...
+  MuxTree sym a ->
+  IO (Pred sym)
+muxTreeEq sym = muxTreeCmpOp sym f
+  where f x y = pure (backendPred sym (x == y))
+
+-- | Compute a less-than predicate on mux trees.
+--
+--   NOTE! This assumes the order relation
+--   defined by `Ord` is the semantic order
+--   relation on `a`.
```
RyanGlScott, line 105:
```suggestion
--   relation on @a@.
```

crucible/src/Lang/Crucible/Utils/MuxTree.hs
```diff
@@ -78,6 +83,78 @@ muxTreeCmpOp sym f xt yt = orOneOf sym folded =<< sequence zs
...
+  MuxTree sym a ->
+  IO (Pred sym)
+muxTreeLt sym = muxTreeCmpOp sym f
+  where f x y = pure (backendPred sym (x < y))
+
+-- | Compute a less-than-or-equal predicate on mux trees.
+--
+--   NOTE! This assumes the order relation
+--   defined by `Ord` is the semantic order
+--   relation on `a`.
```
RyanGlScott, line 119:
```suggestion
--   relation on @a@.
```

crucible/src/Lang/Crucible/Simulator/RegMap.hs
spernsteiner: I'm also wondering about this.  Having to separately call `getCurrentProgramLoc` + `throwIO` seems a little annoying.

spernsteiner:
crux-mir parts look fine to me.

spernsteiner (approved):

crucible/src/Lang/Crucible/Simulator/RegMap.hs
robdockins: `addFailedAssertion` requires the "backend" not just the "sym", and I'm not entirely convinced that adding a proof obligation for situations where the simulator doesn't support some feature is semantically correct anyways.

Maybe it's better to make a little combinator that makes this pattern more concise.

crucible-llvm/test/MemSetup.hs
```diff
@@ -74,11 +79,12 @@ withLLVMCtx mod' action =
           let ?ptrWidth = width
           let ?lc = LLVMTr._llvmTypeCtx ctx
           let ?recordLLVMAnnotation = \_ _ _ -> pure ()
-          action ctx sym
+          action ctx bak
         }}}
   in withIONonceGenerator $ \nonceGen ->
      withHandleAllocator  $ \halloc   -> with nonceGen halloc
 
+data CrucibleLLVMTestData t = CrucibleLLVMTestData
```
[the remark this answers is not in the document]
robdockins: Previously the "backend" data structures occupied the "user data" slot of a What4 expression builder.  We have now lifted those data structures out separately, so now that user data slot is unoccupied; but something has to go there when calling `newExprBuilder`.  Maybe it's better to add a `EmptyUserData` type to `What4.Expr.Builder` instead of separate ones at use sites.  There ended up being more of them than I expected.

crucible-go/src/Lang/Crucible/Go/Simulate.hs
```diff
@@ -73,20 +73,20 @@ setSimulatorVerbosity verbosity sym = do
   return ()
 
 goExtensionEval ::
-  forall sym p ext rtp blocks r ctx.
-  (IsSymInterface sym) =>
-  sym ->
+  forall sym bak p ext rtp blocks r ctx.
+  (IsSymInterface sym, IsBoolSolver sym bak) =>
```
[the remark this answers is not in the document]
robdockins: It's a good question, and I don't know if I have a firm opinion.  For the initial refactoring, it was easier to have them be separate, I think, to help the type checker guide me to places where changes needed to happen. Now, I'm not sure what the best way to arrange this is.  

I'm also not necessarily a huge fan of the names of these constraints.  Any ideas?

crux/src/Crux/Config/Solver.hs
```diff
@@ -29,20 +29,25 @@ data SolverOffline = SolverOnline SolverOnline | Boolector | DReal
   deriving (Eq, Ord, Show)
 
 class HasDefaultFloatRepr solver where
-  withDefaultFloatRepr :: proxy s -> solver -> (forall fm . (WIF.IsInterpretedFloatExprBuilder (WEB.ExprBuilder s CBS.SimpleBackendState (WEB.Flags fm))) => WEB.FloatModeRepr fm -> a) -> a
+  withDefaultFloatRepr ::
+    proxy s ->
+    proxy' st ->
```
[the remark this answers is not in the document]
robdockins: Good question, I'll have to check the call sites to see if that makes sense.

crucible-go/src/Lang/Crucible/Go/Simulate.hs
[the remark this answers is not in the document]
RyanGlScott: I'm generally fond of the `IsSymInterface` approach of bundling all of the common classes up into a type synonym simply because it makes it easier to write code for common cases. This is especially valuable when I write applications on top of `crucible`, since I don't have to think about the various distinctions between each class's expressive power, decreasing my overall cognitive burden.

I can't think of many scenarios where you'd shoot yourself in the foot by picking such a class (is there such a scenario?), so I'd favor offering one. I'm not great at coming up with names, but perhaps `IsSymSolverInterface` would do. I don't think you need to go through the effort of using it everywhere in `crucible`, however, since it's ultimately a convenience for users.

> I'm also not necessarily a huge fan of the names of these constraints. Any ideas?

I struggle to think of better names, primarily since these classes represent a grab-bag collection of ways to interact with expressions and solvers in various ways. I suppose you could object to the use of "`Bool`" in `IsBoolSolver`, but I'm not convinced that a new name would make its purpose more obvious. For better or worse, I think you just have to read the Haddocks to figure out what it's about.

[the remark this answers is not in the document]
robdockins: Thinking about this some more, I think the issue is that there are a lot of code paths where the `bak` type variable is hidden, so it makes it hard to carry around a single combined constraint that mentions both.  That leaves you with a few options 1) two flavors (one that mentions `bak` and one that doesn't) 2) two separate constraints, as here or 3) make `IsBoolSolver` take the constituents of `IsSymInterface` as superclass constraints.

Maybe 3 is best?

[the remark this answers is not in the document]
RyanGlScott: I'd be happy with option (3). The Haddocks for `IsBoolSolver sym bak` even state that `sym` is expected to satisfy the `IsSymInterface` constraints, but this is currently not enforced, which seems odd.

RyanGlScott (approved):

robdockins:
Test failure appears transient.  Merging anyway.

PR #947, opened 2022-01-06
crux-mir: remove default initialValue for crucible::any::Any

spernsteiner:
Previously, locals of type `Any` would be initialized to a (useless) default value.  If the local was declared under a symbolic branch, then the value assigned under that branch would then get muxed with the default value, producing an error ("Attempted to mux ANY values of different runtime type") since the value assigned almost never had the same concrete/runtime type as the default.  This was a major problem for #945, which halts simulation when that error occurs instead of turning it into an assertion failure.

With this branch, we no longer set a default value for `Any`-typed locals.  (I forget the exact mechanism that allows this - probably something involving `MaybeType` - but the same mechanism is already used for various other non-defaultable types, like function pointers.)  Removing the default value eliminates the unnecessary mux and the associated error.

### DIFF
--- a/crux-mir/src/Mir/Trans.hs
+++ b/crux-mir/src/Mir/Trans.hs
@@ -1525,6 +1525,10 @@ initialValue (CTyArray t) = tyToReprM t >>= \(Some tpr) -> case tpr of
 initialValue ty@(CTyBv _sz) = tyToReprM ty >>= \(Some tpr) -> case tpr of
     C.BVRepr w -> return $ Just $ MirExp (C.BVRepr w) $ S.app $ eBVLit w 0
     _ -> mirFail $ "Bv type " ++ show ty ++ " does not have BVRepr"
+-- `Any` values have no reasonable default.  Any default we provide might get
+-- muxed with actual non-default values, which will fail (unless the concrete
+-- type happens to match exactly).
+initialValue CTyAny = return Nothing
 initialValue CTyMethodSpec = return Nothing
 initialValue CTyMethodSpecBuilder = return Nothing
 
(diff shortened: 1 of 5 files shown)


crux-mir/src/Mir/Trans.hs
```diff
@@ -1525,6 +1525,10 @@ initialValue (CTyArray t) = tyToReprM t >>= \(Some tpr) -> case tpr of
 initialValue ty@(CTyBv _sz) = tyToReprM ty >>= \(Some tpr) -> case tpr of
     C.BVRepr w -> return $ Just $ MirExp (C.BVRepr w) $ S.app $ eBVLit w 0
     _ -> mirFail $ "Bv type " ++ show ty ++ " does not have BVRepr"
+-- `Any` values have no reasonable default.  Any default we provide might get
+-- muxed with actual non-default values, which will fail (unless the concrete
+-- type happens to match exactly).
+initialValue CTyAny = return Nothing
```
robdockins, line 1531: I'm a little confused by this change.  At a first glance it seems like this should make no difference, as the pattern doesn't appear previously and the fallthrough case already returns `Nothing`.  Is this a pattern synonym that overlaps with something else?

spernsteiner: `CTyAny` is an pattern synonym for `TyAdt` with a specific name.  Without this case, it falls through to the `M.TyAdt` `Struct` case.  (`crucible::any::Any` is declared as a struct in Rust, but its crucible representation is overridden in crux-mir.)  The `Struct` case happens to produce a crucible `AnyType` value (crux-mir wraps all Rust ADTs in `AnyType`), but the value it produces isn't useful or meaningful, and causes the error described above.

robdockins (approved):
LGTM

CI failure appears unrelated.

PR #948, opened 2022-01-06
Support building with GHC 9.0

RyanGlScott:
Most of the action occurs in the first two commits, whose commit message are reproduced below:

# Support building with GHC 9.0

This contains a variety of fixes needed to make the packages in the `crucible`
repo compile with GHC 9.0:

* GHC 9.0 implements simplified subsumption (see
  [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0?version_id=5fcd0a50e0872efb3c38a32db140506da8310d87#simplified-subsumption)).
  As the packages in the `crucible` repo use quite a bit of nested `forall`s,
  this required a fair bit of code churn to accommodate:
  * In some cases, the easiest fix was to simply eta expand the call sites of
    some functions. See, for example, the call sites for `mainCont` in
    `crucible-concurrency`, `unTranslatedExpr` in `crucible-go`,
    `llvmOverride_def`/`buildRegTypeMap`/etc. in `crucible-llvm`,
    and `traverseFC` in `crucible`.
  * In other cases, it was easier to rearrange the order of `forall`s in type
    signatures to avoid excessive amounts of eta expansion. See, for example,
    the changes to `llvmOverride_def`'s type signatures in
    `crucible-llvm:Lang.Crucible.LLVM.Intrinsics.Common`. Without these, every
    call site to these functions in the module would have needed to be eta
    expanded, which would have been a more intrusive change.

    Similar `forall` surgery was performed in the instance signatures for
    `TraversableFC` instances in `uc-crux-llvm`, as well as some type
    signatures in `crucible-llvm` that end with `OverrideSim`.
* GHC 9.0's pattern-match coverage checker is smarter (see
  [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0?version_id=5fcd0a50e0872efb3c38a32db140506da8310d87#improved-pattern-match-coverage-checker)).
  As a result, it is able to detect that a pattern match on `BitCast` in
  `crucible-llvm`'s `evalGEP` function is redundant. Unfortunately, older GHCs
  aren't smart enough to realize this, so we cannot just remove this case
  without incurring a warning on older GHCs. As a result, I have CPP'd it
  out on GHC 9.0+ only.
* GHC 9.0 is stricter about parsing `@` signs in visible type applications (see
  [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0?version_id=5fcd0a50e0872efb3c38a32db140506da8310d87#whitespace-sensitive-and-)).
  As a result, GHC no longer permits having whitespace after the `@` in a
  visible type application, so an expression like `knownNat @ 42` is rejected—
  one must write `knownNat @42` instead. This affects visible type applications
  in `crux-mir` and `crucible-llvm`.
* GHC's constraint solver now solves constraints in each top-level group
  sooner (see
  [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0?version_id=5fcd0a50e0872efb3c38a32db140506da8310d87#the-order-of-th-splices-is-more-important)).
  This affects `uc-crux-llvm`, as it separates top-level groups with
  `$(return [])` Template Haskell splices. The previous locations of these
  splices made it so that the TH-generated instances in that package were not
  available to any code before the splice, resulting in type errors when
  compiled with GHC 9.0.

  To overcome this, I rearranged the TH-generated instances so that they appear
  before the top-level groups that make use of them.
* GHC 9.0 is pickier about the use of nested `forall`s and contexts in instance
  declarations (see
  [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.0?version_id=5fcd0a50e0872efb3c38a32db140506da8310d87#ghc-is-pickier-about-nested-foralls-and-contexts-in-instance-and-deriving-declarations)).
  This requires removing some extraneous parentheses around derived instances
  in `crucible-symio`, as GHC 9.0 would reject them otherwise.
* Some incorrect uses of CPP in `uc-crux-llvm` were removed.
* Various upper version bounds on `base` were raised to `< 4.16` to allow
  building with `base-4.15.*`, which is bundled with GHC 9.0.

# Fix some warnings discovered by GHC 9.0

* GHC 9.0 now enables `-Wstar-is-type` in `-Wall`, so this patch replaces some
  uses of `*` with `Data.Kind.Type` in `crucible-syntax`. `Data.Kind` requires
  the use of GHC 8.0 or later, so this patch also updates
  `crucible-syntax.cabal`'s lower bounds on `base` to `>= 4.9`. (I'm fairly
  certain that this requirement was already present implicity, but better to
  be explicit about it.)
* The `otherwise` guard in `crucible-concurrency`'s `eventBeforeProc` function
  is redundant, as discovered by GHC 9.0's more intelligent pattern-match
  coverage checker.
* The submodules for `aig` and `golang` were bumped to allow them to build with
  GHC 9.0 without warnings.

-----

Now that we test GHC 9.0 in CI, we should remove the GHC 8.6 configurations at some point to avoid an excessive number of jobs. I will do this in a separate PR.


README.md
```diff
@@ -136,6 +136,31 @@ Testing with coverage tracking is currently only available via
 Use `scripts/stack-test-coverage.sh` to generate a coverage
 report for all test suites.
 
+Notes on Freeze Files
+---------------------
+
+We use the `cabal.GHC-*.config` files to constraint dependency versions in CI.
```
langston-barrett, line 142:
```suggestion
We use the `cabal.GHC-*.config` files to constrain dependency versions in CI.
```

langston-barrett (approved):

robdockins (approved):
Looks OK to me, once we get CI green

.github/workflows/crucible-go-build.yml
```diff
@@ -77,10 +92,16 @@ jobs:
           SOLVER_PKG_VERSION: "snapshot-20210917"
           BUILD_TARGET_OS: ${{ matrix.os }}
 
+      - name: Setup Environment Vars (GHC 8.6.5)
+        if: matrix.ghc == '8.6.5'
```
RyanGlScott, line 96: All the `crucible-go` jobs work on CI now, which is exciting. I'm confused as to why this works, however. The `cachix/install-nix-actions` steps are divided up by whether the GHC version is 9.0.2 or not, but on the other hand, the `Setup Environment Vars` steps are divided up by whether the GHC version is 8.6.5 or not. Why can't they be the same?

.github/workflows/crucible-go-build.yml
```diff
@@ -77,10 +92,16 @@ jobs:
...
+        run: |
+          GHC=haskell.compiler.ghc$(echo ${{ matrix.ghc }} | sed -e s,\\.,,g)
+          echo NS="nix shell github:nixos/nixpkgs/20.09#cabal-install github:nixos/nixpkgs/20.09#$GHC nixpkgs#gmp nixpkgs#zlib nixpkgs#zlib.dev" >> $GITHUB_ENV
       - name: Setup Environment Vars
+        if: matrix.ghc != '8.6.5'
         run: |
           GHC=haskell.compiler.ghc$(echo ${{ matrix.ghc }} | sed -e s,\\.,,g)
-          echo GHC=$GHC >> $GITHUB_ENV
+          echo NS="nix shell github:nixos/nixpkgs/21.05#cabal-install nixpkgs#$GHC nixpkgs#gmp nixpkgs#zlib nixpkgs#zlib.dev" >> $GITHUB_ENV
```
RyanGlScott, line 104: Why is `nixpkgs` `21.05` in particular being used here? Intuitively, I would expect the `nixpkgs` version to match what is being installed in the `cachix/install-nix-actions` steps, which would be `20.09` or `unstable`. Again, though, I might be misunderstanding the relationship between these steps.

kquick: I had to do a bit of experimentation to understand what was happening (thus the different settings you noted in the previous comment), but I believe I've got a handle on it now, so I should be able to do a couple of updates to get everything aligned now.

Basically, the nix tools are evolving from their version 1 forms (`nix-shell`, `nix-build`, `nix-...`) into their version 2 forms (`nix shell`, `nix build`, `nix ...`) which have generally better, but somewhat changed functionality.  In the version 1 form, the installed nixos version would generally dictate which version the packages are being obtained from, so that would have (and used to be) sufficient for determining package sources.  In version 2, the "nixpkgs:" on the command-line causes it to consult the flake registry (`nix registry list`) instead of the `NIX_PATH` set by the installed version (where I was initially assuming the latter).  Thus, the installed version is actually less relevant than the nix shell command-line setting.

It's worth noting (just FYI) that this awkwardness is because we are using a one-line on-the-spot environment generation via `nix shell`, whereas normally the version 2 `nix` commands would look for a flake specification `flake.nix` and a corresponding `flake.lock` file (which form a similar pair to a `.cabal`/`cabal.freeze` file pair) to identify the specific versions of packages desired.

kquick: OK @RyanGlScott , I'm all done and CI should be good now.

RyanGlScott: Fantastic stuff, and I'm sure I'll come to back to study this later. Thanks again!

uc-crux-llvm/exe/Main.hs
```diff
@@ -1,3 +1,7 @@
 module Main (main) where
 
-import RealMain (main)
+-- TODO RGS: Ugh. See https://gitlab.haskell.org/ghc/ghc/-/issues/19397.
```
RyanGlScott, line 3: Oops, I shouldn't need this workaround anymore. Fix incoming.

PR #950, opened 2022-01-10
Drop support for GHC 8.6

RyanGlScott:
Now that `crucible` supports GHC 9.0, it's about time to drop support for 8.6, as we generally only support the previous three stable releases of GHC (in this case, that would be 8.8, 8.10, and 9.0). As a result, lots of 8.6–specific CPP can be removed. I have updated the version bounds of affected packages accordingly.


uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
```diff
@@ -149,17 +149,12 @@ symCreateOverrideFn ::
...
 data UCCruxSimulationResult m arch (argTypes :: Ctx (FullType m)) = UCCruxSimulationResult
   { unsoundness :: Unsoundness,
     explanations :: [Located (Explanation m arch argTypes)]
   }
 
 -- | Based on 'Crux.SimulatorHooks'
---
--- NOTE(lb): The explicit kind signature here is necessary for GHC 8.6
--- compatibility.
 data SimulatorHooks sym m arch (argTypes :: Ctx (FullType m)) r =
```
langston-barrett, line 158: Might as well remove the kind signature if it's no longer needed. (Same comment applies to other places where NOTEs are removed)

uc-crux-llvm/src/UCCrux/LLVM/Run/Result.hs
```diff
@@ -7,7 +7,6 @@ Maintainer   : Langston Barrett <langston@galois.com>
 Stability    : provisional
 -}
 {-# LANGUAGE DataKinds #-}
--- Needed for GHC 8.6
```
langston-barrett, line 10: Maybe remove the language extension as well?

uc-crux-llvm/src/UCCrux/LLVM/Run/Simulate.hs
RyanGlScott: Sounds good to me. I didn't want to apply this change unilaterally since it comes down to a matter of preference, but I'll happily do so now that you've made your preference clear :)

langston-barrett: I appreciate the consideration; I think we should either kind signature everything or nothing for the sake of consistency, and the majority of things lack kind signatures at the moment.

robdockins (approved):

PR #951, opened 2022-01-11
Simple loop fixpoint

andreistefanescu:


crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+    EQF -> compareF (bodyValue x) (bodyValue y)
+    GTF -> GTF
+
+instance OrdF (FixpointEntry sym) => W4.TestEquality (FixpointEntry sym) where
+  testEquality x y = case compareF x y of
+    EQF -> Just Refl
+    _ -> Nothing
+
+instance W4.TestEquality (FixpointEntry sym) => C.EqF (FixpointEntry sym) where
+  eqF x y = isJust $ W4.testEquality x y
```
travitch, line 99: These are slightly scary instances. I suppose the assertion here is that tests for identity are okay, since these definitely don't check for symbolic equality. I think that is worth recording somewhere...

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+  , memFixpointEntryJoinVariable :: W4.SymBV sym w
+  }
+
+instance W4.TestEquality (W4.SymExpr sym) => Eq (MemFixpointEntry sym) where
+  (MemFixpointEntry _ x_join_variable) == (MemFixpointEntry _ y_join_variable)
+    | Just Refl <- W4.testEquality x_join_variable y_join_variable =
+      True
+    | otherwise = False
+
+data FixpointStuff sym blocks = forall args . FixpointStuff
```
travitch, line 112: Can we use a more descriptive name than `Stuff`?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+joinRegEntry ::
+  (?logMessage :: String -> IO (), C.IsSymInterface sym) =>
+  sym ->
+  C.RegEntry sym tp ->
+  C.RegEntry sym tp ->
+  FixpointMonad sym (C.RegEntry sym tp)
+joinRegEntry sym left right = case C.regType left of
+  C.LLVMPointerRepr w
+    | List.isPrefixOf "cmacaw_reg" (show $ W4.printSymNat $ C.llvmPointerBlock (C.regValue left))
+    , List.isPrefixOf "cmacaw_reg" (show $ W4.printSymExpr $ C.llvmPointerOffset (C.regValue left)) -> do
```
travitch, line 143: This looks a bit fragile and scary. Also quite macaw-specific... is there some more structured way we could accomplish this that is resilient to breaking?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+          liftIO $ ?logMessage "SimpleLoopFixpoint.joinRegEntry: LLVMPointerRepr: Nothing"
+          join_varaible <- liftIO $ W4.freshConstant sym (userSymbol' "reg_join_var") (W4.BaseBVRepr w)
+          let join_entry = FixpointEntry
+                { headerValue = C.llvmPointerOffset (C.regValue left)
+                , bodyValue = C.llvmPointerOffset (C.regValue right)
+                }
+          put $ MapF.insert join_varaible join_entry subst
+          return $ C.RegEntry (C.LLVMPointerRepr w) $ C.LLVMPointer (C.llvmPointerBlock (C.regValue left)) join_varaible
+    | otherwise ->
+      fail $
```
travitch, line 175: Is this a `panic` situation?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+        "SimpleLoopFixpoint.joinRegEntry: LLVMPointerRepr: unsupported pointer base join: "
+        ++ show (C.ppPtr $ C.regValue left)
+        ++ " \\/ "
+        ++ show (C.ppPtr $ C.regValue right)
+  C.BoolRepr
+    | List.isPrefixOf "cmacaw" (show $ W4.printSymExpr $ C.regValue left) -> do
```
travitch, line 181: Still scared

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+      join_varaible <- liftIO $ W4.freshConstant sym (userSymbol' "macaw_reg") W4.BaseBoolRepr
+      return $ C.RegEntry C.BoolRepr join_varaible
+  C.StructRepr field_types -> do
+    liftIO $ ?logMessage "SimpleLoopFixpoint.joinRegEntry: StructRepr"
+    C.RegEntry (C.regType left) <$> fmapFC (C.RV . C.regValue) <$> joinRegEntries sym
+      (Ctx.generate (Ctx.size field_types) $ \i ->
+        C.RegEntry (field_types Ctx.! i) $ C.unRV $ (C.regValue left) Ctx.! i)
+      (Ctx.generate (Ctx.size field_types) $ \i ->
+        C.RegEntry (field_types Ctx.! i) $ C.unRV $ (C.regValue right) Ctx.! i)
+  _ -> fail $ "SimpleLoopFixpoint.joinRegEntry: unsupported type: " ++ show (C.regType left)
```
travitch, line 199: Panic, or recoverable?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+      normal_substitution = MapF.filterWithKey
+        (\variable _entry ->
+          Just Refl == W4.testEquality variable (fromJust $ MapF.lookup variable uninterpreted_constant_substitution))
+        substitution
+  in
+  (normal_substitution, uninterpreted_constant_substitution)
+
+
+userSymbol' :: String -> W4.SolverSymbol
+userSymbol' = fromRight (C.panic "SimpleLoopFixpoint.userSymbol'" []) . W4.userSymbol
```
travitch, line 388: Would it be preferrable to use the `safeSymbol` function from what4?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+
+  fixpoint_stuff_ref <- newIORef @(FixpointStuff sym blocks) $ FixpointStuff
+    { fixpointStatus = BeforeFixpoint
+    , fixpointBlockId = undefined
+    , fixpointAssumptionFrameIdentifier = undefined
+    , fixpointSubstitution = undefined
+    , fixpointRegMap = undefined
+    , fixpointMemSubstitution = undefined
+    , fixpointLoopCondition = undefined
+    , fixpointLoopIndexBound = undefined
```
travitch, line 422: We might prefer `error` or `panic` here... at least to let us know where these came from if anyone ever refactors this and introduces a mistake that uses these before they get defined... or are they used? I'm assuming they are filled in eventually

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+        | C.SomeHandle cfgHandle == C.frameHandle (sim_state ^. C.stateCrucibleFrame)
+        , Just Refl <- W4.testEquality
+            (fmapFC C.blockInputs cfgBlockMap)
+            (fmapFC C.blockInputs $ C.frameBlockMap $ sim_state ^. C.stateCrucibleFrame)
+        , Map.member (C.Some block_id) loop_map -> case fixpointStatus fixpoint_stuff of
+
+          BeforeFixpoint -> do
+            ?logMessage $ "SimpleLoopFixpoint: RunningState: " ++ show BeforeFixpoint ++ " -> " ++ show ComputeFixpoint
+            assumption_frame_identifier <- C.pushAssumptionFrame sym
+            let mem_impl = fromJust $ C.lookupGlobal mem_var (sim_state ^. C.stateGlobals)
```
travitch, line 442: Can we make this a panic if the variable is missing? At a minimum, a `HasCallStack` constraint would make this easier to track down if messed up

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+        substitution
+  in
+  (normal_substitution, uninterpreted_constant_substitution)
+
+
+userSymbol' :: String -> W4.SolverSymbol
+userSymbol' = fromRight (C.panic "SimpleLoopFixpoint.userSymbol'" []) . W4.userSymbol
+
+
+simpleLoopFixpoint ::
```
travitch, line 391: It would be really useful to have a description of the intent of this code and an indication as to how a user should use it.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
...
+import qualified Lang.Crucible.LLVM.MemModel.Type as C
+
+
+data FixpointStatus = BeforeFixpoint | ComputeFixpoint | CheckFixpoint | AfterFixpoint
+  deriving (Eq, Ord, Show)
+
+data FixpointEntry sym tp = FixpointEntry
+  { headerValue :: W4.SymExpr sym tp
+  , bodyValue :: W4.SymExpr sym tp
+  }
```
travitch, line 85: If this is part of the interface, we should document what these are

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
andreistefanescu: yes!

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
andreistefanescu: I used `panic` for situations that shouldn't happen, and `fail` for situations that can happen but are not supported. It's possible to handle pointers with symbolic base, it's just out of scope.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
andreistefanescu: recoverable with more work

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
andreistefanescu: I agree, this is a convenient way of checking that a register is locally used inside the loop, but doesn't need to be part of the fixpoint. It can be improved later

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
robdockins: I have been trying to migrate to throwing an `Unsupported` sim error in these situations.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1583,6 +1584,35 @@ buildDisjointRegionsAssertion sym w dest dlen src slen = do
 
   orPred sym diffBlk =<< orPred sym destfirst srcfirst
 
+-- | Build the condition that two memory regions are disjoint, using
+-- subtraction and comparison to zero instead of direct comparison (that is,
+-- 0 <= y - x instead of x <= y). This enables semiring and abstract domain
+-- simplifications. The result if false if any offset is not positive when
+-- interpreted as signed bitvector.
+buildDisjointRegionsAssertionWithSub ::
```
robdockins, line 1592: Is there a reason to preserve both versions of this? Is the condition being check semantically the same?

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/MemLog.hs
```diff
@@ -511,6 +519,92 @@ ppMemState f (BranchFrame _ _ d ms) =
...
+  EmptyMem _a _w chs -> writeRangesMemChanges sym chs
+  StackFrame _a _w _nm chs st ->
+    multiUnion <$> writeRangesMemChanges sym chs <*> writeRangesMemState sym st
+  BranchFrame _a _w chs st ->
+    multiUnion <$> writeRangesMemChanges sym chs <*> writeRangesMemState sym st
+
+-- | Compute the ranges (pairs of the form pointer offset and size) for all
+-- memory writes, indexed by the pointer base. The result is Nothing if the
+-- memory contains any writes with symbolic pointer base, or without size.
+writeRangesMem ::
```
robdockins, line 600: Can you say a little bit about why we want this? I'm thinking ahead to memory model refactoring, so  I'd like to understand what are the critical features of this operation.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
```diff
@@ -0,0 +1,748 @@
+------------------------------------------------------------------------
+-- |
+-- Module           : Lang.Crucible.LLVM.SimpleLoopFixpoint
+-- Description      : Execution feature to compute loop fixpoint
+-- Copyright        : (c) Galois, Inc 2021
+-- License          : BSD3
+-- Stability        : provisional
+------------------------------------------------------------------------
```
robdockins, line 8: I'd like to see some general documentation about what this module does and what its limitations are. I know we talked about this in person, but we should really preserve those thoughts for posterity.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
robdockins: @travitch, I don't understand what you think is scary about this? This is just reflecting the syntactic ordering on `SymExpr` into a lexicographic order on the pair (headerValue, bodyValue)

On the other hand, I think it's interesting to ask why we need these instances at all... do they end up as keys in a map or something?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
robdockins: Agreed, I'd like to know a lot more about what's going on this this type; it seems important.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
robdockins: I recently added a `throwUnsupported` combinator that makes it easier to do this.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
travitch: I see what they do... I just get nervous around instances like that since someone can see them or accidentally use them without the understanding that they are nonce-identity checks and not semantic checks. I'm just imagining potential frustrating future debugging sessions.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
robdockins: It looks like the `fixpointStatus` is being used to encode a state machine here.  I think it would be better to refactor the `FixpointStuff` type with different constructors for the cases so we don't need to have a bunch of fields that are `undefined` until we are ready to fill them in.

robdockins (changes requested):
There's a lot of machinery in `SimpleLoopFixpoint` that I don't know how to evaluate because I'm not actually sure what it is supposed to be doing.

I'll need some more high-level description of the intent and descriptions of the mechanisms here before I'll feel comfortable merging.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpoint.hs
travitch: Well, if those types aren't exported I guess I have no reason to complain... and the underlying types have those instances anyway. Complaint irrelevant.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
robdockins: This version is more restrictive because it requires all the involved numbers to be positive.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/MemLog.hs
robdockins: This computes an overapproximation of all the writes that can occur to some concrete blocks.

PR #952, opened 2022-01-11
Remove `crucible-server`

RyanGlScott:
`crucible-server` is no longer being maintained, and much of its reason for existence has been superseded by projects like `saw-remote-api`. Let's go ahead and remove `crucible-server` from version control, with the understanding that it will continue to exist in the `git` history should we need it again at some point in the future.


robdockins:
I think we can remove the `hpb` submodule as well.  Otherwise looks good.

PR #954, opened 2022-03-02
Export more utilities from `Lang.Crucible.LLVM.SymIO`

RyanGlScott:
This exports the following from `Lang.Crucible.LLVM.SymIO`, which prove useful in a `crucible-symio`–related project where I need to allocate and manage file descriptors through a different way than what the `callOpenFile` API offers:

* The fields of `LLVMFileSystem`
* `allocateFileDescriptor`
* `lookupFileHandle`


crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -61,6 +63,9 @@ module Lang.Crucible.LLVM.SymIO
   , callReadFileHandle
   , writeFileHandle
   , callWriteFileHandle
+  -- * File-related utilities
+  , allocateFileDescriptor
+  , lookupFileHandle
```
travitch, line 68: I recommend augmenting the comment on `lookupFileHandle` to note that its continuation is called multiple times (because it is used in a symbolic branch). This may be surprising to callers if their continuation has side effects. I didn't note it before because users couldn't be surprised by an internal function.

travitch (approved):

PR #962, opened 2022-04-21
Pare down the number of CI configurations

RyanGlScott:
Currently, we have a whopping 36 (!) different number of CI configurations. This is mainly because we test every supported version of GHC across both Linux and macOS, and in the case of `crux-llvm-build`, we even test multiple versions of Linux. I believe this is a bit overkill, so this patch attempts to trim down the number of configurations to something more manageable:

* For each project, we test the latest LTS release of Ubuntu on all supported versions of GHC.
* In addition, we also test the previous LTS release of Ubuntu on the `crux-llvm-build.yml` workflow, but only using the latest supported version of GHC.
* For each project, we test macOS only using the latest supported version of GHC. This is because the turnaround time for macOS builders on GitHub Actions is longer, so we try to avoid bottlenecking CI workflows on macOS builds.

  In the future, we wish to test Windows using the latest supported version of GHC as well.

I've written this up in `.github/workflows/README.md`.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -1,0 +1,18 @@
+# GitHub Actions workflows
+
+We currently build and test a subset of the packages in the `crucible` repo
+on CI, whose CI configurations are located here. We use the following
+conventions when picking with operating systems to test:
+
+* For each project, we test the latest LTS release of Ubuntu on all supported
+  versions of GHC.
+* In addition, we also test the previous LTS release of Ubuntu on the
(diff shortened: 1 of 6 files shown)


kquick (approved):
Looks good.  My only observation is that if we used a script-generated JSON for the matrix (as we do for what4) that we could eliminate duplication, but that 's probably best left for a future change.

RyanGlScott:
Yes, good observation. I've opened #964 to track this idea.

PR #965, opened 2022-04-22
Remove unused submodules and `optional-packages`

RyanGlScott:
Although the `cabal.project` file declares dependencies on `abcBridge`, `blt`, `what4-abc`, and `what4-blt`, none of the projects in `crucible` actually depend on them in practice. This patch removes those dependencies and the associated submodules. This means fewer packages to build with running `cabal build all`, which is all around a good thing.


RyanGlScott:
I think the dependencies on `abcBridge`, `blt`, and friends are a vestige of when `what4-{abc,blt}` used to live in this repo. @robdockins, do sanity-check me to make sure I haven't overlooked something, but the fact that CI passes is an encouraging sign that we can rip this out.

robdockins (approved):

PR #968, opened 2022-04-26
Support building with GHC 9.2

RyanGlScott:
This contains a variety of changes needed to make the libraries in the `crucible` compile with GHC 9.2:

* In `base-4.16.*`, the internals of the `(<=)` type family in `GHC.TypeNats` have changed (see
  [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2?version_id=7e2ce63ba042c1934654c4316dc02028d8d3dd31#changes-to-the-and-type-families)). As a result, GHC is no longer able to infer `0 <= n` for all `n :: Nat`, as this fact must now be assumed as an axiom. This prevents `Crucible.LLVM.Translation.Instruction` from typechecking, so I resolved the issue by making use of the new `leqZero` function from `parameterized-utils`, introduced in GaloisInc/parameterized-utils#128.
* GHC 9.2's pattern-match coverage checker is smarter when it comes to GADT patterns (see [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2?version_id=ad6254b755131de4401bcc73afcb2e6e7542015a#improved-pattern-match-coverage-checker)). As a result, GHC 9.2 uncovers `-Wincomplete-patterns` warnings in `uc-crux-llvm` that GHC 9.0 did not. I resolved the warnings by rearranging the order of patterns, which is one of the mitigations suggested in the GHC 9.2 Migration Guide. A happy consequence of this is that we no longer need to awkwardly CPP out a catch-all wildcard case on GHC 9.0+, as the code no longer produces any warnings on any tested versions of GHC.
* GHC 9.2 now defaults visible, dependent kind arguments to `Type` when `NoPolyKinds` is enabled (see [here](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.2?version_id=b053f50246a371491dd37363516586305ffdabeb#nopolykinds-defaults-visible-dependent-arguments-to-type)). This affects the `Clobber` module in `uc-crux-llvm`'s test suite, so I have resolved the issue by simply enabling `PolyKinds`.
* Due to https://gitlab.haskell.org/ghc/ghc/-/issues/21038, an application of a Template Haskell splice (by way of `($)`) to an argument with a higher-rank type no longer typechecks. Unfortunately, it does not appear that this issue will be fixed in time for a future minor release of GHC 9.2. I have worked around the issue by removing the use of `($)` and surrounding the arguments with parentheses instead.
* The upper version bounds on `base` were raised to allow libraries to build with `base-4.16.*`.
* Various submodules were bumped to allow them to build with GHC 9.2:
  * `abcBridge`: GaloisInc/abcBridge#19
  * `aig`: GaloisInc/aig#12
  * `golang`: GaloisInc/golang#8
  * `jvm-parser`: GaloisInc/jvm-parser#10
  * `llvm-pretty`: elliottt/llvm-pretty#92
  * `llvm-pretty-bc-parser`: GaloisInc/llvm-pretty-bc-parser#181
  * `what4`: GaloisInc/what4#197

In addition, this fixes several warnings that are now a part of `-Wall` in GHC 9.2, including `-Wincomplete-uni-patterns`, `-Wnoncanonical-monad-instances`, and `-Wnoncanonical-monoid-instances`.


langston-barrett (approved):
UC-Crux parts LGTM. Thanks, @RyanGlScott!

robdockins (approved):
I don't see anything controversial here. 👍

PR #972, opened 2022-04-29
crucible-llvm: Handle callbr instruction

langston-barrett:


crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -185,8 +186,11 @@ updateUseSet lab bi bim = if newuse == block_use_set bi then Nothing else Just b
 
   loop [] = mempty
 
+  -- invoke and callbr are a special cases when their return values are assigned
```
RyanGlScott, line 189:
```suggestion
  -- invoke and callbr are a special case when their return values are assigned
```

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -196,6 +200,14 @@ updateUseSet lab bi bim = if newuse == block_use_set bi then Nothing else Just b
             -- invoke is a block terminator, we can ignore the tail of the list
```
RyanGlScott, line 200: Make sure to include this comment in the case for `CallBr` as well.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
```diff
@@ -212,6 +224,10 @@ instrUse from i bim = Set.unions $ case i of
   L.Bit _op x y -> [useTypedVal x, useVal y ]
   L.Conv _op x _tp -> [useTypedVal x]
   L.Call _tailCall _tp f args -> useVal f : map useTypedVal args
+  L.CallBr _tp f args norm ls ->
+    [useVal f, useLabel from norm bim] ++
+      map (\l -> useLabel from l bim) ls ++
+      map useTypedVal args
```
RyanGlScott, line 230: The case for `Invoke` has this comment:

```
-- NB, this is only correct for "invoke" instructions that don't assign the return value
```

I think we'd want a similar comment for `CallBr` as well, given that the indirect labels don't make use of the returned value.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
```diff
@@ -1601,6 +1603,12 @@ generateInstr retType lab defSet instr assign_f k =
       do callFunction defSet instr False fnTy fn args assign_f
          definePhiBlock lab normLabel
 
+    L.CallBr fnTy fn args normLabel otherLabels -> do
+      do callFunction defSet instr False fnTy fn args assign_f
+         _ <- definePhiBlock lab normLabel
+         for_ otherLabels $ \lab' -> void (definePhiBlock lab lab')
+         k
```
RyanGlScott, line 1610: Given that `CallBr` is a block-terminating instruction, I don't think we should invoke `k` here. `Invoke`'s case doesn't invoke `k` either, which makes me especially suspicious.

langston-barrett: Good catch, thanks!

crucible-llvm/src/Lang/Crucible/LLVM/Translation/BlockInfo.hs
langston-barrett: Gotcha, to be frank I didn't quite understand what this comment left, but I'll apply it here as well.

crux-llvm/test-data/golden/T972-fail.at-least-clang12.good
```diff
@@ -0,0 +1,6 @@
+[Crux] Found counterexample for verification goal
+[Crux]   test-data/golden/T972-fail.at-least-clang12.c:8:3: error: in main
+[Crux]   [callFunction] Unsupported function type: PtrTo (FunTy (PrimType Void) [PrimType (Integer 32),PtrTo (PrimType (Integer 8)),PtrTo (PrimType (Integer 8))] False)
```
RyanGlScott, line 3: I'm a bit surprised that it fails with this error message, since this function type doesn't look that strange. (I'd expect it to eventually fail on the `ValAsm` bit, but it doesn't reach that part.) Perhaps the `CallBr` case in `generateInstr` should match on `PtrTo` like the `Call` case does?

The `Invoke` case should do this as well, now that I look closer.

crux-llvm/test-data/golden/T972-fail.at-least-clang12.c
```diff
@@ -0,0 +1,18 @@
+// Adapted from the Clang test suite:
```
RyanGlScott, line 1: Is there a reason not to use an `.at-least-clang9` suffix here? That way, it would be tested with all supported versions of Clang. (I'm hoping that the error message would be the same in each version.)

langston-barrett: I assumed that this was the set of all possible suffixes, is that not true?

https://github.com/GaloisInc/crucible/blob/b558ba2341b11c88e668ed3f772f4d1960275229/crux-llvm/test/Test.hs#L40

RyanGlScott: Ah, I was under the mistaken impression that we were doing something more thorough than that. In that case, disregard my comment above. We might want to make these suffixes more flexible in the future, but it's not terribly important for this patch.

crux-llvm/test-data/golden/T972-fail.at-least-clang12.good
langston-barrett: Good call, fixed!

langston-barrett: Huh, that didn't work for `invoke`:
```
        Exception: user error (Failure encountered while generating a Crucible CFG: at internal: unexpected function type in invoke: FunTy (PtrTo (Alias (Ident "class.std::__cxx11::basic_string"))) [PtrTo (Alias (Ident "class.std::__cxx11::basic_string")),PrimType (Integer 64),PrimType (Integer 64),PtrTo (PrimType (Integer 8)),PrimType (Integer 64)] False)
```
Nonetheless, I've fixed it for `callbr`.

RyanGlScott: Wow, `invoke` gets more mysterious every time I look at it. But in any case, I'm much happier with the way the error message for the `callbr` test case looks now.

RyanGlScott (approved):

crux-llvm/test-data/golden/T972-fail.at-least-clang12.c
kquick: The `tasty-sugar` matching process attempts to match the parameter value if present; if it's not present it will match against the non-parameterized version of the filename.  

In this particular case, if you expect this to compile and build with all versions of clang, simply remove any `clang-range` value from the filename.

For more info, see https://hackage.haskell.org/package/tasty-sugar-1.1.1.0/docs/Test-Tasty-Sugar.html#t:CUBE and https://github.com/kquick/tasty-sugar/tree/master/examples/params

RyanGlScott: The thing is, this shouldn't build with all versions of `clang`, but rather only with Clang 9 and up. If I understand how `tasty-sugar` works correctly, that would require adding a new entry to `"clang-range"` for every version of Clang that we wanted to conditionally test. Certainly doable, although it does make me wonder if there's a way we can have only two `"pre-clang*"` and `at-least-clang*"` wildcard patterns that take care of this legwork once and for all.

kquick: Well, we'd only need to add entries for inflection points we cared about.

Alternatively, we could define the `"clang-range"` parameter as a `Nothing`, which means "accept any value".  Then we could decide on some syntax we wanted to allow the tests to make the right determination.  Something like "preN", "postN", "N", or no field.  When defined as `Nothing`, tasty-sugar can't do any validation of the entries, and you can only treat one of your parameters as a value wildcard like this, but it's certainly doable.

crux-llvm/test-data/golden/T972-fail.at-least-clang12.good
RyanGlScott: Thinking about this some more, I think it's a bug that `invoke`'s function type isn't a `PtrTo` here. I've opened GaloisInc/llvm-pretty-bc-parser#189 for this.

PR #973, opened 2022-04-29
crucible-llvm: Handle empty structs

### LINKED ISSUE #970: crucible-llvm: Out-of-bounds vector access on program with empty structs
```c
typedef struct empty {} empty;
typedef struct empties {
  empty empties[8];
} empties;
const empties glob = {};

int main(void) {
  return 0;
}
```
```
$ cabal v2-run --enable-profiling --enable-library-profiling --enable-executable-profiling "--ghc-options=-fprof-auto-top -threaded -rtsopts" exe:crux-llvm -- empty.c

Up to date
./Data/Vector/Generic.hs:257 ((!)): index out of bounds (-1,0)
  error, called at ./Data/Vector/Internal/Check.hs:87:5 in vector-0.12.3.1-25c9288d0bc754c0b17e5d19103f02bc42e07ac50ae2f16fa925caa19df4be76:Data.Vector.Internal.Check
CallStack (from -prof):
  Data.Vector.Internal.Check.error (Data/Vector/Internal/Check.hs:(86,1)-(87,41))
  Data.Vector.Internal.Check.checkError (Data/Vector/Internal/Check.hs:(100,1)-(103,34))
  Lang.Crucible.LLVM.MemModel.Type.structType (src/Lang/Crucible/LLVM/MemModel/Type.hs:(118,1)-(119,52))
  Lang.Crucible.LLVM.MemModel.Type.mkStructType (src/Lang/Crucible/LLVM/MemModel/Type.hs:(122,1)-(130,23))
  Data.Vector.traverse (Data/Vector.hs:(489,3)-(494,84))
  Lang.Crucible.LLVM.MemModel.toStorableType (src/Lang/Crucible/LLVM/MemModel.hs:(1200,1)-(1214,39))
  Lang.Crucible.LLVM.Globals.populateGlobal (src/Lang/Crucible/LLVM/Globals.hs:(362,1)-(402,48))
  Lang.Crucible.LLVM.Globals.populateGlobals (src/Lang/Crucible/LLVM/Globals.hs:(269,1)-(274,76))
  Lang.Crucible.LLVM.Globals.populateAllGlobals (src/Lang/Crucible/LLVM/Globals.hs:289:1-49)
  Lang.Crucible.LLVM.Types.withPtrWidth (src/Lang/Crucible/LLVM/Types.hs:(75,1)-(77,38))
  Lang.Crucible.LLVM.Translation.Monad.mkLLVMContext (src/Lang/Crucible/LLVM/Translation/Monad.hs:(104,1)-(127,95))
  Lang.Crucible.LLVM.Translation.translateModule (src/Lang/Crucible/LLVM/Translation.hs:(431,1)-(448,24))
  Lang.Crucible.LLVM.Translation.Monad.llvmPtrWidth (src/Lang/Crucible/LLVM/Translation/Monad.hs:87:6-17)
  Crux.LLVM.Simulate.prepLLVMModule (src/Crux/LLVM/Simulate.hs:(242,1)-(256,56))
  Crux.LLVM.Simulate.setupFileSim (src/Crux/LLVM/Simulate.hs:(187,1)-(219,49))
  Crux.LLVM.Simulate.simulateLLVMFile (src/Crux/LLVM/Simulate.hs:(159,1)-(174,12))
  Crux.getSimulatorCallbacks (src/Crux.hs:121:7-27)
  Crux.inFrame (src/Crux.hs:420:5-11)
  Crux.doSimWithResults (src/Crux.hs:(649,1)-(704,78))
  Lang.Crucible.Backend.Online.withOnlineBackend (src/Lang/Crucible/Backend/Online.hs:(206,1)-(216,5))
  Lang.Crucible.Backend.Online.withYicesOnlineBackend (src/Lang/Crucible/Backend/Online.hs:(239,1)-(243,17))
  Crux.withSelectedOnlineBackend' (src/Crux.hs:(389,1)-(417,33))
  Crux.withSelectedOnlineBackend (src/Crux.hs:(355,1)-(375,65))
  Crux.runSimulator (src/Crux.hs:(572,1)-(622,73))
  CruxLLVMMain.mainWithOptions (src/CruxLLVMMain.hs:(81,1)-(87,47))
  Crux.loadOptions (src/Crux.hs:(176,1)-(198,48))
  CruxLLVMMain.withCruxLLVMLogging (src/CruxLLVMMain.hs:(63,1)-(67,6))
  CruxLLVMMain.mainWithOutputConfig (src/CruxLLVMMain.hs:(72,1)-(74,67))
  RealMain.main (exe/unix/RealMain.hs:(22,1)-(25,16))[Crux] Using pointer width: 64 for file crux-build/crux~empty.bc
```
```llvm
; ModuleID = 'empty.c'
source_filename = "empty.c"
target datalayout = "e-m:e-p270:32:32-p271:32:32-p272:64:64-i64:64-f80:128-n8:16:32:64-S128"
target triple = "x86_64-unknown-linux-gnu"

%struct.empties = type { [8 x %struct.empty] }
%struct.empty = type {}

@glob = constant %struct.empties zeroinitializer, align 1

; Function Attrs: noinline nounwind optnone sspstrong uwtable
define i32 @main() #0 {
  %1 = alloca i32, align 4
  store i32 0, i32* %1, align 4
  ret i32 0
}

attributes #0 = { noinline nounwind optnone sspstrong uwtable "correctly-rounded-divide-sqrt-fp-math"="false" "disable-tail-calls"="false" "frame-pointer"="all" "less-precise-fpmad"="false" "min-legal-vector-width"="0" "no-infs-fp-math"="false" "no-jump-tables"="false" "no-nans-fp-math"="false" "no-signed-zeros-fp-math"="false" "no-trapping-math"="true" "stack-protector-buffer-size"="4" "target-cpu"="x86-64" "target-features"="+cx8,+fxsr,+mmx,+sse,+sse2,+x87" "unsafe-fp-math"="false" "use-soft-float"="false" }

!llvm.module.flags = !{!0, !1}
!llvm.ident = !{!2}

!0 = !{i32 1, !"wchar_size", i32 4}
!1 = !{i32 7, !"PIC Level", i32 2}
!2 = !{!"clang version 11.1.0"}
```
The relevent code from the stack trace:

- https://github.com/GaloisInc/crucible/blob/b558ba2341b11c88e668ed3f772f4d1960275229/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs#L119
- https://github.com/GaloisInc/crucible/blob/b558ba2341b11c88e668ed3f772f4d1960275229/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs#L122
- https://github.com/GaloisInc/crucible/blob/b558ba2341b11c88e668ed3f772f4d1960275229/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1200

langston-barrett:
Fixes #970

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
@@ -35,7 +35,6 @@ module Lang.Crucible.LLVM.MemModel.Type
   , ppType
   )  where
 
-import Control.Exception (assert)
 import Control.Lens
 import Control.Monad.State
 import Data.Typeable
@@ -97,7 +96,7 @@ mkStorageType tf = StorageType tf $
     Double -> 8
     X86_FP80 -> 10
     Array n e -> natBytesMul n (storageTypeSize e)
-    Struct flds -> assert (V.length flds > 0) (fieldEnd (V.last flds))
+    Struct flds -> structSize flds
 
 bitvectorType :: Bytes -> StorageType
 bitvectorType w = StorageType (Bitvector w) w
@@ -115,8 +114,7 @@ arrayType :: Natural -> StorageType -> StorageType
 arrayType n e = StorageType (Array n e) (natBytesMul n (storageTypeSize e))
 
 structType :: V.Vector (Field StorageType) -> StorageType
-structType flds = assert (V.length flds > 0) $
-  StorageType (Struct flds) (fieldEnd (V.last flds))
+structType flds = StorageType (Struct flds) (structSize flds)
 
 mkStructType :: V.Vector (StorageType, Bytes) -> StorageType
 mkStructType l = structType (evalState (traverse fldFn l) 0)
@@ -139,8 +137,17 @@ typeEnd a tp = seq a $
     X86_FP80 -> a + 10
     Array 0 _   -> a
     Array n etp -> typeEnd (a + natBytesMul (n-1) (storageTypeSize etp)) etp
-    Struct flds -> typeEnd (a + fieldOffset f) (f^.fieldVal)
-      where f = V.last flds
+    Struct flds ->
+      case V.unsnoc flds of
+        Just (_, f) -> typeEnd (a + fieldOffset f) (f^.fieldVal)
+        Nothing -> a
+
+-- | Returns end of field including padding bytes.
+structSize :: V.Vector (Field StorageType) -> Bytes
+structSize flds =
+  case V.unsnoc flds of
+    Just (_, f) -> fieldEnd f
+    Nothing -> 0
 
 -- | Returns end of field including padding bytes.
 fieldEnd :: Field StorageType -> Bytes
--- a/crux-llvm/test-data/golden/T970.c
+++ b/crux-llvm/test-data/golden/T970.c
@@ -1,0 +1,10 @@
+typedef struct empty {
+} empty;
+
+typedef struct empties {
+  empty empties[8];
+} empties;
+
+const empties glob = {};
+
+int main(void) { return 0; }
--- a/crux-llvm/test-data/golden/T970.good
+++ b/crux-llvm/test-data/golden/T970.good
@@ -1,0 +1,1 @@
+[Crux] Overall status: Valid.


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
```diff
@@ -139,8 +137,19 @@ typeEnd a tp = seq a $
     X86_FP80 -> a + 10
     Array 0 _   -> a
     Array n etp -> typeEnd (a + natBytesMul (n-1) (storageTypeSize etp)) etp
-    Struct flds -> typeEnd (a + fieldOffset f) (f^.fieldVal)
-      where f = V.last flds
```
RyanGlScott, line 143: Hm, the `Array` case appears to assume that the array has length at least 1:

https://github.com/GaloisInc/crucible/blob/79428012dae4e3c922f75c6f41f418c8aacca665/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs#L141

This is admittedly a bit afield of the original scope of the patch, but if the `typeEnd` case for `Struct` is wrong, then it looks wrong for `Array` too.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
```diff
@@ -139,8 +137,19 @@ typeEnd a tp = seq a $
...
+    Struct flds ->
+      if V.length flds == 0
+      then 0
```
RyanGlScott, line 142: Shouldn't this be `a`?

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
```diff
@@ -139,8 +137,19 @@ typeEnd a tp = seq a $
...
+      else
+        let f = V.last flds
+        in typeEnd (a + fieldOffset f) (f^.fieldVal)
+
+-- | Returns end of field including padding bytes.
+structSize :: V.Vector (Field StorageType) -> Bytes
+structSize flds =
+  if V.length flds == 0
+  then 0
+  else fieldEnd (V.last flds)
```
RyanGlScott, line 152: I'm a bit leery about using partial functions like `last`. You could rewrite this function to be total using [`unsnoc`](https://hackage.haskell.org/package/vector-0.12.3.1/docs/Data-Vector.html#v:unsnoc), which should also be constant time.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
langston-barrett: Oof, yes!

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
langston-barrett: Nice, that's much better. Thanks!

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs
langston-barrett: Yeah, it seems like you should be able to make this case underflow using a zero-size array. I'm having a harder time triggering that one. I think `typeEnd` might only be called when you load or store something?

RyanGlScott: > I think `typeEnd` might only be called when you load or store something?

Indeed, and I'm not sure if it's actually possible to load or store a size-0 allocation in a way that invokes `typeEnd`. Nevertheless, I'd sleep better at night knowing that this potential footgun was disarmed. I think we could just add a special case for arrays of size 0 here, much like we now have special cases for empty structs.

langston-barrett: We both missed that there *is* special handling of 0, right above this:
https://github.com/GaloisInc/crucible/blob/a0ddd876d0d8afba913ccee46dbe6e563f759df3/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Type.hs#L140

RyanGlScott: Oops, quite right. Thanks for putting my mind at ease.

RyanGlScott (approved):

PR #976, opened 2022-05-05
uc-crux-llvm: Simplify addConstraint

langston-barrett:
Just a little refactoring to make things clearer.


uc-crux-llvm/src/UCCrux/LLVM/Constraints.hs
```diff
@@ -479,6 +474,16 @@ addConstraint modCtx argTypes constraints =
...
   where
+    fromMaybeL :: a -> Lens.Lens' (Maybe a) a
+    fromMaybeL def = lens (fromMaybe def) (\_old new -> Just new)
+
+    atDefault ::
+      Lens.At c =>
+      Lens.Index c ->
+      Lens.IxValue c ->
+      Lens.Lens' c (Lens.IxValue c)
+    atDefault idx def = at idx . fromMaybeL def
```
travitch, line 485: It isn't technically necessary, but it might be nice to introduce `a` and `c` with explicit `forall`s to make it clear that these are not captured variables (or float to the top level)

travitch (approved):

PR #983, opened 2022-05-12
Prepare for Crux 0.6 release

RyanGlScott:


uc-crux-llvm/CHANGELOG.md
```diff
@@ -1,4 +1,4 @@
-# next
+# 0.2
```
RyanGlScott, line 1: FYI: We're trying out a new convention where we make a note of breaking changes for libraries in their changelogs, even if the libraries themselves don't have conventional Hackage releases. This is mainly done in an effort to help pinpoint when (and what kind of) breaking changes occur for the benefit of downstream projects.

At the moment, the `uc-crux-llvm` changelog is a bit sparse. The only entry here is from a change I made myself a while back, but I'm sure we could be much more detailed than this if we really wanted to. I'm not sure how many projects are using `uc-crux-llvm` as a library dependency right now, so I'm not sure how much more we should write here. Let me know what you think, @langston-barrett.

langston-barrett: That makes sense to me! The only project using UC-Crux as a library that I'm aware of is V-SPELLS, which I'm working on and making all of these breaking changes for. So I don't necessarily feel a need to be too fastidious with this changelog, but  I'm happy to put in some effort on it if others would find it helpful.

RyanGlScott: That makes sense. Given that you're the only person who's using `uc-crux-llvm` in a downstream project at the moment, I'm not bothered by having a sparse `uc-crux-llvm` changelog for now, since you are fully aware of (and are able to adapt to) all breaking changes that come down the pipeline. I suppose the exception would be changes that others make to `uc-crux-llvm`, and since I've already made an effort to write those done, I think we can call this good enough.

crux-llvm/CHANGELOG.md
```diff
@@ -1,6 +1,27 @@
...
+## Bug fixes
+
+* Make `--help` and `--version` respect the `--no-colors` flag.
+
+## Library changes
+
 * `LLVMConfig` now has an `indeterminateLoadBehavior` flag to control the
   `MemOptions` option of the same name. Refer to the `crucible-llvm` changelog
-  for more details.
+-  for more details.
```
langston-barrett, line 25: Intentional?
```suggestion
  for more details.
```

langston-barrett (approved):

crux-llvm/CHANGELOG.md
RyanGlScott: Nope, definitely not. I think this is a leftover from a previous draft that I never fully reverted.

PR #985, opened 2022-05-13
uc-crux-llvm: Improve representation of LLVM types

langston-barrett:
For opaque pointers, replaces value-level representation of type names with value-and-type level representation with a `SymbolRepr`.

For structs, represents packed-ness at the type-level and remove `StructInfo` field. This hs three nice effects:

* Makes `FullTypeRepr` into a true singleton, making its `TestEquality` instances lawful.
* Reflects the fact that packed and non-packed structs are not the same at the type level.
* Reduces redundant storage, instead recomputing `StructInfo`s on demand.

Also introduces a wrapper around Crucible's `DataLayout` with an additional phantom type parameter that marks it as corresponding to a particular LLVM module.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Classify.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Classify.hs
@@ -57,8 +57,6 @@ import qualified What4.ProgramLoc as What4
 import qualified Lang.Crucible.Backend as Crucible
 import qualified Lang.Crucible.Simulator as Crucible
 
-import           Lang.Crucible.LLVM.Bytes (bytesToInteger)
-import           Lang.Crucible.LLVM.DataLayout (DataLayout)
 import qualified Lang.Crucible.LLVM.Errors as LLVMErrors
 import qualified Lang.Crucible.LLVM.Errors.MemoryError as MemError
 import qualified Lang.Crucible.LLVM.Errors.UndefinedBehavior as UB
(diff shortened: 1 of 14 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Classify.hs
```diff
@@ -123,14 +120,13 @@ classifyAssertion _sym predicate loc =
...
-  DataLayout ->
   ModuleTypes m ->
   What4.ConcreteVal (What4.BaseBVType w) ->
   PartTypeRepr m ft ->
   Int
-elemsFromOffset dl mts offset partType =
+elemsFromOffset mts offset partType =
   let pointedTo = asFullType mts partType
-      typeSize = bytesToInteger (memTypeSize dl (toMemType pointedTo))
+      typeSize = sizeInBytes mts pointedTo 1 -- 1 = array length
```
travitch, line 129:
```suggestion
      singletonArrayLength = 1
      typeSize = sizeInBytes mts pointedTo singletonArrayLength
```

uc-crux-llvm/src/UCCrux/LLVM/FullType/Type.hs
```diff
@@ -418,7 +472,9 @@ toFullTypeM memType =
               let ?lc = typeContext mts
               Some ftRepr <-
                 case asMemType' strIdent of
-                  Left opaqueIdent -> pure $ Some (FTOpaquePtrRepr opaqueIdent)
+                  Left (L.Ident opaqueIdent) ->
+                    case someSymbol (Text.pack opaqueIdent) of
+                      Some symRep -> pure $ Some (FTOpaquePtrRepr symRep)
```
travitch, line 477: Is the opaque pointer identifier meaningful here? Are they really distinct nominal types, or are there cases where opaque pointers get fudged?

I guess I'm trying to understand what the symbol is representing, what tradeoffs that implies, and if that is actually a safe/desirable thing to represent here.

uc-crux-llvm/src/UCCrux/LLVM/FullType/MemType.hs
```diff
@@ -24,44 +24,52 @@ import           Data.Parameterized.TraversableFC (toListFC)
...
+    FTUnboundedArrayRepr fullTypeRepr -> ArrayType 0 (toMemType dl fullTypeRepr)
+    FTStructRepr sp fields ->
+      let cdl = crucibleDataLayout dl
+          memFields = toListFC (toMemType dl) fields
+      in StructType (mkStructInfo cdl (structPackedReprToBool sp) memFields)
   where
-    funType :: Maybe (FullTypeRepr m ft) -> Ctx.Assignment (FullTypeRepr m) argTypes -> Bool -> MemType
-    funType maybeRetRepr argsRepr isVarArgs =
+    funType ::
+      DataLayout m ->
```
travitch, line 62: Is this argument necessary? It looks like this function is only called in two places that just pass the (already visible) `dl` to this function

uc-crux-llvm/src/UCCrux/LLVM/Classify.hs
langston-barrett: Thanks for the comment, I've refactored the module containing this helper to have array and non-array helpers, which also clearified the code in a few other places.

uc-crux-llvm/src/UCCrux/LLVM/FullType/MemType.hs
langston-barrett: It's a trade-off between using `ScopedTypeVariables` to match the `m` in this signature with the `m` in the outer signature vs just passing it in here. Happy to change it to used `ScopedTypedVariables` if you prefer it!

travitch: ah, I had assumed scoped type variables was already enabled. I think that using scoped type variables might be less confusing in the long run, as there is no possibility of confusing values.

travitch (approved):

PR #987, opened 2022-05-13
Update llvm-pretty, remove upstreamed debug info utilities

### LINKED ISSUE #696: Handling LLVM debug information
In both `uc-crux-llvm` and SAW (at least), we sometimes need to extract e.g. function argument names or struct field names from LLVM's debug info. Where should the code to do this live?

See discussion here: https://github.com/GaloisInc/crucible/pull/673#discussion_r599103951

langston-barrett:
Fixes #696.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
@@ -37,6 +37,8 @@ import           Control.Lens ((^.), Simple, Lens, lens)
 import           Data.Functor.Const (Const(Const, getConst))
 import qualified Data.Map as Map
 import           Data.Map (Map)
+import qualified Data.IntMap as IntMap
+import           Data.IntMap (IntMap)
 import           Data.Monoid (getFirst, First(First))
 import           Data.Text (Text)
 import qualified Data.Text as Text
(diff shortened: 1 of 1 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Context/Function.hs
```diff
@@ -205,21 +208,23 @@ tryMakeFunctionContext modCtx defnSymb argFullTypes argTypes =
...
                   (Ctx.size argFullTypes)
-                  (fmap (First . Just) (debugInfoArgNames (getModule llvmMod) def))
+                  (fmap
+                     (First . Just . Text.pack)
+                     (debugInfoArgNames (getModule llvmMod) def))
               )
         }
 
-mapToContext ::
+intMapToContext ::
```
kquick, line 219: Do the callers care if it's a Map or an IntMap?  If not, keeping the original generalized name helps immunize the callers against needing to know this detail.

langston-barrett: Presumably, in that they need to pass in differently-typed arguments. There's also only one caller, and it's just above (this function is not exported).

RyanGlScott (approved):

PR #993, opened 2022-05-19
uc-crux-llvm: Abstract type and operations for tracking symbolic value annotations

### LINKED ISSUE #991: uc-crux-llvm: Abstract type and operations for tracking symbolic value annotations
This map type gets used a lot in UC-Crux, we should probably hide it behind a `newtype` and expose appropriate operations over it.
```haskell
  IORef (Map (Some (What4.SymAnnotation sym)) (Some (TypedSelector m arch argTypes)))
```

langston-barrett:
Fixes #991. This map type was used ubiquitously throughout the codebase, but its concrete details aren't important/necessary for most clients. Instead, abstract it behind a newtype and only expose a few operations. Helpfully simplifies some type signatures, and makes their intent more clear.

The first commit is a small, only marginally related simplification. Might be helpful to review the commits separately.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/ExprTracker.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/ExprTracker.hs
@@ -1,0 +1,110 @@
+{-
+Module       : UCCrux.LLVM.ExprTracker
+Description  : Track sources of annotated symbolic expressions
+Copyright    : (c) Galois, Inc 2022
+License      : BSD3
+Maintainer   : Langston Barrett <langston@galois.com>
+Stability    : provisional
+-}
+
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE PolyKinds #-}
+
+module UCCrux.LLVM.ExprTracker
+  ( TypedSelector(..),
+    ExprTracker,
+    empty,
+    insert,
+    lookup,
+    union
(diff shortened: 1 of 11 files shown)


uc-crux-llvm/src/UCCrux/LLVM/ExprTracker.hs
```diff
@@ -0,0 +1,105 @@
...
+--
+-- For example, if the error is an OOB write and classification looks up the
+-- pointer in this map and it appears in an argument, it will attach a
+-- precondition that that part of the argument gets allocated with more
+-- space. If it looks up the pointer and it's not in this map, it's not clear
+-- how this error could be prevented, and the error will be unclassified.
+newtype ExprTracker m sym argTypes =
+  ExprTracker
+    { getExprTracker ::
+        Map (Some (What4.SymAnnotation sym)) (Some (TypedSelector m argTypes))
```
travitch, line 64: I just noticed the type erasure going on here with the key and value having their types erased separately. Can we document why that is important here (i.e., why is this not a `MapF`)?

travitch (approved):
Looks good aside from one RFC

uc-crux-llvm/src/UCCrux/LLVM/ExprTracker.hs
langston-barrett: Yeah good catch, I've added a comment.

Issue #982, opened 2022-05-12
uc-crux-llvm: User-provided specifications for external functions


langston-barrett:
UC-Crux has the ability to skip over execution of declared but not defined functions, e.g., from a third-party library or libc. This is practical, but it has its disadvantages:

- It's not overapproximate (sound): the execution of those functions might modify their pointer arguments or global variables. This means the safety claims made by UC-Crux after functions are skipped are much less strong.
- It's also not underapproximate, meaning that any bugs found by UC-Crux may not actually be feasible in practice.

For libc functions in particular, we could manually implement overrides (in Haskell) for some of them. Sound overrides could live in `crucible-llvm`, and unsound ones in UC-Crux. However, this takes considerable effort and still doesn't work for other third-party libraries.

Instead, we should implement deserialization of function postconditions (#981), probably from YAML, along with specifications of whether the provided postconditions are overapproximate, underapproximate, both, or neither (for the sake of #932 and tracking of unsoundness). This would mostly require making non-GADT versions of UC-Crux's data structures and working out how to convert to/from them.

langston-barrett:
After a bit more reflection I now think that the ideal input language for this feature would be more rich than what UC-Crux currently provides via the `Postcond` type. The ideal is probably closer to SAW's notion of a collection of overrides. For each external function, the user could provide *n* specifications, where a spec consists of:

- A precondition (the existing `Precond` type)
- An optional postcondition (i.e., a summary of the function's effect on the program state)
- A soundness tag, one of
  - Overapproximate
  - Underapproximate
  - Neither
  - Both

The precondition would be a constraint on the program state which specifies when the spec applies (e.g., `f` returns a positive number, but only if given a positive number as an input). Together with these *n* specs, the user could provide the intended behavior if none of their preconditions match, which might be either (1) an error (like SAW does if no overrides match) or (2) fallback to the current unsound function-skipping behavior.

For example, the following is a spec for a division function, stating that the output is symbolic and non-negative if the numerator is non-negative and denominator is positive or if both the numerator is non-positive and denominator is negative.

```yaml
- name: div
  specs:
  - # spec 1: non-negative / positive = non-negative 
    pre:
    - args:
      - # numerator
        signed-greater-or-equal: 0
      - # denominator
        signed-greater: 0
    post:
    - ret:
        symbolic:
          signed-greater-or-equal: 0
  - # spec 2: non-positive / negative = non-negative
    pre:
    - args:
      - # numerator
[...]

PR #994, opened 2022-05-20
uc-crux-llvm: Serialization of data structures

langston-barrett:
Introduce functions for (de)serializing the many type-indexed GADTs in UC-Crux. The reasoning behind the overall shape of the implementation is recorded in [the module comment on `UCCrux.LLVM.View`](https://github.com/GaloisInc/crucible/pull/994/files#diff-9b4d3143d7f87df806a4953d22116d4b59c93f4cc402cb0176a11c9b01a5cc72R1-R28). This will enable features like #982.

This is still a draft because error handling would need to be improved, many of the errors lack sufficient context to really make it easy to see what went wrong during deserialization. However, since this is already fairly large, I wanted to get some initial feedback on the direction before polishing it off.


uc-crux-llvm/src/UCCrux/LLVM/View/Cursor.hs
```diff
@@ -0,0 +1,127 @@
...
+          PP.viaShow idx,
+          "but vector was of length",
+          PP.viaShow len
+        ]
+
+data CursorView
+  = VHere
+  | VDereference Int CursorView
+  | VIndex Natural Natural CursorView
+  | VField Int CursorView
```
travitch, line 75: This set of views is prefixed with `V`, while some of the others are suffixed with `View`. Is there a logic to the name choices?

uc-crux-llvm/src/UCCrux/LLVM/View/Shape.hs
```diff
@@ -0,0 +1,197 @@
...
+  PtrShapeView vtag
+ptrShapeView tag =
+  \case
+    ShapeUnallocated -> VShapeUnallocated
+    ShapeAllocated n -> VShapeAllocated n
+    ShapeInitialized s -> VShapeInitialized (fmap (shapeView tag) s)
+
+viewPtrShape ::
+  ModuleTypes m ->
+  (forall t. FullTypeRepr m t -> vtag -> Either e (tag t)) ->
```
travitch, line 94: Most of the other arguments are pretty straightforward. I think this one is sufficiently complex that it would be worth documenting roughly what this means and what kinds of values a caller might want to pass

uc-crux-llvm/src/UCCrux/LLVM/View/Util.hs
```diff
@@ -0,0 +1,84 @@
...
+-- | The name of an LLVM type
+newtype TypeName
+  = TypeName { getTypeName :: Text }
+  deriving (Eq, Ord, Show, Generic)
+
+instance Aeson.ToJSONKey TypeName where
+  toJSONKey = Aeson.toJSONKeyText getTypeName
+
+instance Aeson.FromJSONKey TypeName where
+  fromJSONKey = Aeson.FromJSONKeyText TypeName
```
travitch, line 76: Presumably these custom-ish instances are important to get "nice"-looking JSON. If that is the case, I think it is worth a note somewhere around here to warn future us

uc-crux-llvm/src/UCCrux/LLVM/View/Constraint.hs
```diff
@@ -0,0 +1,244 @@
...
+import           UCCrux.LLVM.View.FullType (FullTypeReprView, FullTypeReprViewError, fullTypeReprView, viewFullTypeRepr, ppFullTypeReprViewError)
+import           UCCrux.LLVM.View.Shape (ShapeView, ViewShapeError, shapeView, viewShape, ppViewShapeError)
+import           UCCrux.LLVM.View.Util () -- Alignment, ICmpOp instance
+
+-- Helper, not exported
+liftError :: (e -> i) -> Either e a -> Either i a
+liftError l =
+  \case
+    Left e -> Left (l e)
+    Right v -> Right v
```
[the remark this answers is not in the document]
langston-barrett: Thanks, I've noted this in a comment!

uc-crux-llvm/src/UCCrux/LLVM/View/Cursor.hs
langston-barrett: Apparently I decided to use a prefix for sum types and a suffix for record types. Fixed to be consistent!

uc-crux-llvm/src/UCCrux/LLVM/View/Shape.hs
langston-barrett: Added a comment!

travitch (approved):

PR #997, opened 2022-05-26
`crucible-syntax`: Add forward declarations

### LINKED ISSUE #996: `crucible-syntax`: Allow declaring forward declarations for functions
`crucible-syntax`'s `(funcall ...)` atom has a major restriction: it can only be used to invoke functions whose `Handles` are provided ahead of time to the [`top`](https://github.com/GaloisInc/crucible/blob/1d2d95d232eb2aae076d68021e1d75d322968595/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs#L2086-L2088)-level parser. Attempting to invoke any other function will result in a parse error. This has some unfortunate consequences:

* It is not feasible to invoke functions defined in other `.cbl` modules, as their `Handle`s are not known until after they are parsed.
* It is impossible to use a `crucible-syntax` override to invoke a function that does not yet have a `Handle`. This can be useful for scenarios where functions are discovered at simulation time, which is something that happens in the context of `macaw`'s incremental code discovery.

To make such use cases possible, I propose adding a new top-level `crucible-syntax` form for declaring _forward references_ to function definitions. The idea is that a user would declare a forward reference like so:

```scm
(declare @<function-name> ((arg_1 Ty_1) ... (arg_n Ty_n)) RetTy)
```

The syntax for `(declare ...)` is very similar to that of `(defun ...)`, except that there is no function body. After `declare`-ing a function, a user will be able to `(funcall ...)` it like they normally would.

Implementation-wise, each `(declare ...)` would give rise to a fresh `FnHandle` that is not yet associated with an override in the simulator's `FnHandleMap`. Parsing a `.cbl` file containing `(declare ...)` declarations will give rise to a `Map FunctionName SomeFnHandle` mapping each `declare`'d function name to its corresponding handle. The responsibility is on the caller to ensure that each handle is registered in the simulator to an appropriate override; otherwise, invoking the `declare`d function will fail.

RyanGlScott:
This adds a `(declare ...)` form to the `crucible-syntax` language for forward declarations. A forward declaration is like a function, but lacking a body, and is useful for situations where one does not know what the implementation of a function will be until after the `.cbl` file is parsed.

Implementation-wise, this is accomplished by adding a new `parsedProgForwardDecs :: Map FunctionName SomeHandle` field to `ParsedProgram` that maps the name of each forward declaration to its handle. It is the responsibility of the caller to register each handle with a suitable `FnState`, or else invoking the forward declaration will result in a simulation error.

I've added a basic test case to ensure that this plan works as intended. This involved adding another field to `SimulateProgramHooks` (named `resolveForwardDeclarationsHook`) to allow resolving forward declarations before simulation begins.

Note that neither `crucibler` nor `cruces`, two of the current consumers of `crucible-syntax` in the `crucible` repo, support forward declarations at the moment, and trying to use forward declarations in the context of `crucibler` or `cruces` will result in an error. One could imagine adding support, although there would be some UX design questions to work out first.

Fixes #996.

### DIFF
--- a/crucible-concurrency/src/Cruces/CrucesMain.hs
+++ b/crucible-concurrency/src/Cruces/CrucesMain.hs
@@ -14,8 +14,10 @@ Maintainer       : Alexander Bakst <abakst@galois.com>
 module Cruces.CrucesMain (run, defaultCrucesOptions, cruciblesConfig) where
 
 import Control.Lens
+import Control.Monad (unless)
 import Data.Foldable (toList)
 import Data.List (find)
+import qualified Data.Map as Map
 import Data.Text (pack)
 import Data.String (IsString(..))
 import System.Exit
@@ -140,7 +142,10 @@ run (cruxOpts, opts) =
                        Right (ParsedProgram
                                { parsedProgGlobals = gs
                                , parsedProgCFGs = cs
-                               }) ->
+                               , parsedProgForwardDecs = fds
+                               }) -> do
+                         unless (Map.null fds) $
+                           error "Forward declarations not currently supported"
                          return ( findMain (fromString "main") cs
                                 , crucibleSyntaxPrims
                                 , toList gs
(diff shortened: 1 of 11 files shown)


travitch (approved):
I like the new more extensible API to crucible-syntax

travitch:
It would probably be worth getting signoff from @robdockins on this, though I'd be surprised if he was offended

bboston7 (approved):
This looks really useful!

robdockins (approved):
I like it, seems like a nice way to handle access to primitives and "module linking", etc.

PR #998, opened 2022-06-13
Don't treat struct padding as invalid memory with `stable-symbolic`

RyanGlScott:
When `laxLoadsAndStores` + `stable-symbolic` are enabled, all allocations are backed by fresh SMT arrays. When writing a struct to one of these arrays, we traverse through the array byte-by-byte, loading each byte in the struct and updating the array's value accordingly.

Things went awry in the case where the struct has padding, however, as the code in `crucible-llvm` assumed that loading struct padding should always be an error. Not only is this possible with `stable-symbolic`, it has a sensible interpretation: whenever loading a byte of struct padding, just skip updating the array value corresponding to that byte.

See GaloisInc/saw-script#1684 for the motivation behind this bugfix. I haven't found a way to trigger the same bug with `crux-llvm`, but I can trigger it programatically using the `crucible-llvm` API. I've added a test case to the `crucible-llvm` test suite to ensure that the bug remains fixed.


robdockins (approved):
Looks good to me! Tricky bug, good job tracking that down.

PR #1001, opened 2022-06-16
uc-crux-llvm: Factor out constraint-checking code into its own module 

langston-barrett:
This functionality will be useful for implementing #982. Just moving stuff around, no functionality changes.


langston-barrett:
Rebased onto #1002.

PR #1006, opened 2022-06-23
LLVM lazy load

robdockins:
This sets up the `ModuleTranslation` to perform translation to Crucible IR on-demand, instead of eagerly doing all the translation steps in the beginning. Hopefully, this will help with startup times and memory usage when loading large bitcode files.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM.hs
@@ -12,6 +12,7 @@
 {-# LANGUAGE ImplicitParams #-}
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 
@@ -33,7 +34,7 @@ import qualified Text.LLVM.AST as L
 import           Lang.Crucible.Analysis.Postdom
 import           Lang.Crucible.Backend
 import           Lang.Crucible.CFG.Core
-import           Lang.Crucible.FunctionHandle (lookupHandleMap, handleName)
+import           Lang.Crucible.FunctionHandle (lookupHandleMap)
 import           Lang.Crucible.LLVM.Eval (llvmExtensionEval)
 import           Lang.Crucible.LLVM.Extension (ArchWidth)
 import           Lang.Crucible.LLVM.Intrinsics
@@ -45,38 +46,38 @@ import           Lang.Crucible.LLVM.Translation
 import           Lang.Crucible.Simulator.ExecutionTree
 import           Lang.Crucible.Simulator.GlobalState
 import           Lang.Crucible.Simulator.OverrideSim
-import           Lang.Crucible.Utils.MonadVerbosity (showWarning)
-
+import           What4.Interface (getCurrentProgramLoc)
+import           What4.ProgramLoc (plSourceLoc)
 
 
 -- | Register all the functions defined in the LLVM module
 registerModule
    :: (1 <= ArchWidth arch, HasPtrWidth (ArchWidth arch), IsSymInterface sym) =>
+   (LLVMTranslationWarning -> IO ()) ->
    LLVMContext arch ->
    ModuleTranslation arch ->
    OverrideSim p sym LLVM rtp l a ()
-registerModule llvm_ctx mtrans =
-   mapM_ (registerModuleFn llvm_ctx mtrans) (map L.decName (mtrans ^. modTransDefs))
+registerModule handleWarning llvm_ctx mtrans =
+   mapM_ (registerModuleFn handleWarning llvm_ctx mtrans) (map L.decName (mtrans ^. modTransDefs))
 
 registerModuleFn
    :: (1 <= ArchWidth arch, HasPtrWidth (ArchWidth arch), IsSymInterface sym) =>
+   (LLVMTranslationWarning -> IO ()) ->
    LLVMContext arch ->
    ModuleTranslation arch ->
    L.Symbol ->
    OverrideSim p sym LLVM rtp l a ()
-registerModuleFn llvm_ctx mtrans sym =
+registerModuleFn handleWarning llvm_ctx mtrans sym =
   liftIO (getTranslatedCFG mtrans sym) >>= \case
     Nothing ->
       fail $ unlines
         [ "Could not find definition for function"
         , show sym
         ]
-    Just (decl, AnyCFG cfg, _warns) -> do
+    Just (decl, AnyCFG cfg, warns) -> do
       let h = cfgHandle cfg
           s = UseCFG cfg (postdomInfo cfg)
       binds <- use (stateContext . functionBindings)
-      when (isJust $ lookupHandleMap h $ fnBindings binds) $
-        showWarning ("LLVM function handle registered twice: " ++ show (handleName h))
       bindFnHandle h s
       let mvar = llvmMemVar llvm_ctx
       mem <- readGlobal mvar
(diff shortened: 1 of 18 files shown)


crucible-llvm/src/Lang/Crucible/LLVM.hs
```diff
@@ -25,43 +31,157 @@ module Lang.Crucible.LLVM
...
+                                               [ show retTy, show h ]
+                        (Just Refl, Just Refl) ->
+                           do -- Rebind the old handle to call directly into the new handle.
+                              -- This makes sure that even if the handle related to the global symbol
+                              -- has previously been resolved, the actual code for this function
+                              -- is correctly executed.
+                              bindFnHandle h
+                                $ UseOverride
+                                $ mkOverride' fn_name retTy
+                                $ (regValue <$> (callFnVal (HandleFnVal h') =<< getOverrideArgs))
```
RyanGlScott, line 166: I'm trying to figure out what implications this will have on `crux-llvm` profiling reports. If I understand how this works correctly, a profiling report will now display twice as many functions as it used to, since invoking a C function is tantamount to calling two functions on the Crucible side: one named `foo_bootstrap` and one named `foo`. Is that right? That might add some unfortunate clutter.

One alternative to the double-override approach would be to:

1. Give the bootstrapping function that name "`foo`" rather than "`foo_bootstrap`"
2. Inline the parts of `registerModuleFn` that construct a `CFG` for the translated function
3. Re-bind the bootstrapping function handle to a `UseCFG` rather than a `UseOverride`

That way, the profiling reports wouldn't have to mention `_bootstrap` names at all. The first invocation of the function would likely take up more time on the report due to the need to translate the CFG, but perhaps this is an acceptable compromise. Can you think of any other downsides to this?

RyanGlScott (commented):
Wow! This is extremely cool stuff. And if I understand correctly, `crux-llvm` and `uc-crux-llvm` have already been wired up to take advantage of lazy translation, yes? Have you tried this branch out on a large C program that only invokes some of its functions to measure the performance difference?

crucible-llvm/src/Lang/Crucible/LLVM.hs
```diff
@@ -25,43 +31,157 @@ module Lang.Crucible.LLVM
...
+   ModuleTranslation arch ->
+   OverrideSim p sym LLVM rtp l a ()
+registerModule handleWarning llvm_ctx mtrans =
+   mapM_ (registerModuleFn handleWarning llvm_ctx mtrans) (map L.decName (mtrans ^. modTransDefs))
+
+-- | Register a specific named function that is defined in the given
+--   module translation.
+registerModuleFn ::
+   (1 <= ArchWidth arch, HasPtrWidth (ArchWidth arch), IsSymInterface sym) =>
+   (LLVMTranslationWarning -> IO ()) ->
```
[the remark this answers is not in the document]
robdockins: For the lazy version, the translation happens at an arbitrarily-later time, so a continuation seemed like a good solution.  I did the same here to make the types match.

robdockins:
I did a quick-and-dirty job of wiring this into SAW, and it definitely made a pretty big difference on startup times and resident memory usage. It doesn't completely solve the memory pressure for the proof I'm working on, but I think it will help the current CI situation quite a bit.

crux-llvm/src/Crux/LLVM/Simulate.hs
```diff
@@ -276,8 +277,8 @@ checkFun ::
   GlobalVar Mem ->
   OverM personality sym LLVM ()
 checkFun llvmOpts trans memVar =
-  case Map.lookup (fromString nm) mp of
-    Just (_, AnyCFG anyCfg) ->
+  liftIO (getTranslatedCFG trans (fromString nm)) >>= \case
+    Just (_, AnyCFG anyCfg, _warns) ->
```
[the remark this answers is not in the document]
robdockins: probably

crucible-llvm/src/Lang/Crucible/LLVM.hs
robdockins: Yeah, the profiling data will be a bit messier. I think that's probably OK? I guess we could make it a configurable to use lazy or eager loading if users want a more straightforward profile.

I'm not sure the alternative you suggest would be better, but maybe I'm missing some advantage you have in mind?

RyanGlScott: > I'm not sure the alternative you suggest would be better, but maybe I'm missing some advantage you have in mind?

The difference in my approach is that when we invoke the bootstrap function for the first time, instead of creating a new function handle for the translated function, we just simply re-register the bootstrap function's handle with the CFG of the translated function. That way, there's no need to have two handles for each LLVM function. Instead, we would have a single handle per function whose `FnState` will be registered to different things depending on whether you have invoked it once or not.

crucible-llvm/src/Lang/Crucible/LLVM.hs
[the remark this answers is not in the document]
RyanGlScott: Fair enough. I'd overlooked the fact that it's not quite as straightforward to return the `LLVMTranslationWarning`s directly from `registerLazyModuleFn`, which this function invokes.

crucible-llvm/src/Lang/Crucible/LLVM.hs
RyanGlScott: Well, I guess you'd still have two handles for each function, since the CFG has its own handle. But you'd only need to _register_ the handle for the bootstrapping function in the simulator, so from the perspective of the profiler, there is only one thing that shows up in the profiler for each function.

robdockins: I guess we could change the API for translation so that it could take a function handle instead of inventing its own.  That would mean we only have to rebind a single function handle.

I'm still not sure if it is better to hide the lazy loading from profiling or not.  I think having the time spent on loading and translating the code just be silently included in the profile might be pretty confusing, actually.

RyanGlScott: > I think having the time spent on loading and translating the code just be silently included in the profile might be pretty confusing, actually.

True. I'm not sure which of the two approach's downsides would be worse.

A third possible approach: instead of registering boostrap functions for every LLVM function at the start of simulation, don't register anything at all. Later, when running an `LLVM_LoadHandle` in the memory model, there is a check [here](https://github.com/GaloisInc/crucible/blob/b558ba2341b11c88e668ed3f772f4d1960275229/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L440-L468) to see if the loaded handle has been registered in the simulator. Currently, this check will throw an assertion failure if that is not the case, but we could instead perform the lazy translation and registration there. This would ensure both that each function is only registered once and that the time spent during profiling only measures execution time, not translation time. (This is similar to the approach to profiling that we use in a `macaw`-based verifier that also supports lazy translation.)

I haven't fully thought through what code changes would need to happen to make this work. For instance, you might need to plumb through some additional state to the memory model to support lazy translation. But I think this would address both issues, provided it's feasible. If it isn't, I certainly don't want this to hold up the PR.

robdockins:
Sadly, evidence suggests this doesn't help as much as I hoped it would. CI jobs for SAW that were running out of memory still appear to be doing so...

crucible-llvm/src/Lang/Crucible/LLVM.hs
```diff
@@ -25,43 +31,157 @@ module Lang.Crucible.LLVM
...
+import           What4.ProgramLoc (plSourceLoc)
+import           What4.FunctionName (functionNameFromText)
+
+
+-- | Register all the functions defined in the LLVM module.
+registerModule ::
+   (1 <= ArchWidth arch, HasPtrWidth (ArchWidth arch), IsSymInterface sym) =>
+   (LLVMTranslationWarning -> IO ()) ->
+   LLVMContext arch ->
+   ModuleTranslation arch ->
```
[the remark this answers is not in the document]
robdockins: That is a good point; I think we probably don't need to take a separate `LLVMContext`.

robdockins:
After thinking about this some more, I think actually we should not be updating the binding between global symbols and function handles, which lives in the LLVM memory.  The problem here is that this might happen inside a symbolic branch, which would lead to function handle resolution failures later, I think.

Instead, we should set it up so that `getTranslatedCFG` (can?, must?)  take a function handle to use when translating.  Then we only must rebind the function handle (which is an out-of-band operation) instead of updating the global symbol mapping.

RyanGlScott (approved):
Wonderful! I'm much happier with the new approach to lazy registration, as now (1) we only have one function handle per function, and (2) the call that performs translation happens separately from all other calls.

PR #1009, opened 2022-06-29
Bump tasty-sugar to use version 1.3.

kquick:


RyanGlScott:
I think you may need to update the `index-state` in each freeze file to a newer date, as the date it's currently frozen at happened before `tasty-sugar-1.3.0.0`'s release. You may also need to bump the `what4` submodule to bring in GaloisInc/what4#205 so that `what4`'s current upper bounds on `tasty-sugar` in its test suites don't interfere with the build plans here.

RyanGlScott (approved):

PR #1022, opened 2022-08-08
Simple loop invariants

robdockins:
This PR reworks the "simple loop fixpoint" feature to be based more directly on user-provided loop invariants instead.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
@@ -68,6 +68,8 @@
 -- to discover what the values appearing in the invariant correspond
 -- to. This process may well be quite sensitive to changes in the
 -- source code.
+--
+-- Limitiations: currently, this feature is restricted to 64-bit code.
 ------------------------------------------------------------------------
 
 {-# LANGUAGE DataKinds #-}
@@ -120,12 +122,11 @@ import           Data.Parameterized.NatRepr
 import           Data.Parameterized.TraversableF
 import           Data.Parameterized.TraversableFC
 
---import qualified What4.Config as W4
+import qualified What4.Config as W4
 import qualified What4.Interface as W4
 import qualified What4.ProgramLoc as W4
 
 import qualified Lang.Crucible.Analysis.Fixpoint.Components as C
-import qualified Lang.Crucible.Analysis.Postdom as C
 import qualified Lang.Crucible.Backend as C
 import qualified Lang.Crucible.CFG.Core as C
 import qualified Lang.Crucible.CFG.Extension as C
@@ -178,9 +179,7 @@ instance OrdF (W4.SymExpr sym) => OrdF (InvariantEntry sym) where
     GTF -> GTF
 
 instance OrdF (InvariantEntry sym) => W4.TestEquality (InvariantEntry sym) where
-  testEquality x y = case compareF x y of
-    EQF -> Just Refl
-    _ -> Nothing
+  testEquality x y = orderingF_refl (compareF x y)
 
 -- | This datatype captures the state machine that progresses as we
 --   attempt to compute a loop invariant for a simple structured loop.
@@ -191,7 +190,7 @@ data FixpointState p sym ext rtp blocks
     -- | We have encountered the loop head at least once, and are in the process
     --   of converging to an inductive representation of the live variables
     --   in the loop.
-  | ComputeFixpoint (FixpointRecord p sym ext rtp blocks)
+  | ComputeFixpoint C.FrameIdentifier (FixpointRecord p sym ext rtp blocks)
 
     -- | We have found an inductively-strong representation of the live variables
     --   of the loop. We are now executing the from the loop head one final time
@@ -201,17 +200,12 @@ data FixpointState p sym ext rtp blocks
     | CheckFixpoint (FixpointRecord p sym ext rtp blocks)
 
 
-type ArrayTp = W4.BaseArrayType (C.EmptyCtx C.::> W4.BaseBVType 64) (W4.BaseBVType 8)
-
 -- | Data about the loop that we incrementally compute as we approach fixpoint.
 data FixpointRecord p sym ext rtp blocks = forall args r.
   FixpointRecord
   {
     -- | Block identifier of the head of the loop
     fixpointBlockId :: C.BlockID blocks args
-
-    -- | Identifier for the currently-active assumption frame related to this fixpoint computation
-  , fixpointAssumptionFrameIdentifier :: C.FrameIdentifier
 
     -- | Map from introduced widening variables to prestate value before the loop starts,
     --   and to the value computed in a single loop iteration, assuming we return to the
@@ -292,6 +286,8 @@ data MemoryBlockData sym where
     W4.SymBV sym 64 {- ^ length of the allocation -} ->
     MemoryBlockData sym
 
+type ArrayTp = W4.BaseArrayType (C.EmptyCtx C.::> W4.BaseBVType 64) (W4.BaseBVType 8)
+
 -- | A memory substitution gives memory block data for
 --   concrete memory region numbers of writes occurring
 --   in the loop body. This is used to determine where
@@ -308,7 +304,7 @@ fixpointRecord ::
   FixpointState p sym ext rtp blocks ->
   Maybe (FixpointRecord p sym ext rtp blocks)
 fixpointRecord BeforeFixpoint = Nothing
-fixpointRecord (ComputeFixpoint r) = Just r
+fixpointRecord (ComputeFixpoint _ r) = Just r
 fixpointRecord (CheckFixpoint r) = Just r
 
 
@@ -346,18 +342,20 @@ joinRegEntry sym left right = do
  case C.regType left of
   C.LLVMPointerRepr w
 
+      -- TODO! This is a "particularly guesome hack" it would be nice to find some better
+      --  way to handle this situation.
       -- special handling for "don't care" registers coming from Macaw
     | List.isPrefixOf "cmacaw_reg" (show $ W4.printSymNat $ C.llvmPointerBlock (C.regValue left))
     , List.isPrefixOf "cmacaw_reg" (show $ W4.printSymExpr $ C.llvmPointerOffset (C.regValue left)) -> do
-      liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: cmacaw_reg"
+      -- liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: cmacaw_reg"
       return left
 
     | C.llvmPointerBlock (C.regValue left) == C.llvmPointerBlock (C.regValue right)
     , Nothing <- MapF.lookup (C.llvmPointerOffset (C.regValue left)) (varHavoc subst) -> do
-      liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: LLVMPointerRepr"
+      -- liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: LLVMPointerRepr"
       if isJust (W4.testEquality (C.llvmPointerOffset (C.regValue left)) (C.llvmPointerOffset (C.regValue right)))
       then do
-        liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: LLVMPointerRepr: left == right"
+        -- liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: LLVMPointerRepr: left == right"
         return left
       else case MapF.lookup (C.llvmPointerOffset (C.regValue left)) (varSubst subst) of
         Just join_entry -> do
@@ -410,11 +408,11 @@ joinRegEntry sym left right = do
       --   ++ show (W4.printSymExpr $ C.regValue left)
       --   ++ " \\/ "
       --   ++ show (W4.printSymExpr $ C.regValue right)
-      join_varaible <- liftIO $ W4.freshConstant sym (W4.safeSymbol "macaw_reg") W4.BaseBoolRepr
+      join_varaible <- liftIO $ W4.freshConstant sym (W4.safeSymbol "reg_join_var") W4.BaseBoolRepr
       return $ C.RegEntry C.BoolRepr join_varaible
 
   C.StructRepr field_types -> do
-    liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: StructRepr"
+    -- liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: StructRepr"
     C.RegEntry (C.regType left) <$> fmapFC (C.RV . C.regValue) <$> joinRegEntries sym
       (Ctx.generate (Ctx.size field_types) $ \i ->
         C.RegEntry (field_types Ctx.! i) $ C.unRV $ (C.regValue left) Ctx.! i)
@@ -456,7 +454,9 @@ applySubstitutionRegEntry sym substitution entry = case C.regType entry of
           Ctx.generate (Ctx.size field_types) $
           \i -> C.RegEntry (field_types Ctx.! i) $ C.unRV $ (C.regValue entry) Ctx.! i
       }
-  _ -> C.panic "SimpleLoopInvariant.applySubstitutionRegEntry" ["unsupported type: " ++ show (C.regType entry)]
+  _ -> error $ unlines [ "SimpleLoopInvariant.applySubstitutionRegEntry"
+                       , "unsupported type: " ++ show (C.regType entry)
+                       ]
 
 
 loadMemJoinVariables ::
@@ -644,23 +644,23 @@ simpleLoopInvariant ::
   (InvariantPhase -> [Some (W4.SymExpr sym)] -> MapF (W4.SymExpr sym) (InvariantEntry sym) -> IO (W4.Pred sym)) ->
   IO (C.ExecutionFeature p sym ext rtp)
 simpleLoopInvariant sym loopNum cfg@C.CFG{..} mem_var loop_invariant = do
+  -- TODO, can we lift this restriction to 64-bits? I don't think there
+  -- is anything fundamental about it.
   let ?ptrWidth = knownNat @64
 
-  --verbSetting <- W4.getOptionSetting W4.verbosity $ W4.getConfiguration sym
-  --verb <- fromInteger <$> W4.getOpt verbSetting
+  verbSetting <- W4.getOptionSetting W4.verbosity $ W4.getConfiguration sym
+  verb <- fromInteger <$> W4.getOpt verbSetting
 
   loop_map <- computeLoopMap loopNum (C.cfgWeakTopologicalOrdering cfg)
 
   fixpoint_state_ref <- newIORef @(FixpointState p sym ext rtp blocks) BeforeFixpoint
 
-  putStrLn "Setting up simple loop fixpoints feature."
-  putStrLn ("Loop map: " ++ show loop_map)
-  putStrLn ("WTO: " ++ show (C.cfgWeakTopologicalOrdering cfg))
-  putStrLn (show (C.cfgHandle cfg))
-  writeFile "bigcfg.txt" (show (C.ppCFG' True (C.postdomInfo cfg) cfg))
+  --putStrLn "Setting up simple loop fixpoints feature."
+  --putStrLn ("Loop map: " ++ show loop_map)
+  --putStrLn ("WTO: " ++ show (C.cfgWeakTopologicalOrdering cfg))
 
   return $ C.ExecutionFeature $ \exec_state -> do
-    let ?logMessage = \msg -> do -- when (verb >= (3 :: Natural)) $ do
+    let ?logMessage = \msg -> when (verb >= (3 :: Natural)) $ do
           let h = C.printHandle $ C.execStateContext exec_state
           System.IO.hPutStrLn h msg
           System.IO.hFlush h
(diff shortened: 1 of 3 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
+------------------------------------------------------------------------
+-- |
+-- Module           : Lang.Crucible.LLVM.SimpleLoopInvariant
+-- Description      : Execution feature to perform verification of simply
+--                    structured loops via invariants.
+-- Copyright        : (c) Galois, Inc 2021
```
langston-barrett, line 6:
```suggestion
-- Copyright        : (c) Galois, Inc 2022
```

robdockins:
This feature still needs some error checking, but I think the basics are ready for review.

In particular, we still need to check that the indicated loop is single-entry/single-exit, and we need to implement checking that the memory footprint is acceptable (i.e., reaches fixpoint).

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+--   a loop body, new "widening" variables are introduced to stand in
+--   for those locations.  When we introduce such a varible, we
+--   capture what value the variable had when we entered the loop (the
+--   \"header\" value); this is essentially the initial value of the
+--   variable.  We also compute what value the variable should take on
+--   its next iteration assuming the loop doesn't exit and executes
+--   along its backedge.  This \"body\" value will be computed in
+--   terms of the the set of all discovered live variables so far.
+--   We know we have reached fixpoint when we don't need to introduce
+--   and more fresh widening variables, and the body values for each
```
RyanGlScott, line 166:
```suggestion
--   any more fresh widening variables, and the body values for each
```

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+instance OrdF (W4.SymExpr sym) => OrdF (InvariantEntry sym) where
+  compareF x y = case compareF (headerValue x) (headerValue y) of
+    LTF -> LTF
+    EQF -> compareF (bodyValue x) (bodyValue y)
+    GTF -> GTF
+
+instance OrdF (InvariantEntry sym) => W4.TestEquality (InvariantEntry sym) where
+  testEquality x y = case compareF x y of
+    EQF -> Just Refl
+    _ -> Nothing
```
RyanGlScott, line 183: You could simplify this with [`orderingF_refl`](https://hackage.haskell.org/package/parameterized-utils-2.1.5.0/docs/Data-Parameterized-Classes.html#v:orderingF_refl).

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+
+    -- | We have found an inductively-strong representation of the live variables
+    --   of the loop. We are now executing the from the loop head one final time
+    --   to produce the proof obligations arising from the body of the loop,
+    --   the main inductive loop invariant obligation, and any obligations arising
+    --   from code following the loop exit.
+    | CheckFixpoint (FixpointRecord p sym ext rtp blocks)
+
+
+type ArrayTp = W4.BaseArrayType (C.EmptyCtx C.::> W4.BaseBVType 64) (W4.BaseBVType 8)
```
RyanGlScott, line 204: This is a curious definition. What is its purpose?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+
+-- | Data about the loop that we incrementally compute as we approach fixpoint.
+data FixpointRecord p sym ext rtp blocks = forall args r.
+  FixpointRecord
+  {
+    -- | Block identifier of the head of the loop
+    fixpointBlockId :: C.BlockID blocks args
+
+    -- | identifier for the currently-active assumption frame related to this fixpoint computation
```
RyanGlScott, line 213:
```suggestion
    -- | Identifier for the currently-active assumption frame related to this fixpoint computation
```

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+    | List.isPrefixOf "cmacaw" (show $ W4.printSymExpr $ C.regValue left) -> do
+      liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: cmacaw_reg"
+      return left
+    | otherwise -> do
+      -- liftIO $ ?logMessage $
+      --   "SimpleLoopInvariant.joinRegEntry: BoolRepr:"
+      --   ++ show (W4.printSymExpr $ C.regValue left)
+      --   ++ " \\/ "
+      --   ++ show (W4.printSymExpr $ C.regValue right)
+      join_varaible <- liftIO $ W4.freshConstant sym (W4.safeSymbol "macaw_reg") W4.BaseBoolRepr
```
RyanGlScott, line 413: I'm a bit confused by this code. Why would we always name this `macaw_reg`? As far as I can tell, it's possible to hit this case even when neither `left` nor `right` have `macaw`-derived names.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+        , fixpointRegMap = sim_state ^. (C.stateCrucibleFrame . C.frameRegs)
+        , fixpointMemSubstitution = MemSubst mempty
+        , fixpointInitialSimState = sim_state
+        , fixpointImplicitParams = []
+        , fixpointHavocBlock = havoc_blk
+        }
+      return $ C.ExecutionFeatureModifiedState $ C.RunningState (C.RunBlockStart block_id) $
+        sim_state & C.stateGlobals %~ C.insertGlobal mem_var res_mem_impl
+
+    ComputeFixpoint fixpoint_record
```
RyanGlScott, line 773: This is a particularly large case. It might be worth breaking this up into some helper functions to make it more readable.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+-- | This execution feature is designed to allow a limited form of
+--   verification for programs with unbounded looping structures.
+--
+--   It is currently highly experimental and has many limititations.
+--   Most notibly, it only really works properly for functions
+--   consiting of a single, non-nested loop with a single exit point.
+--   Moreover, the loop must have an indexing variable that counts up
+--   from a starting point by a fixed stride amount.
+--
+--   Currently, these assumptions about the loop strucutre are not
```
RyanGlScott, line 627:
```suggestion
--   Currently, these assumptions about the loop structure are not
```

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+      { C.regValue = fmapFC (C.RV . C.regValue) $
+          applySubstitutionRegEntries sym substitution $
+          Ctx.generate (Ctx.size field_types) $
+          \i -> C.RegEntry (field_types Ctx.! i) $ C.unRV $ (C.regValue entry) Ctx.! i
+      }
+  _ -> C.panic "SimpleLoopInvariant.applySubstitutionRegEntry" ["unsupported type: " ++ show (C.regType entry)]
+
+
+loadMemJoinVariables ::
+  (C.IsSymBackend sym bak, C.HasPtrWidth 64, C.HasLLVMAnn sym, ?memOpts :: C.MemOptions) =>
```
RyanGlScott, line 463: Judging from the use of `HasPtrWidth 64`, does this only work for 64-bit? If so, it would be worth documenting this requirement somewhere.

RyanGlScott (commented):
I feel very underqualified to review this, but I've done a quick pass and left some comments on things that stood out to me.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+          , Just loop_body_some_block_ids <- Map.lookup (C.Some fixpoint_block_id) loop_map
+          , JustPausedFrameTgtId true_frame_some_block_id <- pausedFrameTgtId true_frame
+          , JustPausedFrameTgtId false_frame_some_block_id <- pausedFrameTgtId false_frame
+          , C.SomeHandle cfgHandle == C.frameHandle (sim_state ^. C.stateCrucibleFrame)
+          , Just Refl <- W4.testEquality
+              (fmapFC C.blockInputs cfgBlockMap)
+              (fmapFC C.blockInputs $ C.frameBlockMap $ sim_state ^. C.stateCrucibleFrame)
+          , elem true_frame_some_block_id loop_body_some_block_ids /= elem false_frame_some_block_id loop_body_some_block_ids -> do
+
+            (loop_condition, inside_loop_frame, _outside_loop_frame) <-
```
travitch, line 700: Is this just normalizing the order of the conditions? If the outside loop frame is never used, should this just be deleted in this binding?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+
+advanceFixpointState bak mem_var loop_invariant block_id sim_state fixpoint_state fixpoint_state_ref = do
+  let ?ptrWidth = knownNat @64
+  let sym = C.backendGetSym bak
+  loc <- W4.getCurrentProgramLoc sym
+  case fixpoint_state of
+    BeforeFixpoint -> do
+      ?logMessage $ "SimpleLoopInvariant: RunningState: BeforeFixpoint -> ComputeFixpoint " ++ show block_id ++ " " ++ show (pretty (W4.plSourceLoc loc))
+      assumption_frame_identifier <- C.pushAssumptionFrame bak
+      let mem_impl = fromJust $ C.lookupGlobal mem_var (sim_state ^. C.stateGlobals)
```
travitch, line 751: Any `fromJust` always scares me. I guess it has a stack trace now, but it might be nice to panic

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+                       (\ (MapF.Pair _ e) ->
+                            -- filter out the special "noSatisfyingWrite" boolean constants
+                            -- that are generated as part of the LLVM memory model
+                            Set.filter ( \ (C.Some x) ->
+                                           not (List.isPrefixOf "cnoSatisfyingWrite"
+                                                (show $ W4.printSymExpr x))) $
+                            Set.map (\ (C.Some x) -> C.Some (W4.varExpr sym x)) $
+                              (W4.exprUninterpConstants sym (bodyValue e)))
+                       (MapF.toList (varSubst normal_substitution)))
+                    (Set.fromList (MapF.keys (varSubst normal_substitution)))
```
travitch, line 860: This is a good candidate for factoring out with a bit of explanation

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+
+  loop_map <- computeLoopMap loopNum (C.cfgWeakTopologicalOrdering cfg)
+
+  fixpoint_state_ref <- newIORef @(FixpointState p sym ext rtp blocks) BeforeFixpoint
+
+  putStrLn "Setting up simple loop fixpoints feature."
+  putStrLn ("Loop map: " ++ show loop_map)
+  putStrLn ("WTO: " ++ show (C.cfgWeakTopologicalOrdering cfg))
+  putStrLn (show (C.cfgHandle cfg))
+  writeFile "bigcfg.txt" (show (C.ppCFG' True (C.postdomInfo cfg) cfg))
```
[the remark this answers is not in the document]
robdockins: We should definitely not keep this in the final version.

What I'm not sure how to do is establish a good way for the user to see things like the WTO analysis/loop map and the results of the loop-carried dependency analysis when they want to see it (i.e., when first attempting to apply this to a new bit of code, you probably don't know how many/what types those dependencies will have) but not always print a bunch of junk.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
robdockins: We can probably delete the unused outside loop frame.  Previously, the code was using it and that got refactored away, but I never pushed it into here.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
robdockins: Yeah, we can panic here, maybe that's better.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
robdockins: I was using it in some time signatures somewhere, but perhaps all the uses got deleted... I'll check.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+  C.RegEntry sym tp ->
+  C.RegEntry sym tp ->
+  FixpointMonad sym (C.RegEntry sym tp)
+joinRegEntry sym left right = do
+ subst <- get
+ case C.regType left of
+  C.LLVMPointerRepr w
+
+      -- special handling for "don't care" registers coming from Macaw
+    | List.isPrefixOf "cmacaw_reg" (show $ W4.printSymNat $ C.llvmPointerBlock (C.regValue left))
```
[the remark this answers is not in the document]
robdockins: The main problem here, I think, is that even if we factor this out, it's hard to get the information we actually want (i.e., is this actually a manufactured "don't care" variable).  But yes, taking a predicate like that would let us push this particular grossness up into SAW instead.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
robdockins: Yes, only on 64-bit. Agreed, should be documented.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+  forall sym ext p rtp blocks init ret .
+  (C.IsSymInterface sym, C.IsSyntaxExtension ext, C.HasLLVMAnn sym, ?memOpts :: C.MemOptions) =>
+  sym ->
+  Integer {- ^ which of the discovered loop heads to install a loop invariant onto -}  -> 
+  C.CFG ext blocks init ret {- ^ The function we want to verify -} ->
+  C.GlobalVar C.Mem {- ^ global variable representing memory -} ->
+  (InvariantPhase -> [Some (W4.SymExpr sym)] -> MapF (W4.SymExpr sym) (InvariantEntry sym) -> IO (W4.Pred sym)) ->
+  IO (C.ExecutionFeature p sym ext rtp)
+simpleLoopInvariant sym loopNum cfg@C.CFG{..} mem_var loop_invariant = do
+  let ?ptrWidth = knownNat @64
```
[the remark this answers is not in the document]
robdockins: We could maybe generalize this, but I haven't investigated how invasive that change would be.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+              $ C.GenericAssumption loc "loop head invariant"
+                hypothetical_loop_invariant
+
+            -- == set up the state with arbitrary values to run the loop body ==
+
+            writeIORef fixpoint_state_ref $
+              CheckFixpoint
+                FixpointRecord
+                { fixpointBlockId = block_id
+                , fixpointAssumptionFrameIdentifier = undefined --assumption_frame_identifier
```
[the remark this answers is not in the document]
robdockins: Oh, good question.  I think it's safe because that assumption frame only exists when we are constructing the fixpoint, but I don't know if this is actually a good way to do this.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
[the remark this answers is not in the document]
travitch: This is in IO already, so how about an environment variable that is at least easy to toggle on?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+      return $ C.RegEntry C.BoolRepr join_varaible
+
+  C.StructRepr field_types -> do
+    liftIO $ ?logMessage "SimpleLoopInvariant.joinRegEntry: StructRepr"
+    C.RegEntry (C.regType left) <$> fmapFC (C.RV . C.regValue) <$> joinRegEntries sym
+      (Ctx.generate (Ctx.size field_types) $ \i ->
+        C.RegEntry (field_types Ctx.! i) $ C.unRV $ (C.regValue left) Ctx.! i)
+      (Ctx.generate (Ctx.size field_types) $ \i ->
+        C.RegEntry (field_types Ctx.! i) $ C.unRV $ (C.regValue right) Ctx.! i)
+  _ -> fail $ "SimpleLoopInvariant.joinRegEntry: unsupported type: " ++ show (C.regType left)
```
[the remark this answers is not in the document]
robdockins: I'm not sure this should `panic`, as legitimate user input can trigger this condition. The later one should probably also not panic, as I don't see a way to convince myself that's only possible via programmer error.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -0,0 +1,1151 @@
...
+  _ -> C.panic "SimpleLoopInvariant.dropMemStackFrame" ["not a stack frame:", show (C.ppMem $ C.memImplHeap mem)]
+
+
+filterSubstitution ::
+  C.IsSymInterface sym =>
+  sym ->
+  VariableSubst sym ->
+  VariableSubst sym
+filterSubstitution sym (VarSubst substitution havoc) =
+  -- TODO: fixpoint
```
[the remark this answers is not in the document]
robdockins: I think... that this doesn't impact soundness. @andreistefanescu, any thoughts about this? I think this is just for finding join variables that are actually equal, so the worst think that happens is there are multiple equal variables inserted into the loop invariant.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -1107,37 +1135,50 @@ computeMemSubstitution sym fixpoint_record mem_writes =
...
+ where
+   checkEqualMaps str f m1 m2 =
+     do unless (Map.keysSet m1 == Map.keysSet m2)
+               (throwError ("Key sets differ when checking " ++ str))
+        forM_ (Map.assocs m1) $ \ (k,e1) ->
+               case Map.lookup k m2 of
+                 Just e2 -> f k e1 e2
+                 Nothing -> throwError ("Key sets differ when checking " ++ str)
+
+   checkCandidateEqual = checkEqualMaps "memory subsitution" checkMBD (memSubst orig) (memSubst candidate)
```
RyanGlScott, line 1159:
```suggestion
   checkCandidateEqual = checkEqualMaps "memory substitution" checkMBD (memSubst orig) (memSubst candidate)
```

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -644,23 +673,22 @@ simpleLoopInvariant ::
   (InvariantPhase -> [Some (W4.SymExpr sym)] -> MapF (W4.SymExpr sym) (InvariantEntry sym) -> IO (W4.Pred sym)) ->
   IO (C.ExecutionFeature p sym ext rtp)
 simpleLoopInvariant sym loopNum cfg@C.CFG{..} mem_var loop_invariant = do
+  -- TODO, can we lift this restriction to 64-bits? I don't think there
+  -- is anything fundamental about it.
```
RyanGlScott, line 677: File an issue about this?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopInvariant.hs
```diff
@@ -308,10 +304,15 @@ fixpointRecord ::
...
-fixpointRecord (ComputeFixpoint r) = Just r
+fixpointRecord (ComputeFixpoint _ r) = Just r
 fixpointRecord (CheckFixpoint r) = Just r
 
 
+-- The fixpoint monad is used to ease the process of computing variable widenings
+-- and such.  The included "SymNat" is a memory region number guaranteed not
+-- to be a valid memory region; it is used to implement "havoc" registers that
+-- we expect to be junk/scratch space across the loop boundary.
+-- The state component tracks the variable subsitution we are computing.
```
RyanGlScott, line 315:
```suggestion
-- The state component tracks the variable substitution we are computing.
```

PR #1023, opened 2022-08-09
Update CI for nixos-unstable switch from 9.2.3 to 9.2.4.

kquick:


.github/workflows/crucible-go-build.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-20.04]
-        ghc: ["8.10.7", "9.0.2", "9.2.3"]
+        ghc: ["8.10.7", "9.0.2", "9.2.4"]
         include:
           - os: macos-12
             ghc: 9.2.2
```
RyanGlScott, line 33: Shouldn't the OSes listed under the `include`s also be updated to GHC 9.2.4? (This is one of them, but there are others as well.)

.github/workflows/crux-llvm-build.yml
```diff
@@ -95,7 +95,7 @@ jobs:
           case ${{ matrix.ghc }} in
             9.0.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
             9.2.2) GHC_NIXPKGS=github:nixos/nixpkgs/nixos-22.05 ;;
```
RyanGlScott, line 97: Any reason to still keep this line around?

cabal.GHC-9.2.4.config
```diff
@@ -0,0 +1,299 @@
+active-repositories: hackage.haskell.org:merge
```
RyanGlScott, line 1: Now that we have a freeze file for 9.2.4, we should be able to get rid of the ones for 9.2.2 and 9.2.3, no?

.github/workflows/crucible-go-build.yml
kquick: Good catch: I'd just looked for 9.2.3 but hadn't scanned the entire yaml file.  This is now updated.

.github/workflows/crux-llvm-build.yml
kquick: Just a little bit of hoarding instinct: while 9.2 is "in flux", having those historic touchpoints makes me feel safer.   I tend to reference CI build processes of various projects as a fairly definitive and explicit way of building things, and so this might be useful to anyone with similar habits that happens to be trying to use 9.2.2 instead of 9.2.4.

That said, there's always git history (although archaeology is not always the easiest way to find information), so there's no really good reason to keep this if you have a strong preference for its removal.

cabal.GHC-9.2.4.config
kquick: Generally, yes.  The caveat would be people still trying to build with 9.2.2 and 9.2.3 (similar to the previous comment).  My general thinking is that 9.2 roughly represents our "current" target (9.4 is the leading edge, and 8.10 is the trailing edge ... we can support older than 8.10 if convenient, but we aren't committing to it), and I usually look to contract the support range when things move from current to trailing.

.github/workflows/crux-llvm-build.yml
RyanGlScott: OK. I don't feel too strongly about keeping around the configurations for old minor releases—we'll just have to remember to update things like old freeze files.

RyanGlScott (approved):

Issue #888, opened 2021-10-15
crux fails to build golang-0.1.0.0


sp1ff:
New user here, following the "quickstart" guide. `cabal new-build all` fails while building golang-0.1.0.0:

<details>
  <summary>Click to expand!</summary>

```
[5 of 6] Compiling Language.Go.Parser ( src/Language/Go/Parser.hs, /home/mgh/code/projects/crucible/dist-newstyle/build/x86_64-linux/ghc-8.10.7/golang-0.1.0.0/build/Language/Go/Parser.o, /home/mgh/code/projects/crucible/dist-newstyle/build/x86_64-linux/ghc-8.10.7/golang-0.1.0.0/build/Language/Go/Parser.dyn_o )

src/Language/Go/Parser.hs:199:50: error:
    • Couldn't match type ‘Data.Aeson.KeyMap.KeyMap Value’
                     with ‘HM.HashMap k0 Value’
      Expected type: HM.HashMap k0 Value
        Actual type: Object
    • In the second argument of ‘HM.lookup’, namely ‘v’
      In the expression: HM.lookup "value" v
      In the expression:
        case HM.lookup "value" v of
          Just (Object v')
            -> case HM.lookup "type" v' of
                 Just "IOTA"
                   -> return $ IdentExpr pos go_tp Nothing $ Ident IdentNoKind "IOTA"
                 _ -> IdentExpr pos go_tp <$> v .:? "qualifier" <*> v .: "value"
          _ -> fail ""
    |
199 |           "identifier" -> case HM.lookup "value" v of
    |                                                  ^

src/Language/Go/Parser.hs:200:55: error:
    • Couldn't match type ‘Data.Aeson.KeyMap.KeyMap Value’
                     with ‘HM.HashMap k1 a0’
      Expected type: HM.HashMap k1 a0
        Actual type: Object
    • In the second argument of ‘HM.lookup’, namely ‘v'’
      In the expression: HM.lookup "type" v'
      In the expression:
        case HM.lookup "type" v' of
          Just "IOTA"
            -> return $ IdentExpr pos go_tp Nothing $ Ident IdentNoKind "IOTA"
          _ -> IdentExpr pos go_tp <$> v .:? "qualifier" <*> v .: "value"
    |
200 |             Just (Object v') -> case HM.lookup "type" v' of
    |                                                       ^^
Installing   bv-sized-1.0.3 (lib)
Completed    bv-sized-1.0.3 (lib)
[...]

PR #1030, opened 2022-08-30
{uc-,}crux-llvm: Recommend cloning with --recurse-submodules

langston-barrett:
See commentary on #888. Also, --recursive is no longer part of the git clone manpage, so replace with --recurse-submodules.

### DIFF
--- a/crux-llvm/README.md
+++ b/crux-llvm/README.md
@@ -56,7 +56,7 @@ The `crux-llvm` tool can be built by doing the following:
 
 * Clone (incl. the submodules) the `crucible` repository:
 
-        git clone --recursive https://github.com/GaloisInc/crucible.git
+        git clone --recurse-submodules https://github.com/GaloisInc/crucible.git
 
 * Build the `crux-llvm` package:
 
--- a/uc-crux-llvm/README.md
+++ b/uc-crux-llvm/README.md
@@ -139,7 +139,7 @@ The `uc-crux-llvm` tool can be built by doing the following:
 
 * Clone the enclosing `crucible` repository:
 
-        git clone https://github.com/GaloisInc/crucible.git
+        git clone --recurse-submodules https://github.com/GaloisInc/crucible.git
 
 * Change to the `uc-crux-llvm` directory and run the build script:
 


RyanGlScott:
We should also update the [quick start](https://github.com/GaloisInc/crucible/tree/7501d6b62be53fec0a1da4d92749c300c19da23a#quick-start) section of the `README`, since that is where the confusion in #888 arose.

langston-barrett:
> We should also update the [quick start](https://github.com/GaloisInc/crucible/tree/7501d6b62be53fec0a1da4d92749c300c19da23a#quick-start) section of the README, since that is where the confusion in https://github.com/GaloisInc/crucible/issues/888 arose.

@RyanGlScott This PR is intended to address the dicussion following [this comment](https://github.com/GaloisInc/crucible/issues/888#issuecomment-1195013559), which results from not having cloned the submodules (whereas the OP's issue was an actual build failure in `crucible-go`/`golang`). The quickstart section of the README already recommends cloning the submodules (via `scripts/build-sandbox.sh`), so I don't think a change is needed there.

RyanGlScott (approved):
Oh, my bad—I didn't realize that that was the purpose of the `build-sandbox.sh` script. (In my defense, that name is a bit misleading now that it's using `v2`-style `cabal` commands rather than sandboxes, but I digress.) Perhaps we should also update this line to use `--recurse-submodules` as well?

https://github.com/GaloisInc/crucible/blob/7501d6b62be53fec0a1da4d92749c300c19da23a/scripts/build-sandbox.sh#L3

Other than that, LGTM.

langston-barrett:
> Perhaps we should also update this line to use --recurse-submodules as well?

Ah, it's an either-or thing. You can *either* `git clone --recurse-submodules`, which clones and then fetches submodules, *or* you can `git clone http://foo.bar/whatever.git`, `cd whatever.git`, `git submodule update --init --recursive`. Docs for `--recurse-submodules:

> ```
>        --recurse-submodules[=<pathspec>]
>            After the clone is created, initialize and clone submodules within based on the provided
>            pathspec. If no pathspec is provided, all submodules are initialized and cloned. This
>            option can be given multiple times for pathspecs consisting of multiple entries. The
>            resulting clone has submodule.active set to the provided pathspec, or "." (meaning all
>            submodules) if no pathspec is provided.
> 
>            Submodules are initialized and cloned using their default settings. This is equivalent to
>            running git submodule update --init --recursive <pathspec> immediately after the clone is
>            finished. This option is ignored if the cloned repository does not have a
>            worktree/checkout (i.e. if any of --no-checkout/-n, --bare, or --mirror is given)
> ```

RyanGlScott:
Tricky. I see why there are two differently named options now—thanks for teaching me something!

Also, the CI failures aren't your fault—there's a slight misstep in the way the CI signs binaries on forks. I'll open a separate issue for that.

PR #1032, opened 2022-08-30
CI: Sign binaries in separate step, not run on forks

### LINKED ISSUE #1031: CI: Don't sign binary artifacts on PRs from forks
When a pull request is submitted from a fork, it will fail when attempting to sign a binary. See [here](https://github.com/GaloisInc/crucible/runs/8093168017?check_suite_focus=true#step:16:49) for an example:

```
+ sign crux-mir-0.6.0.99-ubuntu-22.04-x86_64.tar.gz
+ set +x
gpg: no valid OpenPGP data found.
```

This is because the OpenPGP secrets are only provided when running PRs from the main `crucible` repo, not from forks. To avoid this, we should skip the signing step when running a PR from a fork, much like how we do in Cryptol (see https://github.com/GaloisInc/cryptol/pull/1409) and `saw-script` (see https://github.com/GaloisInc/saw-script/commit/eedcba1126f7c501ae443ee3fb5b8b07358acbe1 and https://github.com/GaloisInc/saw-script/commit/ac082b0dd95b966a006a26de6e6564b2ffe523e3). This might require some slight reorganization of `.github/ci.sh` to decouple the signing step from the rest of `zip_dist`.

RyanGlScott:
Fixes #1031.


.github/workflows/crux-llvm-build.yml
```diff
@@ -206,13 +206,22 @@ jobs:
...
+        run: |
+          NAME="crux-llvm-${{ needs.outputs.outputs.crux-llvm-version }}-${{ matrix.os }}-x86_64"
+          echo "NAME=$NAME" >> $GITHUB_ENV
+          .github/ci.sh bundle_crux_llvm_files
         if: github.repository == 'GaloisInc/crucible'
         env:
           OS_TAG: ${{ matrix.os }}
+          VERSION: ${{ needs.outputs.outputs.crux-llvm-version }}
+
+      - if: github.event.pull_request.head.repo.fork == false
```
kquick, line 218: I think it would be a good idea to use `- name: Create Binary Bundle (Docker?)` and `- name: Sign bundle` to introduce these steps and add a comment line next to the `if: ` indicating why we only sign bundles on the head repo and not forks.

[In general, I'm a favor of having a `name` for *all* steps that is informative about what the step is trying to accomplish, but naming them all is outside the scope of this PR.]

.github/workflows/crux-mir-build.yml
```diff
@@ -136,13 +136,22 @@ jobs:
         run: .github/ci.sh test crux-mir
 
       - shell: bash
-        run: .github/ci.sh bundle_crux_mir_files
+        run: |
+          NAME="crux-mir-${{ needs.outputs.outputs.crux-mir-version }}-${{ matrix.os }}-x86_64"
+          echo "NAME=$NAME" >> $GITHUB_ENV
```
kquick, line 141: It's unlikely to happen here, but in the spirit of full safety I've learned that `declare -p NAME | cut -d' ' -f2-` is usually better than `echo "NAME=$NAME"` because the former will escape things like double-quotes in the value of NAME.

Same comment for the other yaml file.

.github/workflows/crux-mir-build.yml
```diff
@@ -136,13 +136,22 @@ jobs:
...
+          VERSION: ${{ needs.outputs.outputs.crux-mir-version }}
+
+      - if: github.event.pull_request.head.repo.fork == false
+        shell: bash
+        env:
           SIGNING_PASSPHRASE: ${{ secrets.SIGNING_PASSPHRASE }}
           SIGNING_KEY: ${{ secrets.SIGNING_KEY }}
-          VERSION: ${{ needs.outputs.outputs.crux-mir-version }}
+        run: |
+          .github/ci.sh sign ${NAME}.tar.gz
```
kquick, line 154: And here, defensively add double-quotes to specify `"${NAME}.tar.gz"` just in case NAME's value contains spaces or other things that would create issues.

Same comment for the other yaml file.

kquick (approved):
A couple of safety/useability suggestions, but otherwise it looks good.

.github/workflows/crux-mir-build.yml
RyanGlScott: `f3` seems to do what I want, so I've opted for that instead.

RyanGlScott: Hm, that's [still not quite right](https://github.com/GaloisInc/crucible/runs/8115069198?check_suite_focus=true#step:22:23):

```
+ sign '"crux-llvm-0.6.0.99-ubuntu-20.04-x86_64".tar.gz'
+ set +x
gpg: inserting ownertrust of 6
gpg: using "07FC90D0C93F5F3B56354E1DCBF4BB4143DB4A3A" as default secret key for signing
gpg: can't open '"crux-llvm-0.6.0.99-ubuntu-20.04-x86_64".tar.gz': No such file or directory
gpg: signing failed: No such file or directory
```

PR #1034, opened 2022-08-31
uc-crux-llvm: Use Natural instead of Integer where appropriate

### LINKED ISSUE #692: uc-crux-llvm: Turn some Ints into Naturals
During review of #673 it was suggested that a few `Int`s should be turned into `Natural`s for the sake of expressing the invariant that they're non-negative, and for the more clear semantics of an arbitrary-precision type.

- https://github.com/GaloisInc/crucible/pull/673#discussion_r596551972
- https://github.com/GaloisInc/crucible/pull/673#discussion_r596564855
- https://github.com/GaloisInc/crucible/pull/673#discussion_r596571428

langston-barrett:
Fixes #692. Other uses of types like Int were audited, but I determined that
they are necessary, because they are used to interact with APIs like
Data.Sequence, and so promoting them would just result in unsafe downcasts
elsewhere.

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/FullType/Memory.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/FullType/Memory.hs
@@ -24,6 +24,7 @@ where
 {- ORMOLU_DISABLE -}
 import           Control.Lens ((^.))
 import           Data.BitVector.Sized (mkBV)
+import           Numeric.Natural (Natural)
 
 import           Data.Parameterized.NatRepr (NatRepr, type (+))
 import qualified Data.Parameterized.Vector as PVec
@@ -35,7 +36,7 @@ import qualified What4.Interface as What4
 import qualified Lang.Crucible.Backend as Crucible
 
 -- crucible-llvm
-import           Lang.Crucible.LLVM.Bytes (bytesToInteger)
+import           Lang.Crucible.LLVM.Bytes (bytesToNatural)
 import           Lang.Crucible.LLVM.Extension (ArchWidth)
 import qualified Lang.Crucible.LLVM.MemModel as LLVMMem
 import           Lang.Crucible.LLVM.MemType (memTypeSize)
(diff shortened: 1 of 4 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Setup/Monad.hs
```diff
@@ -276,7 +277,7 @@ malloc ::
   -- | Path to this pointer
   Selector m argTypes inTy ('FTPtr atTy) ->
   -- | Size, as in number of elements. Should be strictly positive.
```
travitch, line 279:
```suggestion
  -- | Size, as in number of elements
```

langston-barrett:
Also requesting review from @badadam for pedagogical purposes. Adam, here are a few thoughts on this PR that may be enlightening about Haskell generally:

- Haskell has many numeric types.
  - `Int` is for signed, word-sized integers (like C's `int`).
  - `Word` is for unsigned, word-sized integers (like C's `unsigned int`), but some people (including myself) consider that the name "word" implies a lower-level view of the type as a bunch of bits, rather than a high-level view of it as an unsigned intger. If I were to need a type that really was a unsigned, word-sized integer, I would consider using a `newtype`.
  - `Integer` is an *unbounded* signed integral type.
  - `Natural` is an *unbounded* unsigned integral type, operations like subtraction are partial (and throw exceptions that can only be caught in the `IO` monad).
- For whatever technical and/or historical reason, many standard Haskell APIs, including those for standard "container" types like `Map`, `[]` (list), `Seq`, `Set`, and `Vector` use `Int` for things like indices and sizes.
- Bounded types like `Word` and `Int` are subject to over- and under-flow.
- Downcasting from unbounded types like `Natural` and `Integer` to bounded types like `Word` and `Int` is unsafe - it might overflow.
- Casting from a signed type to `Natural` can throw an exception.
- Downcasting from unsigned integral types to signed ones (`Word` to `Int`) can be unsafe in the same way that casting a `unsigned int` to an `int` is unsafe in C.
- The function `fromIntegral` can have a lot of types, and can be used for unsafe downcasts.

      ghci> (fromIntegral :: Word -> Int) 0
      0
      ghci> (fromIntegral :: Int -> Integer) 0
      0
      ghci> (fromIntegral :: Integer -> Natural) (-1)
      *** Exception: arithmetic underflow

Hence, this PR:

- I previously used `Integer` where I could have used `Natural`, since all the values are intended to be non-negative. Using `Natural` is more semantically clear, and is safe in this instance because no downcasting or unsafe conversion was required.
- However, there are other places where I used `Int`, and it turned out to be necessary: those `Int`s flowed to use of `Data.Sequence.Seq`, and if I had replaced them with `Natural` I would have had to downcast. Basically, changing `Int` to `Natural` may have been more semantically clear, but it would have just shifted around some unsafety.

uc-crux-llvm/src/UCCrux/LLVM/FullType/Memory.hs
```diff
@@ -52,21 +53,24 @@ import           UCCrux.LLVM.FullType.Type (FullTypeRepr, ModuleTypes, dataLayou
...
   FullTypeRepr m ft ->
   -- | Array length
-  Integer ->
-  Integer
+  Natural ->
+  Natural
 arraySizeInBytes mts ftRepr size = size * sizeInBytes mts ftRepr
 
+natToInt :: Natural -> Integer
+natToInt = fromIntegral  -- safe
```
travitch, line 72:
```suggestion
natToInt = toInteger  -- safe
```

We generally prefer `toInteger` when it fits because it is less polymorphic and always safe. Potential future changes (unlikely here) could cause noisy breakage, which is ideal.

travitch (approved):
One real suggestion in there

uc-crux-llvm/src/UCCrux/LLVM/Setup/Monad.hs
langston-barrett: I think this comment still applies - we don't want to `malloc(0)`.

uc-crux-llvm/src/UCCrux/LLVM/FullType/Memory.hs
langston-barrett: Thanks for the tip!

Issue #907, opened 2021-11-02
uc-crux-llvm: False positives on skipped functions like `getline`


langston-barrett:
Consider this program:
```c
#include <stdio.h>

void do_getline(FILE *f) {
  char loc[256];
  char *line = NULL;
  size_t linesize = 0;
  while (getline(&line, &linesize, f) != -1) {
    printf("%c", line[0]);
  }
}
```
UC-Crux-LLVM thinks there's probably a bug here:
```
[UC-Crux-LLVM] Read from data that concretely wasn't a pointer at do_getline.c:8:5
```
Why? Well, `getline` is a POSIX function. From `getline(3)`:
```
NAME
       getline, getdelim - delimited string input

SYNOPSIS
       #include <stdio.h>
       ssize_t getline(char **restrict lineptr, size_t *restrict n, FILE *restrict stream);

[...]
DESCRIPTION
       getline()  reads  an entire line from stream, storing the address of the
       buffer containing the text into *lineptr.  The buffer is null-terminated
       and includes the newline character, if one was found.

       If *lineptr is set to NULL before the call, then getline() will allocate
       a buffer for storing the line.  This buffer should be freed by the  user
       program even if getline() failed.
```
But UC-Crux doesn't have an override for it, so it just gets skipped by default. This leaves `line` with its original value (`NULL`), leading to a concrete null pointer dereference.

The fix is that the heuristic involving concretely-non-pointer dereferences should be disabled if any functions were skipped during execution.

langston-barrett:
The mechanism that fixed #982 (documented [here](https://github.com/GaloisInc/crucible/blob/555acf77477a02c8b0e101c2c5fcdb9a88338382/uc-crux-llvm/doc/specs.md)) may (eventually, with more work) be useful in such situations: the user could provide specs describing how `getline` writes to its arguments.

langston-barrett:
With #1003, you can provide a spec for such a function. With #1036, UC-Crux will warn you that the result may be a false positive. I think we can consider this fixed.

PR #1036, opened 2022-09-01
uc-crux-llvm: Track more sources of unsoundness

### LINKED ISSUE #932: uc-crux-llvm: Track sources of unsoundness for bug-finding
UC-Crux has a type `Unsoundness` for tracking sources of unsoundness for verification. Safety claims are then appropriately qualified by noting the sources of unsoundness encountered during simulation. However, we're lacking the dual qualification for claims that a program is buggy. Such reports should note any sources of unsoundness for bug-finding (i.e., lack of underapproximation of actual program semantics) that could lead to false/infeasible bugs (such as external functions that were skipped).

langston-barrett:
Fixes #932, #907. The upshot is that in situations like #907, UC-Crux will faithfully report conditions that may lead it to claim that there is a bug when, in fact, there isn't. Here's the new output for the test case in that ticket:
```
[UC-Crux-LLVM] Results for do_getline
[UC-Crux-LLVM] Found likely bugs
[UC-Crux-LLVM] In addition to any assumptions listed above, the following sources of unsoundness may invalidate this claim that there is a bug:
[UC-Crux-LLVM] The following unsound overrides (built-in functions) were used:
[UC-Crux-LLVM]   Execution of the following functions was skipped:
[UC-Crux-LLVM]   - getline
[UC-Crux-LLVM]   The following unsound specifications were applied:
[UC-Crux-LLVM] Read from data that concretely wasn't a pointer at getline.c:8:5
```

### DIFF
--- a/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
@@ -95,9 +95,9 @@ specOverrides ::
   (?memOpts :: MemOptions) =>
   ModuleContext m arch ->
   bak ->
-  -- | Track any unsound specs used
+  -- | Track any unsound specs used. See Note [IORefs].
   IORef (Set SpecUse) ->
-  -- | Origins of created values
+  -- | Origins of created values. See Note [IORefs].
   IORef (ExprTracker m sym argTypes) ->
   -- | Specs of each override, see 'Specs'.
   Map (FuncSymbol m) (SomeSpecs m) ->
@@ -155,9 +155,9 @@ createSpecOverride ::
   (fs ~ 'FS.FuncSig va mft args) =>
   ModuleContext m arch ->
   bak ->
-  -- | Track any unsound specs used
+  -- | Track any unsound specs used. See Note [IORefs].
   IORef (Set SpecUse) ->
-  -- | Origins of created values
+  -- | Origins of created values. See Note [IORefs].
   IORef (ExprTracker m sym argTypes) ->
   -- | Function to be overridden
   FuncSymbol m ->
--- a/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
@@ -9,9 +9,6 @@ Stability    : provisional
 These overrides are useful for describing the behavior of external/library
 functions, or for soundly skipping complex functions even when they happen to be
 defined.
-
-TODO(lb, #932): Track which specs actually execute and whether they are over- or
-under-approximate.
 -}
 
 {-# LANGUAGE DataKinds #-}
--- a/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
+++ b/uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
@@ -22,17 +22,26 @@ under-approximate.
 {-# LANGUAGE ScopedTypeVariables #-}
 
 module UCCrux.LLVM.Overrides.Spec
-  ( specOverrides,
+  ( SpecUse(..),
+    specOverrides,
     createSpecOverride,
   )
 where
 
 {- ORMOLU_DISABLE -}
 import           Control.Lens ((^.))
+import           Control.Monad.IO.Class (liftIO)
 import           Data.IORef (IORef)
+import qualified Data.IORef as IORef
 import           Data.Map (Map)
 import qualified Data.Map as Map
+import           Data.Semigroup (Max(Max, getMax))
+import           Data.Semigroup.Foldable (foldMap1)
+import           Data.Set (Set)
+import qualified Data.Set as Set
 import           Data.Type.Equality ((:~:)(Refl))
+
+import qualified Text.LLVM as L
 
 -- parameterized-utils
 import qualified Data.Parameterized.Context as Ctx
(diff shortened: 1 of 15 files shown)


uc-crux-llvm/src/UCCrux/LLVM/Run/Unsoundness.hs
```diff
@@ -39,39 +46,60 @@ data WithUnsoundness a = WithUnsoundness
...
+        , PP.pretty
+            "Execution of the following functions was skipped:" :
+              bullets
+                (map getSkipOverrideName (Set.toList (unsoundSkipOverridesUsed u)))
+        , PP.pretty
+            "The following unsound specifications were applied:" :
+              bullets specs
+        ]
   where
     bullets = map ((PP.pretty "-" PP.<+>) . PP.pretty)
```
kquick, line 78: Aesthetics opinion: it would be nice if `bullets` printed something like "[None]" if the argument was empty instead of having the header immediately followed by the next header.

uc-crux-llvm/src/UCCrux/LLVM/Run/Unsoundness.hs
```diff
@@ -39,39 +46,60 @@ data WithUnsoundness a = WithUnsoundness
...
+          unsoundSkipOverridesUsed u1 <> unsoundSkipOverridesUsed u2,
+        unsoundSpecsUsed = unsoundSpecsUsed u1 <> unsoundSpecsUsed u2
       }
 
 instance Monoid Unsoundness where
   mempty =
     Unsoundness
       { unsoundOverridesUsed = Set.empty,
-        unsoundSkipOverridesUsed = Set.empty
+        unsoundSkipOverridesUsed = Set.empty,
```
kquick, line 103: Minor suggestion: using `mempty` avoids module prefixing and is robust against changing the underlying container type.  Occurs in lots of other places if you decide to make this change.

uc-crux-llvm/src/UCCrux/LLVM/Soundness.hs
```diff
@@ -51,6 +54,14 @@ stringToSoundness =
     "indefinite" -> Just Indefinite
     _ -> Nothing
 
+soundnessToString :: Soundness -> String
+soundnessToString =
+  \case
+    Precise -> "precise"
+    Overapprox -> "overapprox"
+    Underapprox -> "underapprox"
```
kquick, line 62: Aesthetics opinion: if these strings are displayed to users, I'd recommend full rather than abbreviated text (e.g. "over-approximation" v.s. "overapprox".

Is there a test somewhere that `stringToSoundness . soundnessToString == id` or is this not a needed invariant?

kquick (approved):
Based on a general review, this looks good.  I had some minor comments, and haven't attempted to run and verify this, but it looks reasonable to me.

uc-crux-llvm/src/UCCrux/LLVM/Run/Unsoundness.hs
langston-barrett: > avoids module prefixing

I don't see this as a negative thing, is this an aesthetic preference?

> is robust against changing the underlying container

Agreed, though changing to `mempty` countervails [my preference to make things minimally polymorphic](https://langston-barrett.github.io/style/haskell.html) for the sake of explicitness and readability. I think `mempty` vs. `Set.empty` is pretty low-stakes, but I'll probably just keep it as is for now.

uc-crux-llvm/src/UCCrux/LLVM/Soundness.hs
langston-barrett: > Aesthetics opinion: if these strings are displayed to users, I'd recommend full rather than abbreviated text (e.g. "over-approximation" v.s. "overapprox".

Ah, these strings are *written by* users. I chose these abbreviations because they're fairly unambiguous. The corresponding printing doesn't use these abbreviations.

> Is there a test somewhere that stringToSoundness . soundnessToString == id or is this not a needed invariant?

Documented. Since they're right next to each other and only four constructors, I'll skip a test for now...

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
```diff
@@ -63,13 +72,23 @@ import           UCCrux.LLVM.ExprTracker (ExprTracker)
...
+import           UCCrux.LLVM.Module (FuncSymbol, funcSymbol, getFuncSymbol)
+import           UCCrux.LLVM.Newtypes.FunctionName (FunctionName, functionNameFromString)
 import           UCCrux.LLVM.Overrides.Polymorphic (PolymorphicLLVMOverride, makePolymorphicLLVMOverride)
+import           UCCrux.LLVM.Soundness (Soundness)
 import qualified UCCrux.LLVM.Specs.Apply as Spec
 import           UCCrux.LLVM.Specs.Type (SomeSpecs)
 import qualified UCCrux.LLVM.Specs.Type as Spec
 {- ORMOLU_ENABLE -}
 
+-- | A spec with this soundness was used in place of this function
```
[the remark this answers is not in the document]
langston-barrett: > "A user-supplied spec with the soundness specified here was used in place of analyzing the function directly." 

Applied, thanks for the suggestion!

> Maybe some expanded module-level docs would help here.

Done, the module-level Haddock now refers to the user-facing documentation for this feature, which motivates and explains the notion of "spec".

uc-crux-llvm/src/UCCrux/LLVM/Overrides/Spec.hs
```diff
@@ -79,16 +98,18 @@ specOverrides ::
   (?memOpts :: MemOptions) =>
   ModuleContext m arch ->
   bak ->
+  -- | Track any unsound specs used
+  IORef (Set SpecUse) ->
```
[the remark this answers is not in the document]
langston-barrett: [Done, and documented elsewhere too.](https://github.com/GaloisInc/crucible/pull/1036/commits/56ba530c7aace3abbb8168833d22a969750fe90c)

uc-crux-llvm/src/UCCrux/LLVM/Run/Unsoundness.hs
langston-barrett: Done, good suggestion!

PR #1043, opened 2022-09-13
wasm: Test suite

### LINKED ISSUE #1041: `crucible-wasm` test suite
As Rob so eloquently put it in #282:

> Let's have one!

langston-barrett:
Fixes #1041.


crucible-wasm/test/Test.hs
```diff
@@ -0,0 +1,153 @@
...
+import qualified Crux
+
+import qualified Lang.Crucible.Wasm.Main as C
+
+
+cube :: TS.CUBE
+cube =
+  TS.mkCUBE
+  { TS.inputDirs = ["test-data/"]
+  , TS.rootName = "*.txt"
```
RyanGlScott, line 38: If I'm reading [this part of the WebAssembly spec](https://webassembly.github.io/spec/core/text/conventions.html) correctly, the recommended file extension for WebAssembly code is `.wat`. We should probably use that here as well.

crucible-wasm/test/Test.hs
```diff
@@ -0,0 +1,153 @@
...
+                    ]
+                    -- Highly simplistic "diff" output assumes
+                    -- correlated lines: added or removed lines just
+                    -- cause everything to shown as different from
+                    -- that point forward.
+                  , concatMap dl $ zip el al
+                  , concatMap de $ drop (length al) el
+                  , concatMap da $ drop (length el) al
+                  ]
+    assertFailure $ BSC.unpack (BSC.unlines details)
```
RyanGlScott, line 86: It might be worth factoring out some of this code into a `crux-test` library so that we don't have to duplicate this code. I think it's _just_ involved enough to where having to maintain it in two separate places would become annoying.

crucible-wasm/src/Lang/Crucible/Wasm/Main.hs
```diff
@@ -0,0 +1,103 @@
...
+mainWithOptions (cruxOpts, wasmOpts) =
+  do res <- Crux.runSimulator cruxOpts (simulateWasm cruxOpts wasmOpts)
+     Crux.postprocessSimResult True cruxOpts res
+
+mainWithOutputConfig ::
+  (Maybe OutputOptions -> Crux.OutputConfig Log.CruxLogMessage) ->
+  IO ExitCode
+mainWithOutputConfig mkOutCfg =
+  Crux.withCruxLogMessage $ do
+    Crux.loadOptions mkOutCfg "crux-llvm" version cruxWasmConfig mainWithOptions
```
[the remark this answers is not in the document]
langston-barrett: Yeah, same thing for `crucible-jvm`... Will rename to `crucible-wasm` for now. Thanks for the catch!

crucible-wasm/test/Test.hs
langston-barrett: I agree. Would you like to see that done in this PR, or in a follow-up?

RyanGlScott: I'm fine with leaving that as follow-up work.

langston-barrett: https://github.com/GaloisInc/crucible/issues/1044

RyanGlScott (approved):

PR #1046, opened 2022-09-28
RFC: crucible-llvm: Don't represent architecture at the type level

langston-barrett:
It's necessary to represent the pointer width at the type level in order to statically prevent the formation of ill-typed Crucible/What4 expressions. However, there is no similar motivation for representing architectures at the type level.

The previous state of affairs was incoherent in the sense that we always provided an `ArchRepr` signaling the use of x86, without actually checking that this architecture was in use:

https://github.com/GaloisInc/crucible/blob/041975a5ad4200633613ed30df144c3a9d0e9403/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Monad.hs#L113

We could have fixed this by checking the target triple (functionality recently added to llvm-pretty{,-bc-parser}) and adding constructors for more architectures to `ArchRepr`, but as discussed above there's no motivation to actually do this.

If we want to move forward with this, we should make a follow-up issue for removing all of the `LLVMArch` stuff.

### DIFF
--- a/crucible-llvm/crucible-llvm.cabal
+++ b/crucible-llvm/crucible-llvm.cabal
@@ -95,6 +95,7 @@ library
   other-modules:
     Lang.Crucible.LLVM.Errors.Standards
     Lang.Crucible.LLVM.Extension.Arch
+    Lang.Crucible.LLVM.Extension.PtrSize
     Lang.Crucible.LLVM.Extension.Syntax
     Lang.Crucible.LLVM.Intrinsics.Common
     Lang.Crucible.LLVM.Intrinsics.Libc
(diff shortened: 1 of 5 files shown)


kquick:
I would like to defer a decision on this until I've had a chance to refresh on the unmerged polyglot work (which needs to be merged, but had type-level impact) to see if this is used at the type level for polyglot.

langston-barrett:
> I would like to defer a decision on this until I've had a chance to refresh on the unmerged polyglot work (which needs to be merged, but had type-level impact) to see if this is used at the type level for polyglot.

That sounds fine to me. Could you open an issue with a description of this upcoming feature and a pointer to the implementation as it stands today, and assign it to yourself? That context would help me understand the impact of potential changes like this while that work is pending.

langston-barrett:
@kquick Any updates on the polyglot work?

langston-barrett:
@kquick Any updates on this?

langston-barrett:
@kquick Are there still plans to merge the polyglot work?

langston-barrett:
I'm not sure this is such a great idea. I could see us running into a situation in the future where we would want to parameterize behavior on the architecture, e.g., of target-specific LLVM intrinsics like `@llvm.x86.avx512.*`.

Issue #1053, opened 2022-11-14
`Pair TypeRepr` is needlessly used in various places


RyanGlScott:
`Pair TypeRepr` is used in various spots in `crucible-syntax` and `crucible-concurrency`:

```
$ git grep -n -e "Pair.*TypeRepr" .
crucible-concurrency/src/Cruces/CrucesMain.hs:131:                     , [Pair C.TypeRepr GlobalVar]
crucible-concurrency/src/Cruces/ExploreCrux.hs:61:           , [Pair C.TypeRepr C.GlobalVar]
crucible-concurrency/src/Crucibles/Explore.hs:90:  [Pair C.TypeRepr GlobalVar]  ->
crucible-concurrency/src/Crucibles/Explore.hs:104:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:163:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:216:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:237:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:305:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:343:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:365:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:392:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:430:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:663:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:675:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:691:  [Pair C.TypeRepr GlobalVar] ->
crucible-concurrency/src/Crucibles/Explore.hs:703:  [Pair C.TypeRepr GlobalVar] {-^ Program globals that we might want to inspect -} ->
[...]

PR #1055, opened 2022-11-14
`crucible-syntax`: Add support for `extern`s

### LINKED ISSUE #1054: `crucible-syntax`: Support global variables defined outside of a `.cbl` file
`crucible-syntax`'s `defglobal` feature can be used to define global variables for use within the current `.cbl` file. Currently, there is no way to reference a global variable that is defined _outside_ of the current `.cbl` file, however. This would be useful for situations where global variables from one `.cbl` file is shared among other `.cbl` files.

The function call mechanism already supports something like this: `(declare ...)` can be used to reference functions defined externally. I propose adding a corresponding feature for global variables, which I'll tentatively call `(extern ...)`. `(extern ...)` would be to `(defglobal ...)` as `(declare ...)` would be to `(defun ...)`. `(extern ...)` would have the same syntax as `(defglobal ...)`, e.g.,

```racket
(extern $$global-defined-elsewhere Integer)
```

The rest of the `.cbl` file would be able to use `global-defined-elsewhere` just like any other global variable. The difference between `(extern ...)` and `(defglobal ...)` is that by the time the `.cbl` file is simulated, the value of `global-defined-elsewhere` might have already been set elsewhere, whereas a global arising from `(defglobal ...)` would be undefined at that point.

Implementation-wise, each `(extern ...)` would give rise to a fresh `GlobalVar` that is not yet inserted into the simulator's `SymGlobalState`. The responsibility is on the caller to ensure that each `GlobalVar` is inserted into the `SymGlobalState` with an appropriate `RegValue`; otherwise, attempting to use the `extern`'d variable will fail.

I am not wedded to the name `extern`. Another possible name is`declare-global`, which would make the analogy between `declare`/`defun` more obvious. At the same time, the name might be confusing: a `declare-global` isn't really a complete declaration on its own.

RyanGlScott:
`extern` is to `defglobal` what `declare` is to `defun`. This patch adds the necessary parts to `crucible-syntax` to parse `extern`s, create fresh `GlobalVar`s for each `extern`, and put them into the `ParsedProgram` API so that callers can insert externs into the `SymGlobalState` with appropriate `RegValue`s. I've included a test case that demonstrates this in action.

Fixes #1054.

### DIFF
--- a/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
@@ -177,9 +177,8 @@ data ParserHooks ext = ParserHooks {
        , ?parserHooks :: ParserHooks ext
        -- ParserHooks instance to use recursively when parsing.
        )
-    => m (Pair TypeRepr (Atom s))
-    -- ^ A pair containing a type and an atom of that type from evaluation of
-    -- syntax extension.
+    => m (Some (Atom s))
+    -- ^ The atom computed from evaluating the syntax extension.
 }
 
 -- | A ParserHooks instance that adds no extensions to the crucible-syntax
@@ -189,12 +188,12 @@ defaultParserHooks = ParserHooks empty empty
 
 -- | The results of parsing a program.
 data ParsedProgram ext = ParsedProgram
-  { parsedProgGlobals :: Map GlobalName (Pair TypeRepr GlobalVar)
+  { parsedProgGlobals :: Map GlobalName (Some GlobalVar)
     -- ^ The parsed @defglobal@s.
-  , parsedProgExterns :: Map GlobalName (Pair TypeRepr GlobalVar)
-    -- ^ For each parsed @extern@, map its name to its type and global variable.
-    --   It is the responsibility of the caller to insert each global variable
-    --   into the 'SymGlobalState' alongside an appropriate 'RegValue'.
+  , parsedProgExterns :: Map GlobalName (Some GlobalVar)
+    -- ^ For each parsed @extern@, map its name to its global variable. It is
+    --   the responsibility of the caller to insert each global variable into
+    --   the 'SymGlobalState' alongside an appropriate 'RegValue'.
   , parsedProgCFGs :: [ACFG ext]
     -- ^ The CFGs for each parsed @defun@.
   , parsedProgForwardDecs :: Map FunctionName SomeHandle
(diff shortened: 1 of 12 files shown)


crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -191,6 +191,10 @@ defaultParserHooks = ParserHooks empty empty
 data ParsedProgram ext = ParsedProgram
   { parsedProgGlobals :: Map GlobalName (Pair TypeRepr GlobalVar)
     -- ^ The parsed @defglobal@s.
+  , parsedProgExterns :: Map GlobalName (Pair TypeRepr GlobalVar)
```
RyanGlScott, line 194: As I note in #1053, I'm not quite sure how I feel about the use of `Pair TypeRepr GlobalVar` here, as it really ought to be `Some GlobalVar` instead. On the other hand, `parsedProgGlobals` already uses `Pair TypeRepr GlobalVar`, and the types of `parsedProgGlobals` and `parsedProgExterns` really ought to be the same. I suppose we could fix #1053 now and change `Pair TypeRepr` to `Some` everywhere, but that would be an unforced API change.

bboston7: I agree that the `Some` approach is cleaner, and I'm also sympathetic to not wanting to change the API without a good reason.  That being said, I don't think there are many `crucible-syntax` users. @langston-barrett is the only other person I'm aware of who may be using it, so maybe they can chime in with whether this would be painful for them to accommodate?  Is there anyone else we should check with who might be impacted by changing the type?

I'm happy to go either way on this one.

bboston7 (approved):

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
RyanGlScott: I went ahead and pushed a fix for #1053 as a separate commit. If folks don't want API churn, I can always revert that commit.

langston-barrett: > @langston-barrett is the only other person I'm aware of who may be using it, so maybe they can chime in with whether this would be painful for them to accommodate?

I'm not using it!

RyanGlScott: Ah, good to know. In that case, I will be bold and change the use of `Pair TypeRepr` to `Some`, since I expect the disruption to be minimal.

PR #1056, opened 2022-11-15
crucible-llvm: Show function name in override panic messages

langston-barrett:
I'm hitting this panic and I'd like to know why!

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
@@ -139,53 +139,58 @@ newtype ValTransformer p sym tp tp' =
 
 transformLLVMArgs :: forall m p sym bak args args'.
   (IsSymBackend sym bak, Monad m, HasLLVMAnn sym) =>
+  FunctionName ->
   bak ->
   CtxRepr args' ->
   CtxRepr args ->
   m (ArgTransformer p sym args args')
-transformLLVMArgs _ Ctx.Empty Ctx.Empty =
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -139,53 +139,58 @@ newtype ValTransformer p sym tp tp' =
 
 transformLLVMArgs :: forall m p sym bak args args'.
   (IsSymBackend sym bak, Monad m, HasLLVMAnn sym) =>
+  FunctionName ->
```
RyanGlScott, line 142: It would be worth stating in the Haddocks that this `FunctionName` is only used for error-message purposes.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -139,53 +139,58 @@ newtype ValTransformer p sym tp tp' =
...
+transformLLVMArgs fnName _ _ _ =
   panic "Intrinsics.transformLLVMArgs"
-    [ "transformLLVMArgs: argument shape mismatch!" ]
+    [ "transformLLVMArgs: argument shape mismatch!" 
+    , "in function: " ++ Text.unpack (functionName fnName)
+    ]
 
 transformLLVMRet ::
   (IsSymBackend sym bak, Monad m, HasLLVMAnn sym) =>
+  FunctionName ->
```
RyanGlScott, line 165: Similarly, state in the Haddocks that this `FunctionName` is only used for error-message purposes.

RyanGlScott (approved):

PR #1066, opened 2023-03-13
Split out `crucible-mir` from `crux-mir`

### LINKED ISSUE #1065: Split out `crucible-mir` from `crux-mir`
Currently, `crux-mir` is a mega-library that contains both the code needed to (1) simulate MIR code with `crucible`, and (2) define a `crux` frontent for MIR. This arrangement is somewhat painful for code that wants to do (1) without needing (2), however. I propose that we:

* Create a new `crucible-mir` library, which exports most of the current code in today's `crux-mir`. This library would not have a dependency on `crux`.
* Change `crux-mir` to depend on `crucible-mir`, and have it export any additional functionality that requires `crux`.

One interesting case is [`generateMIR`](https://github.com/GaloisInc/crucible/blob/97be912e7a08699e90f19aba861872022e4ff46d/crux-mir/src/Mir/Generate.hs#L158-L182) in `Mir.Generate`, which runs `mir-json` on the input to generate JSON. This is really only needed for `crux-mir`'s purposes, so this function deserves to live in `crux-mir`, not `crucible-mir`. On the other hand, the `parseMir` and `translateMIR` functions in that module make sense to put in `crucible-mir`, so we should create a new module for them.

See GaloisInc/saw-script#1839 for the corresponding issue on the SAW side.

RyanGlScott:
For the most part, most code was moved wholesale from `crux-mir` with only minor changes (I removed redundant imports to more easily determine which library dependencies could be dropped in `crucible-mir`). The most unusual change was creating a new `Mir.ParseTranslate` module in `crucible-mir`, which contains the `parseMIR` and `translateMIR` functions previously defined in `Mir.Generate` (now a `crux-mir` module that only exports `generateMIR`).

Fixes #1065.


crux-mir/CHANGELOG.md
```diff
@@ -1,3 +1,21 @@
...
+
+* Much of the `crux-mir` library has been split off into a `crucible-mir`
+  library, which removes all of the `crux`-specific code. The following modules
+  were _not_ migrated to `crucible-mir`:
+
+  * `Mir.Generate`
+  * `Mir.Language`
+  * `Mir.Log`
+  * `Mir.Concurrency`
+  * `Mir.Overrides`
```
spernsteiner, line 13: At some point we may want to move certain overrides from `Mir.Overrides` into the crucible-mir library so we can call them from our patched version of the Rust standard library even when verifying with SAW.  But for now I think it's okay to leave it alone - we can move individual overrides later once we know which ones are actually necessary.

spernsteiner (approved):
Looks good to me.  Thanks for doing all this cleanup!

PR #1067, opened 2023-03-15
`crux`: Add `--offline-solver-output` flag

### LINKED ISSUE #1051: `crux`: Add flag for offline solver output
Currently, `crux` has an `--online-solver-output=FILE` option for logging SMT solver interactions in an online solver interaction. We are lacking a counterpart to this flag for offline solver interactions, however. We should add one. A first attempt would be something like this:

```diff
diff --git a/crux/src/Crux/Goal.hs b/crux/src/Crux/Goal.hs
index 4068bfb4..a52b1ec8 100644
--- a/crux/src/Crux/Goal.hs
+++ b/crux/src/Crux/Goal.hs
@@ -24,6 +24,7 @@ import qualified Data.Text as Text
 import           Data.Void
 import           Prettyprinter
 import           System.Exit (ExitCode(ExitSuccess))
+import           System.IO (IOMode(..), withFile)
 import qualified System.Timeout as ST
 
 import What4.Interface (notPred, getConfiguration, IsExprBuilder)
@@ -248,10 +249,10 @@ proveGoalsOffline adapters opts ctx explainFailure (Just gs0) = do
                  -> BoolExpr t
                  -> WS.SolverAdapter st
                  -> IO ([ProgramLoc], ProofResult sym)
-    runOneSolver p assumptionsInScope assumptions goal adapter = do
+    runOneSolver p assumptionsInScope assumptions goal adapter = withFile "solver-output.smt2" WriteMode $ \h -> do
       -- NOTE: We don't currently provide a method for capturing the output
       -- sent to offline solvers.  We would probably want a file per goal per adapter.
-      let logData = WS.defaultLogData
+      let logData = WS.defaultLogData { WS.logHandle = Just h }
       WS.solver_adapter_check_sat adapter (ctx ^. ctxSymInterface) logData [assumptions, goal] $ \satRes ->
         case satRes of
           Unsat _ -> do
```

Where `solver-output.smt2` is a dummy file name—we'd most likely want to put this behind an `--offline-solver-output` flag.

RyanGlScott:
The implementation details for `--offline-solver-output` differ a bit from those of `--online-solver-output` due to the fact that (1) offline solvers are invoked multiple times to handle multiple goals, and (2) users can specify multiple offline solvers. See `withOfflineSolverOutputHandle`.

Fixes #1051.

### DIFF
--- a/crux/src/Crux/Config/Common.hs
+++ b/crux/src/Crux/Config/Common.hs
@@ -174,6 +174,9 @@ data CruxOptions = CruxOptions
   , onlineSolverOutput       :: Maybe FilePath
     -- ^ The file to store the interaction with the online goal solver (if
     -- solving is being performed online)
+  , offlineSolverOutput      :: Maybe FilePath
+    -- ^ A template to use for files to store interactions with offline goal
+    -- solvers (if solving is being performed offline).
 
   , yicesMCSat               :: Bool
     -- ^ Should the MC-SAT Yices solver be enabled (disables unsat cores; default: no)
@@ -295,6 +298,10 @@ cruxOptions = Config
           onlineSolverOutput <-
             sectionMaybe "online-solver-output" stringSpec
             "The file to store the interaction with the online goal solver (if any) (default: none)"
+
+          offlineSolverOutput <-
+            sectionMaybe "offine-solver-output" stringSpec
+            (pack offlineSolverOutputHelp)
 
           simVerbose <-
             section "sim-verbose" numSpec 1
@@ -447,7 +454,7 @@ cruxOptions = Config
       , Option [] ["no-unsat-cores"]
         "Disable computing unsat cores for successful proofs"
         $ NoArg $ \opts -> Right opts { unsatCores = False }
-      
+
       , Option "n" ["get-abducts"]
         "Get these many abducts. Only works with cvc5, 0 otherwise."
         $ ReqArg "ABDUCTS"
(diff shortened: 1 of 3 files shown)


crux/src/Crux/Config/Common.hs
```diff
@@ -477,6 +490,16 @@ cruxOptions = Config
         "The file to store the interaction with the online goal solver (if any) (default: none)"
         $ ReqArg "FILE" $ \f opts -> Right opts { onlineSolverOutput = Just f }
 
+      , Option [] ["offline-solver-output"]
+        (unlines
+          [ "A template to use for files to store interactions with offline goal solvers"
+          , "(if any) (default: none). For example, if the template is `offline-output.smt2`,"
+          , "then each file will be named `offline-output-<goal number>-<solver name>.smt2,"
+          , "where <goal number> is the number of the goal that was proven (starting at 0)"
+          , "and <solver name> is the name of the solver used to dispatch that goal."
```
kquick, line 499: This text duplication makes me sad.

kquick (approved):

crux/src/Crux/Config/Common.hs
RyanGlScott: Yes, a good point. Luckily, these strings are only used within this one module, so it was straightforward to de-duplicate it. (I suppose we should de-duplicate the _other_ strings in this module, but I'm not going to do that here.)

PR #1068, opened 2023-03-17
Require `tasty-sugar-2.0` or greater and address deprecation

RyanGlScott:
The number of test cases reports by `tasty`'s `-l` flag is the same before and after this patch, so I am reasonably confident that `tasty-sugar-2.0` is finding the same test cases as before.

This requires bumping the `what4` submodule to bring in the changes from GaloisInc/what4#229, which also adds lower version bounds of `tasty-sugar >= 2.0` to its test suite.


kquick (approved):
It must be mind numbing to have to make so many identical changes to github actions yml files for the caching, but I'm definitely looking forward to the improvements!

RyanGlScott:
> It must be mind numbing to have to make so many identical changes to github actions yml files for the caching

Indeed, I have often wondered if it would be worthwhile to consolidate some of our GitHub Actions logic (which is no doubt copy-pasted uncountably many times in different repos) into our [`.github`](https://github.com/GaloisInc/.github) repo. Currently, there is only one action defined there, but we could in principle do much more with it. I haven't thought through the consequences of doing so yet (e.g., how would versioning all of these different actions work?), but it is a tempting idea nonetheless.

PR #1074, opened 2023-04-03
Match concrete size array

andreistefanescu:


crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1287,6 +1290,40 @@ loadRaw sym mem ptr valType alignment = do
...
+-- | Load an array with concrete size from memory.
+loadArrayConcreteSizeRaw ::
+  forall sym wptr .
+  (IsSymInterface sym, HasPtrWidth wptr, Partial.HasLLVMAnn sym, ?memOpts :: MemOptions) =>
+  sym ->
+  MemImpl sym ->
+  LLVMPtr sym wptr ->
+  Natural ->
+  Alignment ->
+  IO (Either (Pred sym) (Pred sym, SymArray sym (SingleCtx (BaseBVType wptr)) (BaseBVType 8)))
```
RyanGlScott, line 1302: What is the relationship between the `sz :: Natural` argument and the `SymArray` that is returned? I think this is set up so that the elements of the `SymArray` in the range `[0, sz)` correspond to the elements of the array stored in memory as bytes, and all other elements are set to byte `0`. Worth stating this somewhere in the Haddocks.

RyanGlScott (commented):
The general idea looks reasonable to me. Most of my comments concern documentation, which would help a lot in making this (fairly nuanced) code easier to understand.

RyanGlScott:
Superseded by #1178.

Issue #1077, opened 2023-04-12
Release `crux-llvm` and friends to Hackage


RyanGlScott:
We have a project that depends on `crux-llvm` that we would eventually like to release to Hackage. This issue serves to track preparation that we will need to do on the `crucible` side to prepare for such a release:

* [ ] Remove unused dependencies from `crux-llvm` and its dependencies
* [ ] Check the licenses of dependencies
* [ ] Release `crux-llvm` and dependencies to Hackage, making sure to include this in the PDI package release process in the future

langston-barrett:
This seems done? https://hackage.haskell.org/package/crux-llvm

PR #1078, opened 2023-04-12
Remove unused dependencies detected by `-Wunused-packages`

RyanGlScott:
This prunes unused dependencies caught by the `-Wunused-packages` in pursuit of an eventual Hackage release for `crux-llvm` (see #1077). One of the more noticeable dependencies that we were able to remove is `aig`, allowing us to shed an entire submodule.


kquick (approved):
Yay!

Issue #999, opened 2022-06-14
CI accidentally skips over `crux-llvm` tests on macOS


RyanGlScott:
Recently, I came to wonder why the macOS CI jobs weren't catching #885, which should prevent the `crux-llvm` tests from passing on macOS. It turns out this is because the CI isn't actually running the `crux-llvm` test suite on macOS at all! Take a look at what happens on a typical macOS job like [this one](https://github.com/GaloisInc/crucible/runs/6733176618?check_suite_focus=true#step:19:23):

```
+ /Users/runner/work/crucible/crucible/llvm/bin/clang-withIncl --version
clang-11: error: no input files
+ echo clang version unknown
clang version unknown
+ /Users/runner/work/crucible/crucible/llvm/bin/llvm-link --version
LLVM (http://llvm.org/):
+ cabal v2-test crux-llvm
  LLVM version 11.0.0
  Optimized build.
  Default target: x86_64-apple-darwin19.6.0
  Host CPU: ivybridge
Build profile: -w ghc-9.2.2 -O1
In order, the following will be built (use -v for more details):
 - crux-llvm-0.6.0.99 (test:crux-llvm-test) (first run)
<snip>
Running 1 test suites...
Test suite crux-llvm-test: RUNNING...
Test suite crux-llvm-test: PASS
Test suite logged to:
/Users/runner/work/crucible/crucible/dist-newstyle/build/x86_64-osx/ghc-9.2.2/crux-llvm-0.6.0.99/t/crux-llvm-test/test/crux-llvm-0.6.0.99-crux-llvm-test.log
1 of 1 test suites (1 of 1 test cases) passed.
```

The first suspicious thing here is that the whole test suite takes 12 seconds to run, which is far too short. The second suspicious thing is that `clang-withIncl --version` prints `clang-11: error: no input files` instead of the version number. This is a consequence of how the CI is setting up `clang` on macOS. It assembles a wrapper script named `clang-withIncl` here:

https://github.com/GaloisInc/crucible/blob/4158b6f696c0c88fd6b1b698ffea7aad8bee8597/.github/workflows/crux-llvm-build.yml#L173-L174

There's a mistake here, however: the invocation of `clang` doesn't pass through the arguments (e.g., with `${1+"$@"}`), so no matter what arguments you pass to the `clang-withIncl` script, it will simply invoke `clang` with no input files. We should change this script so that it plumbs through the arguments.

Issue #885, opened 2021-10-13
MacOS missing overrides for symbolic IO


robdockins:
I'm experiencing test suite failures on my local machine that do not occur on the CI runners. The overrides for, e.g., `write` are not found.  This appears to be because clang and/or libc on this system insist on adding a strange prefix to some (?) symbols.  In particular `write` is compiled into `\01_write` (yes, with a leading ASCII character with code `0x01`).  I suppose this is due to some MacOS/XCode nonsense which was added in some version bump between where the runners are and where my machine is (Version 11.5.1 (20G80)).

The obvious solution is to add overrides for both symbols, but the tricky part is understanding when/why this mangling gets done so we can figuring out which symbols need special treatment.

```
crux-llvm
  clang 11.1.0
    write-stdout.c
      z3
        loop
          clang11
            z3 4.8.10
              write-stdout
                write-stdout crux run:    OK (0.45s)
                write-stdout crux output: FAIL
                  test/Test.hs:234:
                  MISMATCH for expected (test-data/golden/stdio/write-stdout.good)
                             and actual (test-data/golden/stdio/write-stdout.z3.out) output:
                      F-expect>|output
                      F-actual>|[Crux] Found counterexample for verification goal
                      F-expect>|[Crux] Overall status: Valid.
                      F-actual>|[Crux]   test-data/golden/stdio/write-stdout.c:9:13: error: in main
                      F-actual>|[Crux]   Failed to load function handle
                      F-actual>|[Crux]   Details:
                      F-actual>|[Crux]     No implementation or override found for pointer: "\SOH_write"
                      F-actual>|[Crux]     The given pointer could not be resolved to a callable function
                      F-actual>|[Crux]     No implementation or override found for pointer: "\SOH_write"
                      F-actual>|[Crux]     Attempting to load callable function with type: i64(i32, i8*, i64)
                      F-actual>|[Crux]       Via pointer: Global symbol "\SOH_write" (1, 0x0:[64])
                      F-actual>|[Crux]     In memory state:
                      F-actual>|[Crux]       Stack frame main
                      F-actual>|[Crux]         No writes or allocations
                      F-actual>|[Crux]       Base memory
                      F-actual>|[Crux]         Allocations:
                      F-actual>|[Crux]           GlobalAlloc 6 0x26:[64] Immutable 1-byte-aligned [global variable  ] .str.1
[...]

PR #1083, opened 2023-04-20
Support up to LLVM 14

RyanGlScott:
This adds support for LLVM versions up to 14 by:

* Bumping the `llvm-pretty` submodule to bring in the changes from https://github.com/elliottt/llvm-pretty/pull/109.
* Bumping the `llvm-pretty-bc-parser` submodule to bring in the changes from https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/219.
* Updating the test output in `crux-llvm` and `uc-crux-llvm`'s test suites accordingly.

-----

In order to adapt to various test cases' output changing with Clang 13 and 14, I needed to do a minor rewrite of the way Clang version ranges are handled in `crux-llvm`'s test suite. See the `crux-llvm/test/README.md` file for a description of the new conventions that are currently in use.


langston-barrett (approved):

kquick (approved):

RyanGlScott:
Somewhat surprisingly, two test configurations fail with these `crux-llvm` test suite changes. Here are my diagnoses:

* macOS 12: My changes to the `crux-llvm` test suite evidently made things more permissive, as we now attempt to run the full test suite on macOS, even though that [we aren't correctly detecting the `clang` version on macOS](https://github.com/GaloisInc/crucible/issues/999). To be honest, I think I like this new behavior better, as we no longer silently discard all of the test cases. Still, we aren't quite equipped to run the `crux-llvm` test suite with macOS yet (see #885 and #999), so I think I will explicitly disable the `crux-llvm` test suite on the macOS CI job for now.
* Ubuntu 20.04: This is a strange one. This is a consequence of the fact that `clang --version` gives different output depending on whether you use Ubuntu 20.04 or 22.04. On Ubuntu 22.04, the `clang --version` output is nice and clean:

  ```
  Ubuntu LLVM version 12.0.1
  ```

  But on Ubuntu 20.04, it's messier:

  ```
  Ubuntu clang version 12.0.0-3ubuntu1~20.04.5
  ```

  The `versions` library has a very hard time parsing things containing tildes, as I've reported in https://github.com/fosskers/versions/issues/62. A consequence of this is that the `crux-llvm` test suite does not identify this Clang as being version 12, causing the wrong `.good` files to be picked.

  I'm not quite sure of the best way to fix this, but perhaps it would suffice to just drop any tilde tags from version numbers like the one above. It's a bit _ad hoc_, but then again, the version tag itself is _ad hoc_.

Issue #1075, opened 2023-04-10
Support opaque pointers in `crucible-llvm`


RyanGlScott:
After upgrading `llvm-pretty` and `llvm-pretty-bc-parser` to support LLVM's opaque `ptr` type (see https://github.com/elliottt/llvm-pretty/issues/102 and https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/177, respectively), we will also need to upgrade `crucible-llvm` and `uc-crux-llvm` to handle them.

## `crucible-llvm`

To a large extent, `crucible-llvm` is already designed to handle opaque pointers, as the `LLVMPtr` type intentionally does not include a pointee type. That being said, there are still various places in `crucible-llvm` that inspect pointee types of LLVM pointers, which will need to be changed to support the `ptr` type. Here are all the places that I have found:

* The `PtrType` data constructor for `MemType` includes a pointee type:

  https://github.com/GaloisInc/crucible/blob/ad4a553487eeb5c6bbb5abf4bde26af905bf0254/crucible-llvm/src/Lang/Crucible/LLVM/MemType.hs#L118

  In the spirit of having [opaque and non-opaque pointers co-exist](https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/177#issuecomment-1497512507), I think we will also need a corresponding `PtrOpaqueType` data constructor that does _not_ have a pointee type.
* Similarly, the LLVM override function parser machinery includes a non-opaque pointer type, but not an opaque pointer type:

  https://github.com/GaloisInc/crucible/blob/ad4a553487eeb5c6bbb5abf4bde26af905bf0254/crucible-llvm/src/Lang/Crucible/LLVM/QQ.hs#L48

  We will want to make the latter available when the user writes `ptr` in an `llvmOvr`.
* Speaking of overrides, we have quite a few overrides that are written with non-opaque-pointer types, such as the `memcpy` override:

  https://github.com/GaloisInc/crucible/blob/ad4a553487eeb5c6bbb5abf4bde26af905bf0254/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs#L86

  We could write a copy of each of these overrides that replaces non-opaque pointers with opaque ones (e.g., `memcpy(ptr, ptr, size_t)`, but this would be extremely tedious to maintain. Better yet would be to tweak the override matcher functionality to accept both opaque and non-opaque pointers alike. That is, in this code:

[...]

Issue #1044, opened 2022-09-13
Duplication across `crucible-wasm` and `crux-llvm` test suite


langston-barrett:
Both test suites have a function for diffing the output of golden tests, and there are more similarities besides. We should factor these out into a small library that they both can use.

RyanGlScott:
There is also some duplication between the test suites for `crux-llvm` and `uc-crux-llvm` after #1085, which copies over the `VersionCheck` machinery from the former into the latter. As a result, this task would now benefit three of the packages in this repo.

PR #1085, opened 2023-05-04
More complete support for opaque pointers (required for LLVM 15+)

RyanGlScott:
This is a collection of various tweaks to `crux-llvm` that allows it to fully support opaque pointers, an alternative representation of pointer types that is enabled by default in LLVM 15 or later. It also tweaks `uc-crux-llvm` to _not_ support opaque pointers for now, as adding support for them in `uc-crux-llvm` would be nontrivial. (See #1075 for more on this.) These changes suffice to make `crux-llvm` and `uc-crux-llvm`'s test suites pass with LLVM 15 and 16.

I have split this patch up into self-contained commits, and for reviewing purposes, I'd recommend viewing each patch in isolation. The general themes of these patches are:

## Adapting to `llvm-pretty` API changes (from https://github.com/elliottt/llvm-pretty/pull/110)

* `Adapt to Load gaining an explicit type`
* `Adapt to GEP/ConstGEP gaining explicit types`

## `crux-llvm` support for opaque pointers

* `crucible-llvm: Add PtrOpaqueType to MemType`
* `crucible-llvm: Make opaque pointers play nicely with overrides`
* `crucible-llvm: Avoid inspecting pointee types during translation`

## `uc-crux-llvm` non-support for opaque pointers

* `uc-crux-llvm: Opaque pointers are unsupported`

## Updating test cases to support Clang 15/16

* `uc-crux-llvm-test: Avoid -Wint-conversion errors with Clang 15+`
* `uc-crux-llvm-test: Avoid -Wimplicit-int errors with Clang 15+`
* `{uc,}-crux-llvm-test: Avoid -Wimplicit-function-declaration errors with Clang 15+`
* `crux-llvm-test: Accept new test output for Clang 15`
* `crux-llvm-test: Accept new test output for Clang 16`
* `{uc,}-crux-llvm: Document support for LLVM 16 in READMEs`

## General test suite/CI tweaks needed to support multiple LLVM ranges

* `Use tasty-sugar v2.2.0.0 for clang-range ranged parameter test filtering.`
* `Remove unneeded imports and functions in crux-llvm tests.`
* `CI: Regenerate cabal.GHC-*.config files`


crux-llvm/test/Test.hs
```diff
@@ -79,28 +88,12 @@ data VersionCheck = VC String (Either T.Text Versioning)
 showVC :: VersionCheck -> String
 showVC (VC nm v) = nm <> " " <> (T.unpack $ either id prettyV v)
 
-vcTag :: VersionCheck -> String
-vcTag v@(VC nm _) = nm <> vcMajor v
-
-vcMajor :: VersionCheck -> String
-vcMajor (VC _ v) = either T.unpack (^. major . to show) v
-
 vcVersioning :: VersionCheck -> Either T.Text Versioning
```
kquick, line 91: With the addition of the VersionCheck module, should this just import that module and use its capabilities instead of duplicating them here, or is there a reason that would be problematic that I'm not seeing?

kquick (approved):
Overall looks fine, just a couple of polishing suggestions.

langston-barrett:
I'm on vacation for the next two weeks, but I approve of the overall direction of this PR. Please feel free to merge without my review.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
```diff
@@ -152,37 +152,40 @@ instance Traversable GEPResult where
...
   [L.Typed L.Value] {- ^ index arguments -} ->
   m (GEPResult (L.Typed L.Value))
 
-translateGEP _ _ [] =
+translateGEP _ _ _ [] =
   throwError "getelementpointer must have at least one index"
 
-translateGEP inbounds base elts =
-  do mt <- liftMemType (L.typedType base)
+translateGEP inbounds baseTy base elts =
```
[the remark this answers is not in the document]
RyanGlScott: They are similar (but still distinct) concepts. The [LLVM Language Reference Manual](https://releases.llvm.org/16.0.0/docs/LangRef.html#getelementptr-instruction) refers to `baseTy` argument as the "type used as the basis for the calculations", and it refers to `base` as "the base address to start from". In this sense, these are both "base" concepts, but I can definitely see how the current code can be confusing.

I've renamed `base` to `basePtr` (which matches the internal naming conventions that LLVM's source code uses) and expanded the Haddocks for `translateGEP` to give a concrete example. Let me know if you find the new version easier to understand.

crux-llvm/test/Test.hs
kquick: Not worth it right now, but something to note for the future.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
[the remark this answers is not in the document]
kquick: That helps a bunch, thanks!

crux-llvm/test/Test.hs
RyanGlScott: I've left a note about this in https://github.com/GaloisInc/crucible/issues/1044#issuecomment-1535335352.

RyanGlScott:
I was surprised to see that the CVC5 abduction-related tests fail on CI, as I observed these tests passing locally. But some further investigation reveals why: I was using CVC5 1.0.5 locally, the CI was using CVC5 1.0.2, and [these tests are sensitive to the choice of CVC5 version](https://github.com/GaloisInc/crucible/issues/1086). Agh!

Issue #988, opened 2022-05-16
Trigger CI, build Docker images when release is tagged


RyanGlScott:
Currently, the CI does not trigger after a `git push --tags`, which means that the CI must be triggered manually for each release. This is especially annoying since the CI is the primary mechanism by which new Docker images are built, tagged, and pushed to ghcr.io, so it would be more convenient if this happened automatically after a `git push --tags`.

To make this happen, we should take inspiration from how SAW's CI does things. In particular:

* SAW has a regex that checks for `git` tag names (see [here](https://github.com/GaloisInc/saw-script/blob/7a8fc9a4b9aa6c35d2db98f2784738b45fcb8f5f/.github/workflows/ci.yml#L9-L11)). For Crux, we'd probably want something like:

  ```yaml
  on:
    push:
      tags: ["crux-v?[0-9]+.[0-9]+(.[0-9]+)?"]
  ```
* We need a mechanism for building Docker images via the CI. See #967, which is largely about the same idea.
* We need to make sure that the build step that generates Docker images runs when a new release is tagged. SAW's CI ([here](https://github.com/GaloisInc/saw-script/blob/7a8fc9a4b9aa6c35d2db98f2784738b45fcb8f5f/.github/workflows/ci.yml#L405)) has an `if: ... || needs.config.outputs.release == 'true'` guard for this purpose.

PR #1093, opened 2023-06-22
Prepare for Crux 0.7 release

RyanGlScott:


langston-barrett (approved):

kquick (commented):
I was expecting to see some changes to the version in the cabal files?

Also: not just Crux.

RyanGlScott:
> I was expecting to see some changes to the version in the cabal files?

This is a good point. Usually, the process for making releases of things like Cryptol and SAW is:

1. Make a release branch
2. On the release branch, bump the version number from a development version (e.g., `0.1.0.99`) to a release version (e.g., `0.2`)
3. Publish the release
4. On the main branch (e.g., `master`), increment the development version number (e.g., from `0.1.0.99` to `0.2.0.99`)

With this workflow, you'd only ever bump the version number for the main branch _after_ a release, which is why I haven't done so yet.

Having said that, the `crucible` repo is a bit weird in that none of the libraries use a development version (e.g., `crux-llvm` currently just uses `0.6`). Arguably, it should. I could either do so now, or I could accept the fact that we're not using the development version workflow in `crucible` and just bump the version numbers here. Do you have a preference as to which approach we use?

> Also: not just Crux.

Yes, this is true. I'm being a bit hand-wavy and lumping everything under the umbrella term "Crux release", since (1) the Crux tools are the only things in the repo that receive a full "product" release, and (2) the timing of Crux releases is a good excuse to bump the other version numbers. I suppose I could call this a "Crucible mega-release" or something like that if you'd like that better :)

kquick:
IIUC, your sequence would be enhanced with the following to cover this particular PR:

-1. Make a branch to update release information pre-release
0. Merge prep branch to master
1. Make a release branch
2. ...

If this is correct, I would suggest an update to this workflow:

1. Make a branch to update release information pre-release
2. Update CHANGELOGS, etc. *and* update the versions in the cabal files to the version being released.  As an independent and final commit on this branch, update the cabal file versions again, this time to the development version.
3. Merge prep branch to master
4. Generate/publish the release.  This would be done from the (now-merged) commit _just before_ the updates to the development versions from the branch in step 2 above.  A release branch isn't actually needed, unless there are additional release-related changes.
5. [If not done as part of step 4] tag the commit that corresponds to the release.
6. If a release branch was created, with additional changes, merge it back to master.

In this procedure, we can ensure that (a) the other updates for a release (CHANGELOGS, etc.) actually match a corresponding package version, (b) master contains an actual released version, and (c) no PRs commits get interspersed with these version updates.

Re: this being more than crux: I agree with doing them all, I was just pointing out it was larger than the name implied.  Not a big deal, but I found more than I was expecting. :-)

RyanGlScott:
> A release branch isn't actually needed, unless there are additional release-related changes.

This is the only thing I'd disagree with, if only for technical reasons: our CI is set up to start a workflow on release branches that, when finished, publishes new versions of our Docker images to ghcr.io. So I'd propose keeping the additional bureaucracy of maintaining a release branch to make deployment simpler.

Other than that, I'm aligned with your suggestion of doing the version bumps as part of the release prep PR. In the case of `crucible`, where we aren't adhering to the development version convention, I suppose that means I'd just need to add a single commit at the end. For something like Cryptol, on the other hand, you'd need one commit to bump the version to 3.0.0, and another commit after that to bump it to 3.0.0.99. Is that right?

kquick:
> > A release branch isn't actually needed, unless there are additional release-related changes.
> 
> This is the only thing I'd disagree with, if only for technical reasons: our CI is set up to start a workflow on release branches that, when finished, publishes new versions of our Docker images to ghcr.io. So I'd propose keeping the additional bureaucracy of maintaining a release branch to make deployment simpler.

I'd agree this is reasonable, although I don't believe that the CI for any of the crucible repos is doing this.  Establishing it as a general procedure to facilitate that is good though.

> Other than that, I'm aligned with your suggestion of doing the version bumps as part of the release prep PR. In the case of `crucible`, where we aren't adhering to the development version convention, I suppose that means I'd just need to add a single commit at the end. For something like Cryptol, on the other hand, you'd need one commit to bump the version to 3.0.0, and another commit after that to bump it to 3.0.0.99. Is that right?

Yup, that's right.

RyanGlScott:
> I'd agree this is reasonable, although I don't believe that the CI for any of the crucible repos is doing this.

Ugh, quite right. But hopefully it will [someday](https://github.com/GaloisInc/crucible/issues/988).

RyanGlScott:
Actually, `crux-llvm` and `crux-mir` _are_ using the development version convention (I just checked), so I'll add a separate commit to bump those to 0.7.0.99 as well.

PR #1094, opened 2023-06-26
Fix Docker issues uncovered during Crux 0.7 release

RyanGlScott:
The `Dockerfile`s used for the `crux-llvm` and `crux-mir` were slightly broken, but unfortunately, we do not currently test these on CI, so I was not aware of this until I was preparing the Crux 0.7. Once this lands, I will push these to the `release-crux-0.7` branch as well.

### DIFF
--- a/.github/Dockerfile-crux-mir
+++ b/.github/Dockerfile-crux-mir
@@ -21,7 +21,7 @@ RUN apt-get update && \
 COPY --from=mir_json /usr/local/cargo /usr/local/cargo
 COPY --from=mir_json /usr/local/rustup /usr/local/rustup
 WORKDIR /usr/local/bin
-RUN curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20220812/ubuntu-22.04-bin.zip"
+RUN curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20220902/ubuntu-22.04-bin.zip"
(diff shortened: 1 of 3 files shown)


langston-barrett (approved):

.github/Dockerfile-crux-mir
```diff
@@ -2,6 +2,8 @@ FROM rust:1.46.0 AS mir_json
 
 ADD dependencies/mir-json /mir-json
 WORKDIR /mir-json
+# Work around https://github.com/rust-lang/cargo/issues/10303
+ENV CARGO_NET_GIT_FETCH_WITH_CLI=true
```
kquick, line 6: Just double checking: this https://github.com/rust-lang/cargo/issues/10303#issuecomment-1241342524 makes it sound like the issue was resolved and this is no longer needed?

kquick (approved):
Not sure if the cargo workaround is still needed, but otherwise LGTM.

.github/Dockerfile-crux-mir
RyanGlScott: It has been fixed in more recent Rust/`cargo` distributions, but we are still using an old Rust nightly that is affected by this issue:

https://github.com/GaloisInc/crucible/blob/5ae12b7598a578b7dd7c42671113572c218a0f0f/.github/workflows/crux-mir-build.yml#L89

(I am working to upgrade the Rust nightly to a 2023 one, but it's not going to be ready in time for the Crux 0.7 release.)

PR #1095, opened 2023-06-30
Nightly Crux Docker builds

### LINKED ISSUE #967: Generate nightly Docker images for `crux-llvm`
Although we have a Docker repository for `crux-llvm`, it is only updated once per new release. We should aspire to also have a nightly build of the `crux-llvm` Docker image like we do for `cryptol` and `saw`.

### LINKED ISSUE #988: Trigger CI, build Docker images when release is tagged

RyanGlScott:
Much of this CI was cargo-culted from Cryptol and SAW's CI setup.

Fixes #967. Fixes #988.

### DIFF
--- a/.github/workflows/crux-llvm-build.yml
+++ b/.github/workflows/crux-llvm-build.yml
@@ -2,8 +2,11 @@ name: crux-llvm
 
 on:
   push:
+    tags: ["crux-v?[0-9]+.[0-9]+(.[0-9]+)?"]
     branches: [master, "release-**"]
   pull_request:
+  schedule:
+    - cron: "0 10 * * *" # 10am UTC -> 2/3am PST
   workflow_dispatch:
 
 # The CACHE_VERSION can be updated to force the use of a new cache if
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Happily, the `build-push-image` jobs work just as I would expect. Here are some manually triggered instances of the job to prove this:

* [`crux-llvm`](https://github.com/GaloisInc/crucible/actions/runs/5426489880)
* [`crux-mir`](https://github.com/GaloisInc/crucible/actions/runs/5426893043)

I'll go ahead and mark this as ready for review.

kquick (approved):
I wish this was a simpler process (in general: not a negative assessment of this PR), but it gets the job done.

Issue #1097, opened 2023-07-08
Update Stack Instructions in README.md


EthanJamesLew:
# Overview

The `stack` path in the Quick Start does not appear to be supported anymore. The `stack-ghc-*.yaml` files referenced don't exist.

## Steps to Reproduce

Following this instruction block in the Quick Start of the README.md,
```shell
ls stack-ghc-*.yaml
# Choose the GHC version you prefer
ln -s stack-ghc-<version>.yaml stack.yaml
./scripts/build-sandbox.sh
stack setup
stack build
```
I received

```
(base) ➜  crucible git:(master) ls stack-ghc-*.yaml

zsh: no matches found: stack-ghc-*.yaml
```

EthanJamesLew:
Awesome, thanks!

Issue #884, opened 2021-10-12
`README`: Note that `cabal` does support coverage with 3.6.0.0 or later


RyanGlScott:
Currently, the `README` [says](https://github.com/GaloisInc/crucible/tree/796d1f595e7cbcd421e68050fa5cc864327b7aab#testing-and-coverage):

> Testing with coverage tracking is currently only available via `stack`, as `cabal new-*` [does not yet support coverage](https://github.com/haskell/cabal/issues/5213). Use `scripts/stack-test-coverage.sh` to generate a coverage report for all test suites.

As pointed out in https://github.com/GaloisInc/crucible/issues/874#issuecomment-934493138, however, the claim that `cabal` doesn't support coverage isn't quite true these days. The issue linked above haskell/cabal#5213, has since been fixed, and the fix is present in `cabal-3.6.0.0` and later. We should change the wording in the `README` to reflect this.

peterbecich:
Are you running `cabal test --enable-coverage all` , `cabal test --enable-coverage --disable-library-coverage all`, or something else?

This may still be unresolved: https://github.com/haskell/cabal/issues/6440
Using Cabal 3.6.2.0 and the latest commit of `crucible`, I get:
```
% cabal test --enable-coverage all
Resolving dependencies...
Error:
    Internal libraries only supported with per-component builds.
    Per-component builds were disabled because program coverage is enabled
    In the package 'tasty-sugar-1.1.0.0'
```

Thanks

RyanGlScott:
Sigh. Admittedly, I hadn't actually tried running `cabal test --enable-coverage` before posting this issue, but you're correct in observing that it doesn't work out of the box just yet. I guess we'll just have to mention the workaround from haskell/cabal#6440 in the `README`.

RyanGlScott:
Done in #1098.

PR #1098, opened 2023-07-10
Remove stack references; cabal is now fully capable for this project.

kquick:


kquick:
Stack support was removed some time ago.  Fixes #1097

README.md
```diff
@@ -111,10 +102,7 @@ Alternately, you can target a more specific sub-package instead of `all`.
 Testing and Coverage
 --------------------
 
-Testing with coverage tracking is currently only available via
-`stack`, as `cabal new-*` [does not yet support coverage](https://github.com/haskell/cabal/issues/5213).
-Use `scripts/stack-test-coverage.sh` to generate a coverage
-report for all test suites.
+Testing with coverage tracking is done via `cabal test --enable-coverage ...`
```
RyanGlScott, line 105: Make sure to mention to mention the workaround for https://github.com/GaloisInc/crucible/issues/884 here.

README.md
```diff
@@ -87,17 +87,8 @@ initialize and recursively update submodules.  You will find it most
...
-ln -s stack-ghc-<version>.yaml stack.yaml
-./scripts/build-sandbox.sh
-stack setup
-stack build
-```
+Crucible can be built with the `cabal` tool (although it is not compatible with
+the old/v1 cabal version).
 
 ```
 ./scripts/build-sandbox.sh
```
RyanGlScott, line 94: As a side note (not necessarily something that you need to do in this PR), I wonder if this script is even worth having. All it does currently is:

https://github.com/GaloisInc/crucible/blob/2cf85d801d806a5f62cfc19403e041169be66cb7/scripts/build-sandbox.sh#L1C1-L3

We could just as well inline this into the `README` instructions. Moreover, I'm not sure how I feel about the `--recursive` part, since that will likely cause submodules to clone _other_ submodules that aren't necessarily for the `crucible` build.

README.md
kquick: +1

RyanGlScott (approved):

PR #1099, opened 2023-07-17
Consistently use same Rust nightly across CI and Docker images

RyanGlScott:
This should fix the nightly `crux-mir` Docker job. While I was in town, I made sure to leave breadcrumbs to the reader such that if they ever update the Rust nightly in one spot, they will remember to update it in the other spot.


RyanGlScott:
A manually triggered CI job with these changes can be found [here](https://github.com/GaloisInc/crucible/actions/runs/5561967032/jobs/10159977329), which verifies that this does in fact fix the `crux-mir` Docker job.

kquick (approved):

PR #1100, opened 2023-07-19
Update license field following SPDX 2.1 license expression standard

frisoft:
The new recommendation is to follow the SPDX 2.1 standard. This allows automatic license verification software to work properly. Reference: https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields


RyanGlScott:
Hi @frisoft.

For some additional context here: the libraries included under `crux-mir/lib` aren't typical Rust libraries, as they're meant to be ingested by our `crux-mir` tool rather than built with a normal Rust compiler. (That means that these libraries will never appear on, say, crates.io.) Moreover, these libraries are taken upstream from the `rustc` source code, with only minor changes applied to make it worth with `crux-mir`.

I do like the initiative to make all `license` fields follow SPDX 2.1, but I wonder if it would be better to patch the upstream source code instead. That way, when we update `crux-mir` to build against a newer version of Rust, we'll inherit all of the `license`-related changes for free. We _could_ deviate from the upstream source code further and change the `license`s here, but then we'd have to record any further changes we've made to the original source code. (See [this file](https://github.com/GaloisInc/crucible/blob/24b84ed809c9298b834cce188895a0d4a2b629b9/crux-mir/lib/Patches.md).) This would be possible, but it would impose an additional maintenance burden on us.

What do you think?

frisoft:
Thank you, @RyanGlScott 

I understand what you mean. If they are taken from an upstream crate, they should be fixed there.
However, I can see that they are already updated in the upstream create.
For example, this is the license for cfg-if: https://github.com/rust-lang/cfg-if/blob/main/Cargo.toml#L5
It's already "MIT OR Apache-2.0".
Why is your copy not in sync with it? How do you keep it in sync?

RyanGlScott:
> Why is your copy not in sync with it? How do you keep it in sync?

We're still using a slightly old copy of the libraries, as we are currently basing our work on the `nightly-2023-01-23` Rust nightly. We will eventually update this to a newer nightly at some point, most likely when the 2024 edition of the language is released. The only reason we haven't done so yet is because upgrading nightlies is a somewhat time-consuming process for `crux-mir`, so we want to make sure we pick a time to upgrade that works well with our time budget.

frisoft:
I see. Thank you for the explanation.
I agree it does not make much sense to update only the license :+1: .
I'll close the PR.
Thank you

Issue #1105, opened 2023-08-21
`Mir.Intrinsics` functions use `SimState` when `SymGlobalState`/`IntrinsicTypes` would suffice


RyanGlScott:
While working on the Rust SAW backend (see https://github.com/GaloisInc/saw-script/issues/1859), I discovered that the `cruible-mir` memory model–related functions in `Mir.Intrinsics` have types that are too specific for SAW's needs. Here is one example of such a type:

https://github.com/GaloisInc/crucible/blob/df3297bb366448dd7207a77eecafefe80baef69d/crucible-mir/src/Mir/Intrinsics.hs#L1139-L1145

SAW needs to write to references as part of the implementation of some of its MIR-related commands, and it ends up calling this function at some point. Or, at the very least, it _should_ call this function, but there is a catch: it requires a `SimState` as an argument. This is inconvenient for SAW's needs, as the setup-related code that it uses to install MIR specifications doesn't have easy access to a `SimState`. (It does have access to a `SimState` when actually simulating the code, but that happens much later in the pipeline.)

What's more, I don't think functions like `writeRefRoot` even need to use `SimState`. If you look at the function's implementation:

https://github.com/GaloisInc/crucible/blob/df3297bb366448dd7207a77eecafefe80baef69d/crucible-mir/src/Mir/Intrinsics.hs#L1146-L1154

The only parts of the `SimState` that the function ever uses are the `SymGlobalState` and `IntrinsicTypes`. Indeed, this is a pattern that holds for nearly every function in `Mir.Intrinsics` that currently uses `SimState`. We could just as well refactor `writeRefRoot` to have this type signature:

```hs
writeRefRoot :: forall sym bak tp.
    (IsSymBackend sym bak) =>
    bak ->
    SymGlobalState sym -> -- new
    IntrinsicTypes sym -> -- new
    MirReferenceRoot sym tp ->
    RegValue sym tp ->
    MuxLeafT sym IO (SymGlobalState sym) -- new
```

If we did this, then it would be much easier for SAW to call into the `crucible-mir` memory model.

RyanGlScott:
I added a small number of `*IO` functions in #1106 for a particular SAW patch, but now I need more for another patch. Perhaps this is a sign that I should bite the bullet and add `*IO` entrypoints for everything in `Mir.Intrinsics` (or at least everything used in `execMirStmt`).

PR #1106, opened 2023-08-22
`Mir.Intrinsics`: Don't pass `SimState` when `SymGlobalState`/`IntrinsicTypes` will suffice

RyanGlScott:
This makes it possible to call into the `crucible-mir` memory model without needing access to a full-blown `SimState`, which is sometimes not available in SAW contexts.

Fixes https://github.com/GaloisInc/crucible/issues/1105.


bboston7 (approved):
Looks good!

PR #1107, opened 2023-08-31
`crucible-mir`: Add `*IO` entrypoints for each memory model operation

RyanGlScott:
A follow-up to #1105/#1106. I found myself needing `subanyMirRefIO`, `subfieldMirRefIO`, and `subjustMirRefIO` in an upcoming SAW patch, but to avoid having to make a bunch of little patches each time I need more things from the `crucible-mir` memory model, I opted to define `*IO` functions for each memory model operation that performs side effects. After this patch, it should be straightforward to call anything in the `crucible-mir` memory model without too much trouble.


spernsteiner (approved):
Seems pretty straightforward

PR #1110, opened 2023-09-28
Implement overrides for `llvm.is.fpclass` and friends

RyanGlScott:
This is primarily motivated by Clang 17 being more eager to compile the `isnan` function into `llvm.is.fpclass`, so this patch is required to make the `isnan.c` test case pass with Clang 17 or later. The machinery used to implement an override for `llvm.is.fpclass` can also be used to implement overrides for the related `isinf` function, so we also do this.

Checks off some boxes in #187.


kquick (approved):
My only qualm is that you have to look pretty carefully to see the difference between `basic_llvm_override Libc.llvm__isinfOverride` and `basic_llvm_override Libc.llvm__isinffOverride`, but they have different type signatures which will help, and we're not likely to be doing many changes for these, so I think it's unlikely this will be a common area for confusion.

Issue #920, opened 2021-11-09
Concrete syntax for the LLVM language extension


langston-barrett:
The only representation of Crucible programs using the LLVM language extension is as an in-memory data structure. Here are a few reasons we might want to have concrete, textual syntax for these programs:

- Tests of LLVM-related functionality are all currently integration tests involving `llvm-pretty{,-bc-parser}`. Not only is this not ideal from an isolation and modularity perspective, it is also slow.
- Furthermore, all of our tests have to decide between including binary blobs (`.bc` files) or depending on tools being present in the environment (`clang` or `llvm-as`), and executing those tools during the test suite (which is also unnecessarily slow).
- Textual test cases benefit from a sizable number of generic development tools (think fuzzers, delta debuggers, and tools like `llvm-lit`). With a some additional investment, very powerful tools like grammar-based fuzzers and test-case reducers become possible (e.g., `llvm-reduce`).
- Higher-level tools like SAW could cache Crucible-LLVM programs to disk for repeated/interactive use.

A reasonable outcome here (i.e., the ability to serialize reasonably-sized Crucible-LLVM programs that result from translation from LLVM modules) might depend on #74.

bboston7:
What a coincidence!  I'm currently working on adding support for arbitrary language extensions to Crucible Syntax.  After that work is done you'll only need to write a small parser hook into Crucible Syntax for the `LLVM` extension.

bboston7:
> What a coincidence! I'm currently working on adding support for arbitrary language extensions to Crucible Syntax. After that work is done you'll only need to write a small parser hook into Crucible Syntax for the `LLVM` extension.

I just opened pull request #923 adding support for arbitrary language extensions.

langston-barrett:
A bare-bones version of this is done, additional work can be tracked in other tickets (e.g., #1117, #1118).

Issue #964, opened 2022-04-21
CI: Use a script to generate the build matrix


RyanGlScott:
Currently, we copy-paste a nearly-identical build matrix across five different CI workflows, which is tedious to maintain. It would likely save us time in the long run to instead generate build matrices using JSON, as described in [this GitHub Actions documentation](https://docs.github.com/en/actions/learn-github-actions/expressions#fromjson). We've already done something similar for `what4` [here](https://github.com/GaloisInc/what4/blob/835238d87500d48c3b0ae9121f2b484b71faa8b5/.github/workflows/test.yml#L51).

langston-barrett:
There are also repeated bits of setup within the CI workflow that would be nice to abstract away, see [this discussion](https://github.com/GaloisInc/crucible/pull/1113) for an example:

https://github.com/GaloisInc/crucible/blob/3bdafe4fe1ed03e49a4255082cfe4d2e4a524f4b/.github/workflows/crux-llvm-build.yml#L153-L157

PR #1113, opened 2023-10-31
crucible-llvm-syntax: Concrete syntax for Crucible-LLVM

langston-barrett:
This is an MVP implementation of syntax for the LLVM language extension. This package would need additional operations to address the use-cases listed in #920, but provides a reasonable base to start building off. It has the additional benefit of giving an in-repo example of usage of the API introduced in #923.


.github/workflows/crux-llvm-build.yml
```diff
@@ -150,26 +150,28 @@ jobs:
...
-          (cd crucible-llvm; with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
-          (cd crux;          with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
-          (cd crux-llvm;     with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
-          (cd uc-crux-llvm;  with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
+          (cd crucible;             with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
+          (cd crucible-llvm;        with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
+          (cd crucible-llvm-syntax; with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
+          (cd crux;                 with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
+          (cd crux-llvm;            with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
+          (cd uc-crux-llvm;         with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean)
```
RyanGlScott, line 158: It might be time to factor out this `with_ghc ghc -o $(setup_bin) $(setup_src) && ./$(setup_bin) clean` bit into a function so that we don't have to repeat ourselves quite so much.

crucible-llvm-syntax/README.md
```diff
@@ -0,0 +1,19 @@
...
+**Types**:
+
+- `(Ptr n)` for a positive numeral `n` represents the type of LLVM pointers that are `n` bits wide. For example `(Ptr 8)` is the type of bytes.
+
+**Statements**:
+
+- `ptr : Nat -> Bitvector w -> Ptr w`: construct a pointer from a block and offset
+- `ptr-block : Ptr w -> Nat`: get the block number of a pointer
+- `ptr-offset : Ptr w -> Bitvector w`: get the offset of a pointer
+- `ptr-ite : Bool -> Ptr w -> Ptr w -> Ptr w`: if-then-else for pointers
```
RyanGlScott, line 17: Do make sure to update this list when you're done.

crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax.hs
```diff
@@ -0,0 +1,182 @@
...
+
+import Lang.Crucible.Syntax.Atoms (Atomic)
+import Lang.Crucible.Syntax.Atoms qualified as Atom
+import Lang.Crucible.Syntax.Concrete (ParserHooks(..), SyntaxState)
+import Lang.Crucible.Syntax.Concrete qualified as Parse
+import Lang.Crucible.Syntax.ExprParse (MonadSyntax)
+import Lang.Crucible.Syntax.ExprParse qualified as Parse
+
+unary :: MonadSyntax Atomic m => m b -> m a -> m a
+unary p0 p = Parse.followedBy p0 (Parse.commit *> Parse.cons p Parse.emptyList) <&> fst
```
RyanGlScott, line 50: This is copy-pasted from `Lang.Crucible.Syntax.Concrete`, right? Is it worth exporting this function somewhere from `crucible-syntax`?

crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax.hs
```diff
@@ -0,0 +1,182 @@
...
+
+    parseMemType :: MonadSyntax Atomic m => m MemType
+    parseMemType = Parse.describe "LLVM type" $ do
+      nm <- Parse.atomName
+      case nm of
+        Atom.AtomName "i8" -> pure (MemType.IntType 8)
+        Atom.AtomName "i16" -> pure (MemType.IntType 16)
+        Atom.AtomName "i32" -> pure (MemType.IntType 32)
+        Atom.AtomName "i64" -> pure (MemType.IntType 64)
+        Atom.AtomName "ptr" -> pure MemType.PtrOpaqueType
```
RyanGlScott, line 181: It would be worth documenting the list of atoms that represent types in the `README` as well.

crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax.hs
```diff
@@ -0,0 +1,182 @@
...
+                  let (Ctx.Empty, ptr) = Ctx.decompose rest
+                  let stmt = Ext.LLVM_Store mvar ptr tyRep storTy align val
+                  Some <$> Parse.freshAtom loc (Reg.EvalExt stmt)
+
+      _ -> empty
+  where
+    parseAlign :: MonadSyntax Atomic m => m Alignment
+    parseAlign = Parse.describe "alignment" $ do
+      s <- Parse.atomName
+      unless (s == Atom.AtomName "none") Parse.cut
```
RyanGlScott, line 170: Worth documenting this `none` atom too.

crucible-llvm-syntax/test-data/ptr.cbl
```diff
@@ -0,0 +1,24 @@
+(defun @test-ptr () (Ptr 64)
+  (start start:
+    (let blk0 (the Nat 0))
+    (let off0 (bv 64 0))
+    (let p0 (ptr 64 blk0 off0))
+    (let p (ptr-ite 64 #t p0 p0))
+    (let blk (ptr-block 64 p))
+    (let off (ptr-offset 64 p))
+    (assert! (equal? blk0 blk) "block numbers equal")
+    (assert! (equal? off0 off) "offsets equal")
```
RyanGlScott, line 10: This test case makes me wonder: should we also add a `ptr-eq` statement as well? This would avoid having to manually check both the block and offset values each time.

RyanGlScott (commented):
This looks exciting!

crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax.hs
langston-barrett: It's close - that function takes a `Keyword` as its first argument, and `Keyword` is not extensible based on the language extension in use. We may want to revisit that in the future, parsing things as `AtomName`s and then matching on particular ones feels like a hack...

crucible-llvm-syntax/test-data/ptr.cbl
langston-barrett: Yes! I've covered just some of the constructors in `LLVM.Extension.Syntax`, but ideally we'd do all of them, and `LLVM_PtrEq` is one of them.

But also, this test is basically just there to ensure we can parse each of these forms as expected, and I thought it'd be nice to have somewhat reasonable runtime behavior that actually checks *something*

.github/workflows/crux-llvm-build.yml
langston-barrett: Do you mind if we defer this to an issue? I'm not super familiar with this setup.

RyanGlScott: Yes, that works for me. Could you add a comment to https://github.com/GaloisInc/crucible/issues/964 about this? (Deduplicating the mess of YAML files in the CI is something we've wanted to do on many separate occasions, so it would be good to consolidate the discussion there.)

crucible-llvm-syntax/test-data/ptr.cbl
langston-barrett: I'll make a follow-up issue to cover the remaining constructors once this is merged.

crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax.hs
langston-barrett: Documented

langston-barrett:
If this is merged, I will make three follow-up issues:

- [ ] Concrete syntax for the remaining `MemType`s
- [ ] Concrete syntax for `Alignment` (other than `none`)
- [ ] Concrete syntax for the additional LLVM syntax formers

.github/workflows/uc-crux-llvm-lint.yaml
```diff
@@ -16,11 +16,16 @@ jobs:
...
         run: |
-          cd uc-crux-llvm/
           curl --location -o hlint.tar.gz \
             https://github.com/ndmitchell/hlint/releases/download/v3.3/hlint-3.3-x86_64-linux.tar.gz
           tar xvf hlint.tar.gz
-          ./hlint-3.3/hlint exe src test
+
+          pushd crucible-llvm-syntax/
+          ../hlint-3.3/hlint src test
+          popd
```
kquick, line 25: I usually recommend using the `(cd tgtdir; something)` syntax over a `pushd tgtdir; something; popd` because it's easier to get the pushd/popd pairs mismatched.

crucible-llvm-syntax/crucible-llvm-syntax.cabal
```diff
@@ -0,0 +1,124 @@
...
+Maintainer:    langston@galois.com
+Build-type:    Simple
+License:       BSD-3-Clause
+License-file:  LICENSE
+Category:      Language
+Synopsis:      A syntax for reading and writing Crucible-LLVM control-flow graphs
+-- Description:
+
+extra-doc-files: README.md
+extra-source-files:
```
kquick, line 14: Your golden test files should be listed here to ensure they are included in any sdist package.

langston-barrett: Nice catch!

kquick: We probably have this issue a lot of places.   We're not sdist-ing these ATM, so it hasn't been a problem, but yes, we should clean this up wherever we can.  It will probably manifest other than sdist at times: I recently discovered that if your cabal.project references a local package path that you will be fine but if it references a source-repository then you will encounter this issue as well.

RyanGlScott: Indeed, I have used `source-repository-package`s instead of `git` submodules on occasion for things under the `GaloisInc` repo, and this issue has bitten me every now and then. It would be good to be consistent about including things in `extra-source-files` going forward, if nothing else.

crucible-llvm-syntax/README.md
```diff
@@ -0,0 +1,37 @@
...
+- `(Ptr n)` for a positive numeral `n` represents the type of LLVM pointers that are `n` bits wide; for example `(Ptr 8)` is the type of bytes
+
+**Primitive atoms**:
+
+- `none : Alignment`: no alignment
+- `i8 : LLVMType`: [LLVM docs][int-type], corresponds to Crucible-LLVM's `IntType 8 :: MemType`
+- `i16 : LLVMType`: [LLVM docs][int-type], corresponds to Crucible-LLVM's `IntType 16 :: MemType`
+- `i32 : LLVMType`: [LLVM docs][int-type], corresponds to Crucible-LLVM's `IntType 32 :: MemType`
+- `i64 : LLVMType`: [LLVM docs][int-type], corresponds to Crucible-LLVM's `IntType 64 :: MemType`
+- `ptr : LLVMType`: [LLVM docs][ptr-type], corresponds to Crucible-LLVM's `IntType 64 :: PtrOpaqueType`
```
RyanGlScott, line 20: Should this be:

```suggestion
- `ptr : LLVMType`: [LLVM docs][ptr-type], corresponds to Crucible-LLVM's `PtrOpaqueType :: MemType`
```

?

RyanGlScott (approved):

PR #1121, opened 2023-11-07
crucible{,-llvm}-syntax: Evaluating programs with syntax extensions

### LINKED ISSUE #1115: crucible-syntax: Allow simulating CFGs with syntax extensions
Despite its name, the `crucible-syntax` also contains utilities for *simulating* Crucible CFGs and printing human-readable output. This is very useful for [testing the simulator](https://github.com/GaloisInc/crucible/tree/e9b1a015a45775ee9a45db125facafef18b5c84e/crucible-syntax/test-data/simulator-tests). It would be great to extend this capability to CFGs using language extensions, such as via `crucible-syntax-llvm`. I believe the main extension would be to [`simulateProgram`](https://github.com/GaloisInc/crucible/blob/e9b1a015a45775ee9a45db125facafef18b5c84e/crucible-syntax/src/Lang/Crucible/Syntax/Prog.hs#L142).

langston-barrett:
Fixes #1115.

- Changes the type of `SimulateProgramHooks` to bind an `ext` variable, so hooks can be extension-specific
- Adds a simulation hook that can run an arbitrary override action before simulation (necessary for initializing LLVM memory global variable)
- Adds a `exec-crucible-llvm` executable based on `crucibler` - has lots of shared code with `crucibler`, but is also substantially different, so not sure if more sharing is feasible
- Adds a test for simulating Crucible-LLVM extension statements


crucible-syntax/src/Lang/Crucible/Syntax/Prog.hs
```diff
@@ -91,19 +92,24 @@ doParseCheck fn theInput pprint outh =
...
 -- | Allows users to hook into the various stages of 'simulateProgram'.
-data SimulateProgramHooks = SimulateProgramHooks
-  { setupOverridesHook ::
-      forall p sym ext t st fs. (IsSymInterface sym, sym ~ (ExprBuilder t st fs)) =>
+data SimulateProgramHooks ext = SimulateProgramHooks
+  { setupHook ::
+      forall p sym rtp a r t st fs. (IsSymInterface sym, sym ~ ExprBuilder t st fs) =>
+        sym ->
+        HandleAllocator ->
+        OverrideSim p sym ext rtp a r ()
```
RyanGlScott, line 100: Add a comment here saying what `setupHook`'s purpose is. It might also be worth citing the specific example of `crucible-llvm-syntax`'s `setupHook` as an example of how it's used in practice.

crucible-syntax/src/Lang/Crucible/Syntax/Prog.hs
```diff
@@ -116,9 +122,10 @@ data SimulateProgramHooks = SimulateProgramHooks
 -- * Sets up no additional overrides.
 --
 -- * Errors out if a program contains one or more forward declarations.
```
RyanGlScott, line 124: Add a bullet point here saying that there is no setup beforehand.

crucible-llvm-syntax/app/Main.hs
```diff
@@ -0,0 +1,151 @@
...
+  Opt.subparser $
+       (Opt.command "check"
+        (Opt.info (CheckCommand <$> parseCheck)
+         (Opt.fullDesc <> Opt.progDesc "Check a file" <> Opt.header "crucibler")))
+    <> (Opt.command "simulate"
+        (Opt.info (SimulateCommand <$> simFile)
+         (Opt.fullDesc <> Opt.progDesc "Simulate a file" <> Opt.header "crucibler")))
+    <> (Opt.command "profile"
+        (Opt.info (ProfileCommand <$> profFile)
+         (Opt.fullDesc <> Opt.progDesc "Simulate a file, with profiling" <> Opt.header "crucibler")))
```
RyanGlScott, line 85: Some vestigial references to `crucibler` that should instead be `exec-crucible-llvm`.

RyanGlScott (commented):
> Adds a `exec-crucible-llvm` executable based on `crucibler` - has lots of shared code with `crucibler`, but is also substantially different, so not sure if more sharing is feasible

I might be missing something, but the code in `exec-crucible-llvm` doesn't seem _that_ different to me. At a glance, the big differences that stand out are:

1. `exec-crucible-llvm` uses a different `ExtensionImpl`
2. `exec-crucible-llvm` uses a different `setupHook`
3. `exec-curcible-llvm` uses a different `ParserHooks`

Am I missing something? If that's it, I could definitely imagine factoring out a large chunk of the functionality that is shared in common between `crucibler` and `exec-crucible-llvm`, including the three things above as arguments to the function.

(As a side note, I've never understand why it's named `crucibler` in the first place. I was going to suggest naming the other tool something like `crucibler-llvm`, but I'm not sure if it's a convention worth keeping.)

langston-barrett:
> Am I missing something? If that's it, I could definitely imagine factoring out a large chunk of the functionality that is shared in common between crucibler and exec-crucible-llvm, including the three things above as arguments to the function.

No, you're right - I guess the real question is: Where should that shared code go, given that it mostly consists of `optparse-applicative` setup? I would say probably not in `crucible-syntax` itself, as there's no reason for library clients to need to build `optparse-applicative` and that module. Do we want to make a new package (`crucible-exec`?) for sharing code between `crucible-syntax` frontends?

langston-barrett:
> I was thinking of moving most of this functionality to Lang.Crucible.Syntax.Prog.

Done.

I don't *love* this solution, as library clients still need to build code they don't need (`Prog.hs`), and the CLIs still need to copy-paste code (`optparse-applicative` setup). But I think iit's better than it was before, so thank you for the suggestion!

In the future, it might be nice to create a `crucible-cli` package or similar that provides the functionality of `Prog.hs`, in addition to the `optparse-applicative` bits, and then have `crucible-cli`, `crucible-llvm-cli`, etc. executables that are all thin wrappers over the `crucible-cli` library that instantiate it with the proper syntax extension. In addition to the current functionality (check/simulate/repl), we could also expose reading and then pretty-printing CFGs from actual code (e.g., LLVM bitcode modules), and perhaps add some debugging-oriented execution features that could be used with `simulate`.

RyanGlScott (approved):
I'm broadly aligned with the goal of factoring out the CLI-related bits, although we certainly don't have to hold up this PR on that.

Everything else looks great!

PR #1123, opened 2023-11-08
crucible-cli: Library for sharing code between Crucible CLI frontends

langston-barrett:
Previously, `crucible-syntax` had some code for setting up and executing the Crucible symbolic simulator. This functionality is very useful for testing and interactive experimentation via a CLI, but is not necessary for most library clients of `crucible-syntax`. Furthermore, creating additional CLI frontends entailed copy-pasting a lot of `optparse-applicative`-related code from `crucible-syntax`'s' `crucibler` executable. Now, both these issues are addressed by factoring this code into a separate `crucible-cli` package.


crucible-cli/app/Main.hs
```diff
@@ -0,0 +1,21 @@
...
+import Lang.Crucible.Syntax.Concrete (defaultParserHooks)
+import Lang.Crucible.Syntax.Overrides (setupOverrides)
+
+import Lang.Crucible.CLI (SimulateProgramHooks(..), defaultSimulateProgramHooks)
+import qualified Lang.Crucible.CLI.Options as Opts
+
+main :: IO ()
+main =
+  do let ?parserHooks = defaultParserHooks
+     Opts.main "crucible" (\_ -> emptyExtensionImpl) simulationHooks
```
RyanGlScott, line 16: Should `"crucible"` be `"crucible-cli"` here?

crucible-cli/src/Lang/Crucible/CLI.hs
```diff
@@ -0,0 +1,275 @@
...
+            , _profOutFile :: FilePath
+            }
+
+data Command
+  = CheckCommand CheckCmd
+  | SimulateCommand SimCmd
+  | ProfileCommand ProfCmd
+  | ReplCommand
+
+execCommand ::
```
RyanGlScott, line 244: This (and possibley the `Command` data type as well) deserve some Haddocks.

crucible-llvm-cli/src/Lang/Crucible/LLVM/CLI.hs
```diff
@@ -0,0 +1,54 @@
...
+
+import Lang.Crucible.LLVM (llvmExtensionImpl)
+import Lang.Crucible.LLVM.DataLayout (EndianForm(LittleEndian))
+import Lang.Crucible.LLVM.Extension (LLVM)
+import Lang.Crucible.LLVM.MemModel (HasPtrWidth, defaultMemOptions, emptyMem, mkMemVar)
+
+import Lang.Crucible.LLVM.Syntax (llvmParserHooks)
+import Lang.Crucible.LLVM.Syntax.TypeAlias (typeAliasParserHooks, x86_64LinuxTypes)
+
+withLlvmHooks ::
```
RyanGlScott, line 33: Haddocks plz

RyanGlScott (approved):

crucible-cli/app/Main.hs
langston-barrett: I actually kind of like it better as `crucible`, but I could be convinced otherwise. I think it's awkward to name a CLI program `*-cli`: the fact that I'm typing its name into my command prompt implies that it's a CLI. Additionally, this is the only binary that interfaces directly with Crucible, so it seems apt to name it `crucible`.

langston-barrett: Note that the executable stanza in the Cabal file matches this naming scheme: 

https://github.com/GaloisInc/crucible/pull/1123/files#diff-b719c603b6f86ffa97657880ea5a3e8a4335c480718149b707e8136b0a560172R137

Issue #1118, opened 2023-11-01
crucible-llvm-syntax: Concrete syntax for remaining LLVM syntax formers


langston-barrett:
#1113 introduced support for parsing syntax that corresponds to some of the constructors of `LLVMExtensionExpr` and `LLVMStmt`. We should extend this parser with syntax for all of these constructors:

[`LLVMExtensionExpr`](https://github.com/GaloisInc/crucible/blob/e9b1a015a45775ee9a45db125facafef18b5c84e/crucible-llvm/src/Lang/Crucible/LLVM/Extension/Syntax.hs#L71)

- [ ] `X86Expr`
- [ ] `LLVM_SideConditions`
- [x] `LLVM_PointerExpr` (#1113)
- [x] `LLVM_PointerBlock` (#1113)
- [x] `LLVM_PointerOffset` (#1113)
- [x] `LLVM_PointerIte` (#1113)

[`LLVMStmt`](https://github.com/GaloisInc/crucible/blob/e9b1a015a45775ee9a45db125facafef18b5c84e/crucible-llvm/src/Lang/Crucible/LLVM/Extension/Syntax.hs#L104):

- [x] `LLVM_PushFrame` (#1402)
- [x] `LLVM_PopFrame` (#1402)
- [x] `LLVM_Alloca` (#1113)
- [x] `LLVM_Load` (#1113)
- [x] `LLVM_Store` (#1113)
- [ ] `LLVM_MemClear`
- [x] `LLVM_LoadHandle` (#1125)
- [x] `LLVM_ResolveGlobal` (#1125)
- [x] `LLVM_PtrEq` (#1402)
- [x] `LLVM_PtrLe` (#1402)
- [x] `LLVM_PtrAddOffset` (#1146)
- [x] `LLVM_PtrSubtract` (#1146)
- [ ] `LLVM_Debug`

langston-barrett:
I think we've now covered most of the syntax formers that we'll want to use in test-cases. Closing this, let's open issues for the remaining ones if and when we need them.

Issue #1016, opened 2022-07-22
Add a proper pretty-printer for `TypeRepr`s


RyanGlScott:
Currently, the `Pretty` instance for `TypeRepr` simply goes through its `Show` instance:

https://github.com/GaloisInc/crucible/blob/5a2d646f3d7bdc7652d228c61bf6a2d673b13dfc/crucible/src/Lang/Crucible/Types.hs#L467-L468

This gets the job done, but the output is not great for showing users in error messages. In the particular case of `crucible-llvm`, you're likely to get error messages involving things like `IntrinsicRepr LLVM_pointer [BVRepr 8]`, which isn't great. It would be better to have a pretty-printer that presents a more compact view of the type involved. We could make the `Pretty` instance do this, but we wouldn't be able to cover all of the possible `IntrinsicRepr`s. Perhaps each library that defines its own `IntrinsicType` (e.g., `crucible-llvm`) could also provide a function that extends the `Pretty` instance to provide special treatment for `IntrinsicRepr`.

langston-barrett:
It would be nice for the pretty-printer to match the syntax provided by the various `crucible-{llvm-,}syntax` packages to the degree possible.

PR #1125, opened 2023-11-08
crucible-llvm-cli: Support for overrides; loading and calling functions

langston-barrett:
- [x] Exports low-level overrides from crucible-llvm that can be registered without having to construct an `L.Declare`
- [x] Uses the above to register an override for `malloc` inside `crucible-llvm-syntax`
- [x] Removes the `L.Type` from the LLVM operation for loading a function handle from a pointer
- [x] Adds concrete syntax for the LLVM syntax extension operation of loading a function handle from a pointer (#1118)
- [x] Adds concrete syntax for resolving LLVM global names to pointers (#1118)
- [x] Adds a simulation test to crucible-llvm-cli that calls the `malloc` override

### DIFF
--- a/crucible-llvm-cli/test-data/ptr.cbl
+++ b/crucible-llvm-cli/test-data/ptr.cbl
@@ -21,4 +21,12 @@
     (assert! (equal? vblk0 vblk) "stored block numbers equal")
     (assert! (equal? voff0 voff) "stored offsets equal")
 
+    (let g (resolve-global "malloc"))
+    (let gblk (ptr-block 64 g))
+    (assert! (not (equal? gblk 0)) "malloc block number nonzero")
+    (let h (load-handle (Ptr 64) ((Ptr 64)) g))
+    (let m (funcall h p0))
+    (let mblk (ptr-block 64 m))
+    (assert! (not (equal? mblk 0)) "malloc'd block number nonzero")
+
     (return ())))
--- a/crucible-llvm/crucible-llvm.cabal
+++ b/crucible-llvm/crucible-llvm.cabal
@@ -67,6 +67,8 @@ library
     Lang.Crucible.LLVM.Extension
     Lang.Crucible.LLVM.Globals
     Lang.Crucible.LLVM.Intrinsics
+    Lang.Crucible.LLVM.Intrinsics.Libc
+    Lang.Crucible.LLVM.Intrinsics.LLVM
     Lang.Crucible.LLVM.MalformedLLVMModule
     Lang.Crucible.LLVM.MemModel
     Lang.Crucible.LLVM.MemModel.CallStack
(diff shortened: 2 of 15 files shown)


crux-llvm/test-data/golden/golden/invoke-test.pre-clang16.z3.good
```diff
@@ -5,7 +5,7 @@
 [Crux]     No implementation or override found for pointer
 [Crux]     The given pointer could not be resolved to a callable function
 [Crux]     No implementation or override found for pointer
-[Crux]     Attempting to load callable function with type: i64({ }*, [3 x i64]*, i64, i8**)
+[Crux]     Attempting to load callable function:
```
langston-barrett, line 8: Note to reviewers: I removed an llvm-pretty type from the LLVM syntax former for loading function handles from pointers, because it was only used to generate this diagnostic message and it seems unnecessary, redundant, and mildly effortful to add a parser for LLVM types to `crucible-llvm-syntax`. This makes this error message less informative than it formerly was. With #1016, we could at least pretty-print the Crucible types involved, but for now this seems like an acceptable trade-off. The LLVM type can always be recovered by looking at the `llvm-dis` output.

crucible-llvm/crucible-llvm.cabal
```diff
@@ -67,6 +67,8 @@ library
     Lang.Crucible.LLVM.Extension
     Lang.Crucible.LLVM.Globals
     Lang.Crucible.LLVM.Intrinsics
+    Lang.Crucible.LLVM.Intrinsics.Libc
+    Lang.Crucible.LLVM.Intrinsics.LLVM
```
langston-barrett, line 71: These modules are now exported to support registering overrides without needing to provide a list of `L.Decl`s, which aren't available in some settings, i.e., in `crucible-llvm-cli`.

langston-barrett:
It's possible that this will need a tweak here or there to pass CI, but it's generally ready for review.

RyanGlScott (approved):
I'm broadly in favor of this, although I'm definitely feeling the itch for more documentation.

crucible-llvm-cli/test-data/ptr.cbl
```diff
@@ -21,4 +21,12 @@
     (assert! (equal? vblk0 vblk) "stored block numbers equal")
     (assert! (equal? voff0 voff) "stored offsets equal")
 
+    (let g (resolve-global "malloc"))
+    (let gblk (ptr-block 64 g))
+    (assert! (not (equal? gblk 0)) "malloc block number nonzero")
+    (let h (load-handle (Ptr 64) ((Ptr 64)) g))
```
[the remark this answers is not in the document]
langston-barrett: Agreed, that was an oversight on my part!

> Speaking of which, the type signature of malloc is i8* malloc(size_t). Does this mean that size_t becomes (Ptr 64) here? I think so, but it would be worth saying as much somewhere.

Yep! These types are at the Crucible-LLVM level, so there is no such thing as `size_t` or `i8*`, just `LLVMPointerType` (written `Ptr`). Will describe that in the README.

PR #1136, opened 2023-11-16
Typed overrides

### LINKED ISSUE #1135: crucible: Typed overrides
Several Crucible frontends ([jvm](https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-jvm/src/Lang/Crucible/JVM/Overrides.hs#L147), [llvm](https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs#L80), [go](https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-go/src/Lang/Crucible/Go/Overrides.hs#L43)) make use of types that include:

- An override "handler", with a type like `forall bak. IsSymBackend sym bak => bak -> Ctx.Assignment (RegValue sym) args -> OverrideSim p sym ext rtp args ret (RegValue sym ret)`
- A `CtxRepr args`
- A `TypeRepr ret`

Several language extensions make use of the LLVM memory model (e.g., LLVM, WASM, and Macaw), and could concievably share override implementations relating to memory. One could also imagine sharing implementations of overrides like `crucible_assert`. To make this task easier, it would be nice to unify these notions of "typed overrides", by adding a type to the base Crucible library and using it in the different frontends.

langston-barrett:
Fixes #1135.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
+++ b/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
@@ -75,7 +75,7 @@ module Lang.Crucible.Simulator.OverrideSim
     -- * Typed overrides
   , TypedOverride(..)
   , SomeTypedOverride(..)
-  , mkTypedOverride
+  , runTypedOverride
     -- * Re-exports
   , Lang.Crucible.Simulator.ExecutionTree.Override
   ) where
@@ -683,14 +683,16 @@ data TypedOverride p sym ext args ret
     , typedOverrideRet :: TypeRepr ret
     }
 
+-- | A 'TypedOverride' with the type parameters @args@, @ret@ existentially
+-- quantified
 data SomeTypedOverride p sym ext =
   forall args ret. SomeTypedOverride (TypedOverride p sym ext args ret)
 
 -- | Create an override from a 'TypedOverride'.
-mkTypedOverride ::
+runTypedOverride ::
   FunctionName ->
   TypedOverride p sym ext args ret ->
   Override p sym ext args ret
-mkTypedOverride nm typedOvr = mkOverride' nm (typedOverrideRet typedOvr) $ do
+runTypedOverride nm typedOvr = mkOverride' nm (typedOverrideRet typedOvr) $ do
   RegMap args <- getOverrideArgs
   typedOverrideHandler typedOvr (fmapFC (RV . regValue) args)
--- a/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
+++ b/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
@@ -72,6 +72,10 @@ module Lang.Crucible.Simulator.OverrideSim
   , IntrinsicImpl
   , mkIntrinsic
   , useIntrinsic
+    -- * Typed overrides
+  , TypedOverride(..)
+  , SomeTypedOverride(..)
+  , mkTypedOverride
     -- * Re-exports
   , Lang.Crucible.Simulator.ExecutionTree.Override
   ) where
@@ -93,6 +97,8 @@ import           Numeric.Natural (Natural)
 import           System.Exit
 import           System.IO
 import           System.IO.Error
+
+import           Data.Parameterized.TraversableFC (fmapFC)
 
 import           What4.Config
 import           What4.Interface
@@ -657,3 +663,34 @@ mkIntrinsic m hdl = mkOverride' (handleName hdl) (handleReturnType hdl) ovr
        sym <- getSymInterface
        (RegMap args) <- getOverrideArgs
        Ctx.uncurryAssignment (m (Proxy :: Proxy r) sym) args
+
+--------------------------------------------------------------------------------
+-- Typed overrides
+
+-- | An action in 'OverrideSim', together with 'TypeRepr's for its arguments
+-- and return values. This type is used across several frontends to define
+-- overrides for built-in functions, e.g., @malloc@ in the LLVM frontend.
+--
+-- For maximal reusability, frontends may define 'TypedOverride's that are
+-- polymorphic in (any of) @p@, @sym@, and @ext@.
+data TypedOverride p sym ext args ret
+  = TypedOverride
+    { typedOverrideHandler ::
+        forall rtp args' ret'.
+        Ctx.Assignment (RegValue' sym) args ->
+        OverrideSim p sym ext rtp args' ret' (RegValue sym ret)
+    , typedOverrideArgs :: CtxRepr args
+    , typedOverrideRet :: TypeRepr ret
+    }
+
+data SomeTypedOverride p sym ext =
+  forall args ret. SomeTypedOverride (TypedOverride p sym ext args ret)
+
+-- | Create an override from a 'TypedOverride'.
+mkTypedOverride ::
+  FunctionName ->
+  TypedOverride p sym ext args ret ->
+  Override p sym ext args ret
+mkTypedOverride nm typedOvr = mkOverride' nm (typedOverrideRet typedOvr) $ do
+  RegMap args <- getOverrideArgs
+  typedOverrideHandler typedOvr (fmapFC (RV . regValue) args)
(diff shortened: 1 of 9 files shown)


crux/src/Crux/Overrides.hs
```diff
@@ -0,0 +1,74 @@
...
+  (C.RegValue' sym stringTy -> OverM p sym ext String) ->
+  C.TypedOverride (p sym) sym ext (C.EmptyCtx C.::> stringTy) (C.BaseToType bty)
+baseFreshOverride bty sty getStr =
+  C.TypedOverride
+  { C.typedOverrideHandler = \(Ctx.Empty Ctx.:> strVal) -> do
+      str <- getStr strVal
+      mkFresh str bty
+  , C.typedOverrideArgs = Ctx.Empty Ctx.:> sty
+  , C.typedOverrideRet = C.baseToType bty
+  }
```
langston-barrett, line 73: This is a good example of the kind of sharing that is made possible by introducing this type. Currently, it's only used by `crux-mir`, but could certainly be used by other frontends, e.g., if crux-jvm wanted to introduce `fresh*` overrides, or by `crux-go` if https://github.com/GaloisInc/crucible/issues/1137 were addressed.

crux/src/Crux/Overrides.hs
```diff
@@ -0,0 +1,88 @@
...
+
+-- | Build an override that takes no arguments and returns a fresh constant a
+-- given name. Generally, frontends should prefer 'baseFreshOverride', to allow
```
RyanGlScott, line 75: This first sentence reads somewhat awkwardly. Perhaps this is what you meant?

```suggestion
-- | Build an override that takes no arguments and returns a fresh constant that uses the
-- given name. Generally, frontends should prefer 'baseFreshOverride', to allow
```

crux/src/Crux/Overrides.hs
```diff
@@ -0,0 +1,88 @@
...
+
+import qualified Lang.Crucible.Backend as C
+import qualified Lang.Crucible.Simulator.RegValue as C
+import qualified Lang.Crucible.Types as C
+import qualified Lang.Crucible.Simulator.OverrideSim as C
+
+import Crux.Types (OverM)
+import Control.Monad.IO.Class (liftIO)
+
+mkFresh ::
```
RyanGlScott, line 27: Haddocks plz

crux/src/Crux/Overrides.hs
```diff
@@ -0,0 +1,88 @@
...
+mkFresh name ty =
+  C.ovrWithBackend $ \bak ->
+    do let sym = C.backendGetSym bak
+       elt <- liftIO (W4.freshConstant sym name ty)
+       loc <- liftIO $ W4.getCurrentProgramLoc sym
+       let ev = C.CreateVariableEvent loc (show name) ty elt
+       liftIO $ C.addAssumptions bak (C.singleEvent ev)
+       return elt
+
+mkFreshFloat ::
```
RyanGlScott, line 41: Haddocks plz

crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
```diff
@@ -657,3 +663,34 @@ mkIntrinsic m hdl = mkOverride' (handleName hdl) (handleReturnType hdl) ovr
...
+    { typedOverrideHandler ::
+        forall rtp args' ret'.
+        Ctx.Assignment (RegValue' sym) args ->
+        OverrideSim p sym ext rtp args' ret' (RegValue sym ret)
+    , typedOverrideArgs :: CtxRepr args
+    , typedOverrideRet :: TypeRepr ret
+    }
+
+data SomeTypedOverride p sym ext =
+  forall args ret. SomeTypedOverride (TypedOverride p sym ext args ret)
```
RyanGlScott, line 687: Haddocks plz

crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
```diff
@@ -657,3 +663,34 @@ mkIntrinsic m hdl = mkOverride' (handleName hdl) (handleReturnType hdl) ovr
...
+
+-- | Create an override from a 'TypedOverride'.
+mkTypedOverride ::
```
RyanGlScott, line 690: The name "`mkTypedOverride`" is somewhat confusing, as this function doesn't return a `TypedOverride` but rather takes one as an argument. I would have almost expected this to be named something like "`runTypedOverride`".

RyanGlScott (commented):
Nice!

It looks like most of these changes involve the internals of libraries, or the APIs of libraries that aren't widely depended on. Is that right? My first inclination was to question if we should change [`LLVMOverride`](https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs#L78-L95) to be defined in terms of a `TypedOverride`, but that would of course result in a breaking API change (and one that would likely matter in practice).

langston-barrett:
> It looks like most of these changes involve the internals of libraries, or the APIs of libraries that aren't widely depended on. Is that right? 

Yep, that's right.

> My first inclination was to question if we should change [LLVMOverride](https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs#L78-L95) to be defined in terms of a TypedOverride, but that would of course result in a breaking API change (and one that would likely matter in practice).

Indeed, and there are two other complications here:

1. `LLVMOverride` also contains an LLVM declaration, so we can't simply *replace* it with `TypedOverride` (though after #1138 we might be able to)
2. The LLVM override implementation (action in `OverrideSim`) additionally takes a `GlobalVar Mem` parameter.

RyanGlScott (approved):
> 1. `LLVMOverride` also contains an LLVM declaration, so we can't simply _replace_ it with `TypedOverride` (though after #1138 we might be able to)

Fair enough. But do we really need to replace all of `LLVMOverride` with `TypedOverride`? I could imagine something like this:

```hs
data LLVMOverride p sym args ret =
  LLVMOverride
  { llvmOverride_declare :: Declare
  , llvmOverride_def :: TypedOverrie p sym LLVM args ret
  }
```

There is precedent for this, as you already do something similar with `JVMOverride` in this patch. Granted, it might be somewhat unsatisfactory to make such a breaking API change now when `llvmOverride_declare` is slated to be removed in #1138, so I could understand not wanting to do this just yet. Either way, it would be worth filing an issue about this as a reminder for later.

> 2. The LLVM override implementation (action in `OverrideSim`) additionally takes a `GlobalVar Mem` parameter.

This doesn't worry me that much, since I imagine that we could simply create closures that close over the `GlobalVar Mem` argument when constructing LLVM `TypedOverride`s. Alternatively, if that proves too annoying, then we can always get and set the underlying `MemImpl` from within the override, similarly to how `crucible-llvm` does it [here](https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L380-L395).

langston-barrett:
> Fair enough. But do we really need to replace all of LLVMOverride with TypedOverride? I could imagine something like this:

Agreed that this would be possible, and that it's probably preferable to delay this until we fix #1138.

> This doesn't worry me that much, since I imagine that we could simply create closures that close over the GlobalVar Mem argument when constructing LLVM TypedOverrides. Alternatively, if that proves too annoying, then we can always get and set the underlying MemImpl from within the override, similarly to how crucible-llvm does it [here](https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L380-L395).

Yeah, the only downside to this approach is that we'd end up with functions `GlobalVar Mem -> TypedOverride`, which makes the type signature of the override inaccessible without a `GlobalVar Mem`, even though it doesn't really depend on it. Might not be a problem in practice.

PR #1147, opened 2023-12-05
`crux-mir`: Support cross-compiling libs to `wasm32` target

qsctr:
- Add an optional argument to the `translate_libs.sh` script to specify a cross-compile target
- Only pass x86 SIMD flags to `memchr` build when supported
- Add the `dlmalloc` lib, which is required for building `std` on `wasm32-unknown-unknown`. Copied from https://github.com/alexcrichton/dlmalloc-rs/tree/0.2.3

`translate_libs.sh` now also depends on (a very simple usage of) `jq`. If that is undesirable I can find a way to remove it.

### DIFF
--- a/crux-mir/lib/dlmalloc/Cargo.toml
+++ b/crux-mir/lib/dlmalloc/Cargo.toml
@@ -1,0 +1,43 @@
+[package]
+name = "dlmalloc"
+version = "0.2.3"
+authors = ["Alex Crichton <alex@alexcrichton.com>"]
+license = "MIT/Apache-2.0"
+readme = "README.md"
+repository = "https://github.com/alexcrichton/dlmalloc-rs"
+homepage = "https://github.com/alexcrichton/dlmalloc-rs"
+documentation = "https://docs.rs/dlmalloc"
+description = """
+A Rust port of the dlmalloc allocator
+"""
+
+[package.metadata.docs.rs]
+features = ['global']
+
+[lib]
+doctest = false
+
+[target.'cfg(all(unix, not(target_arch = "wasm32")))'.dependencies]
+libc = { version = "0.2", default-features = false }
+
+[dependencies]
+# For more information on these dependencies see rust-lang/rust's
+# `src/tools/rustc-std-workspace` folder
+core = { version = '1.0.0', optional = true, package = 'rustc-std-workspace-core' }
+compiler_builtins = { version = '0.1.0', optional = true }
+
+[dev-dependencies]
+rand = "0.3"
+
+[profile.release]
+debug-assertions = true
+
+[features]
+# Enable implementations of the `GlobalAlloc` standard library API, exporting a
+# new `GlobalDlmalloc` as well which implements this trait.
+global = []
+
+# Enable very expensive debug checks in this crate
+debug = []
+
+rustc-dep-of-std = ['core', 'compiler_builtins/rustc-dep-of-std']
--- a/crux-mir/translate_libs.sh
+++ b/crux-mir/translate_libs.sh
@@ -2,21 +2,24 @@
 
 set -e
 
-TARGET=$1
+TARGET=${1:-$(rustc -Z unstable-options --print target-spec-json | jq -r '."llvm-target"')}
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 RLIBS_PARENT="${SCRIPT_DIR}/rlibs_real"
 RLIBS_SYMLINK="${SCRIPT_DIR}/rlibs"
-RLIBS=$(rustc ${TARGET:+ --target "$TARGET"} --print target-libdir --sysroot "${RLIBS_PARENT}")
-if [ "$TARGET" ]; then
-  read -r -d - STD_ENV_ARCH <<< "$TARGET"
-else
-  STD_ENV_ARCH=$(uname -m)
-fi
+RLIBS=$(rustc --target "$TARGET" --print target-libdir --sysroot "${RLIBS_PARENT}")
+
+IFS=- read -r TARGET_ARCH _TARGET_VENDOR TARGET_OS <<< "$TARGET"
+
+STD_ENV_ARCH=$TARGET_ARCH
 export STD_ENV_ARCH
 
+if [ "$TARGET_ARCH" = "x86_64" ]; then
+  IS_X86_64=yes
+fi
+
 translate() {
-    mir-json ${TARGET:+ --target "$TARGET"} -L "${RLIBS}" --out-dir "${RLIBS}" --crate-type rlib --remap-path-prefix "$(pwd -P)=." "$@"
+    mir-json --target "$TARGET" -L "${RLIBS}" --out-dir "${RLIBS}" --crate-type rlib --remap-path-prefix "$(pwd -P)=." "$@"
 }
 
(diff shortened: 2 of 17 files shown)


qsctr:
Hmm it seems like there is a reference to an apparently now-nonexistent flag `--no-model-internal-atomics` for `translate_libs.sh` [here](https://github.com/GaloisInc/crucible/blob/a8839cda3f1570ff921237ce48606a006cd26e0a/crux-mir/Concurrency.md?plain=1#L22). I should do something to prevent people from using it and it getting treated as a target triple by the script, but I'm not sure what is the right thing to do (add the flag back to `translate_libs.sh`? just delete that part of the docs? something else?)

RyanGlScott:
Bummer, I completely missed that `--no-model-internal-atomics` flag when I worked on https://github.com/GaloisInc/crucible/pull/1096. I doubt that it would be a small amount of work to add back to `crux-mir`, so I have opened https://github.com/GaloisInc/crucible/issues/1148 as a reminder to do this at some point in the future. As such, I don't think we need to block this PR on that front, although it would be worth setting `translate_libs.sh` up in such a way that we can pass multiple configuration-related arguments to the script. I'm not strongly opinionated on the best way to do this—perhaps via environment variables? `getopt`? Something else?

crux-mir/translate_libs.sh
```diff
@@ -2,15 +2,24 @@
 
 set -e
 
+TARGET=${1:-$(rustc -vV | sed -n 's/host: //p')}
```
RyanGlScott, line 5: It would be helpful to add some comments describing what these various commands are doing. For instance, I had to stare at this a while (and Google what `rustc -vV`'s output looks like) before I realized that it was scraping the host target triple from `rustc --version --verbose`'s output.

crux-mir/translate_libs.sh
```diff
@@ -2,15 +2,24 @@
...
+
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 RLIBS_PARENT="${SCRIPT_DIR}/rlibs_real"
 RLIBS_SYMLINK="${SCRIPT_DIR}/rlibs"
-RLIBS=$(rustc --print target-libdir --sysroot "${RLIBS_PARENT}")
-STD_ENV_ARCH=$(uname -m)
+RLIBS=$(rustc --target "$TARGET" --print target-libdir --sysroot "${RLIBS_PARENT}")
+
+IFS=- read -r TARGET_ARCH _TARGET_VENDOR TARGET_OS <<< "$TARGET"
```
RyanGlScott, line 12: Similarly, what is this command doing?

crux-mir/translate_libs.sh
```diff
@@ -84,8 +93,14 @@ translate lib/hashbrown/src/lib.rs --edition=2021 --crate-name hashbrown --cfg '
 echo 'Building addr2line...'
 translate lib/addr2line/src/lib.rs  --crate-name addr2line --cfg 'feature="alloc"' --cfg 'feature="compiler_builtins"' --cfg 'feature="core"' --cfg 'feature="rustc-dep-of-std"' --extern "compiler_builtins=${RLIBS}/libcompiler_builtins.rlib" --extern "gimli=${RLIBS}/libgimli.rlib" --extern "alloc=${RLIBS}/liballoc.rlib" --extern "core=${RLIBS}/libcore.rlib"
 
+if [ "$TARGET_ARCH" = "wasm32" ] && [ "$TARGET_OS" != "emscripten" ]; then
+  USE_DLMALLOC=yes
+  echo 'Building dlmalloc...'
+  translate lib/dlmalloc/src/lib.rs --crate-name dlmalloc --cfg 'feature="rustc-dep-of-std"' --extern "compiler_builtins=${RLIBS}/libcompiler_builtins.rlib" --extern "core=${RLIBS}/libcore.rlib"
+fi
```
RyanGlScott, line 100: Worth leaving a comment here saying what `dlmalloc` is and why it is only needed for `wasm32`. (Also, how did you figure out that this was needed?)

crux-mir/translate_libs.sh
```diff
@@ -84,8 +93,14 @@ translate lib/hashbrown/src/lib.rs --edition=2021 --crate-name hashbrown --cfg '
 echo 'Building addr2line...'
 translate lib/addr2line/src/lib.rs  --crate-name addr2line --cfg 'feature="alloc"' --cfg 'feature="compiler_builtins"' --cfg 'feature="core"' --cfg 'feature="rustc-dep-of-std"' --extern "compiler_builtins=${RLIBS}/libcompiler_builtins.rlib" --extern "gimli=${RLIBS}/libgimli.rlib" --extern "alloc=${RLIBS}/liballoc.rlib" --extern "core=${RLIBS}/libcore.rlib"
 
+if [ "$TARGET_ARCH" = "wasm32" ] && [ "$TARGET_OS" != "emscripten" ]; then
```
RyanGlScott, line 96: Why not `emscripten`? (Perhaps this was taken from another build script somewhere which took `emscripten` into account... if so, worth citing it here.)

crux-mir/lib/dlmalloc/Cargo.toml
```diff
@@ -0,0 +1,43 @@
+[package]
+name = "dlmalloc"
+version = "0.2.3"
```
RyanGlScott, line 3: Is this exactly the same as the upstream version of `dlmalloc-0.2.3`, or did you have to apply any patches to make it work with `mir-json`/`crux-mir`? If you did have to patch it, it would be worth putting a high-level description of the changes in `crux-mir/lib/Patches.md`.

RyanGlScott (commented):
Thanks, @qsctr! This mostly looks reasonable, although I would request that you add some additional documentation so that when we next update the `rustc` toolchain that `crux-mir` supports, we have some reminders about why `dlmalloc` needs to be included. (It's easy to overlook this sort of thing otherwise.)

crux-mir/lib/dlmalloc/Cargo.toml
qsctr: Yes it is exactly the same, no patches are required.

crux-mir/translate_libs.sh
qsctr: Heh, I didn't know that `-vV` could be written as `--version --verbose` either. I got it from [here](https://stackoverflow.com/questions/52996949/how-can-i-find-the-current-rust-compilers-default-llvm-target-triple) which got it from [cargo's source code](https://github.com/rust-lang/cargo/blob/000d279306bf30f7c6fc482a89978a51ba10d8e5/src/cargo/util/rustc.rs#L62).

qsctr:
Also, should we test a `wasm32-unknown-unknown` build in CI?

qsctr:
I don't think it's easy to run the test suite with `wasm32`, since it seems to use the binary output of `rustc` as an oracle, which would mean we need to run it in a wasm VM. In any case I don't think there is much runtime (rather than compile time) differences in behavior between targets that might be caused by some bug in `crux-mir`, so I was more thinking of just making sure `TARGET=wasm32-unknown-unknown ./translate_libs.sh` runs successfully.

RyanGlScott:
OK. In that case, I agree that a simple CI action that simply runs `TARGET=wasm32-unknown-unknown ./translate_libs.sh` would be appropriate. Would you be willing to open an issue as a a reminder to do this?

RyanGlScott (approved):

PR #1151, opened 2023-12-07
crucible: Introduce a helper to bind CFGs to their handles

langston-barrett:
A helper for a somewhat verbose but predictable pattern.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
+++ b/crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
@@ -31,6 +31,7 @@ module Lang.Crucible.Simulator.OverrideSim
   , getSymInterface
   , ovrWithBackend
   , bindFnHandle
+  , bindCFG
   , exitExecution
   , getOverrideArgs
   , overrideError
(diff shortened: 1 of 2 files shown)


crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
```diff
@@ -246,6 +247,12 @@ bindFnHandle ::
 bindFnHandle h s =
   stateContext . functionBindings %= FnBindings . insertHandleMap h s . fnBindings
 
+-- | Bind a CFG to its handle.
+--
+-- Computes postdominator information.
+bindCFG :: CFG ext blocks args ret -> OverrideSim p sym ext rtp a r ()
+bindCFG c = bindFnHandle (cfgHandle c) (UseCFG c (postdomInfo c))
```
RyanGlScott, line 254: Perhaps we should also have `bindOverride` (which leverages `UseOverride` instead of `UseCFG`) for symmetry.

RyanGlScott (approved):

crucible/src/Lang/Crucible/Simulator/OverrideSim.hs
RyanGlScott: Fair enough. I have no objections to leaving it out.

PR #1155, opened 2023-12-11
Add overrides for `fma`, `fmaf`, `llvm.fmuladd.*`, and `llvm.fma.*`

### LINKED ISSUE #1154: Add overrides for `fma` and friends
Clang 14 and later enable [`-ffp-contract=on`](https://releases.llvm.org/14.0.0/tools/clang/docs/UsersManual.html#cmdoption-ffp-contract) by default, which which cause combinations of multiplications/additions to be optimized into the [`llvm.fmuladd.*`](https://releases.llvm.org/14.0.0/docs/LangRef.html#llvm-fmuladd-intrinsic) intrinsic, even with `-O0`. Here is an example of a simple program that works when run with `crux-llvm` and Clang 13 and earlier, but fails with Clang 14+ because of this issue:

```c
double fmuladd(double a, double b, double c) {
  return a * b + c;
}

int main(void) {
  double d = fmuladd(1.0, 2.0, 3.0);
  return 0;
}
```
<details>

```
# With Clang 13.0.0
$ cabal run exe:crux-llvm -- test.c -O0
[Crux] Overall status: Valid.

# With Clang 14.0.0
$ cabal run exe:crux-llvm -- test.c -O0
[Crux] *** debug executable: results/test/debug-2
[Crux] *** break on line: 2
[Crux]   test.c:2:16: error: in fmuladd
[Crux]   Details:
[Crux]     Attempting to load callable function with type: double(double, double, double)
[Crux]       Via pointer: Global symbol "llvm.fmuladd.f64" (2, 0x0:[64])
[Crux]     In memory state:
[Crux]       Stack frame fmuladd
[Crux]         Allocations:
[Crux]           StackAlloc 9 0x8:[64] Mutable 8-byte-aligned test.c:1:43
[Crux]           StackAlloc 8 0x8:[64] Mutable 8-byte-aligned test.c:1:33
[Crux]           StackAlloc 7 0x8:[64] Mutable 8-byte-aligned test.c:1:23
[Crux]         Writes:
[Crux]             7 |->   *(7, 0x0:[64]) := 1
[Crux]             8 |->   *(8, 0x0:[64]) := 2
[Crux]             9 |->   *(9, 0x0:[64]) := 3
[Crux]       Stack frame main
[Crux]         Allocations:
[Crux]           StackAlloc 6 0x8:[64] Mutable 8-byte-aligned test.c:6:10
[Crux]           StackAlloc 5 0x4:[64] Mutable 4-byte-aligned internal
[Crux]         Writes:
[Crux]             5 |->   memset (5, 0x0:[64]) 0x0:[8] 0x4:[64]
[Crux]       Base memory
[Crux]         Allocations:
[Crux]           GlobalAlloc 4 0x0:[64] Immutable 1-byte-aligned [defined function ] main
[Crux]           GlobalAlloc 3 0x0:[64] Immutable 1-byte-aligned [defined function ] fmuladd
[Crux]           GlobalAlloc 2 0x0:[64] Immutable 1-byte-aligned [external function] llvm.fmuladd.f64
[Crux]           GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [external function] llvm.dbg.declare
[Crux]     in context:
[Crux]       fmuladd
[Crux]       main
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```
</details>

We should add overrides for `llvm.fmuladd.*`, the related intrinsic [`llvm.fma.*`](https://releases.llvm.org/14.0.0/docs/LangRef.html#int-fma), and the `fma`/`fmaf` functions from `libc`.

RyanGlScott:
Fixes #1154.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
@@ -738,6 +738,30 @@ llvmFloorfOverride =
   [llvmOvr| float @floorf( float ) |]
   (\_memOps sym args -> Ctx.uncurryAssignment (callFloor sym) args)
 
+llvmFmafOverride ::
+     forall sym p
+   . IsSymInterface sym
+  => LLVMOverride p sym
+        (EmptyCtx ::> FloatType SingleFloat
+                  ::> FloatType SingleFloat
+                  ::> FloatType SingleFloat)
+        (FloatType SingleFloat)
+llvmFmafOverride =
+  [llvmOvr| float @fmaf( float, float, float ) |]
+  (\_memOps bak args -> Ctx.uncurryAssignment (callFMA bak) args)
+
+llvmFmaOverride ::
+     forall sym p
+   . IsSymInterface sym
+  => LLVMOverride p sym
+        (EmptyCtx ::> FloatType DoubleFloat
+                  ::> FloatType DoubleFloat
+                  ::> FloatType DoubleFloat)
+        (FloatType DoubleFloat)
+llvmFmaOverride =
+  [llvmOvr| double @fma( double, double, double ) |]
+  (\_memOps bak args -> Ctx.uncurryAssignment (callFMA bak) args)
+
 
 -- math.h defines isinf() and isnan() as macros, so you might think it unusual
 -- to provide function overrides for them. However, if you write, say,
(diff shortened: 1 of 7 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
```diff
@@ -870,6 +894,18 @@ callFloor ::
...
+callFMA ::
+     forall fi p sym bak ext r args ret
+   . IsSymBackend sym bak
+  => bak
+  -> RegEntry sym (FloatType fi)
+  -> RegEntry sym (FloatType fi)
+  -> RegEntry sym (FloatType fi)
+  -> OverrideSim p sym ext r args ret (RegValue sym (FloatType fi))
+callFMA bak (regValue -> x) (regValue -> y) (regValue -> z) = liftIO $
+  iFloatFMA @_ @fi (backendGetSym bak) RNE x y z
```
langston-barrett, line 907: How do we know that `RNE` is the right mode here?

langston-barrett (approved):

kquick (approved):

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Libc.hs
RyanGlScott: IEEE 754 has [this](https://iremi.univ-reunion.fr/IMG/pdf/ieee-754-2008.pdf) to say:

> The roundTiesToEven rounding-direction attribute shall be the default rounding-direction attribute for
results in binary formats. The default rounding-direction attribute for results in decimal formats is language-
defined, but should be roundTiesToEven.

So for most use cases, `RNE` (Round to Nearest Even) will replicate the behavior you'd see most of the time in `libc`. That being said, it would be worth documenting this somewhere, since this is not even the only function where `RNE` is used without explanation. I'll push a separate commit to do that.

We should also make a note somewhere that `crucible-llvm` doesn't allow changing the rounding mode, which is the only scenario (AFAICT) where this behavior would ever differ.

PR #1157, opened 2023-12-12
`crux-mir`: Avoid pointer arithmetic in `vec::IntoIter`

### LINKED ISSUE #1144: `crux_mir`: Iterating over 0-length `Vec` fails with `tried to subtract pointers into different arrays`
When SAW is run on the following function with `count = 0`
```rust
fn sum(count: u32) -> u32 {
    let mut v1 = vec![];
    (0..count).for_each(|i| v1.push(i));
    let mut sum = 0;
    for i in v1 {
        sum += i;
    }
    sum
}
```
```
enable_experimental;

m <- mir_load_module "target/wasm32-unknown-unknown/debug/deps/scratch-c883c2110759b0e9.linked-mir.json";

let sum_spec count = do {
  mir_execute_func [mir_term {{ `count : [32] }}];
  mir_return (mir_term {{ sum [0..<count] : [32] }});
};

mir_verify m "scratch::sum" [] false (sum_spec 0) z3;
```
it fails with
```
[08:31:21.039] Loading file "/home/bretton/Documents/scratch/test_alloc.saw"
[08:31:22.457] Verifying scratch/aace4ee0::sum[0] ...
[08:31:22.465] Simulating scratch/aace4ee0::sum[0] ...
[08:31:22.470] Stack trace:
"mir_verify" (/home/bretton/Documents/scratch/test_alloc.saw:10:1-10:11)
Symbolic execution failed.
  ./lib/alloc/src/vec/into_iter.rs:210:39: 210:47: error: in alloc/0621aa31::vec[0]::into_iter[0]::{impl#5}[0]::size_hint[0]::_inst793e6cb55b43c40a[0]
  tried to subtract pointers into different arrays
```
We might need to do something similar to 69e92cb.

qsctr:
We do the same thing as 69e92cb for `vec::IntoIter`.

Fixes #1144.

### DIFF
--- a/crux-mir/lib/Patches.md
+++ b/crux-mir/lib/Patches.md
@@ -57,6 +57,11 @@ identify all of the code that was changed in each patch.
   in sync with `ptr` and `end`, and avoids the need for pointer
   arithmetic.
 
+* Avoid pointer arithmetic in `vec` `IntoIter` (last applied: December 11, 2023)
+
+  The same problem as above exists for `vec::into_iter::IntoIter::size_hint`,
+  and we fix it in the same way by adding a `len` field to `IntoIter`.
+
 * Implement `str::as_bytes` via `crucible_identity_transmute` (last applied: May 16, 2023)
 
   This is necessary to avoid a gnarly use of `transmute`.
--- a/crux-mir/lib/alloc/src/vec/into_iter.rs
+++ b/crux-mir/lib/alloc/src/vec/into_iter.rs
@@ -43,6 +43,7 @@ pub struct IntoIter<
     pub(super) end: *const T, // If T is a ZST, this is actually ptr+len. This encoding is picked so that
                               // ptr == end is a quick test for the Iterator being empty, that works
                               // for both ZST and non-ZST.
+    pub(super) len: usize,
 }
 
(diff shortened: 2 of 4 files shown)


qsctr:
@RyanGlScott Is there a place to put the test case? I saw there is `crux-mir/test/symb_eval/alloc` but the tests in there seem to be for `allocate` from the `crucible` crate and not the standard Rust `alloc`.

RyanGlScott:
Thanks, @qsctr!

> 

> Is there a place to put the test case? I saw there is `crux-mir/test/symb_eval/alloc` but the tests in there seem to be for `allocate` from the `crucible` crate and not the standard Rust `alloc`.

Indeed, all of the test cases in the `symb_eval` subdirectly rely on symbolic branching in some way. The test case in https://github.com/GaloisInc/crucible/issues/1144#issuecomment-1833765009, however, is entirely concrete. For that reason, I'd favor placing the test case under `crux-mir/tests/conc_eval`. I'm not picky about which `conc_eval` subdirectory to use—something like `conc_eval/vec` should be fine.

RyanGlScott (approved):

PR #1158, opened 2023-12-13
`crucible-mir`: Use correct identifiers for `TransCustom` intrinsics

### LINKED ISSUE #1129: `crucible-mir`: Incorrect `Mir.TransCustom` override for `rotate_left` (and others)
The [`rotate_left`](https://doc.rust-lang.org/std/intrinsics/fn.rotate_left.html) override in `Mir.TransCustom` is wrong:

https://github.com/GaloisInc/crucible/blob/0d26e4a656176d47dcc9fba9503de3c0d8a0219f/crucible-mir/src/Mir/TransCustom.hs#L894-L895

The path should actually be `["core", "intrinsics", "{extern}", "rotate_left"]`. The use of `""` instead of `"{extern}"` is a leftover from a previous `rustc` toolchain.

`rotate_left` is not the only function that suffers from this issue. [`rotate_right`](https://doc.rust-lang.org/std/intrinsics/fn.rotate_right.html), [`wrapping_add`](https://doc.rust-lang.org/std/intrinsics/fn.wrapping_add.html), and [`wrapping_sub`](https://doc.rust-lang.org/std/intrinsics/fn.wrapping_sub.html) are also afflicted by this, and likely several other functions on top of that. We should perform an audit and check any uses of `""` in `Mir.TransCustom`.

RyanGlScott:
This uses the correct identifiers for intrinsics, such as `core::intrinsics::{extern}::rotate_left`. Previously, the `"{extern}"` portion was written as `""` due to an oversight.

Fixes #1129.

### DIFF
--- a/crux-mir/test/conc_eval/intTest/test1129.rs
+++ b/crux-mir/test/conc_eval/intTest/test1129.rs
@@ -1,0 +1,14 @@
+//! Regression test for missing override for the `rotate_left` intrinsic
+
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> u32 {
+    let n: u32 = 0x10000b3u32;
+    let expected = 0xb301;
+    let actual = n.rotate_left(8);
+    assert!(expected == actual);
+    actual
+}
+
+pub fn main() {
+    println!("{:?}", crux_test());
+}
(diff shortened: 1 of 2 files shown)


crux-mir/test/conc_eval/intTest/test1129.rs
```diff
@@ -0,0 +1,12 @@
+#[cfg_attr(crux, crux::test)]
```
spernsteiner, line 1:
```suggestion
//! Regression test for missing override for the `rotate_left` intrinsic

#[cfg_attr(crux, crux::test)]
```
Or something like that, so the reader doesn't have to look up #1129 to understand what this test is for

spernsteiner (approved):

PR #1163, opened 2024-01-10
llvm: Export `CallStack` internals from an internal module

### LINKED ISSUE #1112: llvm: Check if call stack is empty
For the sake of pretty-printing, it would be helpful to be able to check if a [`CallStack`](https://github.com/GaloisInc/crucible/blob/3bdafe4fe1ed03e49a4255082cfe4d2e4a524f4b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/CallStack.hs) is empty.

langston-barrett:
Also fixes #1112.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/CallStack/Internal.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/CallStack/Internal.hs
@@ -33,6 +33,9 @@ newtype CallStack =
 cons :: FunctionName -> CallStack -> CallStack
 cons top (CallStack rest) = CallStack (top Seq.<| rest)
 
+null :: CallStack -> Bool
+null = Seq.null . runCallStack
+
 getCallStack :: MemState sym -> CallStack
 getCallStack =
   \case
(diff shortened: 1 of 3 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/CallStack/Internal.hs
```diff
@@ -0,0 +1,48 @@
...
+  deriving (Eq, Monoid, Ord, Semigroup)
+
+newtype CallStack =
+  CallStack { runCallStack :: Seq FunctionName }
+  deriving (Eq, Monoid, Ord, Semigroup)
+
+cons :: FunctionName -> CallStack -> CallStack
+cons top (CallStack rest) = CallStack (top Seq.<| rest)
+
+null :: CallStack -> Bool
```
RyanGlScott, line 36: Haddocks? (Nothing else in this module has Haddocks yet, unfortunately, but we might as well add them going forward.)

RyanGlScott (approved):
Make sure to include a reference to https://github.com/GaloisInc/crucible/issues/1112 in the message of the commit that introduces `null`.

PR #1165, opened 2024-01-19
Changes

andreistefanescu:


crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -613,6 +626,59 @@ readPartExpr bak (PE p v) msg = do
...
+  m (MapF (SymFnWrapper sym) (SymFnWrapper sym))
+runCHC bak uninterp_inv_fns  = liftIO $ do
+  let sym = backendGetSym bak
+
+  implications <- proofObligationsAsImplications bak
+  clearProofObligations bak
+
+  -- log to stdout
+  let logData = defaultLogData
+        { logCallbackVerbose = \_ -> putStrLn
```
RyanGlScott, line 646: Do we really want to always log things to `stdout`? I would expect the user to have to specify this manually if they want a higher-than-normal log setting.

crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -613,6 +626,59 @@ readPartExpr bak (PE p v) msg = do
...
+        , logReason = "SAW inv"
```
RyanGlScott, line 647: This code lives in Crucible, not SAW... shouldn't the `logReason` mention Crucible instead?

crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -613,6 +626,59 @@ readPartExpr bak (PE p v) msg = do
   addAssertion bak (LabeledPred p (SimError loc msg))
   return v
 
+
+-- | Run the CHC solver on the current proof obligations, and return the
+-- solution as a substitution from the uninterpreted functions to their
+-- definitions.
```
RyanGlScott, line 632: It might be worth mentioning the use of the BV<->LIA transformation somewhere in the Haddocks, since this part wasn't obvious to me until I read the source code carefully.

RyanGlScott (commented):
[Per the CI](https://github.com/GaloisInc/crucible/actions/runs/7583884073/job/20656455761?pr=1165#step:19:457), one `crux-llvm` test case's golden output needs to be updated to account for simplification changes made in https://github.com/GaloisInc/what4/pull/252:

```
     float-cast2.ll
      solver=z3
        loop-merging=loop
          clang-range=pre-clang13
            z3 4.8.14
              float-cast2
                float-cast2 crux run:                    OK (0.04s)
                float-cast2 crux result:                 OK
                float-cast2 crux output:                 FAIL
                  test/Test.hs:310:
                  MISMATCH for expected (test-data/golden/golden/float-cast2.z3.good)
                             and actual (test-data/golden/golden/float-cast2.z3.z3.out) output:
                      F-expect>|bvZext 32 (ite (floatIsNaN cX@3:f) 0x0:[1] 0x1:[1])
                      F-actual>|ite (floatIsNaN cX@3:f) 0x0:[32] 0x1:[32]
                      F        |[Crux] Overall status: Valid.
```

crucible/src/Lang/Crucible/Backend.hs
andreistefanescu: copy-pasta

Issue #1168, opened 2024-01-24
Don't `panic` in `AssumptionStack.popFrame`


langston-barrett:
Consider the following program, written against Crucible's public API:
```haskell
module Main where

import Prelude
import Data.Parameterized.Nonce
import Data.Parameterized.Some
import Lang.Crucible.Backend.AssumptionStack

main :: IO ()
main = do
  Some gen <- newIONonceGenerator
  stk <- initAssumptionStack gen
  frameId <- pushFrame stk
  _frameId <- pushFrame stk
  _result <- popFrame @() frameId stk
  putStrLn "No panic?"
```
This program panics:
```
%< ---------------------------------------------------
  Revision:  bb50c0d16922e13d6358ce9abc7042ed13feccab
  Branch:    master
  Location:  AssumptionStack.popFrame
  Message:   Push/pop mismatch in assumption stack!
             *** Current frame:  0
             *** Expected ident: 1
CallStack (from HasCallStack):
  panic, called at src/Lang/Crucible/Panic.hs:11:9 in crucible-0.6-inplace:Lang.Crucible.Panic
  panic, called at src/Lang/Crucible/Backend/AssumptionStack.hs:204:16 in crucible-0.6-inplace:Lang.Crucible.Backend.AssumptionStack
%< ---------------------------------------------------
```
I would argue that I have *not* found a bug in Crucible. More specifically, the only bug here is that this function calls `panic` when passed invalid arguments, but is exposed via a public API where such arguments may, in fact, be passed. [This panic](https://github.com/GaloisInc/crucible/blob/23cc439654abbc0c3714a7035d026453c234d9c4/crucible/src/Lang/Crucible/Backend/AssumptionStack.hs#L204) (and others like it) should either be reified in the return type (e.g., `Either`), or turned into a dedicated `Exception` instance (which should be noted in the Haddocks). Panics should be reserved for situations that are considered by Crucible to be *literally impossible*.

RyanGlScott:
I agree. It would nice to be able to recover more gracefully in situations like this, so this change would be welcome.

PR #1169, opened 2024-01-24
Refactor `AssumptionStack`

langston-barrett:
Towards #1168.

### DIFF
--- a/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
+++ b/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
@@ -267,15 +267,10 @@ gcAddAssumes :: Monoid asmp => asmp -> GoalCollector asmp goal -> GoalCollector 
 gcAddAssumes as' (CollectingAssumptions as gls) = CollectingAssumptions (as <> as') gls
 gcAddAssumes as' gls = CollectingAssumptions as' gls
 
-{- | Pop to the last push, or all the way to the top, if there were no more pushes.
-If the result is 'Left', then we popped until an explicitly marked push.
-If the result is 'Right', then we popped all the way to the top, and the
-result is the goal tree, or 'Nothing' if there were no goals. -}
-
 -- | A frame, popped by 'gcPop'
 data PoppedFrame asmp goal
   = PoppedFrame
-    { -- | The frame identifier of the popped frame,
+    { -- | The frame identifier of the popped frame
       poppedFrameId :: !FrameIdentifier
       -- | The assumptions that were forgotten by the pop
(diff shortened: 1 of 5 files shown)


langston-barrett:
It seems that pushing/popping operations are exposed via `IsSymBackend`. Ideally, we would only expose operations that automatically enforce the desired matching push/pop condition (basically, just an operation like `inFreshFrame`). However, I'm not really familiar with the code that uses this API (`Lang.Crucible.Backend.Online`), and it's not clear to me how feasible it would be to refactor it to fit this pattern. However, I think the current PR is already an improvement, in that it makes it *possible* to avoid `panic`ing functions in the public API.

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
```diff
@@ -266,22 +268,35 @@ gcAddAssumes as' (CollectingAssumptions as gls) = CollectingAssumptions (as <> a
...
-in that case we return:
-
-    1. the frame identifier of the popped frame,
-    2. the assumptions that were forgotten,
-    3. any proof goals that were generated since the frame push, and
-    4. the state of the collector before the push.
-
+If the result is 'Left', then we popped until an explicitly marked push.
 If the result is 'Right', then we popped all the way to the top, and the
 result is the goal tree, or 'Nothing' if there were no goals. -}
```
RyanGlScott, line 273: I think these Haddocks were meant to annotate `gcPop`, but now they no longer do so since the definitions of `PoppedFrame`/`collectPoppedGoals` were introduced. Can you move these Haddocks down to be next to `gcPop`?

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
```diff
@@ -266,22 +268,35 @@ gcAddAssumes as' (CollectingAssumptions as gls) = CollectingAssumptions (as <> a
...
 
+-- | A frame, popped by 'gcPop'
+data PoppedFrame asmp goal
+  = PoppedFrame
+    { -- | The frame identifier of the popped frame,
```
RyanGlScott, line 278:
```suggestion
    { -- | The frame identifier of the popped frame
```

RyanGlScott (approved):
Truth be told, I don't understand the frame-related code that well myself. But I agree that this is an improvement.

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
langston-barrett: Oh, whoops! Yep.

PR #1170, opened 2024-02-01
Prepare for Crux 0.8 releases

RyanGlScott:
This makes changes in preparation for releases of `crux-llvm-0.8` and `crux-mir-0.8`. For the first time ever, we are also planning to release the following libraries to Hackage:

* `crucible`
* `crucible-symio`
* `crucible-llvm`
* `crux`
* `crux-llvm`

This will make it possible to install `crux-llvm` directly from Hackage. To that end, I have performed some tidying up of these libraries' `.cabal` files to make them suitable for Hackage.


langston-barrett (approved):
I'm out of the office for the next two weeks, but this LGTM so far!

kquick (approved):

Issue #1013, opened 2022-07-18
`special-functions` test case in `crux-llvm` test case does not compile on macOS


RyanGlScott:
@Lukas-Dresel reports that the `special-functions.c` test case in `crux-llvm`'s test suite fails to compile on macOS:

```
loopmerge                                                                                                                                                                                                                          
  at-least-clang12                                                                                                                                                                                                                 
    z3 4.8.17                                                                                                                                                                                                                      
      special-functions                                                                                                                                                                                                            
        special-functions crux run:              OK (0.06s)                                                                                                                                                                        
        special-functions crux output:           FAIL                                                                                                                                                                              
          test/Test.hs:234:                                                                                                                                                                                                        
          MISMATCH for expected (test-data/golden/special-functions.good)                                                                                                                                                          
                     and actual (test-data/golden/special-functions.z3.out) output:                                                                                                                                                
              F-expect>|[Crux] Overall status: Valid.                                                                                                                                                                              
              F-actual>|`clang` compilation failed.                                                                                                                                                                                
              F-actual>|*** Exit code: 1                                                                                                                                                                                           
              F-actual>|*** Standard out:                                                                                                                                                                                          
              F-actual>|*** Standard error:                                                                                                                                                                                        
              F-actual>|   test-data/golden/special-functions.c:31:16: error: implicit declaration of function 'exp10' is invalid in C99 [-Werror,-Wimplicit-function-declaration]                                                 
              F-actual>|     double d04 = exp10(dx);                                                                                                                                                                               
              F-actual>|                  ^                                                                                                                                                                                        
              F-actual>|   test-data/golden/special-functions.c:93:15: error: implicit declaration of function 'exp10f' is invalid in C99 [-Werror,-Wimplicit-function-declaration]                                                
              F-actual>|     float f04 = exp10f(fx);                                                                                                                                                                               
              F-actual>|                 ^                                                                                                                                                                                         
              F-actual>|   test-data/golden/special-functions.c:93:15: note: did you mean 'exp10'?                                                                                                                                 
              F-actual>|   test-data/golden/special-functions.c:31:16: note: 'exp10' declared here                                                                                                                                 
              F-actual>|     double d04 = exp10(dx);                                                                                                                                                                               
              F-actual>|                  ^                                                                                                                                                                                        
              F-actual>|   2 errors generated.                                                                                                                                                                                     
                                                                                                                                                                                                                                   
[...]

PR #1172, opened 2024-02-13
CI: Run crux-llvm test suite in macOS

RyanGlScott:
Previously, we concocted a `clang-withIncl` script for running Clang on macOS CI runners, but this neglected to pass any arguments to Clang (among other issues). This patch replaces the `clang-withIncl` machinery with a much simpler, Homebrew-based setup for install Clang/LLVM. With this in place, we can finally run the `crux-llvm` test suite on macOS.

Fixes https://github.com/GaloisInc/crucible/issues/999.

-----

Running the test suite on macOS reveals several additional issues, among them #885 and #1013. This PR also applies a series of fixes for these issues.

### DIFF
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -70,6 +70,7 @@ install_llvm() {
     echo "CLANG=clang-12" >> "$GITHUB_ENV"
   elif [[ "$RUNNER_OS" = "macOS" ]]; then
     brew install llvm@12
+    echo "$(brew --prefix)/opt/llvm@12/bin" >> "$GITHUB_PATH"
   elif [[ "$RUNNER_OS" = "Windows" ]]; then
     choco install llvm
   else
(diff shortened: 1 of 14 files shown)


kquick (approved):
Removal of all the special case for MacOS testing is nice.  

I assume the generalization of zero  v.s. non-zero in isinf and isnan tests is due to different return values from the different OS environments?

RyanGlScott:
> I assume the generalization of zero v.s. non-zero in isinf and isnan tests is due to different return values from the different OS environments?

Yes. I added this to the commit message:

> On Ubuntu, `isinf(x)` will return `1` if `x` is positive infinity and `-1` if `x` is negative infinity. The C standard doesn't guarantee this, however, as it only requires that `isinf(x)` return _some_ non-zero value if `x` is infinite. More to the point, Apple Clang's version of `isinf` doesn't return the same values that Ubuntu's Clang does, but the `isinf.c` test case in the `crux-llvm` test suite specifically checks for these values, making the test fail on macOS.
> 
> This patch makes the checks in `isinf.c` more general so that they will pass regardless of which non-zero value gets returned from `isinf(x)`.

PR #1166, opened 2024-01-23
Sygus2

andreistefanescu:

PR #1178, opened 2024-03-01
SyGuS, match concrete size array

RyanGlScott:
This is a consolidation of the changes in #1074 and #1166. This also bumps the `what4` submodule to bring in the changes from https://github.com/GaloisInc/what4/pull/256.


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/MemLog.hs
```diff
@@ -743,5 +760,6 @@ concMem ::
   sym ->
   (forall tp. SymExpr sym tp -> IO (GroundValue tp)) ->
   Mem sym -> IO (Mem sym)
-concMem sym conc (Mem endian st) =
-  Mem endian <$> concMemState sym conc st
+concMem sym conc mem = do
+  conc_st <- concMemState sym conc $ mem ^. memState
+  return $ mem & memState .~ conc_st
```
kquick, line 765: Could these two lines just be `return $ mem %~ concMemState sym conc` ?

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpointCHC.hs
```diff
@@ -0,0 +1,1415 @@
...
+  callFrameContextUpdate hdl $
+    \ctx -> ctx { callFrameContextLoopHeaders = bid : callFrameContextLoopHeaders ctx }
+
+callFrameContextPop ::
+  CallFrameHandle init ret blocks ->
+  ExecutionFeatureContext sym wptr ext ->
+  ExecutionFeatureContext sym wptr ext
+callFrameContextPop  hdl =
+  callFrameContextUpdate hdl $
+    \ctx -> ctx { callFrameContextLoopHeaders = tail $ callFrameContextLoopHeaders ctx }
```
kquick, line 295: Since `tail` is partial, do we have any protection that we won't enter here without frame context loop headers?  Maybe a `panic` in that case would be helpful (although GHC 9.10's exception backtraces may finally make these easier to track down).

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpointCHC.hs
```diff
@@ -0,0 +1,1415 @@
...
+  MapF.fromList . catMaybes <$> mapM
+    (\(mem_loc, mem_entry) -> do
+      ptr <- memLocationPtr sym mem_loc
+      case mem_entry of
+        MemStoreFixpointEntry join_variable storage_type -> do
+          val <- C.doLoad bak mem ptr storage_type (C.LLVMPointerRepr $ W4.bvWidth join_variable) C.noAlignment
+          case W4.asNat (C.llvmPointerBlock val) of
+                Just 0 -> return $ Just $  MapF.Pair join_variable $ C.llvmPointerOffset val
+                _ -> fail $ "SimpleLoopFixpoint.loadMemJoinVariables: unexpected val:" ++ show (C.ppPtr val)
+          -- foo <- C.loadRaw sym mem ptr storage_type C.noAlignment
```
kquick, line 602: We should remove this commented-out section of code, or else add a prefix comment indicating what it represents and why it's being kept.

kquick: Nevermind: I see a lot of commented-out code below.  I suppose we should keep it at the moment as "notes" for future refinement rather than just stripping it all out right now; figuring out whether it contains useful information is a fairly large task and I don't want to hold this PR up too much at this stage.

crucible-llvm/src/Lang/Crucible/LLVM/SimpleLoopFixpointCHC.hs
```diff
@@ -0,0 +1,1415 @@
...
+import qualified Lang.Crucible.LLVM.MemModel as C
+import qualified Lang.Crucible.LLVM.MemModel.MemLog as C hiding (Mem)
+import qualified Lang.Crucible.LLVM.MemModel.Pointer as C
+import qualified Lang.Crucible.LLVM.MemModel.Type as C
+-- import qualified Lang.Crucible.LLVM.MemModel.Generic as C (writeArrayMem)
+
+
+-- | When live loop-carried dependencies are discovered as we traverse
+--   a loop body, new "widening" variables are introduced to stand in
+--   for those locations.  When we introduce such a varible, we
```
kquick, line 109: "varible"

kquick (approved):
A couple of small suggestions, but thanks for getting the ball across the line on this large set of changes!

Issue #1138, opened 2023-11-16
llvm: Remove the LLVM declaration from `LLVMOverride`


langston-barrett:
The `LLVMOverride` type has a `L.Declare`:

https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs#L78-L82

This is used to check that an override's "provided" declaration matches the declaration found in the LLVM module under the same name:

https://github.com/GaloisInc/crucible/blob/0834003f70242a65e6850a5e54c8d882c3106b6c/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs#L255-L276

Note, however that the equality check happens *modulo opaque pointers*, meaning that `T*` and `ptr` are treated as equal for all `T`. As LLVM has completed the migration to pointer types, none of the more-specific (non-opaque) in our overrides' `L.Declare`s add any additional type-safety beyond just using `ptr`. We can remove the `L.Declare` from `LLVMOverride` and instead create a function `TypeRepr t -> L.Type` that maps `LLVMPointerType sym w` to LLVM's `ptr`. This would obviate our LLVM quasiquoters, and allow us to specify the types of each override just once (as Crucbile-LLVM types).

langston-barrett:
Here's a first cut of functions to create an LLVM declaration from a Crucible-LLVM type signature:
```haskell
llvmType :: Mem.HasPtrWidth w => C.TypeRepr t -> Maybe L.Type
llvmType =
  \case
    C.AnyRepr {} -> Nothing
    C.BoolRepr -> Just (L.PrimType (L.Integer 1))
    C.CharRepr {} -> Nothing
    C.BVRepr w -> Just (intType w)
    C.ComplexRealRepr {} -> Nothing
    C.FloatRepr {} -> Nothing  -- TODO?
    C.FunctionHandleRepr {} -> Nothing
    C.IEEEFloatRepr {} -> Nothing  -- TODO?
    C.IntegerRepr {} -> Nothing
    C.MaybeRepr {} -> Nothing
    C.NatRepr {} -> Nothing
    C.RealValRepr {} -> Nothing
    C.RecursiveRepr {} -> Nothing
    C.ReferenceRepr {} -> Nothing
    C.SequenceRepr {} -> Nothing
    C.StringRepr {} -> Nothing
    C.StringMapRepr {} -> Nothing
[...]

PR #1184, opened 2024-03-21
crucible-llvm: Generalize `LLVMOverride`'s `ext` parameter

langston-barrett:
Previously, the `OverrideSim` action for an `LLVMOverride` was assumed to have a fixed `ext` parameter, namely, `LLVM`. While this is intuitively appealing, it turns out not to be necessary. This is because there are only a few operations in the `OverrideSim` monad that make essential use of the `ext` parameter, such as registering or calling a CFG in the given language extension. These operations are unlikely to be used when defining, e.g., overrides for libc functions.

The motivation for generalizing the parameter is to enable the reuse of `LLVMOverride`s in other Crucible language extensions that make use of the LLVM memory model, namely, Macaw.


langston-barrett:
Ah, whoops, this breaks something in UC-Crux.

langston-barrett:
Indeed, UC-Crux does actually create `LLVMOverride`s that use `callCFG` to call LLVM CFGs. The way I see it, there are two paths forward:

- Make `ext` a type parameter of `LLVMOverride`. Then the overrides in Crucible-LLVM would be polymorphic over `ext`, while the ones in UC-Crux would fix `ext` to `LLVM`. This is the most flexible solution, as it allows each override to choose a particular extension if it needs it, or not if not.
- Tear this functionality out of UC-Crux. As UC-Crux is not very useful since the introduction of opaque pointers and is unlikely to be modified to be so (see #1075), and the "check" overrides in question aren't a core part of its functionality, this is also fairly appealing.

@RyanGlScott any thoughts?

RyanGlScott:
I am fine with either option, but option (1) sounds slightly more appealing since we want to make `LLVMOverride` more like `TypedOverride` anyway (https://github.com/GaloisInc/crucible/issues/1138), and `TypedOverride` is already parameterized by `ext`.

kquick:
I'd also vote for option 1: I think it's probably too early to give up on uc-crux-llvm's potential.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -95,13 +95,13 @@ data LLVMOverride p sym args ret =
...
          forall rtp args' ret'.
-         OverrideSim p sym LLVM rtp args' ret' (RegValue sym ret)
+         OverrideSim p sym ext rtp args' ret' (RegValue sym ret)
     -- ^ The implementation of the intrinsic in the simulator monad
     -- (@OverrideSim@).
   }
 
 data SomeLLVMOverride p sym =
-  forall args ret. SomeLLVMOverride (LLVMOverride p sym args ret)
+  forall args ret. SomeLLVMOverride (LLVMOverride p sym LLVM args ret)
```
RyanGlScott, line 104: Would it make more sense for `SomeLLVMOverride` to be parameterized by `ext`, just like `LLVMOverride`? We _could_ force it to be `LLVM`, but I worry that someone will come along later and want the same generalizations that this PR is applying to `LLVMOverride`.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -83,7 +83,7 @@ import           Lang.Crucible.LLVM.Translation.Types
 
 -- | This type represents an implementation of an LLVM intrinsic function in
 -- Crucible.
-data LLVMOverride p sym args ret =
+data LLVMOverride p sym ext args ret =
```
RyanGlScott, line 86: Worth leaving a comment here about why `LLVMOverride` is parametric in `ext`, as it's a bit surprising if you aren't aware of the use cases.

RyanGlScott (approved):
Thanks! It would be worth advertising this change in the `crucible-llvm` changelog, as this change will [impact `saw-script`](https://github.com/GaloisInc/saw-script/blob/d56bc186983c4ed071c3cfe97fc9904830f64673/src/SAWScript/Crucible/LLVM/Builtins.hs#L1852-L1856). Otherwise, LGTM.

PR #1186, opened 2024-03-26
crucible-llvm: Make a list of libc overrides

langston-barrett:
Before this change, lists of LLVM overrides (e.g., `declare_overrides`) consisted of libc and LLVM-specific overrides grouped together. Because of the difficulties surrounding "polymorphic" LLVM overrides, they were also only available as `OverrideTemplate`s, which are necessarily dynamic (i.e., are essentially monadic actions in `OverrideSim` that register the override in question, rather than a data structure describing the override).

Macaw-based Crucible frontends might want to reuse the libc overrides in the context of binaries. This commit separates them out and lists them purely as `SomeLLVMOverride`s, facilitating this reuse.


RyanGlScott (commented):
I definitely support this direction of travel. I'm somewhat disappointed that this doesn't include the LLVM intrinsics overrides, however. At the very least, could we make a separate list for the simple ones that can be registered with `basic_llvm_override`?

langston-barrett:
> At the very least, could we make a separate list for the simple ones that can be registered with basic_llvm_override?

Will do!

PR #1188, opened 2024-03-26
crucible-llvm: Generalize pipe-fitting code to any language extension

langston-barrett:
This will be helpful for e.g. Macaw.


RyanGlScott (approved):
Please make a note of these API changes in the `crucible-llvm` changelog. Otherwise, LGTM.

Issue #1190, opened 2024-03-27
crux-mir doesn't handle byte -> char casts


sauclovian-g:
crux-mir doesn't handle byte -> char casts. This came up in Formal VerSo. It isn't a problematic conversion so it seems reasonable to just implement it.

sauclovian-g:
(crux-mir? crucible-mir? actually the latter, not entirely clear on the exact difference/boundary between them)

RyanGlScott:
> (crux-mir? crucible-mir? actually the latter, not entirely clear on the exact difference/boundary between them)

`crucible-mir` is the underlying symbolic execution machinery for MIR code, and `crux-mir` is a standalone tool built on top of `crucible-mir` meant for simulating Rust programs with inline assertions. SAW also depends on `crucible-mir`, but it does not depend on `crux-mir`.

RyanGlScott:
Here is a test case that demonstrates the issue using `crux-mir`:

```rs
pub fn b() -> u8 {
    97
}

#[crux::test]
pub fn test() -> char {
    b() as char
}
```
```
$ cabal run exe:crux-mir -- test.rs 
test test/2604af8f::test[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/2604af8f::test[0] counterexamples ----
[Crux] Found counterexample for verification goal
[Crux]   test.rs:7:5: 7:16: error: in test/2604af8f::test[0]
[Crux]   Translation error in test/2604af8f::test[0]: unimplemented cast: Misc
[Crux]     ty: TyUint B8
[Crux]     as: TyChar

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Invalid.
```

sauclovian-g:
Fixed by #1191.

PR #1191, opened 2024-03-27
Implement byte-to-char casts for crucible-mir.

sauclovian-g:


RyanGlScott:
Thanks, @sauclovian-g! I can believe that this fixes the issue, but I would feel better if we added a regression test alongside it. Can you add [this test case](https://github.com/GaloisInc/crucible/issues/1190#issuecomment-2022567787) to `crux-mir`'s test suite? I'm not too bothered about the exact location where we put the test—somewhere in `test/conc_eval/num` would work, for instance.

RyanGlScott:
@sauclovian-g, do you plan on finishing this?

sauclovian-g:
Yes, we need it for FV, just disorganized/behind on everything :-|

sauclovian-g:
This should be ready to go now.

(I think I should rebase it on master before merging, and probably squash the two commits together too, but probably you should look at it first)

RyanGlScott (approved):
Thanks for pushing this one over the finish line!

sauclovian-g:
I rebased it on master and squashed the commits together. Since it's small and isolated this is pretty harmless, and it's been sitting around for a while.

PR #1192, opened 2024-03-27
crucible-llvm: Return the allocated `FnHandle` from `bind_llvm_func`

langston-barrett:
This is useful in conjunction with
`Lang.Crucible.Simulator.OverrideSim.callOverride`.


langston-barrett:
Perhaps not necessary

PR #1193, opened 2024-03-27
crucible-llvm: Refactor and export override pipe-fitting code

langston-barrett:
The code in question converts between bitvectors and pointers so that it is easy and natural to specify overrides using bitvectors, while at runtime only LLVM pointers are used. The built-in overrides (e.g., for libc) all expect to be wrapped in this casting code. They're all exported, so to make them usable by downstream clients, we also need to export the code that does the pipe-fitting.

- [x] Factor it into its own module
- [x] Make it return errors instead of panicing
- [x] Possibly rename it
- [x] Add haddocks

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Cast.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Cast.hs
@@ -1,6 +1,6 @@
 -- |
 -- Module           : Lang.Crucible.LLVM.Intrinsics.Cast
--- Description      : Cast between bitvectors and pointers in signatuers
+-- Description      : Cast between bitvectors and pointers in signatures
 -- Copyright        : (c) Galois, Inc 2024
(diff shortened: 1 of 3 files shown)


langston-barrett:
I have tested this using a non-Crucible-LLVM frontend, and it appears to work as intended.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Cast.hs
```diff
@@ -0,0 +1,120 @@
+-- |
+-- Module           : Lang.Crucible.LLVM.Intrinsics.Cast
+-- Description      : Cast between bitvectors and pointers in signatuers
```
RyanGlScott, line 3:
```suggestion
-- Description      : Cast between bitvectors and pointers in signatures
```

RyanGlScott (approved):

PR #1197, opened 2024-04-24
llvm: Refactor override matching

langston-barrett:
Figuring out whether a given override should apply to a given `define`
or `declare` in an LLVM module is not as easy as seeing if their names
match. For some C++ overrides, for example, checking if the override
should apply requires demangling the function name, which is an
expensive proposition.

For this reason, override matching happens in two "phases" in
Crucible-LLVM: first, there is an initial, fast, string-based
`TemplateMatcher` that is used to filter out overrides that definitely
*don't* apply. Then, if any remain, there is a second phase that
inspects the declaration in more detail to see if the override matches.

Previously, this second phase consisted of a `RegOverrideM` (i.e.,
`OverrideSim`) action that would actually perform the registration.
This is unnecessarily flexible. The second phase is only really supposed
to perform the inspection of the declaration, construct an override, and
register it. Therefore, it's been replaced by a pure function that
returns a `Maybe (SomeLLVMOverride ...)`, which can then be registered
by client code, if desired.

This more restrictive type should help clarify the intent of the
override matching code, and reduce the potential for superfluous use
of `OverrideSim` effects.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
@@ -24,8 +24,6 @@ module Lang.Crucible.LLVM.Intrinsics.Common
   , llvmSizeT
   , llvmSSizeT
   , OverrideTemplate(..)
-  , TemplateMatcher(..)
-  , stripDarwinAliases
   , callStackFromMemVar'
     -- ** register_llvm_override
   , basic_llvm_override
@@ -49,8 +47,6 @@ import           Control.Lens
 import           Control.Monad.Reader (ReaderT, ask, lift)
 import           Control.Monad.Trans.Maybe (MaybeT)
 import qualified Data.List as List
-import qualified Data.List.Extra as List (stripInfix)
-import           Data.Maybe (fromMaybe)
 import qualified Data.Text as Text
 import           Numeric (readDec)
 import qualified System.Info as Info
@@ -78,6 +74,7 @@ import           Lang.Crucible.LLVM.Globals (registerFunPtr)
 import           Lang.Crucible.LLVM.MemModel
 import           Lang.Crucible.LLVM.MemModel.CallStack (CallStack)
 import qualified Lang.Crucible.LLVM.Intrinsics.Cast as Cast
+import qualified Lang.Crucible.LLVM.Intrinsics.Match as Match
 import           Lang.Crucible.LLVM.Translation.Monad
 import           Lang.Crucible.LLVM.Translation.Types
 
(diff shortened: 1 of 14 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -113,89 +107,36 @@ llvmSizeT = L.PrimType $ L.Integer $ fromIntegral $ natValue $ PtrWidth
...
+        Maybe (SomeLLVMOverride p sym ext)
+    }
+
+-- | Checking if an override applies to a given declaration happens in two
+-- \"phases\".
+--
+-- * An initial, quick, string-based 'Match.TemplateMatcher' checks if an
+--   override might apply to a given declaration, based on its name
+-- * If the 'Match.TemplateMatcher' does indeed match, the slower 'MakeOverride'
+--   performs additional checks and potentially constructs a 'SomeLLVMOverride'.
```
RyanGlScott, line 128: I assume these comments refer to `overrideTemplateMatcher` and `overrideTemplateAction`, respectively? If so, it would make sense just to make these Haddock comments for each field.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -233,58 +174,58 @@ build_llvm_override fnm args ret args' ret' llvmOverride =
...
-                         ovrDecl { L.decName = L.Symbol requestedNm }}
-
-               | otherwise
-               = ovr
-      register_llvm_override ovr'
+      MakeOverride $ \requestedDecl _ _ -> do
+        let L.Symbol requestedNm = L.decName requestedDecl
+        -- If we are on Darwin and the function name contains Darwin-specific
+        -- prefixes or suffixes, change the name of the override to the name
+        -- containing prefixes/suffixes. See Note [Darwin aliases] for an
```
RyanGlScott, line 219: `Note [Darwin aliases]` now lives in a different module, so make sure to say `See Note [Darwin aliases] in Lang.Crucible.LLVM.Intrinsics.Match [...]` to clarify.

crucible-llvm/src/Lang/Crucible/LLVM/SymIO.hs
```diff
@@ -557,7 +557,7 @@ callWriteFileHandle memOps fsVars filedesc buf count =
 symio_overrides
   :: (IsSymInterface sym, HasLLVMAnn sym, HasPtrWidth wptr, wptr ~ ArchWidth arch, ?memOpts :: MemOptions)
   => LLVMFileSystem wptr
-  -> [OverrideTemplate p sym arch rtp l a]
+  -> [OverrideTemplate p sym LLVM arch]
```
RyanGlScott, line 560: Any reason not to make this polymorphic in the `ext` type parameter, similar to the other LLVM overrides?

RyanGlScott (commented):
Thanks for looking into this. The override registration matching code has always made my head spin a bit, and this is a nice simplification.

Of course, do make sure to mention the API changes in the `crucible-llvm` changelog.

RyanGlScott (approved):

Issue #1177, opened 2024-02-21
`crucible-llvm`: Implement `llvm.vector.reduce.*` intrinsics (added in LLVM 12)


RyanGlScott:
Consider this program:

```c
#include <crucible.h>
#include <stddef.h>
#include <stdint.h>

int __attribute__((noinline)) memcmp_alt(const uint8_t *a, const uint8_t *b, size_t len) {
  uint8_t x = 0;

  for (size_t i = 0; i < len; i++) {
    x |= a[i] ^ b[i];
  }

  return x;
}

int main(void) {
  const uint8_t a[8] = {0};
  const uint8_t b[8] = {0};

  return memcmp_alt(a, b, 8);
}
```

`crux-llvm` is able to simulate this program with `-O2` when using Clang 10.0.0:

```
$ PATH=~/Software/clang+llvm-10.0.0/bin:$PATH ~/Software/crux-llvm-0.8/bin/crux-llvm -O2 test.c
[Crux] Using pointer width: 64 for file crux-build/crux~test.bc
[Crux] Simulating function main
[Crux] All goals discharged through internal simplification.
[Crux] Overall status: Valid.
```

If you use Clang 12.0.0 instead, however, then `crux-llvm` fails with:

```
$ PATH=~/Software/clang+llvm-12.0.0/bin:$PATH ~/Software/crux-llvm-0.8/bin/crux-llvm -O2 test.c
[Crux] Using pointer width: 64 for file crux-build/crux~test.bc
[Crux] Simulating function main
[Crux] Attempting to prove verification conditions.
[Crux] *** debug executable: results/test/debug-8
[Crux] *** break on line: 8
[Crux] Found counterexample for verification goal
[Crux]   test.c:8:3: error: in memcmp_alt
[Crux]   Failed to load function handle
[Crux]   Details:
[Crux]     No implementation or override found for pointer
[Crux]     The given pointer could not be resolved to a callable function
[Crux]     No implementation or override found for pointer
[Crux]     Attempting to load callable function with type: i32(<4 x i32>)
[Crux]       Via pointer: Global symbol "llvm.vector.reduce.or.v4i32" (4, 0x0:[64])
[Crux]     In memory state:
[Crux]       Stack frame memcmp_alt
[Crux]         No writes or allocations
[Crux]       Stack frame main
[Crux]         Allocations:
[Crux]           StackAlloc 8 0x8:[64] Mutable 8-byte-aligned test.c:17:3
[...]

PR #1201, opened 2024-05-09
`crucible-llvm`: Add integer-related `llvm.vector.reduce.*` intrinsics

RyanGlScott:
This patch:

* Adds the ability to register polymorphic LLVM overrides involving vector types. (Previously, the polymorphic LLVM machinery was able to handle overrides that vary over a single integer type, but they did not handle vector types with varying sizes.)
* Adds polymorphic overrides for all of the integer-related `llvm.vector.reduce.*` intrinsics, which work over any vector or integer size.
* Adds a `T1177.c` test case to the `crux-llvm` test suite which provides some assurance that the `crucible-llvm` semantics work as expected.

Fixes https://github.com/GaloisInc/crucible/issues/1177.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
@@ -251,6 +251,59 @@ poly1_llvm_overrides =
     )
   ]
 
+-- | An LLVM override that is polymorphic in a single integer argument
+-- (@intSz@) that is used in combination with a vector type, which can be of
+-- varying sizes (@vecSz@).
+newtype Poly1VecLLVMOverride p sym ext
+  = Poly1VecLLVMOverride
+      (forall vecSz intSz
+         . (1 <= intSz)
+        => NatRepr vecSz
+        -> NatRepr intSz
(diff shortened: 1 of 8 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
```diff
@@ -1288,6 +1341,113 @@ llvmX86_SSE2_storeu_dq =
...
+      => NatRepr intSz
+      -> RegEntry sym (VectorType (BVType intSz))
+      -> OverrideSim p sym ext r args ret (SymBV sym intSz))
+     -- ^ The semantics of the override.
+  -> NatRepr vecSz
+     -- ^ The size of the vector type.
+  -> NatRepr intSz
+     -- ^ The size of the integer type.
+  -> LLVMOverride p sym ext
+        (EmptyCtx ::> VectorType (BVType intSz))
```
kquick, line 1361: It may be useful in the future to create a `VectorSizedType vecSz (BVType intSz)` representation to correlate/constrain the arguments, but I don't think it's worth the impact to this PR.

kquick (approved):

PR #1207, opened 2024-06-04
Concretization: Getting concrete `RegValue`s from a model

langston-barrett:
What4 has the capacity to turn symbolic values into plausible "concrete" (/"ground") values, given a model from the SMT solver (`GroundEvalFn`). This commit builds on this feature to enable concretizing more complex Crucible types.

A possible use-case is to present concrete instances where safety assertions fail, e.g. when symbolically executing the following function:

    int f(int x, int y) { return x / (y - 2); }

it would be nice to say which particular values of `x` and `y` would cause either signed underflow (`y < INT_MIN + 2`) or division by zero (`y == 2`). This particular case could be handled by existing What4 functionality, but the same motivation applies to more complex cases involving Crucible-specific types.

### DIFF
--- a/crucible/src/Lang/Crucible/Concretize.hs
+++ b/crucible/src/Lang/Crucible/Concretize.hs
@@ -25,6 +25,7 @@
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
 
 module Lang.Crucible.Concretize
   ( ConcRegValue
@@ -94,18 +95,23 @@ type family ConcRegValue sym tp where
 -- * ConcCtx
 
 -- | Context needed for 'concRegValue'
-data ConcCtx sym =
-  ConcCtx
+--
+-- The @t@ parameter matches that on 'W4GE.GroundEvalFn' and 'Expr', namely, it
+-- is a phantom type brand used to relate nonces to a specific nonce generator
+-- (similar to the @s@ parameter of the @ST@ monad). It also appears as the
+-- first argument to 'ExprBuilder'.
+data ConcCtx sym t
+  = ConcCtx
   { -- | Model returned from SMT solver
-    model :: W4GE.GroundEvalFn sym
+    model :: W4GE.GroundEvalFn t
     -- | How to ground intrinsics
-  , intrinsicConcFuns :: MapF SymbolRepr (IntrinsicConcFn sym)
+  , intrinsicConcFuns :: MapF SymbolRepr (IntrinsicConcFn t)
   }
 
 -- | Helper, not exported
 ground ::
-  ConcCtx sym ->
-  Expr sym tp ->
+  ConcCtx sym t ->
+  Expr t tp ->
   IO (ConcRegValue sym (BaseToType tp))
 ground (ConcCtx (W4GE.GroundEvalFn ge) _) = ge
 
@@ -114,8 +120,8 @@ ground (ConcCtx (W4GE.GroundEvalFn ge) _) = ge
 
 -- | Helper, not exported
 ite ::
-  (SymExpr sym ~ Expr sym) =>
-  ConcCtx sym ->
+  (SymExpr sym ~ Expr t) =>
+  ConcCtx sym t ->
   W4I.Pred sym ->
   a ->
   a ->
@@ -126,8 +132,8 @@ ite ctx p t f = do
 
 -- | Helper, not exported
 iteIO ::
-  (SymExpr sym ~ Expr sym) =>
-  ConcCtx sym ->
+  (SymExpr sym ~ Expr t) =>
+  ConcCtx sym t ->
   W4I.Pred sym ->
   IO a ->
   IO a ->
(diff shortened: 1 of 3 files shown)


crucible/src/Lang/Crucible/Concretize.hs
```diff
@@ -0,0 +1,380 @@
...
+--
+-- Type families cannot appear partially applied.
+type ConcRegValue' :: Type -> CrucibleType -> Type
+newtype ConcRegValue' sym tp = ConcRegValue' { getConcRegValue' :: ConcRegValue sym tp }
+
+-- | Defines the \"concrete\" interpretations of 'CrucibleType' (as opposed
+-- to the \"symbolic\" interpretations, which are defined by 'RegValue'), as
+-- returned by 'concRegValue'.
+type ConcRegValue :: Type -> CrucibleType -> Type
+type family ConcRegValue sym tp where
```
RyanGlScott, line 74: I am very surprised that we have to pass a `sym` type parameter here. In my mind, `ConcRegValue` should exist independently of the symbolic backend, much like `what4`'s `GroundValue` does.

crucible/src/Lang/Crucible/Concretize.hs
```diff
@@ -0,0 +1,380 @@
...
+import qualified Lang.Crucible.Simulator.RegValue as RV
+import qualified Lang.Crucible.Simulator.SymSequence as SymSeq
+import qualified Lang.Crucible.Utils.MuxTree as MuxTree
+import           Lang.Crucible.Types
+
+-- | Newtype to allow partial application of 'ConcRegValue'.
+--
+-- Type families cannot appear partially applied.
+type ConcRegValue' :: Type -> CrucibleType -> Type
+newtype ConcRegValue' sym tp = ConcRegValue' { getConcRegValue' :: ConcRegValue sym tp }
```
RyanGlScott, line 68: Perhaps this should be:

```suggestion
newtype ConcRegValue' sym tp = ConcRV' { unConcRV' :: ConcRegValue sym tp }
```

To mirror the existing naming conventions that `RegValue'`/`RV`/`unRV` uses? (This is a matter of personal taste, so up to you.)

crucible/src/Lang/Crucible/Concretize.hs
```diff
@@ -0,0 +1,380 @@
...
+    (ReferenceRepr _, _) ->
+      concMux ctx v
+    (SequenceRepr tp', _) ->
+      concSymSequence ctx tp' v
+    (StringMapRepr tp', _) ->
+      concStringMap ctx tp' v
+    (VariantRepr tps, _) ->
+      concVariant ctx tps v
+
+    -- Incomplete clases
```
RyanGlScott, line 379:
```suggestion
    -- Incomplete cases
```

crucible/src/Lang/Crucible/Concretize.hs
```diff
@@ -0,0 +1,380 @@
...
+      traverse (fmap ConcRegValue' . concRegValue ctx tp') v
+
+    -- Cases with helper functions
+    (MaybeRepr tp', _) ->
+      concPartialWithErr ctx tp' v
+    (FunctionHandleRepr args ret, _) ->
+      concFnVal ctx args ret v
+    (IntrinsicRepr nm tyCtx, _) ->
+      case tryConcIntrinsic ctx nm tyCtx v of
+        Nothing -> fail "Missing concretization function for intrinsic!"
```
RyanGlScott, line 368: Perhaps report the name of the intrinsic here.

crucible/src/Lang/Crucible/Concretize.hs
langston-barrett: It's needed in this case:
```haskell
  ConcRegValue sym (FloatType fi) = W4GE.GroundValue (SymInterpretedFloatType sym fi)
```
I think the issue is that Crucible has "interpreted" floats, whereas What4's "base" floats are somehow different?

langston-barrett: A Haddock says:
```
   -- | A type index for floating point numbers, whose interpretation
   --   depends on the symbolic backend.
```
which makes it sound like this might be unavoidable. I could note the cause in the Haddock, though?

RyanGlScott: Ah, I missed the fact that `FloatType` is parameterized by a `FloatInfo`. In the case, I agree that passing a `sym` here is likely the most direct way of making this work. (It would be nice to comment that this is the one reason why we have to pass a `sym`.)

For context: both [`SymFloat`](https://hackage.haskell.org/package/what4-1.6/docs/What4-Interface.html#t:SymFloat) and [`SymInterpretedFloat`](https://hackage.haskell.org/package/what4-1.6/docs/What4-InterpretedFloatingPoint.html#t:SymInterpretedFloat) are What4 concepts. `SymFloat` always uses an IEEE-754 interpretation of symbolic floats, whereas `SymInterpretedFloat` can use IEEE-754, real numbers, or uninterpreted functions depending on how the symbolic backend is configured. The latter is more flexible, which is why Crucible uses it.

RyanGlScott (approved):
This looks like a nice utility to have when concretizing Crucible values. We could potentially go even further and offer feature parity with the API in `What4.Expr.GroundEval`, but perhaps we should wait until someone requests this before adding it.

PR #1210, opened 2024-06-07
Make `cabal sdist` work for `crucible-cli{,-llvm}`

### LINKED ISSUE #1209: `Error: cabal: sdist of crucible-cli-0.1: invalid file glob`, Cannot install `crux-mir`
I try to install `crux-mir`. But I get errors when trying to run the command:

```sh
$ cabal v2-install exe:crux-mir --overwrite-policy=always
Wrote tarball sdist to
/home/zhuyt/Repos/RUST/crucible/dist-newstyle/sdist/crucible-0.7.0.99.tar.gz
Error: cabal: sdist of crucible-cli-0.1: invalid file glob
'test-data/**/*.cbl'. Using the double-star syntax requires 'cabal-version:
2.4' or greater. Alternatively, for compatibility with earlier Cabal versions,
list the included directories explicitly.
```
However, `cabal v2-build` works well.

Then, I try to switch `LLVM`(from 12 to 16) and `GHC`, none of them help. Thanks for helping me to solve this problem.

RyanGlScott:
* Neither package's `.cabal` file contains a `README.md` file, so remove mention of this file name from the `extra-doc-files`.

* `crucible-cli.cabal` uses double-star syntax (`test-data/**/*.cbl`) in its `extra-source-files`, which requires a `cabal-version` or `2.4` or greater. Make sure that `crucible-cli.cabal` properly declares this version dependency.

Fixes #1209.

### DIFF
--- a/.github/workflows/crucible-go-build.yml
+++ b/.github/workflows/crucible-go-build.yml
@@ -112,6 +112,9 @@ jobs:
       - name: Configure
         shell: bash
         run: .github/ci.sh configure crux-go
+      - name: Generate source distribution
+        shell: bash
+        run: cabal sdist crucible-go
       - name: Build
         shell: bash
         run: .github/ci.sh build exe:crux-go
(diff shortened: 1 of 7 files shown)


RyanGlScott:
> I wonder if there's a way we could catch such mistakes in CI?

The most direct way would be to run `cabal sdist` on each package, which should be relatively cheap to do. Let me see how difficult it would be to add that to the current CI setup...

PR #1213, opened 2024-06-07
Use `Seq` instead of lists when concretizing `SymSequence`

### LINKED ISSUE #1212: Improve concretization of sequences
Concretization of sequences currently uses `++`, which is slow:

https://github.com/GaloisInc/crucible/blob/976a16e44668fdc81ea3d2327a79541da4762c04/crucible/src/Lang/Crucible/Concretize.hs#L297

We should switch the following case to use `Seq`:

https://github.com/GaloisInc/crucible/blob/976a16e44668fdc81ea3d2327a79541da4762c04/crucible/src/Lang/Crucible/Concretize.hs#L98

The same problem applies here:

https://github.com/GaloisInc/crucible/blob/976a16e44668fdc81ea3d2327a79541da4762c04/crucible/src/Lang/Crucible/Simulator/SymSequence.hs#L402

We should make that function use `Seq` as well. It's exported, but its name has a typo, so we should make the `Seq` version with the correctly-spelled name, and add a deprecation warning to the `[]` version with the incorrect name. We can then use this function in the implementation of concretization of sequences.

langston-barrett:
To avoid the slow `++`. Fixes #1212.

### DIFF
--- a/crucible/CHANGELOG.md
+++ b/crucible/CHANGELOG.md
@@ -1,3 +1,7 @@
+# next
+
+* Deprecate `concreteizeSymSequence` in favor of `concretizeSymSequence`
+
 # 0.7 -- 2024-02-05
 
 * Add `TypedOverride`, `SomeTypedOverride`, and `runTypedOverride` to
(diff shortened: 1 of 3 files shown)


RyanGlScott (approved):
Make sure to mention `concreteizeSymSequence`'s deprecation in the `crucible` changelog. Otherwise, LGTM.

PR #1214, opened 2024-06-10
Work around LLVM's reltable lookup optimization

### LINKED ISSUE #1174: `crucible-llvm`: Support string tables with Clang 14.0.0 + optimizations
Consider this code, which was minimized from a real-world example:

```c
const char *F(int tag) {
  static const char *const table[] = {
    "A",
    "B",
  };

  return table[tag];
}

int main() {
  return 0;
}
```

If you run this through `crux-llvm` with optimizations and Clang 14.0.0, it will crash:

```
$ ~/Software/crux-llvm-0.8/bin/crux-llvm -O1 test.c
user error (Encountered error while processing global @reltable.F: Illegal operation applied to pointer argument)
```

It will pass if:

* Optimizations are disabled, or
* You use Clang 12.0.0 instead of 14.0.0

The problematic bitcode in the Clang 14.0.0 + optimizations version is:

```ll
@reltable.F = internal unnamed_addr constant [2 x i32] [i32 trunc (i64 sub (i64 ptrtoint ([2 x i8]* @.str to i64), i64 ptrtoint ([2 x i32]* @reltable.F to i64)) to i32), i32 trunc (i64 sub (i64 ptrtoint ([2 x i8]* @.str.1 to i64), i64 ptrtoint ([2 x i32]* @reltable.F to i64)) to i32)], align 4
@.str = private unnamed_addr constant [2 x i8] c"A\00", align 1
@.str.1 = private unnamed_addr constant [2 x i8] c"B\00", align 1
```

Without optimizations (or with Clang 12.0.0), we instead have:

```ll
@F.table = internal constant [2 x i8*] [i8* getelementptr inbounds ([2 x i8], [2 x i8]* @.str, i32 0, i32 0), i8* getelementptr inbounds ([2 x i8], [2 x i8]* @.str.1, i32 0, i32 0)], align 16, !dbg !0
@.str = private unnamed_addr constant [2 x i8] c"A\00", align 1
@.str.1 = private unnamed_addr constant [2 x i8] c"B\00", align 1
```

RyanGlScott:
Clang 14.0.0+ introduce a relative lookup table optimization that produces code that subtracts two pointers from different allocation regions, which is something that `crucible-llvm` is ill equipped to reason about. For lack of a viable alternative, we work around this issue by "undoing" this optimization before translating the code to Crucible. For the full details, see the new `Note [Undoing LLVM's relative table lookup conversion pass]` in `Lang.Crucible.LLVM.Globals`.

Fixes #1174.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Globals.hs
@@ -46,7 +46,7 @@ import           Control.Monad (foldM)
 import           Control.Monad.IO.Class (MonadIO(..))
 import           Control.Monad.Except (MonadError(..))
 import           Control.Lens hiding (op, (:>) )
-import           Data.List (foldl')
+import           Data.List (foldl', genericLength, isPrefixOf)
 import           Data.Map.Strict (Map)
 import qualified Data.Map.Strict as Map
 import qualified Data.Set as Set
@@ -107,7 +107,7 @@ makeGlobalMap :: forall arch wptr. (?lc :: TypeContext, HasPtrWidth wptr)
               => LLVMContext arch
               -> L.Module
               -> GlobalInitializerMap
-makeGlobalMap ctx m = foldl' addAliases globalMap (Map.toList (llvmGlobalAliases ctx))
+makeGlobalMap ctx m = foldl' addAliases globalMap1 (Map.toList (llvmGlobalAliases ctx))
 
   where
    addAliases mp (glob, aliases) =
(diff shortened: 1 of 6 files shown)


kquick (approved):
There's a lot of very specific pattern recognition in `buildLoadRelConstInitMap`, which is good because it limits the functionality to a very specific case and prevents capturing unrelated scenarios, but for the same reasons this is also very brittle and I feel like we haven't seen the last of this.  Nonetheless, I agree that this seems like the best way I can think of to handle this LLVM shortcut.

langston-barrett (approved):

PR #1215, opened 2024-06-18
Helpers for checking proof obligations

langston-barrett:
We check proof obligations that come from Crucible a lot:

https://github.com/GaloisInc/crucible/blob/f9760bdb9921523848efc9889fbd72bd6dea506e/crucible-symio/tests/TestMain.hs#L151C3-L151C31

https://github.com/GaloisInc/macaw/blob/645754178a034c22e97aac49a962c99b387aaa01/symbolic/src/Data/Macaw/Symbolic/Testing.hs#L330

https://github.com/GaloisInc/surveyor/blob/96b6748d811bc2ab9ef330307a324bd00e04819f/crux-dbg/src/Crux/Debug/LLVM/Overrides.hs#L161

https://github.com/GaloisInc/mss/blob/ca97f1bc6b296b30bcccad3b4098544d91427103/grackle/src/Grackle/Overrides/Utils.hs#L84

https://github.com/GaloisInc/ambient-verifier/blob/eab04abb9750825a25ec0cbe0379add63f05f6c6/src/Ambient/Verifier/Prove.hs#L137

As seen above, most of these instances use somewhat low-level What4 functions, and require thinking about the encoding of proof assumptions and obligations into a SMT-compatible format (in particular, they inline the material conditional). This PR introduces more high-level helpers. Doing so required a bit of refactoring.

### DIFF
--- a/crucible/crucible.cabal
+++ b/crucible/crucible.cabal
@@ -72,6 +72,9 @@ library
 
   hs-source-dirs: src
 
+  other-modules:
+    Lang.Crucible.Backend.Assumptions
+
   exposed-modules:
     Lang.Crucible.Analysis.DFS
     Lang.Crucible.Analysis.ForwardDataflow
--- a/crucible/crucible.cabal
+++ b/crucible/crucible.cabal
@@ -74,7 +74,6 @@ library
 
   other-modules:
     Lang.Crucible.Backend.Assumptions
-    Lang.Crucible.Backend.Prove
 
   exposed-modules:
     Lang.Crucible.Analysis.DFS
@@ -87,6 +86,7 @@ library
     Lang.Crucible.Backend.AssumptionStack
     Lang.Crucible.Backend.ProofGoals
     Lang.Crucible.Backend.Online
+    Lang.Crucible.Backend.Prove
     Lang.Crucible.Backend.Simple
     Lang.Crucible.Concretize
     Lang.Crucible.CFG.Common
--- a/crucible/crucible.cabal
+++ b/crucible/crucible.cabal
@@ -74,6 +74,7 @@ library
 
   other-modules:
     Lang.Crucible.Backend.Assumptions
+    Lang.Crucible.Backend.Prove
 
   exposed-modules:
     Lang.Crucible.Analysis.DFS
--- a/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
+++ b/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
@@ -18,7 +18,6 @@ module Lang.Crucible.Backend.ProofGoals
   , traverseGoals, traverseOnlyGoals
   , traverseGoalsWithAssumptions
   , traverseGoalsSeq
-    -- ** Proving goals
 
     -- * Goal collector
   , FrameIdentifier(..), GoalCollector
--- a/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
+++ b/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
@@ -46,14 +46,6 @@ import           Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import           Data.Word
 
-import qualified What4.Interface as W4
-import qualified What4.Expr as WE
-import qualified What4.Config as W4C
-import qualified What4.Solver.Yices as W4Y
-import qualified What4.Solver.Adapter as WSA
-import qualified What4.SatResult as W4R
-import qualified What4.Partial as W4
-
 -- | A proof goal consists of a collection of assumptions
 --   that were in scope when an assertion was made, together
 --   with the given assertion.
--- a/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
+++ b/crucible/src/Lang/Crucible/Backend/ProofGoals.hs
@@ -18,6 +18,7 @@ module Lang.Crucible.Backend.ProofGoals
   , traverseGoals, traverseOnlyGoals
   , traverseGoalsWithAssumptions
   , traverseGoalsSeq
+    -- ** Proving goals
 
     -- * Goal collector
   , FrameIdentifier(..), GoalCollector
@@ -44,6 +45,14 @@ import           Data.Functor.Const
 import           Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import           Data.Word
+
+import qualified What4.Interface as W4
+import qualified What4.Expr as WE
+import qualified What4.Config as W4C
+import qualified What4.Solver.Yices as W4Y
+import qualified What4.Solver.Adapter as WSA
+import qualified What4.SatResult as W4R
+import qualified What4.Partial as W4
 
 -- | A proof goal consists of a collection of assumptions
 --   that were in scope when an assertion was made, together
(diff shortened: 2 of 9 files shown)


crucible/crucible.cabal
```diff
@@ -72,6 +72,9 @@ library
 
   hs-source-dirs: src
 
+  other-modules:
+    Lang.Crucible.Backend.Assumptions
```
kquick, line 76: Is there a reason not to expose this module?

langston-barrett: Not really, I don't have a strong sense of when to make modules public vs. private, just opted for private because it exposes less of a surface area for people to rely on (e.g., keeping it private lets us reorganize the code further, so long as we keep exporting it from `Backend`).

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -0,0 +1,135 @@
...
+  let goalPred = goal ^. CB.labeledPred
+  asmsPred <- CB.assumptionsPred sym asms
+  notGoal <- W4.notPred sym goalPred
+  let goal' = CB.ProofGoal asms goal
+  WSA.solver_adapter_check_sat adapter sym ld [asmsPred, notGoal] $
+    k .
+      \case
+        W4R.Sat (gfn, binds) -> Disproved goal' gfn binds
+        W4R.Unsat () -> Proved goal'
+        W4R.Unknown -> Unknown goal'
```
RyanGlScott: I'd leave some comments here about how we're going about proving the goal, since it may be somewhat counterintuitive at first to see the `Sat` case correspond to `Disproved` and the `Unsat` case correspond to `Proved`. Specifically, the approach is:

1. Gather all of the assumptions currently in scope.
2. Negate the goal that were are trying to prove.
3. Attempt to prove the negated goal in the current scope. If this negated goal is satisfiable, then there exists a counterexample that makes the original goal false, so we have disproven the goal. If the negated goal is unsatisfiable, on the other hand, then the original goal is proven.

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -0,0 +1,135 @@
...
+     -- | The goal was disproved, and a model that falsifies it is available.
+     --
+     -- Corresponds to 'W4R.Sat'.
+   | Disproved (Goal sym) (WE.GroundEvalFn t) (Maybe (WE.ExprRangeBindings t))
+     -- | The SMT solver returned \"unknown\".
+     --
+     -- Corresponds to 'W4R.Unknown'.
+   | Unknown (Goal sym)
+
+-- | Prove a single goal.
```
RyanGlScott, line 50: Perhaps say:

```suggestion
-- | Prove a single goal (i.e., an 'CB.Assertion') under the supplied 'Assumptions'.
```

It took me a bit to figure out what the difference between `proveGoal` and `proveProofGoal` is: it's the data types involved. Calling out the data types in the Haddocks makes this difference a little more obvious, in my opinion.

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -0,0 +1,135 @@
...
+import           Lang.Crucible.Backend.Assumptions (Assumptions)
+import           Lang.Crucible.Backend.ProofGoals (traverseGoalsWithAssumptions)
+
+type Goal sym = CB.ProofGoal (CB.Assumptions sym) (CB.Assertion sym)
+
+-- | The result of attempting to prove a goal with an SMT solver.
+--
+-- The constructors of this type correspond to those of 'W4R.SatResult'.
+data ProofResult sym t
+   = -- | The goal was proved
```
RyanGlScott, line 37:
```suggestion
   = -- | The goal was proved.
```

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
```diff
@@ -18,6 +18,7 @@ module Lang.Crucible.Backend.ProofGoals
   , traverseGoals, traverseOnlyGoals
   , traverseGoalsWithAssumptions
   , traverseGoalsSeq
+    -- ** Proving goals
```
RyanGlScott, line 22: There's nothing under this Haddock section header. Is this intentional?

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -0,0 +1,132 @@
...
+
+-- | Prove a collection of 'CB.Goals'.
+proveGoals ::
+  (Monoid m, sym ~ WE.ExprBuilder t st fs) =>
+  W4.IsSymExprBuilder sym =>
+  sym ->
+  WSA.LogData ->
+  WSA.SolverAdapter st ->
+  CB.Goals (CB.Assumptions sym) (CB.Assertion sym) ->
+  -- | Continuation to process the 'ProofResult'.
```
RyanGlScott, line 95: I think this is processing multiple `ProofResult`s, right? Should this be:

```suggestion
  -- | Continuation to process each 'ProofResult'.
```

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -0,0 +1,132 @@
...
+      goals
+
+-- | Prove a collection of 'CB.ProofObligations'.
+proveObligations ::
+  (Monoid m, sym ~ WE.ExprBuilder t st fs) =>
+  sym ->
+  WSA.LogData ->
+  WSA.SolverAdapter st ->
+  CB.ProofObligations sym ->
+  -- | Continuation to process the 'ProofResult'.
```
RyanGlScott, line 111: Should this be:

```suggestion
  -- | Continuation to process each 'ProofResult'.
```

langston-barrett:
This PR moves in the direction of *deduplicating* code that proves goals. In the future, it would be nice to *improve* these helpers by:

- Allowing "fast failure" on conjunctions of goals (i.e., not attempting to prove the RHS if the LHS is disproved)
- Parallelization and timeouts of solver queries (see the ambient-verifier link above)
- Use of `push`/`pop` in an online solver, rather than distributing assumptions into goals

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -0,0 +1,132 @@
...
+    Just goals -> proveGoals sym ld adapter goals k
+
+-- | Prove a the current collection of 'CB.ProofObligations' associated with the
+-- symbolic backend (retrieved via 'CB.getProofObligations').
+proveCurrentObligations ::
+  (Monoid m, CB.IsSymBackend sym bak, sym ~ WE.ExprBuilder t st fs) =>
+  bak ->
+  WSA.LogData ->
+  WSA.SolverAdapter st ->
+  -- | Continuation to process the 'ProofResult'.
```
RyanGlScott, line 126: Should this be:

```suggestion
  -- | Continuation to process each 'ProofResult'.
```

RyanGlScott (approved):
This looks like a nice cleanup. I only have minor suggestions.

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
langston-barrett: Whoops, no, that's an oversight! Thanks

PR #1216, opened 2024-06-20
Make obligation checking more configurable

langston-barrett:
- Enable the use of online solvers
- Enable clients to select between "failing fast" vs. "keeping going"
- Timeouts for offline solvers
- ~~Parallel proving~~

### DIFF
--- a/crucible/src/Lang/Crucible/Backend/Prove.hs
+++ b/crucible/src/Lang/Crucible/Backend/Prove.hs
@@ -133,6 +133,9 @@ consumeGoalsWithAssumptions onGoal onConj goals =
 -- | The result of attempting to prove a goal with an SMT solver.
 --
 -- The constructors of this type correspond to those of 'W4R.SatResult'.
+--
+-- * @sym@ is the symbolic backend, usually 'WE.ExprBuilder'
+-- * @r@ is the return type of the eventual 'ProofConsumer'
 data ProofResult sym t
    = -- | The goal was proved.
      --
@@ -143,6 +146,9 @@ data ProofResult sym t
      -- The 'WE.GroundEvalFn' is only available for use during the execution of
      -- a 'ProofConsumer'. See 'WSA.SolverAdapter'.
      --
+     -- The @'Maybe' 'WE.ExprRangeBindings'@ are 'Just' when using
+     -- 'offlineProve' and 'Nothing' when using 'onlineProve'.
+     --
      -- Corresponds to 'W4R.Sat'.
    | Disproved (WE.GroundEvalFn t) (Maybe (WE.ExprRangeBindings t))
      -- | The SMT solver returned \"unknown\".
@@ -151,6 +157,11 @@ data ProofResult sym t
    | Unknown
 
 -- | A 'ProofStrategy' dictates how results are proved.
+--
+-- * @sym@ is the symbolic backend, usually 'WE.ExprBuilder'
+-- * @m@ is the monad in which the 'Prover' and 'Combiner' run
+-- * @t@ is the parameter to 'WE.Expr'
+-- * @r@ is the return type of the eventual 'ProofConsumer'
 data ProofStrategy sym m t r
   = ProofStrategy
     { -- | Generally 'offlineProver' or 'onlineProver'
@@ -162,6 +173,9 @@ data ProofStrategy sym m t r
 --
 -- If the result is 'Disproved', then this function must consume the
 -- 'WE.GroundEvalFn' before returning. See 'WSA.SolverAdapter'.
+--
+-- * @sym@ is the symbolic backend, usually 'WE.ExprBuilder'
+-- * @t@ is the parameter to 'WE.Expr'
 newtype ProofConsumer sym t r
   = ProofConsumer (CB.ProofObligation sym -> ProofResult sym t -> IO r)
 
@@ -178,6 +192,9 @@ data SubgoalResult r
   deriving Functor
 
 -- | How to combine results of proofs, used as part of a 'ProofStrategy'.
+--
+-- * @m@ is the monad in which the 'Prover' and 'Combiner' run
+-- * @r@ is the return type of the eventual 'ProofConsumer'
 newtype Combiner m r
   = Combiner
     { getCombiner ::
--- a/crucible/src/Lang/Crucible/Backend/Prove.hs
+++ b/crucible/src/Lang/Crucible/Backend/Prove.hs
@@ -71,7 +71,6 @@ module Lang.Crucible.Backend.Prove
     -- *** Offline
   , offlineProve
   , offlineProveWithTimeout
-  , offlineProveWithTimeoutIO
   , offlineProver
     -- *** Online
   , onlineProve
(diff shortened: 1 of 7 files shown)


crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -3,32 +3,212 @@ Module      : Lang.Crucible.Backend.Prove
...
+  solvers ('onlineProver') can share state as assumptions are added. See the
+  top-level README for What4 for further discussion of this choice.
+* Failing fast ('failFast') vs. keeping going ('keepGoing')
+* Timeouts: Proving with timeouts ('offlineProveWithTimeout') vs. without
+  ('offlineProve')
+
+Once an appropriate strategy has been selected, it can be passed to entrypoints
+such as 'proveObligations' to dispatch proof obligations.
+
+== Parallelism
```
langston-barrett, line 24: Having worked on it for a bit, I'm not sure it actually makes sense to export helpers for running proofs in parallel from this module. The upsides are:

- Parallel proving is complex, it's nice to have a single, centralized implementation.
- The approach used here will probably work for a variety of clients.

But there are some downsides as well:

- There are some knobs that are currently not exposed. This means this implementation won't work for everybody, but exposing them would make the API more complex. Examples include the delay when polling threads, and the maximum number of threads.
- The implementation is complex, so perhaps hard to understand, extend, and maintain.
- The implementation has some limitations (e.g., it doesn't support failing fast, in part because this would require a more complex interface that wouldn't match the single-threaded version).

More featureful clients might just want to roll their own custom traversal of `Goals` which incorporates whatever features they need.

When reviewing, keep in mind that I'd be aligned with simply reverting the commits having to do with parallelism.

sauclovian-g: The following are vague generalities you might want to just ignore but might be somewhat useful for thinking about this:

- It seems that it should be possible to roll config settings (like delay values) into the API in a way that allows basic users to mostly ignore them. Haskell doesn't make this easy, but it should still be possible...
- In general providing an incomplete/special case library that some users will need to reimplement doesn't seem all that desirable: it incurs most of the costs (both immediate and long-term) without actually solving the problem for keeps. Furthermore, the reimplementations will probably end up being cut-paste-and-edit which isn't good for anyone in the long term. I think, though, this doesn't mean "don't do it", it means "go ahead and do it right".
- If doing it right involves more work than you're prepared to do right now, the right positioning is probably not "some clients will need to reimplement this" but "for some users we will need to extend this"; document the limitations and a sketch of what's entailed in lifting them.
- If none of the above is feasible it's probably also better to split the just parts that would need to be rewritten into a separate module (or even a separate library) so that it's clear where the boundaries are.

It also might make sense to split the parallel bits into a separate branch so they're more readily separable later on.

langston-barrett: Yep, I'm convinced! Reverted the bits about parallelism.

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -40,110 +220,377 @@ data ProofResult sym t
...
      -- Corresponds to 'W4R.Sat'.
    | Disproved (WE.GroundEvalFn t) (Maybe (WE.ExprRangeBindings t))
      -- | The SMT solver returned \"unknown\".
      --
      -- Corresponds to 'W4R.Unknown'.
    | Unknown
 
--- | Prove a single goal ('CB.Assertion') under the supplied 'Assumptions'.
+-- | A 'ProofStrategy' dictates how results are proved.
+data ProofStrategy sym m t r
```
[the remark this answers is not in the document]
langston-barrett: Good call!

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -3,147 +3,452 @@ Module      : Lang.Crucible.Backend.Prove
...
+
+---------------------------------------------------------------------
+-- * Strategy
 
 -- | The result of attempting to prove a goal with an SMT solver.
 --
 -- The constructors of this type correspond to those of 'W4R.SatResult'.
+--
+-- * @sym@ is the symbolic backend, usually 'WE.ExprBuilder'
+-- * @r@ is the return type of the eventual 'ProofConsumer'
```
RyanGlScott, line 138: `ProofResult` doesn't have an `r` type parameter. Should this be `t`?

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -3,147 +3,452 @@ Module      : Lang.Crucible.Backend.Prove
...
      -- Corresponds to 'W4R.Unknown'.
    | Unknown
 
--- | Prove a single goal ('CB.Assertion') under the supplied 'Assumptions'.
+-- | A 'ProofStrategy' dictates how results are proved.
 --
--- The overall approach is:
+-- * @sym@ is the symbolic backend, usually 'WE.ExprBuilder'
+-- * @m@ is the monad in which the 'Prover' and 'Combiner' run
+-- * @t@ is the parameter to 'WE.Expr'
```
RyanGlScott, line 163: It might be worth clarifying that `t` is a What4 `BaseType`.

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -3,147 +3,452 @@ Module      : Lang.Crucible.Backend.Prove
...
--- * Attempt to prove the conjunction of the assumptions and the negated goal.
+-- If the result is 'Disproved', then this function must consume the
+-- 'WE.GroundEvalFn' before returning. See 'WSA.SolverAdapter'.
 --
--- If this goal is satisfiable ('W4R.Sat'), then there exists a counterexample
--- that makes the original goal false, so we have disproven the goal. If the
--- negated goal is unsatisfiable ('W4R.Unsat'), on the other hand, then the
--- original goal is proven.
+-- * @sym@ is the symbolic backend, usually 'WE.ExprBuilder'
+-- * @t@ is the parameter to 'WE.Expr'
```
RyanGlScott, line 178: It might be worth clarifying that `t` is a What4 `BaseType`.

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -3,147 +3,452 @@ Module      : Lang.Crucible.Backend.Prove
...
+newtype ProofConsumer sym t r
```
RyanGlScott, line 179: `r` isn't described in the Haddocks above. Should it be?

RyanGlScott (approved):

crucible/src/Lang/Crucible/Backend/Prove.hs
langston-barrett: It actually [is *not* a `BaseType`](https://hackage.haskell.org/package/what4-1.6/docs/What4-Expr.html#t:Expr), which is perhaps itself worth clarifying.

crucible/src/Lang/Crucible/Backend/Prove.hs
langston-barrett: Whoops, yep!

langston-barrett:
(This is just waiting for me to test this in my downstream codebase to confirm it all works as expected there)

PR #1220, opened 2024-07-12
`crucible-llvm`: Improve error messages in a few pointer-to-bitvector casts

langston-barrett:
Also, discourage the use of `projectLLVM_bv`, which provides an unhelpful, generic message when its assertion fails. Provide `ptrToBv`, an alternative that takes an error message as an argument.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -1136,7 +1136,8 @@ strLen bak mem = go (BV.zero PtrWidth) (truePred sym)
         do ast <- impliesPred sym cond loadok
            assert bak ast $ AssertFailureSimError "Error during memory load: strlen" ""
            v <- unpackMemValue sym (LLVMPointerRepr (knownNat @8)) llvmval
-           test <- bvIsNonzero sym =<< Partial.projectLLVM_bv bak v
+           let err = AssertFailureSimError "Found pointer in string passed to `strlen`" ""
+           test <- bvIsNonzero sym =<< Partial.ptrToBv bak err v
            iteM bvIte sym
              test
              (do cond' <- andPred sym cond test
@@ -1171,7 +1172,8 @@ loadString bak mem = go id
   go f _ (Just 0) = return $ f []
   go f p maxChars = do
      v <- doLoad bak mem p (bitvectorType 1) (LLVMPointerRepr (knownNat :: NatRepr 8)) noAlignment
-     x <- Partial.projectLLVM_bv bak v
+     let err = AssertFailureSimError "Found pointer when loading string" ""
+     x <- Partial.ptrToBv bak err v
      case BV.asUnsigned <$> asBV x of
        Just 0 -> return $ f []
        Just c -> do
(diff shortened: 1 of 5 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1135,7 +1136,8 @@ strLen bak mem = go (BV.zero PtrWidth) (truePred sym)
         do ast <- impliesPred sym cond loadok
            assert bak ast $ AssertFailureSimError "Error during memory load: strlen" ""
            v <- unpackMemValue sym (LLVMPointerRepr (knownNat @8)) llvmval
-           test <- bvIsNonzero sym =<< Partial.projectLLVM_bv bak v
+           let err = AssertFailureSimError "Found pointer in string passed to `strlen`" ""
```
RyanGlScott, line 1139: Perhaps says:

```suggestion
           let err = AssertFailureSimError "Found pointer instead of byte in string passed to `strlen`" ""
```

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -1170,7 +1172,8 @@ loadString bak mem = go id
   go f _ (Just 0) = return $ f []
   go f p maxChars = do
      v <- doLoad bak mem p (bitvectorType 1) (LLVMPointerRepr (knownNat :: NatRepr 8)) noAlignment
-     x <- Partial.projectLLVM_bv bak v
+     let err = AssertFailureSimError "Found pointer when loading string" ""
```
RyanGlScott, line 1175: Perhaps say:

```suggestion
     let err = AssertFailureSimError "Found pointer instead of byte when loading string" ""
```

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs
```diff
@@ -262,16 +263,32 @@ annotateME sym mop rsn p =
        (BBMemoryError (MemoryError mop rsn))
      return p'
 
--- | Assert that the given LLVM pointer value is actually a raw bitvector and extract its value.
-projectLLVM_bv ::
+-- | Assert that the given LLVM pointer value is actually a raw bitvector and
+-- extract its value.
+ptrToBv ::
```
RyanGlScott, line 268: I wonder if we should include a disclaimer that in general, you should prefer tagging your pointers with the relevant `UndefinedBehavior` check (e.g., `ComparePointerToBV`) rather than circumventing it with `ptrToBv`. (I think there are scenarios where `ptrToBv` is unavoidable, of course.)

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs
```diff
@@ -262,16 +263,32 @@ annotateME sym mop rsn p =
...
   do let sym = backendGetSym bak
      p <- natEq sym blk =<< natLit sym 0
-     assert bak p $ AssertFailureSimError "Pointer value coerced to bitvector" ""
+     assert bak p err
      return bv
 
+-- | Assert that the given LLVM pointer value is actually a raw bitvector and extract its value.
+--
+-- Note that this assertion has a very generic message, which can be unhelpful
+-- to users when it fails. Consider using 'ptrToBv' instead.
```
RyanGlScott, line 284: A similar suggestion about recommending undefined behavior checking here.

RyanGlScott (approved):

Issue #1224, opened 2024-07-18
`crucible-mir` fails to parse MIR JSON that casts array reference to pointer


RyanGlScott:
`crucible-mir` should be able to simulate this program and conclude that `foo` returns `42`:

```rs
// test.rs
type T = u64;
const COUNT: usize = 16;

pub fn output_array(src: &[T; COUNT], dst: *mut T) {
    unsafe {
        std::ptr::copy(src as *const T, dst, COUNT);
    }
}

#[crux::test]
pub fn foo() -> T {
    let src: &[T; COUNT] = &[42; COUNT];
    let dst: &mut [T; COUNT] = &mut [0; COUNT];
    output_array(src, dst as *mut T);
    dst[0]
}
```

However, `crucible-mir` fails to even parse the MIR JSON file that one gets when compiling this program:

```
$ crux-mir test.rs 
user error (JSON Decoding of test.all.mir failed: Error in $.fns[0].body.blocks[0].block.data[8].rhs.type: bad CastKind: Just (String "Pointer(ArrayToPointer)"))
```

We should ensure that the `ArrayToPointer` `CastKind` is supported.

Issue #1223, opened 2024-07-18
`crucible-mir`: Split out the `Misc` `CastKind` into more precise constructors 


RyanGlScott:
Currently, the [`Misc`](https://github.com/GaloisInc/crucible/blob/c310fcb7f7a67fa78af368d2bf8f4e15d1c1820e/crucible-mir/src/Mir/Mir.hs#L464) `CastKind` is a catch-all for several unrelated `CastKind`s, as can be seen in this code:

https://github.com/GaloisInc/crucible/blob/c310fcb7f7a67fa78af368d2bf8f4e15d1c1820e/crucible-mir/src/Mir/JSON.hs#L405-L416

As a result, it is difficult to distinguish between all of these `CastKind`s in `Mir.Trans.evalCast'`. I propose that we delete the `Misc` constructor and replace it with more precise constructors that reflect the structure of the [`CastKind`](https://doc.rust-lang.org/stable/nightly-rustc/rustc_middle/mir/syntax/enum.CastKind.html) enum in the `rustc` API. This will require some detective work to figure out which existing uses of `Misc` are covered by which `CastKind` variants.

RyanGlScott:
Note that as of https://github.com/rust-lang/rust/commit/4630d1b23b996793e9188490cc99b1576176233f (which debuted in Rust 1.81.0), `rustc` enforces that `ArrayToPointer` and `MutToConstPointer` casts never make it to runtime MIR (they're only used for borrowck purposes). Instead, these types of casts are normalized to `PtrToPtr`. As such, we can safely leave out `ArrayToPointer` and `MutToConstPointer` from `crucible-mir`'s AST and only include `PtrToPtr`.

PR #1225, opened 2024-07-18
`crucible-mir`: Properly parse `ArrayToPointer` casts

RyanGlScott:
Although `crucible-mir` had support for translating these sorts of casts (`ArrayToPointer` casts), it did not parse them properly. This patch ensures that `ArrayToPointer` casts are parsed as `Misc`, which allows them to be handled by the code in `Mir.Trans`.

Fixes https://github.com/GaloisInc/crucible/issues/1224.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -1,3 +1,8 @@
+# next -- TBA
+
+* Fix a bug in which `crucible-mir` would fail to parse MIR JSON code involving
+  casts from array references to pointers.
+
 # 0.2 -- 2024-02-05
 
 * `crucible-mir` now supports the `nightly-2023-01-23` Rust toolchain. Some of
--- a/crucible-mir/src/Mir/JSON.hs
+++ b/crucible-mir/src/Mir/JSON.hs
@@ -407,6 +407,7 @@ instance FromJSON CastKind where
                                                -- https://github.com/GaloisInc/crucible/issues/1223
                                                Just (String "PointerExposeAddress") -> pure Misc
                                                Just (String "PointerFromExposedAddress") -> pure Misc
+                                               Just (String "Pointer(ArrayToPointer)") -> pure Misc
                                                Just (String "DynStar") -> pure Misc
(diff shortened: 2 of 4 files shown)


RyanGlScott:
After my changes, the `maybe_uninit_array_cast` test case [now fails](https://github.com/GaloisInc/crucible/actions/runs/9996732271/job/27631745693?pr=1225#step:17:1563):

```
      mem
        maybe_uninit_array_cast: FAIL (1.39s)
          Compiling and running oracle program                           (0.09s)
          Oracle output: [1, 2]                                          (1.30s)
          Crux output: 
          failures:
          
          ---- maybe_uninit_array_cast/984deb76::crux_test[0] counterexamples ----
          [Crux]   test/conc_eval/mem/maybe_uninit_array_cast.rs:7:19: 7:45: error: in maybe_uninit_array_cast/984deb76::crux_test[0]
          [Crux]   Translation error in maybe_uninit_array_cast/984deb76::crux_test[0]: unimplemented cast: Misc
          [Crux]     ty: TyRawPtr (TyArray (TyInt B32) 2) Mut
          [Crux]     as: TyRawPtr (TyInt B32) Mut
          
            test/Test.hs:123:
            crux doesn't match oracle
```

Fair enough, since I changed the relevant code which casts from an array reference (or, in this case, an array pointer) to a pointer to look for `ArrayToPointer` instead of `PtrToPtr`, the latter of which is lumped under the `Misc` umbrella. I _could_ add a dedicated `PtrToPtr` constructor, but the problem is that there are many different `evalCast'` cases that could now equally apply to both `Misc` and `PtrToPtr`. It would be nice to clean up the code to make this sort of thing easier to express (#1223), but I am inclined not to that just now, since this is intended to be a simple bugfix.

As such, I think I will do the simplest possible thing and map `ArrayToPointer` to `Misc`. This is a bit ugly, but it avoids the possibility of introducing regressions like the one seen in the `maybe_uninit_array_cast` test case. I'll defer the larger task of getting rid of `Misc` to #1223.

mccleeary-galois (approved):
Seems like a reasonable option for a quick fix.

PR #1234, opened 2024-08-02
Support building with GHC 9.8

RyanGlScott:
This PR contains a variety of fixes needed to build the libraries in the `crucible` repo with GHC 9.8:

* I've bumped the `jvm-parser` submodule to bring in the changes from https://github.com/GaloisInc/jvm-parser/pull/12.
* Various upper version bounds on `base` and `th-abstraction` were raised to allow building with GHC 9.8.
* GHC 9.8 adds the `-Wx-partial` warning to `-Wall`, which warns on uses of the partial `head` and `tail` functions. I have made an effort to remove uses of these functions where possible. Where this is not possible, I have rewritten uses of `head` and `tail` into pattern matches with explicit `panic`s in the partial cases.
* I have updated the CI to test GHC 9.4, 9.6, and 9.8. We now drop GHC 9.2 from CI. This means that the Dockerfiles now use GHC 9.4 instead of 9.2.


crucible/src/Lang/Crucible/Simulator/Profiling.hs
```diff
@@ -76,6 +76,7 @@ import           What4.SatResult
 
 import           Lang.Crucible.Backend
 import           Lang.Crucible.CFG.Core
+import           Lang.Crucible.Panic
```
langston-barrett, line 79: Perhaps prefer a more explicit import?
```suggestion
import           Lang.Crucible.Panic (panic)
```

langston-barrett (approved):
Excellent, thanks for putting this together!

crux-llvm/test/Test.hs
```diff
@@ -124,8 +143,10 @@ getZ3Version =
   let getVer (Right inp) =
         -- example inp: "Z3 version 4.8.7 - 64 bit"
         let w = words inp
-        in if and [ length w > 2, head w == "Z3" ]
-           then w !! 2 else "?"
+        in if |  ("Z3":_:verNum:_) <- w
```
kquick, line 146: I suppose these could also have avoided the warnings by changing `head w` to `w !! 0`, but the pattern matching approach is fine as well.  I'd suggest further simplifying by replacing the `let w = words inp in` with `if | [ ... ] <- words inp`, but it's up to you.

kquick (approved):

crux-llvm/test/Test.hs
RyanGlScott: Good point. In fact, we can get rid of the use of `MultiWayIf` and just use plain-old mattern matching here, which makes the code even simpler.

PR #1239, opened 2024-08-16
cli: Allow client libraries to use `IO` to create `ExtensionImpl`

langston-barrett:
This is needed in Macaw.

### DIFF
--- a/crucible-cli/app/Main.hs
+++ b/crucible-cli/app/Main.hs
@@ -13,7 +13,7 @@ import qualified Lang.Crucible.CLI.Options as Opts
 main :: IO ()
 main =
   do let ?parserHooks = defaultParserHooks
-     Opts.main "crucible" (\_ -> emptyExtensionImpl) simulationHooks
+     Opts.main "crucible" (\_ -> pure emptyExtensionImpl) simulationHooks
   where
         simulationHooks =
           defaultSimulateProgramHooks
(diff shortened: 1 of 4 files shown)


RyanGlScott (commented):
I am a bit surprised that you'd need `IO` just to create an `ExtensionImpl`. Can you say more about how it's used in Macaw? (I'm sure there is a good reason, but I'm having trouble seeing it without further context.)

langston-barrett:
Yeah! The problem is that to create a `macawExtensions`, you need a `MacawArchEvalFn`, which, at least for x86_64, requires something called `SymFns`, which you create with `newSymFns ::~ sym -> IO SymFns`.

langston-barrett:
Also, to create a `MemModelConfig`, I'll need access to `bak`, not just `sym` (see most recent commit).

langston-barrett:
... and in fact, [`newGlobalMemoryWith`](https://github.com/GaloisInc/macaw/blob/3dfc22bf4108db1ec1c9d5bb3a7b834e0eb8686d/symbolic/src/Data/Macaw/Symbolic/Memory.hs#L208) even needs to know `sym ~ WEB.ExprBuilder t st fs`...

langston-barrett:
See https://github.com/GaloisInc/macaw/pull/423 for what I'm trying to do with this.

RyanGlScott (approved):
Ah, right.

PR #1242, opened 2024-08-22
CI: Consistently use `macos-14` in all workflows

RyanGlScott:
The `crux-llvm` and `crux-mir` CI workflows were using `macos-14` (i.e., ARM64), but the remaining workflows (`crucible-go`, `crucible-jvm`, and `crucible-wasm`) were using `macos-12` (i.e., X64). This updates the latter ones to use `macos-14` so that we are consistent.


mccleeary-galois (approved):
Thanks for catching this Ryan!

PR #1243, opened 2024-08-23
Rearrange the handling of constant slices.

sauclovian-g:
The idea is to split the slice into a reference and a separate static allocation for the body the reference points to. This then allows us to treat the references as normal references to memory, which in turn avoids problems in SAW that arise when it goes to try to enforce disjointness.

Requires accompanying changes in mir-json that produce the new output we read. (These are in [mir-json #73](https://github.com/GaloisInc/mir-json/pull/73).)

See [saw-script #2064](https://github.com/GaloisInc/saw-script/issues/2064) for background.

### DIFF
--- a/crucible-mir/src/Mir/Generator.hs
+++ b/crucible-mir/src/Mir/Generator.hs
@@ -167,7 +167,7 @@ data PtrMetadata s =
 
 instance Show (PtrMetadata s) where
     show NoMeta = "NoMeta"
-    show (SliceMeta _) = "SliceMeta"
+    show (SliceMeta len) = "SliceMeta " ++ show len
 
 ---------------------------------------------------------------------------------
 
--- a/crucible-mir/src/Mir/Generator.hs
+++ b/crucible-mir/src/Mir/Generator.hs
@@ -164,10 +164,7 @@ data MirPlace s where
 data PtrMetadata s =
       NoMeta
     | SliceMeta (R.Expr MIR s UsizeType)
-
-instance Show (PtrMetadata s) where
-    show NoMeta = "NoMeta"
-    show (SliceMeta len) = "SliceMeta " ++ show len
+  deriving Show
 
 ---------------------------------------------------------------------------------
 
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -496,7 +496,6 @@ data FloatLit
 data ConstVal =
     ConstFloat FloatLit
   | ConstInt IntLit
-  | ConstSliceBody [ConstVal]
   | ConstSliceRef DefId Int
   | ConstStrBody B.ByteString
   | ConstStrRef DefId Int
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -498,7 +498,6 @@ data ConstVal =
   | ConstInt IntLit
   | ConstSliceRef DefId Int
   | ConstStrBody B.ByteString
-  | ConstStrRef DefId Int
   | ConstBool Bool
   | ConstChar Char
   | ConstVariant DefId
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -496,9 +496,10 @@ data FloatLit
 data ConstVal =
     ConstFloat FloatLit
   | ConstInt IntLit
-  | ConstSlice [ConstVal]
-  | ConstStr B.ByteString
-  | ConstByteStr B.ByteString
+  | ConstSliceBody [ConstVal]
+  | ConstSliceRef DefId Int
+  | ConstStrBody B.ByteString
+  | ConstStrRef DefId Int
   | ConstBool Bool
   | ConstChar Char
   | ConstVariant DefId
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -496,7 +496,11 @@ data FloatLit
 data ConstVal =
     ConstFloat FloatLit
   | ConstInt IntLit
+  -- | A reference to a static allocation for a slice (type is `&[T]` or `&str`)
+  -- The DefId can lead to either a ConstArray or a ConstStrBody.
   | ConstSliceRef DefId Int
+  -- | A string body, of type [u8]. Currently only arises from string literals,
+  -- but might in the future be used for all [u8]s.
   | ConstStrBody B.ByteString
   | ConstBool Bool
   | ConstChar Char
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -312,8 +312,8 @@ instance Pretty Substs where
 instance Pretty ConstVal where
     pretty (ConstFloat i)   = pretty i
     pretty (ConstInt i)     = pretty i
-    pretty (ConstStr i)     = dquotes (viaShow i)
-    pretty (ConstByteStr i) = viaShow i
+    pretty (ConstStrBody i) = dquotes (viaShow i)
+    pretty (ConstStrRef a len) = pretty "&" <> pr_id a <> (viaShow len)
     pretty (ConstBool i)    = pretty i
     pretty (ConstChar i)    = pretty i
     pretty (ConstVariant i) = pr_id i
@@ -328,7 +328,8 @@ instance Pretty ConstVal where
     pretty (ConstRawPtr a) = pretty a
     pretty (ConstStruct fs) = pretty "struct" <> list (map pretty fs)
     pretty (ConstEnum v fs) = pretty "enum" <> list ((pretty "variant" <+> pretty v) : map pretty fs)
-    pretty (ConstSlice cs)  = list (map pretty cs)
+    pretty (ConstSliceBody cs)  = list (map pretty cs)
+    pretty (ConstSliceRef a len) = pretty "&" <> pr_id a <> (viaShow len)
     pretty (ConstFnPtr i)   = pretty "fn_ptr" <> brackets (pretty i)
 
 instance Pretty AggregateKind where
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -313,7 +313,6 @@ instance Pretty ConstVal where
     pretty (ConstFloat i)   = pretty i
     pretty (ConstInt i)     = pretty i
     pretty (ConstStrBody i) = dquotes (viaShow i)
-    pretty (ConstStrRef a len) = pretty "&" <> pr_id a <> (viaShow len)
     pretty (ConstBool i)    = pretty i
     pretty (ConstChar i)    = pretty i
     pretty (ConstVariant i) = pr_id i
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -328,7 +328,6 @@ instance Pretty ConstVal where
     pretty (ConstRawPtr a) = pretty a
     pretty (ConstStruct fs) = pretty "struct" <> list (map pretty fs)
     pretty (ConstEnum v fs) = pretty "enum" <> list ((pretty "variant" <+> pretty v) : map pretty fs)
-    pretty (ConstSliceBody cs)  = list (map pretty cs)
     pretty (ConstSliceRef a len) = pretty "&" <> pr_id a <> (viaShow len)
     pretty (ConstFnPtr i)   = pretty "fn_ptr" <> brackets (pretty i)
 
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -327,7 +327,7 @@ instance Pretty ConstVal where
     pretty (ConstRawPtr a) = pretty a
     pretty (ConstStruct fs) = pretty "struct" <> list (map pretty fs)
     pretty (ConstEnum v fs) = pretty "enum" <> list ((pretty "variant" <+> pretty v) : map pretty fs)
-    pretty (ConstSliceRef a len) = pretty "&" <> pr_id a <> (viaShow len)
+    pretty (ConstSliceRef a _len) = pretty "&" <> pr_id a
     pretty (ConstFnPtr i)   = pretty "fn_ptr" <> brackets (pretty i)
 
 instance Pretty AggregateKind where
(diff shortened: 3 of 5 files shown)


sauclovian-g:
Note that this shouldn't get merged without a commit that bumps the mir-json submodule reference, but that should wait until the mir-json changes are ready to merge.

sauclovian-g:
oh, of course none of the tests are going to work without the new mir-json...

crucible-mir/src/Mir/Trans.hs
```diff
@@ -359,6 +391,26 @@ staticPlace did = do
             MirPlace (G.globalType gv) <$> globalMirRef gv <*> pure NoMeta
         Nothing -> mirFail $ "cannot find static variable " ++ fmt did
 
+-- variant of staticPlace for slices
+-- tpr is the element type; len is the length
+staticPlaces :: HasCallStack => C.TypeRepr tp -> Int -> M.DefId -> MirGenerator h s ret (MirPlace s)
+staticPlaces tpr len did = do
```
RyanGlScott, line 397: Perhaps rename this to something like "`staticSlicePlace`" to reflect its intended purpose.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -149,33 +149,64 @@ transConstVal _ty (Some (UsizeRepr)) (M.ConstInt i) =
...
+--      (with globalMirRef rather than constMirRef, that's the point of
+--      all this)
+--    * apply subindexRef as above
+--    * cons up the length
+--    * call mkStruct
+--    * cons up the final MirExp
+--
+-- staticPlaces does the first four of these actions; addrOfPlace
+-- does the last two. Note that addrOfPlace uses mkSlice instead
+-- of mkStruct as above, but as far as I can tell it's equivalent.
```
RyanGlScott, line 177: Yes, slices are encoded as structs with very particular field types, so it's perfectly fine to use `mkSlice` in place of `mkStruct`.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -359,6 +391,26 @@ staticPlace did = do
...
+            Refl <- testEqualityOrFail (MirVectorRepr tpr) tpr_found $
+                "staticPlaces: wrong type: expected vector of " ++
+                show tpr ++ ", found " ++ show tpr_found
+            ref <- globalMirRef gv
+            ref' <- subindexRef tpr ref (R.App $ usizeLit 0)
+            -- XXX we seem to need to cons up a Crucible Atom (see
+            -- crucible/src/Lang/Crucible/CFG/Reg.hs) from len to put in here.
+            --let len' = error "oops"
+            let len' = R.App $ usizeLit $ fromIntegral len
+            return $ MirPlace tpr ref' (SliceMeta len')
```
RyanGlScott, line 411: I don't think these comments are relevant anymore, since there is now a suitable `Atom` to use for the `SliceMeta` length. Delete these comments?

crucible-mir/src/Mir/Trans.hs
```diff
@@ -149,33 +149,64 @@ transConstVal _ty (Some (UsizeRepr)) (M.ConstInt i) =
...
+--
+-- Before the slice code got split into separate body and reference steps,
+-- it did the following after the code that's equivalent to the ConstArray
+-- case (which is where constant slice bodies get handled now):
+--
+--vecRef <- constMirRef vec_tpr vec
+--ref <- subindexRef tpr vecRef (R.App $ usizeLit 0)
+--let len = R.App $ usizeLit $ fromIntegral $ length cs
+--let struct = S.mkStruct (mirSliceCtxRepr tpr) (Ctx.Empty Ctx.:> ref Ctx.:> len)
+--return $ MirExp (MirSliceRepr tpr) struct
```
RyanGlScott, line 162: I think it's fine to include a historical note about the previous approach used to translate constant slices, but I worry that including the full code (verbatim) will be a bit much if you don't have the full context for what all of these variables are. (For instance, `vec` no longer appears in the new version of the code—I think you'd be better off going through the `git` history if you want the previous implementation.)

My personal preference would be to:

* Describe the steps needed to turn the looked-up `DefId` into a `MirExp` (as you currently do in the bulleted list below).
* Note that MIR slice references used to be translated in a way such that they would be turned into references to constant vectors, but this prevented them from being eligible for equality or overlap checks (citing https://github.com/GaloisInc/saw-script/issues/2064). The current approach instead translates slice references into references to separate global allocations (citing the `ConstStrBody` body case below).

crucible-mir/src/Mir/Trans.hs
```diff
@@ -149,33 +149,64 @@ transConstVal _ty (Some (UsizeRepr)) (M.ConstInt i) =
...
+    return $ MirExp vec_tpr mirVec
+--
+-- This is the rest of the code that was here before the reference and
+-- body steps got split:
+--
+--vecRef <- constMirRef vec_tpr mirVec
+--ref <- subindexRef u8Repr vecRef (R.App $ usizeLit 0)
+--let len = R.App $ usizeLit $ fromIntegral $ BS.length bs
+--let struct = S.mkStruct knownRepr (Ctx.Empty Ctx.:> ref Ctx.:> len)
+--return $ MirExp (MirSliceRepr u8Repr) struct
```
RyanGlScott, line 199: Similarly, I'd favor not including the full code from the previous implementation here. I think it suffices to say that this case it exactly like the `ConstSliceRef` above, but specialized to strings.

RyanGlScott:
> oh, of course none of the tests are going to work without the new mir-json...

Indeed, I think it would be worth bumping the `mir-json` submodule commit to the current state of https://github.com/GaloisInc/mir-json/pull/73 so that we can verify that everything works. When https://github.com/GaloisInc/mir-json/pull/73 lands, we can bump the submodule to the merge commit on `mir-json`'s `master` branch.

Relatedly, I think it would be worth noting somewhere in the commit message that this depends on the changes from https://github.com/GaloisInc/mir-json/pull/73.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -359,6 +391,26 @@ staticPlace did = do
             MirPlace (G.globalType gv) <$> globalMirRef gv <*> pure NoMeta
         Nothing -> mirFail $ "cannot find static variable " ++ fmt did
 
+-- variant of staticPlace for slices
+-- tpr is the element type; len is the length
+staticPlaces :: HasCallStack => C.TypeRepr tp -> Int -> M.DefId -> MirGenerator h s ret (MirPlace s)
+staticPlaces tpr len did = do
+    sm <- use $ cs.staticMap
+    case Map.lookup did sm of
+        Just (StaticVar gv) -> do
```
spernsteiner, line 400: Would it be possible / reasonable for this to call `staticPlace` instead of duplicating the lookup code?  It looks like the `MirPlace` returned by `staticPlace` contains all the pieces you need

crucible-mir/src/Mir/Generator.hs
```diff
@@ -167,7 +167,7 @@ data PtrMetadata s =
 
 instance Show (PtrMetadata s) where
     show NoMeta = "NoMeta"
-    show (SliceMeta _) = "SliceMeta"
+    show (SliceMeta len) = "SliceMeta " ++ show len
```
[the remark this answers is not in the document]
sauclovian-g: Hmm indeed. Someone must have done it that way on purpose then. Wonder why... looks like the contents have changed since so it's probably no longer relevant.

Updated in d4bd0cd0ff03b11e463ef51b795d004c8eaa802e

crucible-mir/src/Mir/PP.hs
```diff
@@ -328,7 +328,7 @@ instance Pretty ConstVal where
     pretty (ConstRawPtr a) = pretty a
     pretty (ConstStruct fs) = pretty "struct" <> list (map pretty fs)
     pretty (ConstEnum v fs) = pretty "enum" <> list ((pretty "variant" <+> pretty v) : map pretty fs)
-    pretty (ConstSlice cs)  = list (map pretty cs)
+    pretty (ConstSliceRef a len) = pretty "&" <> pr_id a <> (viaShow len)
```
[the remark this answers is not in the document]
sauclovian-g: There was some reason I did that rather than just ignore the value, but I can't remember what and I agree the resulting output is nonsense. I think, since this is for prettyprinting rather than ast-dumping, that it should just skip the length. Updated in 3444637b455edf15c14963403aa69cd2f4192f5e

crucible-mir/src/Mir/PP.hs
```diff
@@ -312,8 +312,8 @@ instance Pretty Substs where
 instance Pretty ConstVal where
     pretty (ConstFloat i)   = pretty i
     pretty (ConstInt i)     = pretty i
-    pretty (ConstStr i)     = dquotes (viaShow i)
-    pretty (ConstByteStr i) = viaShow i
+    pretty (ConstStrBody i) = dquotes (viaShow i)
+    pretty (ConstStrRef a len) = pretty "&" <> pr_id a <> (viaShow len)
```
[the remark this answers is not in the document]
sauclovian-g: Obsolete! :-)

crucible-mir/src/Mir/Trans.hs
sauclovian-g: It got that way because originally it was different from ```staticPlace``` specifically by handling a length instead of a single element... but it's not that simple any more. Updated in 3444637b455edf15c14963403aa69cd2f4192f5e

crucible-mir/src/Mir/Trans.hs
sauclovian-g: Yes, the question was whether it's exactly the same or if there was something special/different about the mkStruct call such that it wasn't using mkSlice for a reason. Which there apparently wasn't. But it wasn't obvious without expanding/looking up a bunch of things.

crucible-mir/src/Mir/Trans.hs
sauclovian-g: They are not. Removed in 3444637b455edf15c14963403aa69cd2f4192f5e

crucible-mir/src/Mir/Trans.hs
sauclovian-g: That comment's there as a semi-proof to myself that I haven't accidentally changed the behavior.

There's a lot of things here where you can write the same action as either ```a``` or ```A``` (so to speak) and then assorted combinations of ```Ab```, ```Bc```, ```bC```, etc. such that if you started with ```((Ab,cD),ef)``` and want to change it to ```(a, (bCd), EF)``` or similar you want to be pretty careful that what you've changed it to is actually equivalent, and neither missing a step nor replacing ```c``` with ```C'``` instead of ```C``` when you didn't mean to. The first couple times I got it wrong (even laying it out explicitly), hence the high level of paranoia.

Now that it's running maybe it should just be flushed, though.

Let me chew on it over the weekend.

crucible-mir/src/Mir/Trans.hs
sauclovian-g: This one's now gone entirely anyway.

crucible-mir/src/Mir/Trans.hs
sauclovian-g: It could, but it seems the wrong way to abstract things (these are two different ways to look up a static and construct a Place for it, and having the more complex one call the simpler one and then munge the results doesn't seem right)

So... I'd rather not I think

sauclovian-g:
> I think it would be worth bumping the `mir-json` submodule commit to the current state of [GaloisInc/mir-json#73](https://github.com/GaloisInc/mir-json/pull/73) so that we can verify that everything works. When [GaloisInc/mir-json#73](https://github.com/GaloisInc/mir-json/pull/73) lands, we can bump the submodule to the merge commit on `mir-json`'s `master` branch.

I have done this.

> Relatedly, I think it would be worth noting somewhere in the commit message that this depends on the changes from [GaloisInc/mir-json#73](https://github.com/GaloisInc/mir-json/pull/73).

Now each of the descriptions references the other one.

crucible-mir/src/Mir/Trans.hs
spernsteiner: FWIW I think one interpretation of `staticSlicePlace` is that it obtains a reference to the array `did` (e.g. `&[u8; 10]`), which is exactly the output of `staticPlace`, and then converts that array reference to a slice reference (`&[u8]`).  But implementation-wise I think it's fine either way.

crucible-mir/src/Mir/Mir.hs
```diff
@@ -496,9 +496,8 @@ data FloatLit
 data ConstVal =
     ConstFloat FloatLit
   | ConstInt IntLit
-  | ConstSlice [ConstVal]
-  | ConstStr B.ByteString
-  | ConstByteStr B.ByteString
+  | ConstSliceRef DefId Int
```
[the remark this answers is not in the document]
spernsteiner: > `ConstSliceRef` is a reference to a static allocation for a slice (of type `[T]`) or a string (or type `str`).

To be clear, the type of `ConstSliceRef` is `&[T]` or `&str`

> In the former case, the allocation is backed by a constant array (i.e., a `ConstArray`), and in the latter case, the allocation is backed by a constant string (i.e., a `ConstStrBody`).

IMO there's no need to mandate this if we view `ConstStrBody` as shorthand for a `ConstArray` of bytes.  There's no issue with allowing `ConstStrBody` for `&[u8]` or `&[u8; N]`.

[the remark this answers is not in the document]
RyanGlScott: > To be clear, the type of `ConstSliceRef` is `&[T]` or `&str`

Oh, certainly! I didn't mean to imply that the overall type of the reference was `[T]` or `str`. I was more describing how what type of constant the static allocation would be creating internally.

> IMO there's no need to mandate this if we view `ConstStrBody` as shorthand for a `ConstArray` of bytes. There's no issue with allowing `ConstStrBody` for `&[u8]` or `&[u8; N]`.

That's fair. That being said, in practice we only use `ConstStrBody` as part of creating static allocations for `&str`s, correct? It's possible that we might change this in the future, but I think it's helpful to make a note of under what circumstances `ConstStrBody` can arise today. (It wasn't obvious to me at a first glance.)

crucible-mir/src/Mir/Trans.hs
sauclovian-g: That is also true. hrm...

I guess at this point the path of least resistance is to not make more changes...

crucible-mir/src/Mir/Trans.hs
sauclovian-g: Conclusion: it should just be flushed. It is only of interest if the new code turns out not to work in some corner case, and in that case fetching things out of history is fine.

crucible-mir/src/Mir/Mir.hs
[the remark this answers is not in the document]
sauclovian-g: That is correct. I've added some notes.

RyanGlScott (approved):

sauclovian-g:
I force-pushed to squash the reference to the pre-merged version in mir-json. Perhaps not strictly necessary (since that version still exists) but it's untidy to leave it lying around.

PR #1249, opened 2024-09-06
Revise Haddocks for Crucible backends

langston-barrett:
The fact that these backends use an `AssumptionStack` is an implementation detail that is not exposed via their interface, instead clients interact with `getProofObligations` is an implementation detail that is not exposed via their interface, instead clients interact with `getProofObligations`.

Also, point to a module that helps with dispatching such obligations.


RyanGlScott (approved):
LGTM. I think the commit message accidentally copy-pasted too much, however, as the phrase "implementation detail that is not exposed via their interface" appears twice (making the sentence appear to a run-on sentence).

langston-barrett:
Good catch, thanks @RyanGlScott! Fixed the commit message.

PR #1254, opened 2024-09-11
llvm: Print out the failed predicate in a panic

langston-barrett:


RyanGlScott (approved):
Looks reasonable. Perhaps it's worth printing out the predicate in the `NoErr` case just above it as well?

langston-barrett:
> Perhaps it's worth printing out the predicate in the NoErr case just above it as well?

Perhaps? It's not really the predicate that's the problem in that case, but rather, the shape of the data (which is already printed).

PR #1260, opened 2024-10-04
G/C unused Monoid instances in crucible-mir

### LINKED ISSUE #1259: `crucible-mir`: Remove vestigial `?mirCS` and `?mirLib` implicit params, as well as `Monoid` instances for `Collection`/`CollectionState`/`RustModule`
In #1253, we propose to make `crucible-mir` respect the `mir-json` schema version. As part of implementing this, it would be convenient to store the version number in the parsed `Collection`, as this could be useful to report to the user in various situations. However, there is an obstacle that stands in our way of doing so: `Collection` has a `Monoid` instance, and it's very unclear how to implement `mempty` and `mappend` for version numbers. We certainly don't want to conjure up a version number out of thin air in `mempty`, and it's unclear how we would combine two different version numbers.

Why does `Collection` have a `Monoid` instance anyway? It ultimately dates back to an old approach that `crux-mir` used for translating `Collection`s in commit 12df2891b51076a0c79924f23c1431a2c4f7d3ea. In that approach, one would call `translateMIR` multiple times and combine the results using `mappend`. However, that approach was removed in commit b0708a0512bce265ceb6d85ce9a4638e7a01dfeb. After that commit, the only time we ever use the `Monoid` instance for `Collection` is to call `mappend (mempty :: Collection) x`. This could just as well be simplified to `x`, thereby avoiding the need for the `Monoid` instance in the first place.

Similarly, we could remove:

* The `Monoid` instances for `CollectionState` and `RustModule` (which depend on the `Monoid Collection` instance, but are similarly unused)
* The `?mirCS` and `?mirLib` implicit parameters (which are only ever instantiated with `mempty`).

sauclovian-g:
Closes #1259.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -1,3 +1,13 @@
+# next -- TBA
+
+* The calling sequence of ```translateMIR``` has changed: the first argument,
+  which should always have been passed as ```mempty```, has been removed.
+  This will require adjustments in any downstream callers.
+* The corresponding implicit argument in the ```Pass``` type has been removed.
+* The Semigroup and Monoid instances for Collection, CollectionState, and
+  RustModule have been removed. It is not expected that there are any
+  downstream users.
+
 # 0.3 -- 2024-08-30
 
 * Implement byte-to-char casts.
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -586,18 +586,6 @@ makeLenses ''NamedTy
 makeWrapped ''Substs
 
 --------------------------------------------------------------------------------------
--- Other instances for ADT types
---------------------------------------------------------------------------------------
-
-instance Semigroup Collection where
-  (Collection f1 a1 a1' t1 s1 v1 n1 tys1 r1) <> (Collection f2 a2 a2' t2 s2 v2 n2 tys2 r2) =
-    Collection (f1 <> f2) (a1 <> a2) (a1' <> a2') (t1 <> t2) (s1 <> s2) (v1 <> v2) (n1 <> n2) (tys1 <> tys2) (r1 <> r2)
-instance Monoid Collection where
-  mempty  = Collection mempty mempty mempty mempty mempty mempty mempty mempty mempty
-
-
-
---------------------------------------------------------------------------------------
 --- aux functions ---
 --------------------------------------------------------------------------------------
 
--- a/crucible-mir/src/Mir/Generator.hs
+++ b/crucible-mir/src/Mir/Generator.hs
@@ -395,19 +395,6 @@ $(return [])
 
 -- ** Operations and instances
 
-instance Semigroup RustModule where
-  (RustModule cs1 cm1 ex1) <> (RustModule cs2 cm2 ex2) =
-    RustModule (cs1 <> cs2) (cm1 <> cm2) (ex1 <> ex2)
-instance Monoid RustModule where
-  mempty  = RustModule mempty mempty mempty
-
-instance Semigroup CollectionState  where
-  (CollectionState hm1 vm1 sm1 dm1 chm1 col1) <> (CollectionState hm2 vm2 sm2 dm2 chm2 col2) =
-      (CollectionState (hm1 <> hm2) (vm1 <> vm2) (sm1 <> sm2) (dm1 <> dm2) (Map.unionWith (<>) chm1 chm2) (col1 <> col2))
-instance Monoid CollectionState where
-  mempty  = CollectionState mempty mempty mempty mempty mempty mempty
-
-
 instance Show (MirExp s) where
     show (MirExp tr e) = (show e) ++ ": " ++ (show tr)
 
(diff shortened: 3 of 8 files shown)


RyanGlScott:
Thanks for working on this, @sauclovian-g!

This does change the API of `crucible-mir`, so we should check that SAW (the primary consumer of `crucible-mir` other than `crux-mir`) is still buildable with these changes. From a quick glance, [this part](https://github.com/GaloisInc/saw-script/blob/55359cc0b7c9b47cf55c42d620a465710007df77/src/SAWScript/Crucible/MIR/Builtins.hs#L537) of `crucible-mir-comp` will be impacted (at a minimum).

sauclovian-g:
Good point, hadn't thought about that. I'm trying that now. Looks like we'll need to bump what4 in saw, but that shouldn't hurt anything...

sauclovian-g:
ok, draft saw-script pull request here: https://github.com/GaloisInc/saw-script/pull/2133

crucible-mir/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # 0.3 -- 2024-08-30
 
+* The calling sequence of ```translateMIR``` has changed: the first argument,
+  which should always have been passed as ```mempty```, has been removed.
+  This will probably require adjustments in downstream callers.
```
RyanGlScott, line 5: We already assigned a date (2024-08-30) to the 0.3 release, so let's put this bullet point under a new section (with the head `# next -- TBA`) above this one.

RyanGlScott, line 5: More than just "probably": it _will_ require adjustment in downstream callers!

crucible-mir/src/Mir/Mir.hs
```diff
@@ -592,8 +592,6 @@ makeWrapped ''Substs
 instance Semigroup Collection where
   (Collection f1 a1 a1' t1 s1 v1 n1 tys1 r1) <> (Collection f2 a2 a2' t2 s2 v2 n2 tys2 r2) =
     Collection (f1 <> f2) (a1 <> a2) (a1' <> a2') (t1 <> t2) (s1 <> s2) (v1 <> v2) (n1 <> n2) (tys1 <> tys2) (r1 <> r2)
```
RyanGlScott, line 594: Thinking about this some more, I think we should remove the `Semigroup` instances for `Collection` and friends as well. The reasoning is similar: when we add a schema version number to `Collection`, it's not clear how one would implement `(<>)` for two version numbers. (The only uses of `(<>)` in `crucible-mir` have already been removed in this PR.)

crucible-mir/CHANGELOG.md
RyanGlScott: Also, make sure to advertise the removal of the `Semigroup` and `Monoid` instances, since those are also part of the public API.

sauclovian-g: I've moved the entry in the right place and added the instances (but there are not expected to be any downstream users -- are these part of the public API because you can't hide instances, or were they intended? (And is there any better way to determine what's in the public API than wade through the export lists of every source file? Is the ```Pass``` type in Mir.Pass, which has one fewer implicit argument now, also part of the public API because it's apparently exported and all the source files are exposed?)

sauclovian-g: Well, yes, I think I meant to write "downstream code" that might or might not actually call it (although I suppose it would be pretty odd if something used this and didn't call that...)

crucible-mir/src/Mir/Mir.hs
sauclovian-g: You compose data blobs of different schema versions by promoting/converting the old one to the new version (or demoting the new one to the old version if it can be encoded in the old version, depending on requirements) and then you have a single version again.

But we aren't set up to do that and it's probably not worthwhile to try...

crucible-mir/CHANGELOG.md
RyanGlScott: > are these part of the public API because you can't hide instances, or were they intended?

Instances are always exported, so they're always part of the public API. More specifically...

> And is there any better way to determine what's in the public API than wade through the export lists of every source file? 

My measuring stick for what constitutes a publicly-facing API change is to use the [Haskell Package Versioning Policy (PVP)](https://pvp.haskell.org/). According to the PVP, removing an instance is always a breaking change.

> Is the `Pass` type in Mir.Pass, which has one fewer implicit argument now, also part of the public API because it's apparently exported and all the source files are exposed?

While the PVP doesn't say anything about type synonyms per se, since we are changing the definition of `Pass`, I think one could argue that that is a breaking change per the PVP. We could mention this in the changelog to be on the safe side.

RyanGlScott (approved):
I'm happy with the way this looks. If you'd like, you could mention the changes to `Pass` in the changelog, but I think we've covered the important changes.

sauclovian-g:
Do you think I should squash this one before merging it?

RyanGlScott:
I think squashing makes sense here, yes.

sauclovian-g:
Yeah, me too.

PR #1265, opened 2024-10-25
Dockerfiles: Properly set up non-root users

### LINKED ISSUE #1261: Docker images: Don't default to `root` user
Currently, the `crux-llvm` and `crux-mir` Docker images default to a `root` user, which makes it impractical for unprivileged users or rootless container environments. We should instead make the Docker images default to unprivileged users, just like we do in our Cryptol and SAW Docker images.

RyanGlScott:
Previously, the `crux-{llvm,mir}` Docker images were set up to use `root` accounts, and all files that were created in the Docker images were given `root` permissions. This is impractical for running rootless container environments, as noted in #1261.

This rewrites the images to set the default users to unprivileges `crux-{llvm,mir}` users. This is surprisingly involved, and it requires a fair bit of refactoring to accomplish:

1. We cannot easily base the `crux-mir` image on the official `rust` image, as the latter has `cargo`, `rustup`, and related tools installed under a `root` account. Instead, we change to an unprivileged user and install these Rust tools using rustup.sh's installer script.
2. Previously, the `crux-llvm` image was supporting Unicode character printing by changing the locale, which requires root access to accomplish. This is overkill, however, as it suffices to simply set the `LANG` and `LC_ALL` environment variables. I've opted for the latter approach, which is much simpler and avoids the need for root access.
3. I needed to reorganize the order of commands in the `base` layers of each image in order to ensure that the relevant files are `chown`ed from the unprivileged user accounts.

Fixes #1261.

### DIFF
--- a/.github/Dockerfile-crux-llvm
+++ b/.github/Dockerfile-crux-llvm
@@ -14,83 +14,86 @@ RUN apt-get update && \
       # LLVM toolchain
       clang-${LLVM_VER} llvm-${LLVM_VER}-tools \
       # Miscellaneous
-      locales unzip wget
-RUN sed -i '/en_US.UTF-8/s/^# //g' /etc/locale.gen && \
-    locale-gen
-ENV LANG en_US.UTF-8
-ENV LANGUAGE en_US:en
-ENV LC_ALL en_US.UTF-8
+      unzip wget
 
-WORKDIR /usr/local/bin
+ARG CRUX_BUILD_DIR=/crux-llvm/build
+RUN mkdir -p ${CRUX_BUILD_DIR}
+ADD crucible ${CRUX_BUILD_DIR}/crucible
+ADD crucible-llvm ${CRUX_BUILD_DIR}/crucible-llvm
+ADD crucible-symio ${CRUX_BUILD_DIR}/crucible-symio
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Successful Docker CI jobs for [`crux-llvm`](https://github.com/GaloisInc/crucible/actions/runs/11520798864/job/32072978608) and [`crux-mir`](https://github.com/GaloisInc/crucible/actions/runs/11520801062/job/32072985008).

mccleeary-galois (approved):

kquick (approved):
I think these could be a bit more efficient (each `ADD` and `COPY` creates a layer, so we can compress and re-order), but that's unrelated to the changes being effected here, so we can address that separately.

sauclovian-g (commented):
There's stuff you might want to clean up further in the future, like it sets LANG and LC_ALL more than once. Also I'm surprised you need to put /usr/local/lib in LD_LIBRARY_PATH (or maybe you don't actually).

(also if C.UTF-8 works, that's great, I just remember complaining to I think Debian when it didn't work and being told that's not a bug.)

sauclovian-g:
Hrm, maybe you do. N/M.

C.UTF-8 _ought_ to work. It just didn't at one point in the relatively recent past (2018ish?)

PR #1266, opened 2024-11-26
`crucible-syntax`: Allow exotic characters in fresh atom names

### LINKED ISSUE #1024: `crucible-syntax`: `let ... fresh` fails to parse when identifier contains a hyphen
Consider the following `crucible-syntax` program, which does work:

```racket
(defun @test () Unit
  (start start:
    (let i1 (fresh Bool))
    (let i-2 #t)
    (return ())))
```

This shows that you can (1) initialize values with `fresh` symbolic values, and (2) define names with hyphens. If you combine these two things, however:

```racket
(defun @test () Unit
  (start start:
    (let i-3 (fresh Bool))
    (return ())))
```

Then it fails to parse for some mysterious reason:

```
$ cabal run exe:crucible -- check test.cbl
At test.cbl:3:14, expected an atom but got (fresh Bool)
```

RyanGlScott:
Previously, we used `userSymbol` to create fresh atom names, but `userSymbol` will fail if given any characters outside of the very narrow range of characters allowed in What4 solver names. Instead, we now use `safeSymbol`, which will Z-encode any unallowable characters into something that is allowed in a What4 solver name. This ensures (among other things) that we can write things like `(let i-3 (fresh Bool))`, as hyphens are not allowed in What4 solver names without Z-encoding them.

Fixes #1024.

### DIFF
--- a/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
@@ -1564,21 +1564,24 @@ atomSetter (AtomName anText) =
 
     fresh =
       do t <- reading (unary Fresh isType)
-         describe "user symbol" $
-           case userSymbol (T.unpack anText) of
-             Left err -> describe (T.pack (show err)) empty
-             Right nm ->
-               do loc <- position
-                  case t of
-                    Some (FloatRepr fi) ->
-                      Some <$>
-                        freshAtom loc (FreshFloat fi (Just nm))
-                    Some NatRepr ->
-                      Some <$> freshAtom loc (FreshNat (Just nm))
-                    Some tp
-                      | AsBaseType bt <- asBaseType tp ->
-                          Some <$> freshAtom loc (FreshConstant bt (Just nm))
-                      | otherwise -> describe "atomic type" $ empty
+         -- Note that we are using safeSymbol below to create a What4 symbol
+         -- name, which will Z-encode names that aren't legal solver names. This
+         -- includes names that include hyphens, which are very common in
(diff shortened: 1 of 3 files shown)


crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -1563,22 +1563,26 @@ atomSetter (AtomName anText) =
...
+         case t of
+           Some (FloatRepr fi) ->
+             Some <$>
+               freshAtom loc (FreshFloat fi (Just nm))
+           Some NatRepr ->
+             Some <$> freshAtom loc (FreshNat (Just nm))
+           Some tp
+             | AsBaseType bt <- asBaseType tp ->
+                 Some <$> freshAtom loc (FreshConstant bt (Just nm))
+             | otherwise -> describe "atomic type" $ empty
```
sauclovian-g, line 1585: I don't think you want two layers of describe here...

sauclovian-g: ...although looking more carefully I notice this isn't new behavior. Might want to fix it up anyway.

sauclovian-g (approved):

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
RyanGlScott: Good point. I think calling `describe` twice made more sense back when this function had two separate ways to fail, with one of the failures being `userSymbol`-related. Now that the only way to fail is by constructing a non-atomic type, I think that should be the only place we call `describe`. I've pushed a new version which does this.

PR #1274, opened 2025-01-22
syntax: `Pretty` instance for `ExprError`

langston-barrett:


crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -160,6 +161,26 @@ instance Semigroup (ExprErr s) where
...
 
+instance PP.Pretty (ExprErr s) where
+  pretty =
+    \case
+      TrivialErr p ->
+        "Trivial error at" PP.<+> PP.viaShow p
+      Errs e1 e2 ->
+        PP.vcat ["Multiple errors:" , PP.pretty e1 , PP.pretty e2]
+      DuplicateAtom p a ->
+        "Duplicate atom `" PP.<> PP.pretty a PP.<> "` at" PP.<+> PP.viaShow p
```
RyanGlScott, line 172: A minor thing, but I'd prefer using the [`enclose`](https://hackage.haskell.org/package/prettyprinter-1.7.1/docs/Prettyprinter.html#v:enclose) combinator rather than manually enclosing `a` with backticks, i.e.,

```suggestion
      DuplicateAtom p a ->
        "Duplicate atom" PP.<+> PP.enclose "`" "`" (PP.pretty a) PP.<+> "at" PP.<+> PP.viaShow p
```

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -160,6 +161,26 @@ instance Semigroup (ExprErr s) where
...
+      DuplicateLabel p l ->
+        "Duplicate label `" PP.<> PP.pretty l PP.<> "` at" PP.<+> PP.viaShow p
```
RyanGlScott, line 174: Similarly, use `enclose` here. (We might consider defining `` backticks = enclose "`" "`" `` for symmetry with the existing `parens` and `brackets` combinators in `prettyprinter`.)

RyanGlScott (approved):

PR #1275, opened 2025-01-22
cli: Pretty-print syntax errors

langston-barrett:


crucible-cli/src/Lang/Crucible/CLI.hs
```diff
@@ -139,8 +140,8 @@ simulateProgramWithExtension mkExt fn theInput outh profh opts hooks =
             let hdls = [ (SomeHandle h, p) | (FnBinding h _,p) <- ovrs ]
             parseResult <- top ng ha hdls $ prog v
             case parseResult of
-              Left (SyntaxParseError e) -> T.hPutStrLn outh $ printSyntaxError e
-              Left err -> hPutStrLn outh $ show err
+              Left e ->
+                T.hPutStrLn outh (PP.renderStrict (PP.layoutPretty PP.defaultLayoutOptions (PP.pretty e)))
```
RyanGlScott, line 144: There are some similar pieces of code in `crucible-concurrency` and `crucible-syntax` as well:

* https://github.com/GaloisInc/crucible/blob/0f1b3b80e39e3aba9af3edeb5967471c47306b20/crucible-concurrency/src/Cruces/CrucesMain.hs#L141-L142
* https://github.com/GaloisInc/crucible/blob/0f1b3b80e39e3aba9af3edeb5967471c47306b20/crucible-syntax/src/Lang/Crucible/Syntax/Prog.hs#L75-L76

Should these be updated too?

RyanGlScott (approved):

crucible-cli/src/Lang/Crucible/CLI.hs
langston-barrett: Good catch, yes!

PR #1277, opened 2025-01-23
Fix `crux-mir` Docker image entrypoint

### LINKED ISSUE #1276: crux-mir image won't run
There is no longer a `/usr/local/bin/cargo`. It's been moved to `/home/crux-mir/.cargo`.

RyanGlScott:
This fixes an oversight introduced in #1265 where the location of `cargo` was moved to `/home/crux-mir/.cargo/bin/cargo` in the `crux-mir` Docker image, but the image entrypoint was not updated accordingly.

Fixes #1276.


kquick (approved):
At some point it might be nice to add a test that the docker entrypoint works for some minimal input.  https://github.com/GaloisInc/crucible/issues/1278

PR #1295, opened 2025-02-06
crucible: Replace `viaShow` with a real Pretty` instance for `TypeRepr`

### LINKED ISSUE #1016: Add a proper pretty-printer for `TypeRepr`s

langston-barrett:
Fixes #1016.


crucible/src/Lang/Crucible/Types.hs
```diff
@@ -455,8 +457,47 @@ instance HashableF TypeRepr where
...
+      BoolRepr -> "Bool"
+      NatRepr -> "Nat"
+      IntegerRepr -> "Integer"
+      RealValRepr -> "RealVal"
+      ComplexRealRepr -> "ComplexReal"
+      BVRepr n -> parens ("Bitvector" <+> viaShow n)
+      IntrinsicRepr name tys ->
+        parens (pretty (symbolRepr name) <+> prettyCtx tys)
+      RecursiveRepr name tys ->
+        parens (pretty (symbolRepr name) <+> prettyCtx tys)
```
RyanGlScott, line 482: We use the same pretty-printing logic to print `IntrinsicRepr`s and `RecursiveRepr`s, but I worry that it will be easy to confuse the two. Is it worth thinking of a convention to distinguish the two?

crucible/src/Lang/Crucible/Types.hs
```diff
@@ -455,8 +457,47 @@ instance HashableF TypeRepr where
...
+      AnyRepr -> "Any"
+      UnitRepr -> "Unit"
+      BoolRepr -> "Bool"
+      NatRepr -> "Nat"
+      IntegerRepr -> "Integer"
+      RealValRepr -> "RealVal"
+      ComplexRealRepr -> "ComplexReal"
+      BVRepr n -> parens ("Bitvector" <+> viaShow n)
+      IntrinsicRepr name tys ->
+        parens (pretty (symbolRepr name) <+> prettyCtx tys)
```
RyanGlScott, line 480: This part of the patch gives me pause, since (if I'm reading it correctly) types like `LLVMPointerRepr 64` will be printed as something like `(LLVM_pointer (Bitvector 64))`. That is better than the `Show` output, I suppose, but it's a bit inconsistent with how other `*Repr`s are printed. Then again, it's hard to do better here, since this instance has no knowledge that specific `Intrinsic`s should be printed in certain ways.

I wonder if in addition to defining this `Pretty` instance, we should also provide a `ppTypeRepr` function that allows users to customize how `IntrinsicRepr`s are printed. Then, we could also add a wrapper function in `crucible-llvm` that builds on top of `ppTypeRepr` to print `LLVMPointerRepr`s in a prettier way. (Similarly for other `Intrinsic`s as well.) What do you think?

RyanGlScott (commented):
It would be worth advertising these changes in the `crucible` changelog, just in case anyone is relying on the precise `Pretty` output of a `TypeRepr` in their code.

crucible/src/Lang/Crucible/Types.hs
langston-barrett: Yeah! Maybe we prefix recursive types with `Rec`? So you'd get `(Rec <name> args)`?

RyanGlScott: I like the `(Rec <name> args)` suggestion.

crucible/src/Lang/Crucible/Types.hs
langston-barrett: > types like LLVMPointerRepr 64 will be printed as something like (LLVM_pointer (Bitvector 64)).

Yes, that's right.

> Then again, it's hard to do better here, since this instance has no knowledge that specific Intrinsics should be printed in certain ways.

Right.

> I wonder if in addition to defining this Pretty instance, we should also provide a ppTypeRepr function that allows users to customize how IntrinsicReprs are printed. Then, we could also add a wrapper function in crucible-llvm that builds on top of ppTypeRepr to print LLVMPointerReprs in a prettier way. (Similarly for other Intrinsics as well.) What do you think?

100% agree! Will do.

crucible/src/Lang/Crucible/Types.hs
```diff
@@ -455,8 +460,84 @@ instance HashableF TypeRepr where
...
+-- | Pretty-print a type.
+--
+-- Attempts to be consistent with the syntax provided in the @crucible-syntax@
+-- package.
+ppTypeRepr ::
+  Monad f =>
+  -- | How to print 'IntrinsicRepr', see 'ppIntrinsicDefault'
+  (forall s ctx. SymbolRepr s -> CtxRepr ctx -> f (Doc ann)) ->
+  TypeRepr tp ->
+  f (Doc ann)
```
RyanGlScott, line 485: What is the use case for making this function monadic?

langston-barrett: You might want to fail if you find an intrinsic that you don't know about / expect

langston-barrett: e.g., maybe you have a `Map` from `SymbolRepr`s to printers for those intrinsic types, and you want to return `Nothing` when a lookup in that `Map` fails

RyanGlScott: OK. This wasn't obvious to me when reading the Haddocks—could you add this justification?

crucible/src/Lang/Crucible/Types.hs
```diff
@@ -455,8 +460,84 @@ instance HashableF TypeRepr where
...
+-- The following specialization is used in the 'Pretty' instance and doesn't
+-- need to generate code for (>>=), so it seems worth specializing for it.
+{-# SPECIALIZE
+  ppTypeRepr :: 
+    (forall s ctx. SymbolRepr s -> CtxRepr ctx -> Identity (Doc ann)) ->
+    TypeRepr tp ->
+    Identity (Doc ann) #-}
```
RyanGlScott, line 492: You probably want to specialize `prettyCtx` as well.

RyanGlScott (approved):

crucible/src/Lang/Crucible/Types.hs
langston-barrett: I will add a comment!

PR #1297, opened 2025-02-07
crux: Add a `--debug` flag to drop into the Crucible debugger

langston-barrett:


crux/crux.cabal
```diff
@@ -34,6 +34,8 @@ library
     containers,
     contravariant >= 1.5 && < 1.6,
     crucible,
+    crucible-debug,
+    crucible-syntax,
```
RyanGlScott, line 38: Worth mentioning in the changelog that `crux` now incurs some extra dependencies (`crucible-debug` and `crucible-syntax`).

crux/src/Crux/Config/Common.hs
```diff
@@ -364,6 +367,10 @@ cruxOptions = Config
 
           onlineProblemFeatures <- pure noFeatures
 
+          debug <-
+            section "debug" yesOrNoSpec False
+            "Enable the Crucible debugger"
```
RyanGlScott, line 372: It would be worth clarifying that the debugger starts before simulation begins (similarly for the other places where "Enable the Crucible debugger" is documented).

RyanGlScott (approved):

PR #1305, opened 2025-02-11
`crucible-llvm`: Add an `LLVMValString` case for `unpackMemValue`

### LINKED ISSUE #1304: Missing `unpackMemValue` case for `LLVMValString`
The `unpackMemValue` function has a case for `LLVMValArray`, but not its cousin `LLVMValString`:

https://github.com/GaloisInc/crucible/blob/850f88fd146603a3c8f9bbe402ff689534438f04/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1538-L1539

This means that if you call `unpackMemValue` on an `LLVMValString`, it will trigger this panic:

https://github.com/GaloisInc/crucible/blob/850f88fd146603a3c8f9bbe402ff689534438f04/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1555-L1560

I do not know how to trigger this issue with `crux-llvm` alone, but I have seen this panic arise in the context of a large SAW proof. Fortunately, the fix is simple: add a case for `LLVMValString` that calls [`explodeStringValue`](https://github.com/GaloisInc/crucible/blob/850f88fd146603a3c8f9bbe402ff689534438f04/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs#L413-L420) on it and unpacks the resulting `LLVMValArray`.

RyanGlScott:
This is largely inspired by a similar change needed for https://github.com/GaloisInc/saw-script/pull/2149.

Fixes #1304.

### DIFF
--- a/crucible-llvm/CHANGELOG.md
+++ b/crucible-llvm/CHANGELOG.md
@@ -1,3 +1,8 @@
+# next -- TBA
+
+* Fix a bug in which the memory model would panic when attempting to unpack
+  constant string literals.
+
 # 0.7 -- 2024-08-30
 
 * Add support for GHC 9.8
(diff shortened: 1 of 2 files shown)


sauclovian-g (approved):
Seems reasonable

langston-barrett (approved):

Issue #1253, opened 2024-09-09
`crucible-mir`: Support multiple `mir-json` JSON schema versions (within reasonable limits)


RyanGlScott:
Currently, `crucible-mir` is only able to parse and reason about a single snapshot of the `mir-json` JSON schema. This is a somewhat extreme limitation from a user perspective, however, as it means that tools which depend on `crucible-mir` (i.e., `crux-mir` and SAW) must know exactly which JSON schema their tool supports and install a matching version of `mir-json`. This gets even more unwieldy if you try to install nightly versions of `crux-mir` and SAW simultaneously, as they may have different, incompatible JSON schema requirements. (See https://github.com/GaloisInc/saw-script/issues/2111 for an example where this has bitten in the wild.)

(Note that at the moment, `mir-json`'s JSON schema is implicitly versioned. Starting with https://github.com/GaloisInc/mir-json/pull/75, I hope to make this versioning explicit.)

Note that `crucible-llvm` is more flexible: it uses `llvm-pretty-bc-parser`, which is capable of parsing multiple LLVM versions and reifying them all into a single, consistent AST. While this does require more work on our end to maintain support for multiple versions, the payoff is worth it for users, as it gives them much more flexibility in picking LLVM versions. It would be great if we had similar flexibility with `mir-json`, especially given that it may change more frequently than LLVM does.

That being said, it is unclear whether we want to support all `mir-json` schema versions that have ever existed, especially given that `mir-json` may change versions more frequently. As such, I propose that we only support `mir-json` versions within a certain window—let's say the previous three `mir-json` schema versions. This should strike a nice balance of flexibility while not adding an especially undue amount of maintenance burden.

Once we have a `mir-json` JSON schema support window established, we can teach `crucible-mir` to check the schema version in a MIR JSON file, and if the version falls outside the support window, it should throw a proper error message rather than giving a confusing error message later down the line (as seen in https://github.com/GaloisInc/saw-script/issues/2111).

sauclovian-g:
My suggestion would be to tie it to saw and/or crux releases: support the versions produced by the mir-json included with (or implicitly included with) supported saw and crux releases. Possibly with a proviso that schema versions that never appeared in a release (and thus only appeared in current/nightly builds) can be dropped after some fixed period of time, maybe a month.

[...]

PR #1309, opened 2025-02-12
`crucible-mir`: Explicitly check `mir-json` schema version

RyanGlScott:
This ensures that `crucible-mir` is only ever loading MIR JSON files that are
known to be supported.

Fixes https://github.com/GaloisInc/crucible/issues/1253.


crucible-mir/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # next -- TBA
 
+This release supports [version
+1](https://github.com/GaloisInc/mir-json/blob/c52b16bf26af2f5b98157ebf9975aa0021982bbc/SCHEMA_CHANGELOG.md#1)
```
sauclovian-g, line 4: These links should probably point to the head rather than to a specific blob (even though the blob should be stable enough to point to), both for tidiness and because if we ever clarify something it would be better for people to get pointed at the clarified version.

crux-mir/README.md
```diff
@@ -19,6 +19,11 @@ Next, navigate to the `crucible/dependencies/mir-json` directory and install
 `mir-json` according to the instructions in [the `mir-json`
 README][mir-json-readme].
 
+Currently, `crux-mir` supports [version
+1](https://github.com/GaloisInc/mir-json/blob/c52b16bf26af2f5b98157ebf9975aa0021982bbc/SCHEMA_CHANGELOG.md#1)
+of `mir-json`'s schema. As a general policy, `crux-mir` strives to support the
+`mir-json` schema versions corresponding to the last two `crux-mir` releases.
```
sauclovian-g, line 25: shouldn't that be the last two mir-json releases?

sauclovian-g (approved):
I can't remember what the sticking point was before but this all seems fine now...

crucible-mir/CHANGELOG.md
RyanGlScott: To clarify, are you requesting that it point to this URL?

https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#1

Or this URL?

https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md

I would have a slight preference towards the former, since I think pointing directly to the schema version number in question would be what most people are interested in, but I could be convinced of pointing to the latter if you feel strongly.

crux-mir/README.md
RyanGlScott: I'm not sure! There is a real question of how far back the support window should be. I thinking was to support the schema versions used in last two `crux-mir` releases so that when someone upgrades `crux-mir`, they have a smoother upgrade path. By contrast, `mir-json` schemas don't really have "releases" in the traditional sense (at least, not currently), so it's less clear how many `mir-json` schema version bumps might occur between `crux-mir`/SAW releases.

crucible-mir/CHANGELOG.md
sauclovian-g: I was thinking the former, yeah.

crux-mir/README.md
sauclovian-g: Ah, hmm. I thought it was just a typo because as written it reads like a circular definition. This isn't so easy to express coherently, especially without promising that random never-released schema versions will be supported... let me see.

Maybe:

We expect that the schema versions produced by `mir-json` will change irregularly and somewhat unpredictably as dictated by internal requirements and upstream changes. To help smooth this over, we intend that once `crux-mir` introduces support for any given schema version, it will retain that support across at least two releases.

sauclovian-g: Although we might want to waffle even harder such that we don't promise that across changes in the supported Rust version. Since virtually any new Rust version is bound to change _something_ that leads to a schema change, and we've already decided we don't want to try supporting multiple Rust versions at once...

crucible-mir/CHANGELOG.md
RyanGlScott: I've altered the changelogs and the README to use the former link.

crux-mir/README.md
RyanGlScott: I've updated the wording. Does this look better?

crux-mir/README.md
```diff
@@ -19,6 +19,22 @@ Next, navigate to the `crucible/dependencies/mir-json` directory and install
...
 README][mir-json-readme].
 
+Currently, `crux-mir` supports [version
+1](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#1) of
+`mir-json`'s schema. Note that the schema versions produced by `mir-json` can
+change over time as dictated by internal requirements and upstream changes. To
+help smooth this over, we intend that once:
+
+* `crux-mir` introduces support for any given schema version, it will retain
+  that support across at least two releases.
```
RyanGlScott, line 29: Upon a closer read, this would be naturally structured as:

```suggestion
help smooth this over:

* We intend that once `crux-mir` introduces support for any given schema version, it will retain
  that support across at least two releases.
```

crux-mir/README.md
sauclovian-g: Yes (with the amendment)

PR #1312, opened 2025-02-20
`crucible-mir`: Improve pretty-printing of ZST constants

RyanGlScott:
`ConstZST` (i.e., constant values with zero-sized types) do not include any information about what the zero-sized type is. In particular, function constants are represented as `ConstZST`s, and the only way to determine what function the `ConstZST` corresponds to is to look at its type (`FnDef`), which stores the `DefId` of the function name itself.
    
Previously, the `crucible-mir` pretty-printer was printing `ConstZST`s as `<ZST>` because it did not have access to any type information when pretty-printing `ConstVal`s. This results in very suboptimal pretty-printed output (see the example below). This is also true for other forms of `ConstZST`s, although currently `mir-json` only turns function constants into `ConstZST`s. (In the future, this could conceivably change.)
    
To fix this issue, we add special cases in the `Pretty Constant` instance (which does have access to type information for `ConstVal`s) such that `ConstZST`s are printed using their type information. In particular, if the `ConstZST` has type `TyFnDef`, then the function `DefId` stored in the `TyFnDef` is printed. All other forms of `ConstZST`s are printed like `<ZST: [ty]>`, where `[ty]` is the pretty-printed type.

As an example of the improvements in this patch, consider this program:

```rs
pub fn foo() {}
pub fn bar() { foo() }
```

Before the changes in this patch, the MIR for `bar()` would be pretty-printed as:

```rs
fn test/f3b25c00::bar[0]() -> () {
   let mut _0 : ();
   bb0: {
      call(_0 = <ZST>(), bb1)
   }
   bb1: {
      return;
   }
}
```

(Note the call to `<ZST>()`, which gives no indication whatsoever about what function was called.)

After the changes in this patch, it is pretty-printed as:

```rs
fn test/4717ccf0::bar[0]() -> () {
   let mut _0 : ();
   bb0: {
      call(_0 = test/4717ccf0::foo[0](), bb1)
   }
   bb1: {
      return;
   }
}
```

Fixes GaloisInc/mir-json#79.

### DIFF
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -249,7 +249,27 @@ instance Pretty Operand where
     pretty (Temp c) = pretty c
 
 instance Pretty Constant where
-    pretty (Constant _a b) = pretty b
+    pretty (Constant a b) =
+      case (a, b) of
+        -- We include two special cases for ConstZST that leverage its type to
+        -- improve the pretty-printing:
+        --
+        -- - If the type is TyFnDef, then we pretty-print the function's DefId.
+        --   (This is crucial to ensure that pretty-printing function
+        --   applications reveals the name of the applied function.)
+        --
+        -- - Otherwise, we print <ZST: [ty]>, where [ty] is the pretty-printed
+        --   type.
+        --
+        -- There is also a case for ConstZST in the `Pretty ConstVal` instance,
+        -- but the output there (<ZST>) is much less informative, as it cannot
+        -- take the type of the ConstZST into account.
+        (TyFnDef defId, ConstZST) ->
+          pr_id defId
+        (_, ConstZST) ->
+          pretty "<ZST:" <+> pretty a <> pretty ">"
+        _ ->
+          pretty b
 
 instance Pretty NullOp where
     pretty SizeOf = pretty "sizeof"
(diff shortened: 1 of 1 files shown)


crucible-mir/src/Mir/PP.hs
```diff
@@ -323,7 +331,14 @@ instance Pretty ConstVal where
     pretty (ConstFunction a)   = pr_id a
     pretty (ConstInitializer a) = pr_id a
     pretty (ConstStaticRef a) = pretty "&" <> pr_id a
-    pretty ConstZST = pretty "<ZST>"
+    pretty ConstZST =
+      -- A ConstZST value represents a function definition, but because this is
```
spernsteiner, line 335: >A ConstZST value represents a function definition

I think this comment is overly specific.  It's currently the case that we use `ConstZST` only for `TyFnDef`, but it would be reasonable (IMO) for mir-json to use it for zero-sized structs/enums, empty tuples and closures, and zero-length arrays.  It looks like the code in `Trans.hs` that actually handles `ConstZST` actually happens to support all of these cases already.

Given that all the useful information is in the type in all of these cases, maybe it would be better to have the `Pretty Constant` instance always print the type if the expression is `ConstZST` (with a special case for `TyFnDef` if the `Pretty Ty` output doesn't look good).

spernsteiner (approved):

PR #1313, opened 2025-02-20
Use `*WithWarnings` functions from `llvm-pretty-bc-parser`

RyanGlScott:
This bumps the `llvm-pretty-bc-parser` submodule to bring in the changes from https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/286. It also updates the code to use the `*WithWarnings` variants of bitcode parsing functions.


RyanGlScott:
Oops, the commit message should mention https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/282, not https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/286.

crucible-llvm/test/Tests.hs
```diff
@@ -112,11 +113,13 @@ assemble (LLVMAssembler llvm_as) !inputFile !outputFile =
...
 parseLLVM :: FilePath -> IO (Either String Module)
 parseLLVM !file =
-  parseBitCodeFromFile file >>=
+  parseBitCodeFromFileWithWarnings file >>=
     \case
       Left err -> pure $ Left $ "Couldn't parse LLVM bitcode from file"
                                 ++ file ++ "\n" ++ show err
-      Right m  -> pure $ Right m
+      Right (m, warnings) -> do
+        IO.hPrint IO.stderr $ ppParseWarnings warnings
```
sauclovian-g, line 121: Like the one in saw this might want a check for whether warnings is empty.

RyanGlScott: Good catch. Done.

crux-llvm/src/Crux/LLVM/Simulate.hs
```diff
@@ -256,6 +264,13 @@ prepLLVMModule llvmOpts halloc bak bcFile memVar = do
                  =<< initializeAllMemory bak llvmCtxt llvmMod
     return $ PreppedLLVM llvmMod (Some trans) memVar mem
 
+sayParseWarnings ::
+  Log.SupportsCruxLLVMLogMessage msgs =>
+  Crux.Logs msgs =>
+  Seq ParseWarning -> IO ()
+sayParseWarnings =
+  Log.sayCruxLLVM . Log.ParseWarning . Text.pack . show . ppParseWarnings
```
sauclovian-g, line 272: this probably also wants a null check (sorry I didn't notice it as well yesterday)

sauclovian-g (approved):

crux-llvm/src/Crux/LLVM/Simulate.hs
RyanGlScott: I tested this locally, and interestingly enough, the logging framework does not emit a newline if you load a bitcode file without any warnings. Perhaps there is some sort of empty string check somewhere in the logging framework?

In any case, this is pretty non-obvious, so I'll add an explicit null check here just to be on the safe side.

PR #1316, opened 2025-03-05
Rename crux-mir's test suite

sauclovian-g:


RyanGlScott (approved):
I've also been mildly annoyed by this, but never enough to act on it. Thank you for doing so!

(For what it's worth, `crux-mir-comp`'s test suite is [also named `test`](https://github.com/GaloisInc/saw-script/blob/ca335befff7e70e85aea6057d6468d08974afe33/crux-mir-comp/crux-mir-comp.cabal#L78), so we may want to rename that as well for consistency.)

sauclovian-g:
Yes, that's how I came to discover this :-) Renaming it is in the pipeline over there...

RyanGlScott:
Note that a corresponding change is being made on the SAW side in https://github.com/GaloisInc/saw-script/pull/2280, where `crux-mir-comp`'s test suite is being renamed from `test` to `crux-mir-comp-test`.

PR #1318, opened 2025-03-07
Additional tests and documentation for `Backend`

langston-barrett:


crucible/test/helpers/Main.hs
```diff
@@ -1,20 +1,84 @@
...
+backendTests :: TestTree
+backendTests =
+  testGroup
+  "Backend"
+  [ TTH.testCase "push/pop nothing" $ do
+      Some (SomeBackend bak) <- mkBackend
+      asmps <- LCB.popAssumptionFrame bak =<< LCB.pushAssumptionFrame bak
+      p <- LCB.assumptionsPred (LCB.backendGetSym bak) asmps
+      Just True TTH.@=? W4I.asConstantPred p
+  , TTH.testCase "push/pop assumptions" $ do
```
RyanGlScott, line 52: Could you leave a high-level comment next to each test case that describes the property it is testing for? I'll admit that the code is dense enough that it's hard to tell just by looking at it...

crucible/test/helpers/Main.hs
```diff
@@ -1,20 +1,84 @@
...
+  , TTH.testCase "push/pop obligations" $ do
+      Some (SomeBackend bak) <- mkBackend
+      let sym = LCB.backendGetSym bak
+      a <- W4I.freshConstant sym (W4I.safeSymbol "a") W4I.BaseBoolRepr
+      assumePred bak a
+      frm <- LCB.pushAssumptionFrame bak
+      b <- W4I.freshConstant sym (W4I.safeSymbol "b") W4I.BaseBoolRepr
+      assumePred bak b
+      c <- W4I.freshConstant sym (W4I.safeSymbol "c") W4I.BaseBoolRepr
+      LCB.assert bak c (GenericSimError "")
```
RyanGlScott, line 72: Worth giving a more descriptive string than "" here?

crucible/test/helpers/Main.hs
```diff
@@ -1,20 +1,84 @@
...
 
+mkBackend :: IO (Some SomeBackend)
+mkBackend = do
+  Some nonceGen <- newIONonceGenerator
+  sym <- newExprBuilder FloatIEEERepr EmptyExprBuilderState nonceGen
+  Some . SomeBackend <$> newSimpleBackend sym
+
+assumePred :: LCB.IsSymBackend sym bak => bak -> W4I.Pred sym -> IO ()
+assumePred bak p =
+  LCB.addAssumption bak (LCB.GenericAssumption W4P.initializationLoc "" p)
```
RyanGlScott, line 41: Worth giving a more descriptive string than "" here?

RyanGlScott (approved):

crucible/test/helpers/Main.hs
langston-barrett: I'm actually a bit surprised to hear that you found the code challenging to read! It says:

1. Make a fresh predicate `a`
2. Assume `a`
3. Push a frame
4. Make a fresh predicate `b`
5. Assume `b`
6. Pop the frame
7. Assert that the popped assumptions are equal to `b`

To me, that's a pretty direct reading (basically one step per line) of lines 55-62.

I will add a comment that says "check that popping returns assumptions from the popped frame".

crucible/test/helpers/Main.hs
langston-barrett: Do you have any suggestions? The predicates being asserted in these tests are just fresh symbolic constants, they don't mean anything.

crucible/test/helpers/Main.hs
langston-barrett: I'm not exactly sure what to put considering that this is asserting a fresh symbolic constant. I could put something like `"c"` or `"test"`, but I'm not sure how much of an improvement that would be.

crucible/test/helpers/Main.hs
RyanGlScott: I was just hoping to have some sort of unique, easily searchable message in the event that there is a bug and these messages are reported as part of a test assertion failure.

langston-barrett: Fair enough. While I remain somewhat skeptical of the value added, I'm more than willing to defer and add some strings.

crucible/test/helpers/Main.hs
langston-barrett: Added `"asserting c"`

PR #1319, opened 2025-03-07
`crux-mir`: Migrate modified Rust standard libraries to `mir-json`

### LINKED ISSUE #1252: Maintain modified Rust standard libraries and `translate_libs.sh` in a separate repo
Currently, we vendor in our modified copies of the Rust standard libraries (as well as the associated `translate_libs.sh` script, which runs `mir-json` on these modified standard libraries to compile them) as a subdirectory in `crux-mir`. Historically, this made sense, as `crux-mir` was the sole consumer of these files. It is unclear if this arrangement makes as much sense nowadays, however, for the following reasons:

1. Both `crux-mir` and SAW (the latter of which now has a MIR backend) need to make use of `translate_libs.sh`. Telling SAW users to download `crux-mir` for the sole purpose of running `translate_libs.sh` doesn't make much sense, given that `crux-mir` is otherwise unrelated to SAW.
2. The vendored-in copies of the Rust standard libraries are actually quite large (on the order of several megabytes), which substantially increases the amount of time it takes to clone the `crucible` repo. This is a bit of a shame, considering that many `crucible` developers aren't going to be using the MIR-related parts.

To address both of these concerns, I propose that we factor out the modified Rust standard libraries (and `translate_libs.sh`) into a separate repo. I'll tentatively call this repo "`mir-json-libs`" until I can think of something better. That way, `mir-json-libs` development can live outside of `crucible` (thereby avoiding further bloating of the `crucible` repo), and there is now a central location for further standard library modifications needed to support `crux-mir` and SAW.

RyanGlScott:
This patch:

* Removes the modified versions of the Rust standard libraries that were vendored in to `crux-mir/lib`, as well as the associated `crux-mir/translate_libs.sh` script. These were migrated to the `mir-json` repo as part of https://github.com/GaloisInc/mir-json/pull/85.
* Adjusts the various references to the standard libraries to ensure that `crux-mir`'s documentation and CI remain up to date.

Fixes #1252.


.github/ci.sh
```diff
@@ -158,7 +158,6 @@ bundle_crux_mir_files() {
   setup_dist
   extract_exe crux-mir dist/bin
   cp crux-mir/README.md dist/doc
-  cp -r crux-mir/rlibs dist
```
RyanGlScott, line 161: This is perhaps one of the more mysterious changes. I _think_ the intent here is to include the compiled Rust standard libraries as part of a `crux-mir` bindist so that `crux-mir` users can make use of them. That being said, this was broken in practice—`rlibs` is a symlink, so this `cp` invocation did nothing but copy a broken symlink. (Now that the `translate_libs.sh` is no longer located under `crux-mir/`, I decided to just remove this line rather than updating it to point to its new home in `mir-json/`.)

We should probably have a think about whether we want to continue to bundle the compiled standard libraries with tools like `crux-mir` and/or SAW in the future, as I think there is a case to be made that they should be distributed separately. I can defer this to another issue, if folks are amenable to that.

sauclovian-g: I think it makes sense to bundle them with mir-json; it is effectively the compiler and they are effectively its stdlib. Plus you'll be installing it if and only if you're working with Rust.

qsctr (commented):
Should these be updated as well?
https://github.com/GaloisInc/crucible/blob/5bdbe074e5bde207433449753f39f40c9fe9f5f9/crux-mir/README.md?plain=1#L70
https://github.com/GaloisInc/crucible/blob/5bdbe074e5bde207433449753f39f40c9fe9f5f9/crux-mir/README.md?plain=1#L113-L116

.github/Dockerfile-crux-mir
```diff
@@ -39,7 +39,9 @@ ENV LANG=C.UTF-8 \
 WORKDIR ${CRUX_BUILD_DIR}/dependencies/mir-json
 RUN curl https://sh.rustup.rs -sSf | bash -s -- -y --profile minimal --default-toolchain ${RUST_TOOLCHAIN}
 RUN rustup component add --toolchain ${RUST_TOOLCHAIN} rustc-dev
-RUN cargo install --locked
+RUN cargo install --locked && \
+    ./translate_libs.sh
+ENV CRUX_RUST_LIBRARY_PATH=$(pwd)/rlibs
```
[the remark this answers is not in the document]
RyanGlScott: This is indeed for `crux v2-test crux-mir`. (The `CRUX_RUST_LIBRARY_PATH` is in a different Docker stage and is a separate matter entirely.) I'll leave a comment on the top-most `CRUX_RUST_LIBRARY_PATH` definition that it is defined for the benefit of `crux v2-test crux-mir`.

PR #1322, opened 2025-03-13
ci: Use `cabal-ghc-compat` action

langston-barrett:
Testing out https://github.com/GaloisInc/.github/pull/13


langston-barrett:
We should instead do something like https://github.com/GaloisInc/parameterized-utils/pull/191

PR #1325, opened 2025-03-18
Bump submodules before release

sauclovian-g:
(llvm-pretty, what4, mir-json)


dependencies/what4
marsella, line 1: This library has tagged releases but the latest ([1.6.3](https://github.com/GaloisInc/what4/tags)) isn't on [master](https://github.com/GaloisInc/llvm-pretty/commits/master/). I'm not sure what the intended structure is there so I don't know whether to recommend using that tagged version or not.

marsella (commented):
I'm confused about the versioning for what4, and my question on [llvm-pretty-bc-parser](https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/289) also applies here.

dependencies/what4
langston-barrett: It's not uncommon for projects to create branches for releases, and it looks like What4 has this structure - the commit tagged 1.6.3 lives on the 1.6 branch. This structure allows "backporting" fixes to previous releases (the commit message mentions this idea). So, for instance, if What4 `master` progressed and we released 1.7.0, we might discover a critical bug and want to release both 1.7.1 and 1.6.4. Having a 1.6 branch around makes this process simpler, as it doesn't have any of the other features or bugfixes added between 1.6.3 and 1.7.0.

I don't actually know What4's branching / releasing strategy, but this is my best guess as to what's going on.

langston-barrett (approved):

langston-barrett:
Is this in advance of making a Hackage release for Crucible? It would be a good idea to write some developer documentation on this process, in part so that reviewers can know what to look for.

sauclovian-g:
Yeah, this is the first step (do all material submodule changes/updates) just in case one blows up. Then the ones during the actual release process only involve administrative updates and are relatively safe.

Regarding what4, what seems to have happened is that we decided to push the timeout and exit fixes onto a stable branch branching off from 1.6.2 and make a point release without the ```ConjPred``` change. Then the version number also got bumped on the trunk without being entirely clear about why, so the history is slightly confusing.

Crucible has been using the ```ConjPred``` change, but didn't yet have the exit fixes.

sauclovian-g:
Re developer documentation, that is in the works. In the meantime, the goal is to be able to have each release pointing at the release versions of things it depends on, so at this stage the important questions are (a) is it broken? and (b) are all the submodules and deps pointing at a commit that's the same as the release version except for administrative changes?

If you want to crosscheck (b) for me that would be great but it will be a lengthy pain in the neck. 😐

sauclovian-g:
Freeze files will get done separately.

PR #1327, opened 2025-03-20
Bump parameterized-utils to new release 2.1.10.0

sauclovian-g:


cabal.project
```diff
@@ -23,6 +23,7 @@ packages:
   crux-llvm/
   crux-mir/
   uc-crux-llvm/
+  https://hackage.haskell.org/package/parameterized-utils-2.1.10.0/candidate/parameterized-utils-2.1.10.0.tar.gz
```
langston-barrett, line 26: Is this still necessary? I see 2.1.10.0 listed at https://hackage.haskell.org/package/parameterized-utils (though it appears its Haddocks are broken...)

langston-barrett: Ah, not "broken", I spoke too soon. It says "Docs pending".

RyanGlScott (commented):
No particular objections here, but like with https://github.com/GaloisInc/cryptol/pull/1827, it might be better to upgrade `parameterized-utils` and `what4` in one go.

cabal.project
sauclovian-g: Sorry, the commentary's in the corresponding cryptol pull request that Ryan linked. It's not supposed to be (that's why I removed it, which is why this comment no longer seems to make sense) but apparently I didn't do things right.

Also hackage says normally hackage generates the haddocks and you don't have to upload them explicitly, so they should appear at some point.

sauclovian-g:
Yeah, same plan

sauclovian-g:
It seems that the freeze files don't pin the what4 version. Is that a bug, and if so should it be fixed now or later?

sauclovian-g:
aargh, I screwed it up. There are too many moving parts in this thing

sauclovian-g:
Ok, for reference what is actually in this branch is:
- bump the parameterized-utils version in the cabal freeze files
- also bump the index date so that the new release is visible

What was here temporarily was:
- refer directly to the candidate parameterized-utils release to make sure it works before releasing it

Plus I got myself confused about what4 (because I'd missed that Cryptol gets it via Hackage and then thought the situation here was the same) and there were some unnecessary changes for trying to test the what4 release candidate that have since been reverted.

It should all be ok now.

RyanGlScott:
Would it be worth bumping the `what4` submodule to include the upcoming `what4` release? Or do you plan to do that in a separate PR?

sauclovian-g:
It already has everything except the administrative release changes, and those are in the release-prep branch.

RyanGlScott (approved):

PR #1330, opened 2025-03-20
doc: Add developer documentation on GHC versions

langston-barrett:
Like https://github.com/GaloisInc/parameterized-utils/pull/178, but for Crucible.

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -1,0 +1,30 @@
+# Developer documentation
+
+## GHC versions
+
+We support the three most recent versions of GHC.
+We try to support new versions as soon as they are supported by the libraries that we depend on.
+
+### Adding a new version
+
+The following checklist enumerates the steps needed to support a new version of GHC.
+When performing such an upgrade, it can be helpful to copy/paste this list into the MR description and check off what has been done, so as to not forget anything.
+
+- [ ] Allow the [new version of `base`][base] in the Cabal `build-depends`
+- [ ] Run `cabal {build,test,haddock}`, bumping dependency bounds as needed
+- [ ] Fix any new warnings from [`-Wdefault`][wdefault]
+- [ ] Run [`cabal freeze`][freeze] and rename `cabal.freeze` to `cabal.GHC-X.Y.Z.config`
+- [ ] Add the new GHC version to the matrix in the GitHub Actions workflows
+- [ ] Add the new version to the code that sets `GHC_NIXPKGS` in the CI config
+- [ ] Use the new GHC version in the Dockerfiles in `.github/`
+- [ ] Follow the below steps to remove the oldest GHC version
+
+[base]: https://www.snoyman.com/base/
+[freeze]: https://cabal.readthedocs.io/en/stable/cabal-commands.html#cabal-freeze
+[wdefault]: https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Wdefault
+
+### Removing an old version
+
+- [ ] Remove the old version from the matrix in the GitHub Actions configuration
+- [ ] Remove outdated CPP `ifdef`s that refer to the dropped version
+- [ ] Remove outdated `if` stanzas in the Cabal files


doc/dev.md
```diff
@@ -0,0 +1,30 @@
...
+We support the three most recent versions of GHC.
+We try to support new versions as soon as they are supported by the libraries that we depend on.
+
+### Adding a new version
+
+The following checklist enumerates the steps needed to support a new version of GHC.
+When performing such an upgrade, it can be helpful to copy/paste this list into the MR description and check off what has been done, so as to not forget anything.
+
+- [ ] Allow the [new version of `base`][base] in the Cabal `build-depends`
+- [ ] Run `cabal {build,test,haddock}`, bumping dependency bounds as needed
```
RyanGlScott, line 14: I think this checklist needs to say "update the submodules to include any necessary changes to support newer GHC versions" (or something to that effect). That ends up taking quite a bit of time when adding support for newer GHC versions.

doc/dev.md
```diff
@@ -0,0 +1,30 @@
...
+- [ ] Fix any new warnings from [`-Wdefault`][wdefault]
+- [ ] Run [`cabal freeze`][freeze] and rename `cabal.freeze` to `cabal.GHC-X.Y.Z.config`
+- [ ] Add the new GHC version to the matrix in the GitHub Actions workflows
+- [ ] Add the new version to the code that sets `GHC_NIXPKGS` in the CI config
+- [ ] Use the new GHC version in the Dockerfiles in `.github/`
```
RyanGlScott, line 19: This might deserve some further elaboration. In particular, it appears that we adhere to the convention that the oldest supported GHC version is what is used in the Dockerfiles. As such, part of the process is _bumping_ the GHC versions in the Dockerfiles, not using the newest one (as the text here might imply if you're not reading carefully).

doc/dev.md
langston-barrett: Just so I understand are you saying that we should mention that some of our deps are submodules and so need to be bumped using `git` rather than in the Cabal file? Or are you saying that some of our deps are also Galois-owned and will need to support the new GHC version before Crucible can? Or both? (Feel free to suggest wording, or just answer the above and I can give it a go.)

kquick: Both, IMO.

doc/dev.md
```diff
@@ -0,0 +1,30 @@
...
+- [ ] Follow the below steps to remove the oldest GHC version
+
+[base]: https://www.snoyman.com/base/
```
kquick, line 22: It would be better to use the GHC version information directly rather than a third party's version: https://gitlab.haskell.org/ghc/ghc/-/wikis/commentary/libraries/version-history

doc/dev.md
```diff
@@ -0,0 +1,30 @@
+# Developer documentation
+
+## GHC versions
+
+We support the three most recent versions of GHC.
+We try to support new versions as soon as they are supported by the libraries that we depend on.
```
kquick, line 6: Here again, we should say something about continuing to support older GHC versions as long as it's not a significant burden to do so.

doc/dev.md
```diff
@@ -0,0 +1,30 @@
...
+- [ ] Add the new version to the code that sets `GHC_NIXPKGS` in the CI config
+- [ ] Use the new GHC version in the Dockerfiles in `.github/`
+- [ ] Follow the below steps to remove the oldest GHC version
+
+[base]: https://www.snoyman.com/base/
+[freeze]: https://cabal.readthedocs.io/en/stable/cabal-commands.html#cabal-freeze
+[wdefault]: https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Wdefault
+
+### Removing an old version
+
```
kquick, line 27: Add para:  This enumerates the steps for removing the support for an older version of GHC that we decide will no longer be supported.  We will continue to support older versions until the point where the maintenance becomes burdensome (e.g. significant `#ifdef` in the code, `if` clauses in the cabal file, etc.), although we commit to supporting the latest three versions regardless of the burden.

kquick (changes requested):

doc/dev.md
langston-barrett: Good call, thanks for the link!

doc/dev.md
langston-barrett: See [this discussion](https://github.com/GaloisInc/crucible/pull/968/files#diff-9ad82307f9850d8d79f949bbd805f3f53ca211fd33b2d808bb9aed7dcde52aecL47) for deciding what we want the policy to be.

doc/dev.md
RyanGlScott: > We noted that we had some clients that use (packages that depend on) `parameterized-utils`, but have trouble updating to modern OSes/GHC versions. Is something similar true for Crucible?

_The_ reason that `crucible` is on Hackage at all right now (as far as I am aware) is for the benefit of [`copilot-verifier`](https://hackage.haskell.org/package/copilot-verifier). Much like other Galois projects, `copilot-verifier` only supports three recent-ish GHC versions. As such, I don't think there is as much pressure to support old GHC versions in `crucible`.

Contrast this with [`copilot-theorem`](https://hackage.haskell.org/package/copilot-theorem) (a library which depends on `what4` and `parameterized-utils`, but not `crucible`), which does have a much wider GHC support window.

RyanGlScott (approved):

PR #1332, opened 2025-03-21
MIR: Add custom override for `bitreverse` intrinsic

RyanGlScott:
This contains the necessary changes needed to allow `crucible-mir` to symbolically execute the [`bitreverse`](https://doc.rust-lang.org/std/intrinsics/fn.bitreverse.html) compiler intrinsic, which in turn powers the [`reverse_bits`](https://doc.rust-lang.org/std/primitive.u32.html#method.reverse_bits) family of functions. This will form the basis of a fix for GaloisInc/saw-script#2257.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -11,6 +11,9 @@ This release supports [version
 * The Semigroup and Monoid instances for Collection, CollectionState, and
   RustModule have been removed. It is not expected that there are any
   downstream users.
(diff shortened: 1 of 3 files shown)


qsctr (approved):
I'm not familiar with the internals of crucible, but this seems reasonable to me.

sauclovian-g (approved):
Looks plausible but this is all very hard to follow even when one has seen the Crucible bits before...

PR #1333, opened 2025-03-21
Prepare crux 0.10 release

sauclovian-g:


sauclovian-g:
It would be helpful if people could crosscheck whether my version bumps are right (major vs. minor) -- I am pretty sure all the changelogs are complete but not absolutely certain about the implications of all of them.

crucible-debug/crucible-debug.cabal
```diff
@@ -78,14 +80,14 @@ common shared
     -Werror=warnings-deprecations
     -Werror=wrong-do-bind
 
-  ghc-prof-options: -O2 -fprof-auto-top
+  -- ghc-prof-options: -O2 -fprof-auto-top
```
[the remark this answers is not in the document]
langston-barrett: We could probably set this in the top-level `cabal.project` instead of the individual Cabal files to reduce duplication and avoid the warning. Agreed that we should do it at a different time.

crucible-llvm-cli/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# next
+
+Nothing yet.
+
 # 0.1 -- 2024-02-05
```
langston-barrett, line 5: Should this be 0.1.0.0 (same on other `CHANGELOG`s)?

crucible-syntax/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# next
+
+Nothing yet.
+
+# 0.4.1 -- 2025-03-21
+
+* Internal changes only.
```
langston-barrett, line 7: This should probably list

https://github.com/GaloisInc/crucible/pull/1274

https://github.com/GaloisInc/crucible/pull/1266

crucible-syntax/crucible-syntax.cabal
```diff
@@ -85,12 +85,14 @@ common shared
       -Werror=type-equality-out-of-scope
       -Werror=type-equality-requires-operators
 
-  ghc-prof-options: -O2 -fprof-auto-top
+  --ghc-prof-options: -O2 -fprof-auto-top
   default-language: Haskell2010
 
 library
+  default-language: Haskell2010
```
langston-barrett, line 92: nit: This could go in `shared` to reduce duplication. No need to fix necessarily.

sauclovian-g: A thing I forgot to mention explicitly because it was late last night: it _is_ in ```shared``` but the library isn't using ```shared```, which is almost certainly a mistake but this didn't seem like a good point to try to fix it.

langston-barrett: Thanks for noticing this, I've filed an issue and assigned myself: https://github.com/GaloisInc/crucible/issues/1334

crucible-llvm-cli/CHANGELOG.md
sauclovian-g: It looks like most of the actual package releases (in the cabal files as well as the changelogs) don't have trailing zeros. So probably the cabal files that have 0.1.0 or 0.1.0.0 are the ones that should be changed. But sticking .99 on all of them (and noting what the practice about trailing zeros is in the release docs, which is now on my list) effectively attends to that going forward.

So I could change all of those so they're only 0.1 as of the release commit, but I kinda think it isn't worthwhile.

sauclovian-g:
The versioning policy on Hackage is here: https://pvp.haskell.org/

(it is basically semantic versioning with two-component major versions, and doesn't allow 0.x versions to be special)

langston-barrett:
I'm aware of the PVP (though I appreciate the link), what I wasn't sure of was (1) is that the versioning policy adopted by this project (or is it SemVer? or something else?), and if not, or if the answer is unclear, then (2) is that the policy that you're requesting that we check against? Seems like the answer to (2) is "yes", so I can refresh my memory on the PVP and read through these changes again to check whether I agree with the new version numbers you selected.

For (1), I assume it's not strictly SemVer nor PVP, or at least we add something on top of those, considering the `.99` thing. I probably sound like a broken record, but it would be nice (for both ourselves and consumers of our API) to document what our intended policy is. Not asking you to do that now, just an idea for the future.

crucible-llvm-cli/CHANGELOG.md
langston-barrett: :+1:

RyanGlScott:
`crucible` adheres to the PVP. A version number like `0.1.0.0.99` is PVP-compliant. From [the PVP specification](https://pvp.haskell.org/#version-numbers):

> A package version number **SHOULD** have the form _A.B.C_, and **MAY** optionally have any number of additional components, for example 2.1.0.4 (in this case, _A_=2, _B_=1, _C_=0). This policy defines the meaning of the first three components _A_-_C_, the other components can be used in any way the package maintainer sees fit.

langston-barrett:
Thanks for clarifying that, @RyanGlScott! I assume that applies to the other packages in the repo as well? I'll add this to the README in a separate PR.

RyanGlScott:
> I assume that applies to the other packages in the repo as well?

Yes, that's right.

langston-barrett (approved):
I reviewed the PVP and agree with your version number assignment; the vast majority of the packages had no breaking changes.

crucible-concurrency/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
 # next
 
+Nothing yet.
+
+# 0.1 -- as of 2025-03-21
+
 * TODO: Describe API changes here
```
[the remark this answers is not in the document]
sauclovian-g: Yeah, may as well.

crucible-debug/crucible-debug.cabal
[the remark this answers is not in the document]
sauclovian-g: Yeah, I'm reverting these.

sauclovian-g:
> I reviewed the PVP and agree with your version number assignment; the vast majority of the packages had no breaking changes.

Thank you!

crucible-syntax/CHANGELOG.md
sauclovian-g: I added these, lmk if you have better wordings/improvements.

RyanGlScott (approved):

crucible-syntax/CHANGELOG.md
langston-barrett: LGTM, thanks!

PR #1335, opened 2025-03-21
syntax: Import Cabal `shared` stanza in `library` stanza, fix warnings

### LINKED ISSUE #1334: syntax: Cabal `library` stanza does not `import` the `shared` stanza
This is almost certainly just an oversight.

This

https://github.com/GaloisInc/crucible/blob/0d90f74d11c5cb74672537700c3825b4d5d8b4db/crucible-syntax/crucible-syntax.cabal#L91

should do this

https://github.com/GaloisInc/crucible/blob/0d90f74d11c5cb74672537700c3825b4d5d8b4db/crucible-syntax/crucible-syntax.cabal#L122

to import this

https://github.com/GaloisInc/crucible/blob/0d90f74d11c5cb74672537700c3825b4d5d8b4db/crucible-syntax/crucible-syntax.cabal#L18

langston-barrett:
Fixes #1334. Will conflict with #1333, I will deconflict and rebase after that lands.

### DIFF
--- a/crucible-syntax/src/Lang/Crucible/Syntax/ExprParse.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/ExprParse.hs
@@ -34,8 +34,9 @@ import Control.Monad.IO.Class (MonadIO)
 import Control.Monad.Reader (MonadReader(..), ReaderT(..))
 import qualified Control.Monad.State.Strict as Strict
 
+import Data.Containers.ListUtils (nubOrd)
 import Data.Foldable as Foldable
-import Data.List
+import qualified Data.List as List
 import qualified Data.List.NonEmpty as NE
 import Data.List.NonEmpty (NonEmpty(..))
 import Data.String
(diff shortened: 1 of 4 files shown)


crucible-syntax/src/Lang/Crucible/Syntax/ExprParse.hs
```diff
@@ -34,8 +34,9 @@ import Control.Monad.IO.Class (MonadIO)
 import Control.Monad.Reader (MonadReader(..), ReaderT(..))
 import qualified Control.Monad.State.Strict as Strict
 
+import Data.Containers.ListUtils (nubOrd)
```
RyanGlScott, line 37: Since we're now using this, make sure to bump the lower version bounds on `containers` to `>= 0.6.0.1` in the `.cabal` file.

RyanGlScott (approved):

sauclovian-g (approved):

PR #1339, opened 2025-03-22
CI: copy the crux-mir rlibs tree with tar

### LINKED ISSUE #1338: Ubuntu crux-mir builds are broken
The Ubuntu crux-mir binary artefacts contain only a symlink in place of the rlibs tree:
```
tar -tzvf ...
  :
lrwxrwxrwx  0 runner docker        0 Mar 21 18:57 crux-mir-0.10-ubuntu-20.04-X64/rlibs -> /home/runner/work/crucible/crucible/crux-mir/rlibs_real/lib/rustlib/x86_64-unknown-linux-gnu/lib
```

This is because ci.sh is copying the rlibs tree with `cp -r`, and that's not portable in its symlink handling. Fix is to copy with tar instead.

sauclovian-g:
cp -r is producing inconsistent (and sometimes broken) results.

Fixes #1338.

### DIFF
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -158,7 +158,14 @@ bundle_crux_mir_files() {
   setup_dist
   extract_exe crux-mir dist/bin
   cp crux-mir/README.md dist/doc
-  cp -r crux-mir/rlibs dist
+  # rlibs is a symlink into rlibs_real; we can't just copy the symlink
+  # though because (currently at least) it's an absolute path that
+  # won't work in the output artefact.
+  #
+  # Use tar to copy (not cp -r, which isn't very safe) and descend
+  # into rlibs explicitly first as a cheap way to follow that symlink.
+  mkdir dist/rlibs
+  (cd crux-mir/rlibs && tar -cf - .) | (cd dist/rlibs && tar -xvf -)
   # It's less fragile to have users install mir-json themselves
   # (cd dependencies/mir-json && cargo install --locked --force --root ../../dist)
   VERSION=${VERSION:-$DATE}


sauclovian-g:
That seems to have done the trick, the crux-mir build artefacts are all about 60M and the new Ubuntu ones contain the same files as the previous MacOS ones.

RyanGlScott:
I accidentally ran into the same issue in https://github.com/GaloisInc/crucible/pull/1319/files#r1987985211, although I resolved it in a different way by just not copying them at all.

Alternatively, if we want to restore the previous behavior, we could implement a simpler solution: run `cp -Lr` instead of `cp -r` to follow symlinks. (This is what I plan to use in `mir-json`'s CI—see https://github.com/GaloisInc/mir-json/pull/85/commits/3f91f8c166a6cb88305c6f8d2643ee1429c0ccbb.) I think that would be much simpler than using `tar` as an intermediate step, as is done here.

kquick (approved):

sauclovian-g:
The standard/traditional way to copy a directory tree is tar-to-tar, which behaves reliably (unlike cp, which does not) so it should be this, not cp. cp options vary a lot and don't always behave consistently...

sauclovian-g:
(anyway for the moment we need a bindist that has the rlibs in it, even though #1319 is the way forward)

RyanGlScott:
> unlike cp, which does not

Can you elaborate more on this? Under what circumstances does `cp -L` not work reliably? I could certainly believe that it is not portable across operating systems, although I haven't seen as much in my CI testing in https://github.com/GaloisInc/mir-json/commit/3f91f8c166a6cb88305c6f8d2643ee1429c0ccbb.

RyanGlScott (approved):

sauclovian-g:
The MacOS man page suggests (but doesn't clearly state) that -L only works with -R. We could probably use -LR (rather than -Lr) since we don't need to care about running on legacy systems. But tar is really the standard way to do this...

PR #1344, opened 2025-03-25
README: Describe versioning scheme

### LINKED ISSUE #1337: README: Document the version numbering scheme
Crucible's packages use [the PVP](https://pvp.haskell.org/) for released versions. We should add this information to the README so that users know what to expect.

We also use a versioning numbering scheme that extends the PVP and is compliant with it, where we bump the version number by adding `.99` to the end immediately after each release. It would be nice but not strictly necessary to document this practice as well.

See also https://github.com/GaloisInc/parameterized-utils/issues/190, because I suspect that the scheme in use is the same.

langston-barrett:
Fixes #1337.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,10 +162,12 @@ thereof) in each.
 
 In addition to the release versions described above, we use *development
 versions* for each package. Immediately after release, we add a fifth component
-to the version number of each package. This component starts at `.99` and
-is incremented occasionally thereafter. These allow downstream packages to
-indicate in their Cabal files that they depend on unreleased functionality. Such
-five-component versions are never uploaded to Hackage.
+to the version number of each package. This component starts at `.99` and is
+incremented occasionally thereafter. These make it possible for downstream
+packages to indicate in their Cabal files that they depend on unreleased
+functionality, and for developers and users users to identify binaries compiled
+from development versions (via the `--version` flag). Such five-component
+versions are never uploaded to Hackage.
 
 Acknowledgements
 ----------------
--- a/README.md
+++ b/README.md
@@ -156,15 +156,15 @@ share a version number.
 
 [pvp]: https://pvp.haskell.org/.
 
-The Crux packages (`crux{,-llvm,-mir}`) all share the same version number. The
-major version of Crux is bumped on *every* release, regardless of the changes
-(or lack thereof) in these individual packages.
+`crux-llvm` and `crux-mir` share the same version number. The major version of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -147,6 +147,26 @@ unix
...
+
+The Crux packages (`crux{,-llvm,-mir}`) all share the same version number. The
+major version of Crux is bumped on *every* release, regardless of the changes
+(or lack thereof) in these individual packages.
+
+In addition to the release versions described above, we use *development
+versions* for each package. Immediately after release, we add a fifth component
+to the version number of each package. This component starts at `.99` and
+is incremented occasionally thereafter. These allow downstream packages to
+indicate in their Cabal files that depend on unreleased functionality. Such
```
RyanGlScott, line 167:
```suggestion
indicate in their Cabal files that they depend on unreleased functionality. Such
```

README.md
```diff
@@ -147,6 +147,26 @@ unix
...
+
+All of the library components in this repository conform to the Haskell [Package
+Versioning Policy][pvp]. Every executable has a corresponding library, and they
+share a version number.
+
+[pvp]: https://pvp.haskell.org/.
+
+The Crux packages (`crux{,-llvm,-mir}`) all share the same version number. The
+major version of Crux is bumped on *every* release, regardless of the changes
+(or lack thereof) in these individual packages.
```
RyanGlScott, line 161: It might be worth mentioning that an exception to this versioning rule is `crux`, which is a library-only package that is shared in common between all of the other `crux-*` packages.

RyanGlScott (approved):

README.md
```diff
@@ -147,6 +147,26 @@ unix
...
+
+`crux-llvm` and `crux-mir` share the same version number. The major version of
+these packages is bumped on *every* release, regardless of the changes (or lack
+thereof) in each.
+
+In addition to the release versions described above, we use *development
+versions* for each package. Immediately after release, we add a fifth component
+to the version number of each package. This component starts at `.99` and
+is incremented occasionally thereafter. These allow downstream packages to
+indicate in their Cabal files that they depend on unreleased functionality. Such
```
sauclovian-g, line 167: They also allow identifying when you're looking at a devel version. (To the extent that the version number is available from compiled versions, anyway.) This is probably the most important property of labeling devel versions differently...

sauclovian-g (approved):

kquick (approved):

README.md
langston-barrett: Noted, thanks!

PR #1359, opened 2025-03-28
Remove UC-Crux-LLVM

### LINKED ISSUE #1336: uc-crux-llvm: Deprecate and remove in favor of GREASE
From [the blog post introducing GREASE](https://www.galois.com/articles/introducing-grease):

> UC-Crux was our previous tool for under-constrained symbolic execution of LLVM. Unfortunately, we made two fundamental mistakes in the design of UC-Crux: (1) its core is specialized to analysis of LLVM, limiting its potential applicability to binaries, and (2) its analysis makes use of the types of LLVM pointers. LLVM has phased such types out in favor of the “opaque pointer” type ptr, and accordingly it would be difficult to adapt UC-Crux to work well on recent versions of LLVM. We plan to deprecate UC-Crux in favor of GREASE.

(See #1075 for the latter.)

We should deprecate/remove UC-Crux-LLVM.

langston-barrett:
It hasn't been under active development for a while, and GREASE does the same thing, but better. Let's stop building and maintaing this.

- https://www.galois.com/articles/introducing-grease
- https://galoisinc.github.io/grease/
- https://github.com/GaloisInc/grease

Fixes #1336.

After this is merged, I'll go ahead and close [the issues labeled `uc-crux`](https://github.com/GaloisInc/crucible/issues?q=is%3Aissue%20state%3Aopen%20label%3Auc-crux).


RyanGlScott (approved):
Make sure to also update this comment in `crux-llvm`:

https://github.com/GaloisInc/crucible/blob/b55a3bc6a397d435ff9218d0279b3c11f4ffeba6/crux-llvm/src/Crux/LLVM/Simulate.hs#L309

Otherwise, LGTM.

kquick (approved):
LGTM, modulo conflicts.

Issue #1360, opened 2025-03-28
ci: Build packages (and modules) in parallel


langston-barrett:
Consider the following snippet from the Crux-LLVM CI build:

https://github.com/GaloisInc/crucible/blob/2744ee5b174232d229b808410bcb9beb9f3e5d1f/.github/workflows/crux-llvm-build.yml#L124-L137

This is suboptimal for two reasons:

1. We are manually maintaining some kind of topological sort of the dependency DAG of our packages. Cabal already knows which things depend on which other things, we should use that knowledge.
2. We are not taking full advantage of build parallelism.

We could fix both of these by invoking `cabal build` just once, with all of our packages on the command line.

On (2): We may be building modules in parallel (I'm not sure if Cabal does this by default), but we're certainly not building *packages* in parallel, as we're not passing `--jobs` to Cabal, and [the default is `--jobs=1`](https://cabal.readthedocs.io/en/3.14/cabal-project-description-file.html#cfg-field-jobs). We could even use [`--semaphore`](https://cabal.readthedocs.io/en/3.14/cabal-project-description-file.html#cfg-flag---semaphore) so that Cabal and GHC coordinate their use of parallelism for maximally fast builds.

Of course, parallelism begets nondeterminism, but parallel compiler invocations are fairly standard practice and shouldn't be *too* problematic. The trade-off for build speed is likely worth it.

langston-barrett:
> I don't think this is quite right

Thank you for the explanation! I find that Cabal behavior surprising, so perhaps it would be good to either add a comment or to explicitly add `--jobs`.

> This sounds interesting, but with the caveat that we'd have to selectively enable this on GHC 9.8 or later.

Right, this would depend on the completion of #1308.

langston-barrett:
This was done in #1362.

PR #1362, opened 2025-03-28
ci: Make use of parallelism, and don't retry builds

### LINKED ISSUE #1361: ci: Don't retry builds
Our CI setup can retry `cabal build` up to six (!) times.

https://github.com/GaloisInc/crucible/blob/2744ee5b174232d229b808410bcb9beb9f3e5d1f/.github/ci.sh#L48-L54

The origins of this retry logic are in this commit from five years ago: https://github.com/GaloisInc/crucible/commit/d2da4133d2db9594f05e3dcce0cad3ffbf492908. The commit message does not indicate why retrying was considered necessary, nor does it provide further information about the mysterious OSX-related bug. I would go so far as to say it is likely not necessary, in that there are several projects that are downstream of Crucible that do not include such retry logic in their CI builds, and they appear to work fine.

Let's remove this `build` function entirely and just run `cabal build` inline.

langston-barrett:
Fixes #1361, #1360.

### DIFF
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -42,7 +42,6 @@ configure() {
   ghc_ver="$(ghc --numeric-version)"
   cp cabal.GHC-"$ghc_ver".config cabal.project.freeze
   cabal configure "$@" -j2 --enable-tests --minimize-conflict-set
-  #tee -a cabal.project > /dev/null < cabal.project.ci
 }
 
 test() {
--- a/.github/ci.sh
+++ b/.github/ci.sh
@@ -45,14 +45,6 @@ configure() {
   #tee -a cabal.project > /dev/null < cabal.project.ci
 }
 
-build() {
-  if ! retry cabal build "$@" && [[ "$RUNNER_OS" == "macOS" ]]; then
-    echo "Working around a dylib issue on macos by removing the cache and trying again"
-    cabal clean
-    retry cabal build "$@"
-  fi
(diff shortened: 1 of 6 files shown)


langston-barrett:
Ah, it appears that building with parallelism is more complex than I anticipated in a multi-package repo. I tried to split the build of dependencies (which should have their packages built in parallel but modules built in serial, which is the current behavior for *all* packages) from that of local packages (which should be built serially but with their modules compiled in parallel, which latter part would be new as of this PR), but that's hard when you have local packages depending on one another. For instance, `cabal build --only-dependencies exe:crux-go` builds `crucible-go` (which is local). Thus, I will instead content myself with building local *packages* in parallel until we can use `--semaphore`.

langston-barrett:
This does [appear to compile local packages in parallel](https://github.com/GaloisInc/crucible/actions/runs/14135254127/job/39605345942):
```
[20 of 20] Compiling Data.LLVM.Internal
[38 of 58] Compiling Lang.Crucible.LLVM.MemModel 
```
Does that actually make things faster? It appears so, that build took 4:53, whereas [this recent one on `master](https://github.com/GaloisInc/crucible/actions/runs/14133060378/job/39598050502) took 5:33. That could just be a fluke, but there's no way that this is *slower*, so I think it's safe to merge.

RyanGlScott (approved):

PR #1363, opened 2025-03-28
READMEs: Add logos

langston-barrett:


RyanGlScott:
Should we add the Crux logo to the `crux-mir` README as well?

RyanGlScott (approved):

PR #1365, opened 2025-04-04
Dockerfiles: Support both amd64 and arm64 builds

RyanGlScott:
Ensure that the `what4-solvers` and `ghcup` binaries correspond to the appropriate architecture when downloading them.

Addresses one half of #1364.


RyanGlScott:
Manually triggered CI jobs that demonstrate that the Docker builds still work on x86-64:

* `crux-llvm`: https://github.com/GaloisInc/crucible/actions/runs/14264564006/job/39983353955
* `crux-mir`: https://github.com/GaloisInc/crucible/actions/runs/14264566341/job/39983361650

(I verified that the Docker images build with ARM64 locally.)

PR #1371, opened 2025-04-10
ci: Don't install LLVM in workflows that don't need it

### LINKED ISSUE #1370: ci: Installing LLVM in builds where it is not needed
Most of our workflows use a function called `install_system_deps` in `ci.sh`:

https://github.com/GaloisInc/crucible/blob/a6d03962bfe8449797afb577d0813c9214f0fd48/.github/workflows/crucible-go-build.yml#L71

However, that function calls `install_llvm`:

https://github.com/GaloisInc/crucible/blob/a6d03962bfe8449797afb577d0813c9214f0fd48/.github/ci.sh#L103

This is likely unnecessary for `crucible-{go,jvm,wasm}`, which should probably call `install_solvers` directly.

langston-barrett:
Fixes #1370. Not quite sure if crux-mir will end up needing it, but removed until proven necessary.

### DIFF
--- a/.github/workflows/crucible-go-build.yml
+++ b/.github/workflows/crucible-go-build.yml
@@ -68,7 +68,7 @@ jobs:
             ${{ env.CACHE_VERSION }}-cabal-${{ matrix.os }}-${{ matrix.ghc }}-${{ hashFiles(format('cabal.GHC-{0}.config', matrix.ghc)) }}-
 
       - shell: bash
-        run: .github/ci.sh install_system_deps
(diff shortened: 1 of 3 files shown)


langston-barrett:
Huh, removing the LLVM installation made the `crux-mir` tests fail with
```
        test/coverage/coverage_try.crux.log: withFile: user error (Could not find: z3)
```
which is odd, because we're still installing the solvers... Might just revert for now anyway.

RyanGlScott (approved):

PR #1372, opened 2025-04-10
ci: Use `haskell-ci` for `crucible-{go,jvm,wasm}` build

langston-barrett:
Replace these largely copy/pasted workflows with a reference to a unified "reusable" workflow that does all the same things but generally slightly better.

See https://github.com/GaloisInc/.github?tab=readme-ov-file#haskell-ci.

Depends on https://github.com/GaloisInc/.github/pull/30. Progress on #1373, though `crux-{llvm,mir}` still share a cache.

### DIFF
--- a/.github/workflows/crucible-go-build.yml
+++ b/.github/workflows/crucible-go-build.yml
@@ -5,24 +5,11 @@ on:
   pull_request:
   workflow_dispatch:
 
-# The CACHE_VERSION can be updated to force the use of a new cache if
-# the current cache contents become corrupted/invalid.  This can
-# sometimes happen when (for example) the OS version is changed but
-# older .so files are cached, which can have various effects
-# (e.g. cabal complains it can't find a valid version of the "happy"
-# tool).
-#
-# This also periodically happens on MacOS builds due to a tar bug
-# (symptom: "No suitable image found ... unknown file type, first
-# eight bytes: 0x00 0x00 0x00 0x00 0x00 0x00 0x00 0x00")
 env:
-  CACHE_VERSION: 1
+  CI_TEST_LEVEL: "1"
 
 jobs:
   build:
-    runs-on: ${{ matrix.os }}
-    env:
-      CI_TEST_LEVEL: "1"
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


.github/workflows/crucible-go-build.yml
```diff
@@ -37,117 +24,35 @@ jobs:
...
+      check: false
+      ghc: ${{ matrix.ghc }}
+      os: ${{ matrix.os }}
+      pre-hook: |
+        cp "cabal.GHC-${{ matrix.ghc }}.config" cabal.project.freeze
+
+        env \
+          SOLVER_PKG_VERSION="snapshot-20250326" \
+          BUILD_TARGET_OS="${{ matrix.os }}" \
+          BUILD_TARGET_ARCH="${{ matrix.os == macos-14 && 'ARM64' || 'X64' }}" \
```
langston-barrett, line 51: Note for reviewers: Unfortunately, the `runner` context is not available here. I think ideally, `what4-solvers` could host a script that could detect the current OS and architecture, and download and install the correct snapshot into a configurable directory. But IMO this local workaround is OK for right now.

langston-barrett:
Fun: GitHub Actions doesn't like the output of `pwd` being used as a `working-directory` on later commands.
```
Error: An error occurred trying to start process 'C:\Program Files\Git\bin\bash.EXE' with working directory '/d/a/crucible/crucible'. The directory name is invalid.
```
I'll need to figure out how to make haskell-ci properly Windows-compatible in this regard.

.github/ci.sh
```diff
@@ -85,9 +85,16 @@ install_llvm() {
...
+  snapshot=20250326
+  curl \
+    --fail \
+    --location \
+    --proto '=https' \
+    --show-error \
+    --silent \
+    --tlsv1.2 \
+    "https://github.com/GaloisInc/what4-solvers/blob/d8d6849dd63a4655e1ff07a824c7ec4f234c91a6/scripts/install.sh" | \
+    bash -s -- "${snapshot}" "${BIN}"
```
langston-barrett, line 97: Currently just testing https://github.com/GaloisInc/what4-solvers/pull/66. Will need to be updated to take `SOLVER_PKG_VERSION` before merge. Callsites will need to be modified to stop specifying OS/Arch.

.github/workflows/crucible-go-build.yml
langston-barrett: `$RUNNER_ARCH` is available, so using that.

PR #1375, opened 2025-04-11
crux-llvm: Remove `build-stack.sh` script

langston-barrett:
We don't explicitly support building with Stack. This also refered to
non-existent files. Remove to avoid confusion.


langston-barrett:
CI is failing due to https://github.com/GaloisInc/.github/issues/31.

Come to think of it, not exactly sure why this isn't affecting the Linux builds - those should be getting a cache hit off the `restore-key` and the cache from `master`...

langston-barrett:
Nice, the problem appears to be fixed with `haskell-ci@v1.4.1`.

RyanGlScott (approved):

PR #1380, opened 2025-04-16
ci: Simplify Cabal/GHC compatibility check

### LINKED ISSUE #1368: ci: Simplify GHC/Cabal compatibility check
Like was done in https://github.com/GaloisInc/parameterized-utils/pull/191

langston-barrett:
Fixes #1368. Also, set default shell to `bash` while we're in the neighborhood.

### DIFF
--- a/.github/workflows/crucible-go-build.yml
+++ b/.github/workflows/crucible-go-build.yml
@@ -24,7 +24,7 @@ jobs:
             cabal: 3.10.3.0
             ghc: 9.8.2
     name: crucible-go - GHC v${{ matrix.ghc }} - ${{ matrix.os }}
-    uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1.4
+    uses: GaloisInc/.github/.github/workflows/haskell-ci.yml@v1.4.2
     with:
       build-targets: pkg:crucible-go
       cabal: ${{ matrix.cabal }}
--- a/.github/workflows/crucible-go-build.yml
+++ b/.github/workflows/crucible-go-build.yml
@@ -24,7 +24,7 @@ jobs:
             cabal: 3.10.3.0
             ghc: 9.8.2
     name: crucible-go - GHC v${{ matrix.ghc }} - ${{ matrix.os }}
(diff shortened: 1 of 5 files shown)


langston-barrett:
Huh, the `crucible-{go,jvm,wasm}` builds seem to be hitting https://github.com/GaloisInc/.github/issues/31. This is odd, because that was fixed and the fix [was released as v1.4.1](https://github.com/GaloisInc/.github/commit/45097ccf006a9fe62f3df44138b38287a4c461f1). It appears that, in contrast to pre-packaged actions like `actions/checkout@v4`, GHA does *not* fetch the most recent compatible release/tag of reusable workflows, but just *precisely* the tag that was specified. Should be easily fixable by pinning `haskell-ci@v1.4.2`, but it's a bit annoying to have to manually bump the version numbers for bug fixes. Dependabot could help here.

RyanGlScott:
If I'm understanding https://github.com/actions/toolkit/blob/main/docs/action-versioning.md#compatibility correctly, the way that GitHub Actions' versioning scheme works is that it will only pick up minor releases if you target `@vN`, where `N` is the primary version number. If you specify a secondary version number or beyond, it will assume that you are targeting a specific release (and indeed, there is a specific `GaloisInc/.github@v1.4` release that is older than `v1.4.2`).

We may want to target `GaloisInc/.github@v1` instead.

langston-barrett:
Oh no, it appears I was mistaken about how version pinning works in general on GHA. It appears that the maintainers of actions such as `actions/checkout` just manually move tags like `v4` to point at the lastest point-release 😱. We should likely do the same in the `haskell-ci` repo.

langston-barrett:
@RyanGlScott Further on in those docs: https://github.com/actions/toolkit/blob/main/docs/action-versioning.md#recommendations

Apparently, the content after the `@` is not parsed by GHA at all - use of more generic numbers is just supported by maintainers moving those tags around. So we could pin `1.4` if we also moved the `1.4` tag around to match the latest patch-release.

But yeah, we should probably pin `@v1` here.

RyanGlScott (approved):

PR #1383, opened 2025-04-17
ci: Use `haskell-ci`'s' `pre-test-hook` when installing solvers

langston-barrett:
This installs solvers a bit later in the CI pipeline. This way, if the build fails, we skip installing the solvers and save some time.


.github/workflows/crucible-jvm-build.yml
```diff
@@ -44,12 +44,12 @@ jobs:
...
       pre-hook: |
         cp "cabal.GHC-${{ matrix.ghc }}.config" cabal.project.freeze
-
+      pre-test-hook: |
         env \
           SOLVER_PKG_VERSION="snapshot-20250326" \
           BUILD_TARGET_OS="${{ matrix.os }}" \
           BUILD_TARGET_ARCH="${RUNNER_ARCH}" \
-          .github/ci.sh \
+          "${GITHUB_WORKSPACE}/.github/ci.sh" \
```
RyanGlScott, line 52: What goes wrong if you don't specify `${GITHUB_WORKSPACE}` here? I ask since the `pre-hook` runs `cp "cabal.GHC-${{ matrix.ghc }}.config" cabal.project.freeze` instead of `cp "${GITHUB_WORKSPACE}/cabal.GHC-${{ matrix.ghc }}.config" "${GITHUB_WORKSPACE}/cabal.project.freeze"`, so I'm wondering if there is a potential pitfall here that `pre-hook` might be susceptible to.

RyanGlScott (approved):

.github/workflows/crucible-jvm-build.yml
langston-barrett: The `pre-test-hook` is run inside the `sdist` if such was made, so a path relative to the repo root doesn't work here. `pre-hook` runs before any other steps, so a repo-root-relative path is fine there.

RyanGlScott: Tricky! I wonder if the documentation for `pre-test-hook` should clarify that it may be run in a different directory from `GITHUB_WORKSPACE`, as I could imagine people accidentally shooting themselves with this footgun if they're not aware of this difference.

PR #1389, opened 2025-04-24
`crucible-mir`: Always store source positions in `Statement`s/`Terminator`s

### LINKED ISSUE #1388: `crucible-mir`: Preserve more source positions in `Statement`s/`Terminator`s
All `mir-json` [`Statement`s](https://doc.rust-lang.org/1.69.0/nightly-rustc/rustc_middle/mir/struct.Statement.html) and [`Terminator`s](https://doc.rust-lang.org/1.69.0/nightly-rustc/rustc_middle/mir/terminator/struct.Terminator.html) include source positions. Despite this, only a subset of `crucible-mir` `Statement`s/`Terminator`s preserve source positions (indicated by the fields below with a "`pos"` suffix):

https://github.com/GaloisInc/crucible/blob/4c3b83133bdbf3620a20d6e5a47d5e302c7d3e1b/crucible-mir/src/Mir/Mir.hs#L293-L301

https://github.com/GaloisInc/crucible/blob/4c3b83133bdbf3620a20d6e5a47d5e302c7d3e1b/crucible-mir/src/Mir/Mir.hs#L364-L400

This can sometimes lead to `crucible-mir` error messages not having source positions at all, which `crucible` prints as `internal`. For example:

https://github.com/GaloisInc/crucible/blob/4c3b83133bdbf3620a20d6e5a47d5e302c7d3e1b/crux-mir/test/symb_eval/crux/early_fail.good#L6-L15

Note how the second counterexample has a source position, while the first one only has `internal`. This makes it unclear where the first counterexample arises from.

To rectify this, we should store source positions for _all_ `Statement`s and `Terminator`s and ensure that we always call `setCurrentPosition` on the source positions when translating them. We can take inspiration from MIR and represent them like:

```hs
data Statement = Statement { _spos :: Text, _skind :: StatementKind }
data Terminator = Terminator { _tpos :: Text, _tkind :: TerminatorKind }
```

Where `StatementKind` and `TerminatorKind` are what we currently call `Statement` and `Terminator` inside of `crucible-mir`.

RyanGlScott:
Fixes #1388.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -1,6 +1,15 @@
-# next
+# next -- TBA
 
-Nothing yet.
+This release supports [version
+1](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#1) of
+`mir-json`'s schema.
+
+* Split out the `StatementKind` and `TerminatorKind` data types from `Statement`
+  and `Terminator`, respectively. `{Statement,Terminator}` now contain a pair
+  of `{Statement,Terminator}Kind` and its source position.
+* Improve source position tracking for `Statement`s and `Terminator`s during
+  the translation to Crucible. This should result in more precise error messages
+  in certain situations.
 
 # 0.4 -- 2025-03-21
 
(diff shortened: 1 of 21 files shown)


spernsteiner (approved):

sauclovian-g (approved):
Seems reasonable, and definitely desirable to track positions correctly...

sauclovian-g (approved):
Seems reasonable, and definitely desirable to track positions correctly...

PR #1391, opened 2025-04-30
`crucible-mir`: erase pointee type from `MirReferenceType` ("opaque pointers")

### LINKED ISSUE #1385: `crucible-mir`: erase pointee type from `MirReferenceType` ("opaque pointers")
Currently, `MirReferenceType` has a type index indicating the Crucible type of the location it points to.  For example, the Rust type `*const i32` is represented by `MirReferenceType (BVType 32)`.  The pointee type is required to match the type of the pointee location, so a reference whose `MirReferencePath` points to a location of type `BVType 32` must have type `MirReferenceType (BVType 32)`.

This design means we cannot support Rust casts that change the pointee type, such as `*const i32 -> *const ()`, since this would need to produce a value of type `MirReferenceType (StructType [])` (`*const ()`) that actually points to a `BVType 32` location.  This is becoming a problem because the Rust standard library has begun using casts to `*mut ()`, `*const u8`, and similar in more places.  For example, see [uses of `ptr::from_raw_parts`](https://github.com/rust-lang/rust/blob/d6c1e454aa8af5e7e59fbf5c4e7d3128d2f99582/library/core/src/ptr/const_ptr.rs#L125) or [the new `CloneToUninit` trait](https://github.com/rust-lang/rust/blob/d6c1e454aa8af5e7e59fbf5c4e7d3128d2f99582/library/core/src/clone.rs#L437) (which is already being used in `Box::clone` and `Rc::make_mut`).

The general pattern here is that a pointer to a location of type `T` is cast to a generic / erased / `void*`-equivalent type like `*mut ()`, and is cast back to `*mut T` before use.  We could support this under our existing memory model by removing the requirement that every `*mut ()` value points to a location of type `StructType []`.

The idea of opaque pointers is to remove the type index, so `*mut u8` and `*mut ()` and all other pointers are simply `MirReferenceType`.  The actual pointee type would be erased from the static type, but would still be tracked at run time (pointer values might need to gain a `Some TypeRepr` field, or perhaps it can be derived from the `MirReferencePath`) and checked when accessing the pointer.  Then we could support code that casts e.g. `*mut i32 -> *mut ()`, as long as it casts back to `*mut i32` before dereferencing.  This would expand our unsafe support to cover the patterns now appearing in the standard library.

spernsteiner:
Currently, `MirReferenceType` has a type index indicating the Crucible type of the location it points to.  For example, the Rust type `*const i32` is represented by `MirReferenceType (BVType 32)`.  The pointee type is required to match the type of the pointee location, so a reference whose `MirReferencePath` points to a location of type `BVType 32` must have type `MirReferenceType (BVType 32)`.  This prevents us from implementing certain casts, such as `*const i32 -> *const ()`: `MirReferenceType (BVType 32)` and `MirReferenceType UnitType` require pointees of type `BVType 32` and `UnitType` respectively, and thus it's impossible to have two such references that point to the same location.

This branch removes the type index from `MirReferenceType`, so all Rust pointers and references are simply `MirReferenceType` with no further (static) distinction.  This allows crucible-mir to support a variety of raw-pointer casts that are currently unsupported, such as casting `*mut T -> *mut () -> *mut T` to temporarily type-erase a pointer.  Casts like this appear in several places in the standard library, notably in the implementation of `RawVec`.

With this change, a few operations, notably `writeMirRef`, now require a `TypeRepr`, since the Crucible type of the pointee can no longer be inferred from the index of the `MirReferenceType`.  In most cases, the `TypeRepr` was already available at the call site for other reasons.  In a few places, only the `Mir.Ty` was available, so I had to add a new call to `tyToRepr` to get the `TypeRepr`; calling `tyToRepr` on a pointee type will fail if the pointee is dynamically sized (a slice or `dyn Trait`), but I think all the places where I added it are only reachable for sized pointees (`&[_]` and `&dyn Trait` take different code paths from sized `&T` in most of the translation).

Fixes #1385

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -1030,20 +1030,20 @@ instance OrdFC MirStmt where
 
 instance TypeApp MirStmt where
   appType = \case
-    MirNewRef tp    -> MirReferenceRepr
+    MirNewRef _    -> MirReferenceRepr
     MirIntegerToRef _ -> MirReferenceRepr
-    MirGlobalRef gv -> MirReferenceRepr
-    MirConstRef tp _ -> MirReferenceRepr
+    MirGlobalRef _ -> MirReferenceRepr
+    MirConstRef _ _ -> MirReferenceRepr
     MirReadRef tp _ -> tp
     MirWriteRef _ _ _ -> UnitRepr
     MirDropRef _    -> UnitRepr
-    MirSubanyRef tp _ -> MirReferenceRepr
-    MirSubfieldRef ctx _ idx -> MirReferenceRepr
-    MirSubvariantRef _ ctx _ idx -> MirReferenceRepr
-    MirSubindexRef tp _ _ -> MirReferenceRepr
-    MirSubjustRef tp _ -> MirReferenceRepr
-    MirRef_VectorAsMirVector tp _ -> MirReferenceRepr
-    MirRef_ArrayAsMirVector btp _ -> MirReferenceRepr
+    MirSubanyRef _ _ -> MirReferenceRepr
+    MirSubfieldRef _ _ _ -> MirReferenceRepr
+    MirSubvariantRef _ _ _ _ -> MirReferenceRepr
+    MirSubindexRef _ _ _ -> MirReferenceRepr
+    MirSubjustRef _ _ -> MirReferenceRepr
+    MirRef_VectorAsMirVector _ _ -> MirReferenceRepr
+    MirRef_ArrayAsMirVector _ _ -> MirReferenceRepr
     MirRef_Eq _ _ -> BoolRepr
     MirRef_Offset _ _ -> MirReferenceRepr
     MirRef_OffsetWrap _ _ -> MirReferenceRepr
@@ -1182,6 +1182,12 @@ dropRefRoot _bak _gs (Const_RefRoot tpr _) =
     leafAbort $ GenericSimError $
         "Cannot drop Const_RefRoot (of type " ++ show tpr ++ ")"
 
+-- | Helper for defining a `MuxLeafT` operation that works only for
+-- `MirReference`s with a specific pointee type `tp`.  If the `MirReference`
+-- argument is a valid reference (not `MirReference_Integer`) with pointee type
+-- `tp`, this calls `k` on the reference's parts; otherwise, this fails.
+-- `desc` is a human-readable description of the operation, which is used in
+-- the `leafAbort` error message.
 typedLeafOp ::
     Monad m =>
     String ->
@@ -2070,7 +2076,7 @@ subindexMirRefIO bak iTypes tpr ref x =
 mirRef_offsetSim :: IsSymInterface sym =>
     TypeRepr tp -> MirReferenceMux sym -> RegValue sym IsizeType ->
     OverrideSim m sym MIR rtp args ret (MirReferenceMux sym)
-mirRef_offsetSim tpr ref off =
+mirRef_offsetSim _tpr ref off =
     ovrWithBackend $ \bak ->
       modifyRefMuxSim (\ref' -> mirRef_offsetWrapLeaf bak ref' off) ref
 
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -347,26 +347,26 @@ isizeToUsize _wrap = id
 
 
 --------------------------------------------------------------
--- * A MirReference is a Crucible RefCell paired with a path to a sub-component
+-- * A MirReference is a Crucible RefCell (or other root) paired with a path to a sub-component
 --
--- We use this to represent mutable data
+-- We use this to represent pointers and references.
 
 type MirReferenceSymbol = "MirReference"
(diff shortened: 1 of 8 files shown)


crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1032,25 +1030,25 @@ instance OrdFC MirStmt where
 
 instance TypeApp MirStmt where
   appType = \case
-    MirNewRef tp    -> MirReferenceRepr tp
-    MirIntegerToRef tp _ -> MirReferenceRepr tp
-    MirGlobalRef gv -> MirReferenceRepr (globalType gv)
-    MirConstRef tp _ -> MirReferenceRepr tp
+    MirNewRef tp    -> MirReferenceRepr
```
RyanGlScott, line 1033: I believe this `tp` variable is no longer used, so I'd prefer to replace it with a wildcard pattern `_`. Similarly for the other unused variables below in this definition.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1184,195 +1182,200 @@ dropRefRoot _bak _gs (Const_RefRoot tpr _) =
     leafAbort $ GenericSimError $
         "Cannot drop Const_RefRoot (of type " ++ show tpr ++ ")"
 
+typedLeafOp ::
```
RyanGlScott, line 1185: Could you leave a Haddock comment on what this function does? It's not entirely obvious at a glance.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1112,17 +1118,23 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
+    doRef (M.TySlice ty') | MirSliceRepr <- tpr = doSlice ty' ref
+    doRef M.TyStr | MirSliceRepr <- tpr = doSlice (M.TyUint M.B8) ref
     doRef (M.TyDynamic _) | DynRefRepr <- tpr = do
         dynRef <- readMirRef tpr ref
         return $ MirPlaceDynRef dynRef
-    doRef _ | MirReferenceRepr tpr' <- tpr = do
+    doRef ty' | MirReferenceRepr <- tpr = do
+        -- This use of `tyToRepr` is okay because `TyDynamic` and other unsized
+        -- cases are handled above.
+        Some tpr' <- tyToReprM ty'
```
RyanGlScott, line 1129:
```suggestion
        -- This use of `tyToReprM` is okay because `TyDynamic` and other unsized
        -- cases are handled above.
        Some tpr' <- tyToReprM ty'
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1112,17 +1118,23 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
         MirPlace tpr' <$> readMirRef tpr ref <*> pure NoMeta
     doRef _ = mirFail $ "deref: bad repr for " ++ show ty ++ ": " ++ show tpr
 
-    doSlice tpr' ref' = do
-        slice <- readMirRef (MirSliceRepr tpr') ref'
+    doSlice ty' ref' = do
+        -- This use of `tyToRepr` is okay because we know the element type of a
+        -- slice is always sized.
+        Some tpr' <- tyToReprM ty'
```
RyanGlScott, line 1136:
```suggestion
    doSlice ty' ref' = do
        -- This use of `tyToReprM` is okay because we know the element type of a
        -- slice is always sized.
        Some tpr' <- tyToReprM ty'
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -122,9 +122,6 @@ pattern CTyBvSize128 <- M.TyAdt _ $(M.explodedDefIdPat ["crucible", "bitvector",
 pattern CTyBvSize256 <- M.TyAdt _ $(M.explodedDefIdPat ["crucible", "bitvector", "_256"]) (M.Substs [])
 pattern CTyBvSize512 <- M.TyAdt _ $(M.explodedDefIdPat ["crucible", "bitvector", "_512"]) (M.Substs [])
 pattern CTyBv t <- M.TyAdt _ $(M.explodedDefIdPat ["crucible", "bitvector", "Bv"]) (M.Substs [t])
-pattern CTyBv128 <- CTyBv CTyBvSize128
-pattern CTyBv256 <- CTyBv CTyBvSize256
-pattern CTyBv512 <- CTyBv CTyBvSize512
```
RyanGlScott, line 127: Worth noting the removal of these pattern synonyms in the `crucible-mir` changelog. (I doubt anyone is using them, but better safe than sorry.)

RyanGlScott (approved):
This looks great! It would be worth adding an entry to the `crucible-mir` changelog that summarizes these changes. (Although all the API changes are too numerous to enumerate exhaustively, we could perhaps mention the fact that `MirReference` and `MirReferenceMux` now lack a `tp` type parameter, which is the thing that ended up causing SAW the most grief in practice.)

RyanGlScott:
Also: please mention "fixes #1385" somewhere in the commit message.

spernsteiner:
I've pushed fixes for all of the comments, and added the issue number to the PR description.

PR #1395, opened 2025-05-08
crucible-mir: `ctpop` intrinsic returns `u32` unconditionally

samcowger:
This matches Rust's type for this function.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -38,6 +38,7 @@ module Mir.Trans(transCollection,transStatics,RustModule(..)
                 , callExp
                 , derefExp, readPlace, addrOfPlace
                 , transmuteExp
+                , extendUnsignedBV
                 ) where
 
 import Control.Monad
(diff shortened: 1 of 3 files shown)


spernsteiner:
Can you add a small test case under `crux-mir/test/conc_eval/num/`?  See `crux-mir/test/conc_eval/cell/slice.rs` for an example of what `conc_eval` tests should look like.  These tests get run under crux-mir and also under rustc to check that the test function returns the same results, which helps catch arithmetic bugs and such in addition to translation errors.

samcowger:
> Can you add a small test case under `crux-mir/test/conc_eval/num/`? See `crux-mir/test/conc_eval/cell/slice.rs` for an example of what `conc_eval` tests should look like. These tests get run under crux-mir and also under rustc to check that the test function returns the same results, which helps catch arithmetic bugs and such in addition to translation errors.

Done. I also used `crux-mir/test/conc_eval/enum/mixed_discrs.rs` as an example of how to refer to intrinsics in test cases, which I hope is appropriate.

PR #1396, opened 2025-05-12
`crux-mir`: Support `nightly-2025-02-16`

RyanGlScott:
This is a fairly large overhaul of `crux-mir`'s and `crucible-mir`'s internals to support a more recent Rust nightly (`nightly-2025-02-16`). The work is split between myself, @samcowger, @qsctr, and @spernsteiner. This piggybacks off of the corresponding `mir-json` changes in https://github.com/GaloisInc/mir-json/pull/98, a fix for https://github.com/GaloisInc/mir-json/issues/78.

There is a lot happening here, so if you want to learn specifics, I'd encourage you to look at individual commits. Some of the broad themes in the patchset include:

* Support auto-generated shim functions.
* Handle lang item `DefId`s of the form `$lang::Option` (rather than `core::option::Option` as they were before). This allows us to refer to lang item `DefId`s directly from `crucible-mir` without needing to know their disambiguator values, which would otherwise vary from compilation to compilation. (For instance, we use this to wire in the `OrderingEnum` type into `crucible-mir`.)
* Adapt to `CheckedBinaryOp` being merged into `CheckedOp`.
* Handle closure-to-function-pointer casts.
* Add a custom override for the [`bswap`](https://doc.rust-lang.org/std/intrinsics/fn.bswap.html) function.


RyanGlScott:
While running this toolchain on the test case from https://github.com/GaloisInc/mir-json/issues/94, I discovered that `crucible-mir` doesn't currently support the [`Subtyping`](https://doc.rust-lang.org/1.86.0/nightly-rustc/rustc_middle/mir/type.PlaceElem.html#variant.Subtype) `PlaceElem`. Here is a minimized reproducer, based on [this `rustc` test case](https://github.com/rust-lang/rust/blob/3148e6a9933b17b28ed6c7b8d8bd6c8e49fe4a50/tests/mir-opt/mir_subtyping.rs):

```rs
struct Foo<T: 'static>(T);

fn useful<'a>(x: &'a u8) -> &'a u8 {
    x
}

struct Wrapper(for<'b> fn(&'b u8) -> &'b u8);

static X: u8 = 42;

#[crux::test]
pub fn f() -> u8 {
    let wrapper = Wrapper(useful);
    let hr_fnptr = Box::new(Foo::<for<'a> fn(&'a u8) -> &'a u8>(wrapper.0));
    let lr_fnptr = hr_fnptr as Box<Foo<fn(&'static u8) -> &'static u8>>; // This line requires a Subtype
    *((*lr_fnptr).0)(&X)
}
```
```
user error (JSON Decoding of test.all.mir failed: Error in $.fns[1].body.blocks[1].block.data[0].rhs.usevar.data.data[0]: bad PlaceElem: Just (String "Subtype"))
```

Based on the documentation for `Subtype`, it has no impact on runtime behavior, so simulating it should be straightforward.

PR #1397, opened 2025-05-16
`crucible-mir`: More robust handling for `&dyn Trait`

### LINKED ISSUE #1092: `crucible-mir`: Robustly handle `&dyn` references
MIR JSON code occasionally produces code of the form `&*x`, where `x: &dyn Trait`. This is an awkward case for `crucible-mir`, which doesn't easily permit representing `*x` as a `MirPlace (lvalue). Specifically, `MirPlace` contains a Crucible `Expr` of `MirReferenceType`, but a `&dyn Trait` does not directly contain a `MirReferenceType`. Instead, it contains an `AnyType` that wraps a `MirReferenceType t` for some unknown `t`.

Two possible ways to add support for `&*x` code are:

1. The proper fix: modify the representation of `&dyn Trait` to use `MirReferenceType AnyType` instead of an `AnyType` wrapping `MirReference t`.

   This requires extending `MirReferenceType` to support a new `erase` operation that turns a `MirReferenceType t` into a `MirReferenceType AnyType` and a new `unerase` operation that does the reverse (failing if the new `t` doesn't match the original).  I think `erase` can be implemented by adding a new `Erase_RefPath` variant to `MirReferencePath`, and `unerase` can be implemented by popping the last `Erase_RefPath` from the path (if the types match).  I don't think we would need an `Unerase_RefPath`, since the only legal operation on an erased reference is to unerase it.

   This would also require updating the code that generates vtable shims and calls.  These shims are responsible for unwrapping the `AnyType` and passing the resulting reference to the actual trait method.

2. The quick hack: add a second variant to `MirPlace` that takes an `AnyType` instead of a `MirReferenceType`.  This would essentially be a special case specifically for the `&*x` pattern.  We would need to add cases for the new variant in several places, but nearly all of them would simply call `mirFail` (there are very few operations on unsized lvalues that are legal in MIR).

Option (1) is likely more work, but brings `crucible-mir`'s behavior more in line with `rustc`'s and helps future-proof against e.g. `self: Rc<Self>` methods in trait objects. For the time being, I am going to go with option (2), but we should revisit this.

(Much of what I have written above is paraphrased from @spernsteiner.)

RyanGlScott:
This PR:

* Translates the data field of trait objects to a `MirReference` instead of an `Any` value.
* Removes the `MirPlaceDynRef` hack.

Fixes #1092.

### DIFF
--- a/crucible-mir/src/Mir/Generator.hs
+++ b/crucible-mir/src/Mir/Generator.hs
@@ -138,21 +138,6 @@ data MirExp s where
 -- result of lvalue evaluation.
 data MirPlace s where
     MirPlace :: C.TypeRepr ty -> R.Expr MIR s MirReferenceType -> PtrMetadata s -> MirPlace s
-    -- | This is a hack used to support @&dyn Trait@ references. Unlike other
-    -- reference types, trait objects do not use 'MirReferenceType', instead
-    -- using a custom 'AnyType' representation that wraps a
-    -- @'MirReferenceType' t@ (for some unknown @t@). The only 'MirPlaceDynRef'
-    -- operation that is currently supported is @addrOfPlace@, which supports
-    -- code that looks like @&*x@ (where @x: &dyn Trait@). This sort of
-    -- operation does arise in @rustc@-compiled MIR, even in the standard
-    -- libraries, so we must have /some/ level of support for it.
-    --
-    -- All other 'MirPlaceDynRef' operations (e.g., @readPlace@ and
-    -- @evalPlaceProj@) are unsupported and will throw an exception if
-    -- encountered. To implement these operations, we will need to change the
-    -- encoding of trait objects to use 'MirReferenceType' in a more standard
-    -- way. See <https://github.com/GaloisInc/crucible/issues/1092>.
-    MirPlaceDynRef :: R.Expr MIR s DynRefType -> MirPlace s
 
 -- | MIR supports a notion of "unsized places" - for example, it generates code
 -- like `(*s)[i]` where `s` is a slice.  To handle this, we attach the metadata
@@ -164,7 +149,8 @@ data MirPlace s where
 -- we may need to add `PtrMetadata` to `MirExp`s at some point as well.
 data PtrMetadata s =
       NoMeta
-    | SliceMeta (R.Expr MIR s UsizeType)
+    | SliceMeta (R.Expr MIR s UsizeType) -- ^ The slice length
+    | DynMeta (R.Expr MIR s Core.AnyType) -- ^ The trait object's vtable
   deriving Show
 
 ---------------------------------------------------------------------------------
@@ -400,7 +386,6 @@ instance Show (MirExp s) where
 
 instance Show (MirPlace s) where
     show (MirPlace tr e m) = show e ++ ", " ++ show m ++ ": & " ++ show tr
-    show (MirPlaceDynRef e) = "dyn reference " ++ show e
 
 instance Show MirHandle where
     show (MirHandle _nm sig c) =
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -411,6 +411,7 @@ derefExp pointeeTy (MirExp MirSliceRepr e) = do
     let ptr = getSlicePtr e
     let len = getSliceLen e
     return $ MirPlace tpr ptr (SliceMeta len)
+-- TODO RGS: Do we need a case for trait objects?
 derefExp pointeeTy (MirExp tpr _) = mirFail $ "don't know how to deref " ++ show tpr
 
 readPlace :: HasCallStack => MirPlace s -> MirGenerator h s ret (MirExp s)
@@ -418,16 +419,15 @@ readPlace (MirPlace tpr r NoMeta) = MirExp tpr <$> readMirRef tpr r
 readPlace (MirPlace tpr _ meta) =
     mirFail $ "don't know how to read from place with metadata " ++ show meta
         ++ " (type " ++ show tpr ++ ")"
-readPlace MirPlaceDynRef{} =
-    -- See https://github.com/GaloisInc/crucible/issues/1092
-    mirFail "readPlace not supported for dyn references"
 
 addrOfPlace :: HasCallStack => MirPlace s -> MirGenerator h s ret (MirExp s)
 addrOfPlace (MirPlace tpr r NoMeta) = return $ MirExp MirReferenceRepr r
 addrOfPlace (MirPlace tpr r (SliceMeta len)) =
     return $ MirExp MirSliceRepr $ mkSlice r len
-addrOfPlace (MirPlaceDynRef dynRef) =
-    return $ MirExp DynRefRepr dynRef
+addrOfPlace (MirPlace _tpr r (DynMeta vtable)) =
+    return $ MirExp DynRefRepr $
+      R.App $ E.MkStruct DynRefCtx $
+      Ctx.Empty Ctx.:> r Ctx.:> vtable
 
 
 -- Given two bitvectors, extend the length of the shorter one so that they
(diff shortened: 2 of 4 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -411,23 +411,23 @@ derefExp pointeeTy (MirExp MirSliceRepr e) = do
     let ptr = getSlicePtr e
     let len = getSliceLen e
     return $ MirPlace tpr ptr (SliceMeta len)
+-- TODO RGS: Do we need a case for trait objects?
```
RyanGlScott, line 414: I'm not sure if we need a `derefExp` case for trait object–like things or not.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1122,7 +1119,10 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
     doRef M.TyStr | MirSliceRepr <- tpr = doSlice (M.TyUint M.B8) ref
     doRef (M.TyDynamic _) | DynRefRepr <- tpr = do
         dynRef <- readMirRef tpr ref
-        return $ MirPlaceDynRef dynRef
+        let dynRefData = S.getStruct dynRefDataIndex dynRef
+        let dynRefVtable = S.getStruct dynRefVtableIndex dynRef
+        -- TODO RGS: Is AnyRepr the correct type here?
+        return $ MirPlace C.AnyRepr dynRefData (DynMeta dynRefVtable)
```
RyanGlScott, line 1125: I'm not quite sure what type to put here. I waffled between `MirReferenceRepr` (the type of the trait object's data) and `AnyRepr` (the type of the trait object's vtable), and I've arbitrarily picked one here for no particular reason. I'm welcome advice on how to pick the "right" type here.

crucible-mir/src/Mir/Trans.hs
spernsteiner: Might be nice to have in case we ever encounter `y = &*x` where `x: &dyn Trait`.  IIRC this sort of code is the main reason why `derefExp` needed a slice case.

crucible-mir/src/Mir/Trans.hs
spernsteiner: Some background: `MirExp tpr expr` is the Crucible translation of a MIR rvalue: `tpr` is the value's `TypeRepr`, and `expr` is a Crucible expression of that type.  For example, if `x: i32`, then the translation of the rvalue `x` will be `MirExp (BVRepr 32) <some expr>`. `MirPlace tpr ref meta` is the Crucible translation of a MIR place (lvalue): `tpr` is the `TypeRepr` of the value stored at that place, `ref` is a `MirReferenceType` that points to the place, and `meta` is the pointer metadata.  The translation of the place `x` will be `MirPlace (BVRepr 32) <address of x> NoMeta`.

For any sized type `T`, the `MirPlace` will have the form `MirPlace T <some expr of type *T> NoMeta`.  For unsized types, the story is more complex: the expr is just the data pointer, and the slice length / dyn vtable is stored in the metadata instead.  A place of type `[U]` would look like `MirPlace U <expr : *U> (SliceMeta len)` - *not* `MirPlace [U] <expr : *[U]> meta`.  For `dyn Trait`, the story is a little more complicated - there isn't really any good choice of `TypeRepr` for the `tpr` field of the `MirPlace`.  But I don't think anything should ever use that `TypeRepr`, since there's no way to directly access the data of a `dyn Trait` (e.g. you can't access `(*x).field` if `x: &dyn Trait`).  So it should be okay to fill it in with a dummy value like `AnyType` or maybe `UnitType`.

It would also probably be good to document the `evalPlaceProj M.Deref` case a little better, specifically to clarify the double indirection: the input is `MirPlace (*T) <expr : **T> NoMeta`, and the output (for sized types) is `MirPlace T <expr : *T> NoMeta`.

crucible-mir/src/Mir/Trans.hs
RyanGlScott: One thing that wasn't obvious to me until I read this code more closely: unless I'm missing something, `derefExp` is _not_ used in the implementation of `&*x`. Rather, the `*x` part is powered by the `Deref` case of the `evalPlaceProj` function, and the `&` part is powered by the `addrOfPlace` function, both of which now have cases for trait objects.

`derefExp` is only used in one very specific spot: in the custom override for the [`discriminant_value`](https://doc.rust-lang.org/std/intrinsics/fn.discriminant_value.html) intrinsic here:

https://github.com/GaloisInc/crucible/blob/fb06f32e70de358c4affc5f12e54b0cbb248b5ce/crucible-mir/src/Mir/TransCustom.hs#L925-L939

As such, I'm not sure if it's even possible to call `derefExp` with a trait object. Perhaps we don't need to worry about this after all? (Then again, `derefExp` has a case for slices, which _also_ appears impossible to reach from `discriminant_value`'s override...)

crucible-mir/src/Mir/Trans.hs
RyanGlScott: I think I came to the same conclusion about the choice of type being a dummy value, but it's helpful to have you confirm my hunch. In that case, I'll remove this TODO and document the conventions here more carefully.

crucible-mir/src/Mir/Trans.hs
spernsteiner: In that case I think it would be reasonable to delete the slice case and note that `derefExp` is for sized things only.  Then there's less duplication between `derefExp` and the deref projection case.

crucible-mir/src/Mir/Generator.hs
```diff
@@ -164,7 +149,8 @@ data MirPlace s where
 -- we may need to add `PtrMetadata` to `MirExp`s at some point as well.
 data PtrMetadata s =
       NoMeta
-    | SliceMeta (R.Expr MIR s UsizeType)
+    | SliceMeta (R.Expr MIR s UsizeType) -- ^ The slice length
+    | DynMeta (R.Expr MIR s Core.AnyType) -- ^ The trait object's vtable
```
spernsteiner, line 153:
```suggestion
    | DynMeta (R.Expr MIR s C.AnyType) -- ^ The trait object's vtable
```
For consistency with other mentions of Crucible types throughout the module

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1139,6 +1160,28 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
+    -- Types like `dyn Trait` are unsized, so we handle them differently. Given
+    -- a MirPlace input of the form:
+    --
+    --   MirPlace (*dyn Trait) <expr: **dyn Trait> NoMeta
+    --
+    -- We produce output of the form:
+    --
+    --   MirPlace AnyType <data: *?> (DynMeta vtable)
+    --
+    -- Where `data` is the trait object's data points (a reference to some
```
spernsteiner, line 1172:
```suggestion
    -- Where `data` is the trait object's data pointer (a reference to some
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2202,7 +2239,7 @@ transVirtCall colState intrName methName dynTraitName methIndex
                 (C.FunctionHandleRepr (C.AnyRepr <: argTys) retTy)
 
         -- Call it
-        G.tailCall vtsFH (recvData <: args)
+        G.tailCall vtsFH (R.App (E.PackAny MirReferenceRepr recvData) <: args)
```
spernsteiner, line 2242: At some point we should change vtable methods to take `MirReferenceType` instead of `AnyType` for the receiver argument, so this wrapping won't be needed.  But the vtable code is pretty gnarly, so that's probably best left to a separate PR.

spernsteiner (approved):

crucible-mir/src/Mir/Trans.hs
RyanGlScott: I've opened https://github.com/GaloisInc/crucible/issues/1747 to track this task.

Issue #1308, opened 2025-02-12
Support GHC 9.12


langston-barrett:
https://www.haskell.org/ghc/download_ghc_9_12_1.html

- [x] Fix https://github.com/GaloisInc/llvm-pretty/issues/156
- [x] Fix https://github.com/GaloisInc/what4/issues/286
- [ ] Allow the [new version of `base`](https://wiki.haskell.org/Base_package#Versions) in the Cabal `build-depends`, build and test
- [ ] Fix any new warnings from [`-Wdefault`](https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Wdefault)
- [ ] Add a new [`cabal freeze`](https://cabal.readthedocs.io/en/stable/cabal-commands.html#cabal-freeze) file
- [ ] Add the new GHC version in CI configuration
- [ ] Update the GHC version used in any Dockerfiles
- [ ] Add the new GHC version to the Cabal [`tested-with` field](https://cabal.readthedocs.io/en/stable/cabal-package-description-file.html#pkg-field-tested-with), if present
- [ ] Update any mentions of GHC versions in the documentation (e.g., README), if present
- [ ] Optionally repeat the above steps, but removing / deprecating an older GHC version

RyanGlScott:
Done in https://github.com/GaloisInc/crucible/pull/1828.

Issue #1061, opened 2023-02-15
`crucible-llvm`: Support loading mostly symbolic strings


RyanGlScott:
Currently, the LLVM memory model in `crucible-llvm` comes with the restriction that all strings must be concrete. Here, when I say "string", I really mean "C string", i.e., an array of bytes that is followed by a null terminator. This restriction impacts several parts of the `crucible-llvm` codebase, including our `printf` and `strlen` overrides and the various parts of `crucible-symio` that interface with strings.

This issue tracks possible ways to this lift this restriction. The restriction manifests itself in this part of `loadString`:

https://github.com/GaloisInc/crucible/blob/994b4f944630c561b9390f2a5f8d53d40f7318e5/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1219-L1230

Here, the only place where we actually make use of the string being concrete is in the null terminator check. However, this is not an all-or-nothing proposition: it is possible for a string to contain some symbolic characters but still end with a concrete null terminator character. To help distinguish between these types of strings, I will refer to strings that entirely symbolic characters (including the null terminator) as _fully symbolic_ strings, while I will refer to strings that some symbolic characters followed by a concrete null terminator as _mostly symbolic_ strings.

Handling fully symbolic strings is definitely not possible for Crucible. That would be tantamount to asking Crucible to handle strings of symbolic length, which is something that Crucible is fundamentally unequipped to handle. Handling mostly symbolic strings, however, is definitely possible, as that would allow Crucible to say with certainty where strings start and end. It doesn't matter so much whether the characters preceding the null terminator are symbolic.

One proposal for handling mostly symbolic strings is to remove the `Nothing` case in `loadString.go` above, and generalize the `Just c` case to handle all characters that are not concretely equal to the null terminator. (This would require removing the use of `bvLit`.) I've successfully employed this technique on a downstream project, and it should work nicely with Crux's [`crucible_string`](https://github.com/GaloisInc/crucible/blob/0a21fc4222795acbefa035c4c7ea28ac07bd17d0/crux-llvm/c-src/includes/crucible.h#L24-L31) function, which allocates mostly symbolic strings.

[...]

PR #1405, opened 2025-05-22
llvm-syntax: Overrides for manipulating C-style null-terminated strings

### LINKED ISSUE #1393: `crucible-llvm-syntax`: Add utilities for C-style strings
A common pattern in C programs is to read and write strings to pointers. Accomplishing this in C code is straightforward enough, but doing the same thing in S-expression programs proves tedious due to the fact that `crucible-syntax`'s `String` type is represented very differently from C's strings, which are usually represented by a type like `char*` (i.e., a sequence of byte-sized characters). It is not obvious how one would read a `String` from a `Ptr`, for instance.

We should take inspiration from [`ambient-verifier`](https://github.com/GaloisInc/ambient-verifier/blob/eab04abb9750825a25ec0cbe0379add63f05f6c6/README.rst#built-in-overrides) and add some built-in functions to `crucible-llvm-syntax` for bridging the gap between `crucible-syntax` `String`s and C-style strings. The point in the design space that `ambient-verifier` chose was to add the following built-ins:

* `read-c-string :: Ptr n -> String Unicode` reads a null-terminated, UTF-8–encoded, concrete string from the `Pointer` and converts it to a `String`.
* `write-c-string :: Ptr n -> String Unicode -> Unit` writes a UTF-8–encoded, concrete string to a ``Pointer``, including a null terminator.

For situations where you have non-UTF-8–encoded data, `ambient-verifier` also offers the following built-ins:

* `read-bytes :: Ptr n -> Vector (Bitvector 8)` reads a null-terminated, sequence of bytes from a `Ptr`. The null terminator at the end of the sequence of bytes will be concrete, but the preceding bytes may be symbolic. Unlike `read-c-string`, this function reads the raw bytes without converting to a particular text encoding.
* `write-bytes :: Vector (Bitvector 8) -> Ptr n -> Unit` writes a sequence of bytes to a `Ptr`, including a null terminator (which does not need to be in the `Vector`). The null terminator written at the end will be concrete, but the preceding bytes may be symbolic. Unlike `write-c-string`, this function writes the raw bytes without converting to a particular text encoding.

langston-barrett:
Fixes #1393, see that issue for motivation.

Along the way:

- Create a new Crucible-LLVM module for housing string-manipulation functions. Currently just adds one function beyond what was already available, but would provide a convenient home for future functionality around strings, e.g., `strcpy`/`strcmp`-like functions. Re-export existing string-manipulating code there, with the idea of moving it out of the mega-module `MemModel` in the future.
- Simplify the `crucible-cli` API by merging the "hook" that resolves forward declarations into the "setup hook". For LLVM, this means that we can bind forward declarations to overrides that use the memory model.
- New helpers for `TypedOverride`
- Finally, port the `ambient-verifier` S-expression string-manipulation overrides and wire them through `crucible-llvm-{cli,syntax}`.

### DIFF
--- a/crucible-llvm-syntax/CHANGELOG.md
+++ b/crucible-llvm-syntax/CHANGELOG.md
@@ -1,6 +1,7 @@
 # next
 
-Nothing yet.
+* Add overrides for string-manipulation `{read,write}-bytes` and
+  `{read,write}-c-string`.
 
 # 0.1 -- 2024-02-05
 
--- a/crucible-llvm-syntax/CHANGELOG.md
+++ b/crucible-llvm-syntax/CHANGELOG.md
@@ -1,6 +1,6 @@
 # next
 
-* Add overrides for string-manipulation `{read,write}-bytes` and
+* Add overrides for string-manipulation: `{read,write}-bytes` and
   `{read,write}-c-string`.
 
 # 0.1 -- 2024-02-05
--- a/crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax/Overrides/String.hs
+++ b/crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax/Overrides/String.hs
@@ -1,0 +1,199 @@
+{-# LANGUAGE DataKinds #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ImplicitParams #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeOperators #-}
+
+-- | Overrides related to C strings
+module Lang.Crucible.LLVM.Syntax.Overrides.String
+  ( readBytesOverride
+  , readCStringOverride
+  , writeBytesOverride
+  , writeCStringOverride
+  ) where
+
+import Control.Monad.IO.Class (liftIO)
+import Data.BitVector.Sized qualified as BV
+import Data.ByteString qualified as BS
+import Data.Parameterized.Context qualified as Ctx
+import Data.Parameterized.NatRepr qualified as DPN
+import Data.Text qualified as DT
+import Data.Text.Encoding qualified as DTE
+import Data.Text.Encoding.Error qualified as DTEE
+import Data.Vector qualified as Vec
+import Lang.Crucible.Backend qualified as LCB
+import Lang.Crucible.LLVM.MemModel qualified as LCLM
+import Lang.Crucible.LLVM.MemModel.Strings qualified as LCLMS
+import Lang.Crucible.Simulator qualified as C
+import Lang.Crucible.Simulator qualified as LCS
+import Lang.Crucible.Types qualified as LCT
+import What4.Interface qualified as WI
+
+-- | Override for @read-bytes@.
+--
+-- The loaded string must be concrete. If a symbolic character is encountered
+-- while loading, this function will generate an assertion failure.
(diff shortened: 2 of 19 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -83,9 +83,12 @@ module Lang.Crucible.LLVM.MemModel
...
   , doArrayConstStoreUnbounded
-  , loadString
-  , loadMaybeString
-  , strLen
+  , {-# DEPRECATED "Exported from Crucible.LLVM.MemModel.Strings instead" #-}
+    loadString
+  , {-# DEPRECATED "Exported from Crucible.LLVM.MemModel.Strings instead" #-}
+    loadMaybeString
+  , {-# DEPRECATED "Exported from Crucible.LLVM.MemModel.Strings instead" #-}
+    strLen
```
langston-barrett, line 91: NTS: Deprecating exports is only available in GHC 9.8, these deprecations will have to wait for #1308.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -0,0 +1,52 @@
+-- TODO(TODO: file an issue, put number here): Move the definitions of the
```
langston-barrett, line 1: Reviewers: I will file this issue and update this TODO if you like the idea of moving the string-related functions into this new module.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -2,12 +2,51 @@
...
+import qualified Lang.Crucible.LLVM.DataLayout as CLD
+import qualified Lang.Crucible.LLVM.MemModel as LCLM
 import qualified Lang.Crucible.LLVM.MemModel as Mem
+import qualified What4.Interface as WI
 
+-- | Store a string to memory, adding a null terminator at the end.
+storeString ::
+  forall sym bak w.
+  ( LCB.IsSymBackend sym bak
+  , WI.IsExpr (WI.SymExpr sym)
```
RyanGlScott, line 32: `IsExpr (SymExpr sym)` is a superclass of `IsSymBackend sym bak`, so the former constraint is redundant.

crucible-llvm-syntax/CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # next
 
-Nothing yet.
+* Add overrides for string-manipulation `{read,write}-bytes` and
```
RyanGlScott, line 3: Perhaps add a colon or other form of punctuation to make this sentence flow a little more nicely, e.g.,

> * Add overrides for string-manipulation: `{read,write}-bytes` and `{read,write}-c-string`.

crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax/Overrides/String.hs
```diff
@@ -0,0 +1,199 @@
...
+import Lang.Crucible.Simulator qualified as C
+import Lang.Crucible.Simulator qualified as LCS
+import Lang.Crucible.Types qualified as LCT
+import What4.Interface qualified as WI
+
+-- | Override for @read-bytes@.
+--
+-- The loaded string must be concrete. If a symbolic character is encountered
+-- while loading, this function will generate an assertion failure.
+readBytesOverride ::
```
RyanGlScott, line 40: It might be worth adding a TODO here (and near `readBytes`) citing #1061. Although `readBytesOverride` and `readBytes` are currently limited to loading concrete strings, I think this restriction would be lifted if #1061 were resolved.

RyanGlScott (approved):

PR #1407, opened 2025-05-23
Add `TraversableF` instances for `CrucibleEvent` and `CrucibleAssumption{,s}`

RyanGlScott:
I found these instances useful when debugging `ProofGoal`s. The module already defines most of what is needed to traverse over these data structures anyway, so implementing these instances proves straightforward.

### DIFF
--- a/crucible/src/Lang/Crucible/Backend/Assumptions.hs
+++ b/crucible/src/Lang/Crucible/Backend/Assumptions.hs
@@ -48,6 +48,7 @@ import           Control.Lens (Traversal, folded)
 import           Data.Kind (Type)
 import           Data.Functor.Identity
 import           Data.Functor.Const
+import qualified Data.Parameterized.TraversableF as TF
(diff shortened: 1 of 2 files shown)


crucible/src/Lang/Crucible/Backend/Assumptions.hs
```diff
@@ -48,6 +48,7 @@ import           Control.Lens (Traversal, folded)
 import           Data.Kind (Type)
 import           Data.Functor.Identity
 import           Data.Functor.Const
+import           Data.Parameterized.TraversableF
```
langston-barrett, line 51: No need to block on this change, but generally I'd rather see qualified / explicit imports for all the reasons listed here: https://stackoverflow.com/questions/2386714/why-is-import-bad
```suggestion
import qualified Data.Parameterized.TraversableF as TF
```

langston-barrett (approved):

PR #1415, opened 2025-06-05
`crucible-mir`: Remove `AnyType` from struct/enum representations

### LINKED ISSUE #1414: crucible-mir: remove AnyType from struct/enum representation
Currently, a type such as `struct S { a: i32, b: u8 }` is represented as `AnyType`, with the inner type of the `Any` being the actual struct type `StructType [ BVType 32, BVType 8 ]`.  Enum types are similarly wrapped in `AnyType`.  This is a historical artifact that we'd like to remove to simplify the representation and [enable custom DST support](https://github.com/GaloisInc/crucible/issues/1399#issuecomment-2942082237).  Specifically, we'd like for `tyToRepr` to produce `StructType` for struct ADTs and `RustEnumType` for enum ADTs.

There were two reasons for the use of `AnyType`, both of which are now obsolete.  First, `tyToRepr` originally didn't take the MIR `Collection` as an argument, so it had no way of looking up the fields or variants of an ADT.  The `col` argument was added to `tyToRepr` in e5c768c3c7f78f28ccedf454b8d5ef34577f9d79.  Second, when `MirReferenceType` indexed by a pointee type, the use of `AnyType` broke an infinite recursion for recursive types like `struct Node { val: i32, next: *const Node }`.  Previously, this would have produced `tyToRepr Node = StructType [ BVType 32, MirReferenceType (tyToRepr Node) ]`.  The type index was removed from `MirReferenceType` in #1391.  With this change, `tyToRepr Node = StructType [ BVType 32, MirReferenceType ]`, which is non-recursive.  The same should be true of all recursive types accepted by rustc, since Rust allows recursion only behind a pointer (otherwise `size_of::<T>()` would be infinite).

The general approach here involves (1) updating `tyToRepr` to return `StructType`/`RustEnumType` instead of `AnyType` for `TyAdt` and (2) fixing field/variant accesses and ADT construction.  This could probably be done separately for structs first and then for enums (enums are more complex).  Once this is done, it should be possible to remove `Any_RefPath` and the associated `MirSubanyRef` / `Generator.subanyRef` / etc.

There are some uses of `AnyType` that are not related to ADTs and should not be removed: (1) in the representation and operations of `CTyAny` / `crucible::any::Any`, which exposes `AnyType` to Rust; (2) as a placeholder type for DSTs, where the `MirPlace` needs a `TypeRepr` but we don't have a more specific one available; and (3) in `DynRefType` / vtables / vtable methods.

RyanGlScott:
We now translate struct or enum types directly to `StructType` or `RustEnumType`, respectively, instead of using `AnyType`. As a result of this change, we no longer need `Any_RefPath`, `MirSubanyRef`, or `subanyRef`, so these have been removed.

Fixes #1414.

### DIFF
--- a/crucible-mir/src/Mir/TransTy.hs
+++ b/crucible-mir/src/Mir/TransTy.hs
@@ -56,7 +56,7 @@ import qualified Debug.Trace as Debug
 
 import           Mir.Generator
     ( MirExp(..), MirPlace(..), PtrMetadata(..), MirGenerator, mirFail
-    , subanyRef, subfieldRef, subvariantRef, subjustRef
+    , subfieldRef, subvariantRef, subjustRef
     , mirVector_fromVector
     , cs, collection, discrMap, findAdt, mirVector_uninit, arrayZeroed )
 import           Mir.Intrinsics
@@ -78,9 +78,8 @@ import           Mir.Intrinsics
 -- Arrays and slices are both crucible vectors; no difference between them.
 -- Tuples are crucible structs.
 
--- Non-custom ADTs are encoded as Any.  The underlying type is either a Struct
--- or a Variant of Structs, depending on whether the Rust type is a struct or
--- enum.
+-- Non-custom ADTs are encoded either as a Struct or a variant of Structs,
+-- depending on whether the Rust type is a struct or enum.
 --
 -- Custom type translation is on the bottom of this file.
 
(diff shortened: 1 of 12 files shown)


crucible-mir/src/Mir/TransTy.hs
```diff
@@ -78,9 +78,8 @@ import           Mir.Intrinsics
 -- Arrays and slices are both crucible vectors; no difference between them.
 -- Tuples are crucible structs.
 
--- Non-custom ADTs are encoded as Any.  The underlying type is either a Struct
--- or a Variant of Structs, depending on whether the Rust type is a struct or
--- enum.
+-- Non-custom ADTs are encoded either as Any a Struct or a variant of Structs,
```
spernsteiner:
```suggestion
-- Non-custom ADTs are encoded either as a Struct or a variant of Structs,
```

PR #1420, opened 2025-06-10
Publish multi-platform docker image

### LINKED ISSUE #1419: Build multi-platform docker images
The current [CI workflow](https://github.com/GaloisInc/crucible/blob/master/.github/workflows/crux-mir-build.yml) builds binaries for Linux and Mac, but the published docker image is `amd64` only, so it is really slow when used on modern Macs.

Build and push multi platform docker images (x64/arm64), initially at least for `crux-mir`.

An example how this can be achieved can be found [here](https://github.com/rems-project/cn/blob/main/.github/workflows/docker.yml#L25), we built multiplatform images using appropriate github actions.

podhrmic:
Fixes #1419

### DIFF
--- a/.github/workflows/crux-mir-build.yml
+++ b/.github/workflows/crux-mir-build.yml
@@ -214,7 +214,7 @@ jobs:
           path: crux-mir-*.tar.gz*
           name: crux-mir-${{ matrix.os }}-${{ matrix.ghc }}
 
-  build-push-image:
+  build-image:
     runs-on: ubuntu-24.04
     needs: [config]
     if: (github.event_name == 'schedule' || github.event_name == 'workflow_dispatch' || needs.config.outputs.release == 'true') && github.repository_owner == 'GaloisInc'
@@ -225,6 +225,9 @@ jobs:
           - file: .github/Dockerfile-crux-mir
             image: ghcr.io/galoisinc/crux-mir
             cache: ghcr.io/galoisinc/cache-crux-mir
+        platform:
+          - linux/amd64
+          - linux/arm64
     steps:
       - uses: actions/checkout@v4
         with:
@@ -241,6 +244,11 @@ jobs:
           sudo rm -rf /opt/ghc
           sudo rm -rf "/usr/local/share/boost"
           sudo rm -rf "$AGENT_TOOLSDIRECTORY" # Python installations
+
+      - name: Prepare
+        run: |
+          platform=${{ matrix.platform }}
+          echo "PLATFORM_PAIR=${platform//\//-}" >> $GITHUB_ENV
 
       - uses: rlespinasse/github-slug-action@v3.x
 
@@ -263,13 +271,16 @@ jobs:
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
 
-      - uses: docker/build-push-action@v2
+      - uses: docker/build-push-action@v6
+        id: build
         with:
-          context: .
-          tags: ${{ matrix.image }}:${{ steps.common-tag.outputs.common-tag }}
+          platforms: ${{ matrix.platform }}
           labels: ${{ steps.labels.outputs.labels }}
-          load: true
-          push: false
+          tags: ${{ matrix.image }}:${{ steps.common-tag.outputs.common-tag }}
+          outputs: type=image,push-by-digest=true,name-canonical=true,push=true
+          #context: .
+          #load: true
+          #push: false
           file: ${{ matrix.file }}
           build-args: ${{ matrix.build-args }}
           cache-from: |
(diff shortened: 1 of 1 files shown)


podhrmic:
@RyanGlScott not sure what is the best way to test the updated pipeline - looks like I would need to temporarily change when those jobs are run. Open to other ideas!

.github/workflows/crux-mir-build.yml
```diff
@@ -214,7 +214,7 @@ jobs:
           path: crux-mir-*.tar.gz*
           name: crux-mir-${{ matrix.os }}-${{ matrix.ghc }}
 
-  build-push-image:
+  build-image:
     runs-on: ubuntu-24.04
```
RyanGlScott: Surely we'd want to run this on both `ubuntu-24.04` and `ubuntu-24.04-arm`? (My impression is that building the `arm64` image on x86-64 would be needlessly slow.)

.github/workflows/crux-mir-build.yml
```diff
@@ -225,6 +225,9 @@ jobs:
           - file: .github/Dockerfile-crux-mir
             image: ghcr.io/galoisinc/crux-mir
             cache: ghcr.io/galoisinc/cache-crux-mir
+        platform:
+          - linux/amd64
+          - linux/arm64
```
RyanGlScott: If we ran this job on both `ubuntu-24.04` and `ubuntu-24.04-arm`, then I don't think we'd need to bother with configuring the platform explicitly.

.github/workflows/crux-mir-build.yml
```diff
@@ -285,16 +296,75 @@ jobs:
...
+      #     docker tag ${{ matrix.image }}:$COMMON_TAG ${{ matrix.image }}:nightly
+      #     docker push ${{ matrix.image }}:nightly
+
+      # - if: needs.config.outputs.release == 'true'
+      #   name: ${{ matrix.image }}:${{ needs.config.outputs.crux-mir-version }}
+      #   run: |
+      #     docker tag ${{ matrix.image }}:$COMMON_TAG ${{ matrix.image }}:${{ needs.config.outputs.crux-mir-version }}
+      #     docker push ${{ matrix.image }}:${{ needs.config.outputs.crux-mir-version }}
+      #     docker tag ${{ matrix.image }}:$COMMON_TAG ${{ matrix.image }}:latest
+      #     docker push ${{ matrix.image }}:latest
```
RyanGlScott: Delete this?

.github/workflows/crux-mir-build.yml
```diff
@@ -263,13 +271,16 @@ jobs:
...
-          tags: ${{ matrix.image }}:${{ steps.common-tag.outputs.common-tag }}
+          platforms: ${{ matrix.platform }}
           labels: ${{ steps.labels.outputs.labels }}
-          load: true
-          push: false
+          tags: ${{ matrix.image }}:${{ steps.common-tag.outputs.common-tag }}
+          outputs: type=image,push-by-digest=true,name-canonical=true,push=true
+          #context: .
+          #load: true
+          #push: false
```
RyanGlScott: Delete this?

.github/workflows/crux-mir-build.yml
```diff
@@ -285,16 +296,75 @@ jobs:
...
+      - name: Export digest
+        run: |
+          mkdir -p ${{ runner.temp }}/digests
+          digest="${{ steps.build.outputs.digest }}"
+          touch "${{ runner.temp }}/digests/${digest#sha256:}"
+
+      - name: Upload digest
+        uses: actions/upload-artifact@v4
+        with:
+          name: digests-${{ env.PLATFORM_PAIR }}
```
RyanGlScott: I think it would suffice to use `digests-${{ runner.arch }}` here instead—no need for any `PLATFORM_PAIR` business that way.

.github/workflows/crux-mir-build.yml
```diff
@@ -285,16 +296,75 @@ jobs:
...
+
+      # - if: needs.config.outputs.release == 'true'
+      #   name: ${{ matrix.image }}:${{ needs.config.outputs.crux-mir-version }}
+      #   run: |
+      #     docker tag ${{ matrix.image }}:$COMMON_TAG ${{ matrix.image }}:${{ needs.config.outputs.crux-mir-version }}
+      #     docker push ${{ matrix.image }}:${{ needs.config.outputs.crux-mir-version }}
+      #     docker tag ${{ matrix.image }}:$COMMON_TAG ${{ matrix.image }}:latest
+      #     docker push ${{ matrix.image }}:latest
+
+  # TODO: configure the image tag for release/event-schedule
```
RyanGlScott: I think copying this line from the `build-image` job should suffice:

https://github.com/GaloisInc/crucible/blob/3f9656edeaf9ef8e0ff443af5d3423ad0916e705/.github/workflows/crux-mir-build.yml#L220

.github/workflows/crux-mir-build.yml
```diff
@@ -285,16 +296,75 @@ jobs:
...
+  merge-push-image:
+    runs-on: ubuntu-latest
+    needs:
+      - build-image
+    env:
+      Image: ghcr.io/galoisinc/crux-mir
```
RyanGlScott: Don't do this. Use `${{ matrix.image }}` instead, as the workflow already goes through the effort of defining it above.

.github/workflows/crux-mir-build.yml
```diff
@@ -285,16 +296,75 @@ jobs:
...
+      - name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v5
+        with:
+          images: $Image
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
```
RyanGlScott: I'm not sure if these tags are configured correctly. (I think these might just be copy-pasted from [here](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners), not sure.)

At a minimum, we will need the following:

* A `type` for the `nightly` tag. I forget what the exact `docker/metadata-action` syntax is, but I think it's something like `type=raw,value=nightly,enable=...`, where the `...` should be something that is predicated on `needs.config.outputs.event-schedule == 'true'`, similarly to what the current CI YAML file does:

  https://github.com/GaloisInc/crucible/blob/17507cc6a231a0881b52074ec03fee901da9671b/.github/workflows/crux-mir-build.yml#L288-L292
* A `type` for the `crux-mir-version` tag that is predicated on `needs.config.outputs.release == 'true'`, similar to this:

  https://github.com/GaloisInc/crucible/blob/17507cc6a231a0881b52074ec03fee901da9671b/.github/workflows/crux-mir-build.yml#L294-L300
* A `type` for the `latest` tag that is also predicated on `needs.config.outputs.release == 'true'`, similar to this:

  https://github.com/GaloisInc/crucible/blob/17507cc6a231a0881b52074ec03fee901da9671b/.github/workflows/crux-mir-build.yml#L294-L300

.github/workflows/crux-mir-build.yml
```diff
@@ -285,16 +280,60 @@ jobs:
...
+
+      - name: Create manifest list and push
+        working-directory: ${{ runner.temp }}/digests
         run: |
-          docker tag ${{ matrix.image }}:$COMMON_TAG ${{ matrix.image }}:nightly
-          docker push ${{ matrix.image }}:nightly
+          docker buildx imagetools create $(jq -cr '.tags | map("-t " + .) | join(" ")' <<< "$DOCKER_METADATA_OUTPUT_JSON") \
+            $(printf '$Image@sha256:%s ' *)
```
langston-barrett: This looks a little scary. Perhaps add a commenty saying what it's doing and why? Maybe even with example inputs and outputs?

RyanGlScott: To be honest, I was going to hold off on making suggestions for this part of the CI configuration until the first batch of reviews had been resolved. :)

As scary as this code is (and I agree that it could be helpful to comment it), it actually isn't scary _enough_, as it's missing a step. In order to ensure that the image renders properly in the GitHub Container Registry's container page, you have to copy over manifest-related annotations. This CI workflow doesn't do this just yet, but `grease`'s does ([here](https://github.com/GaloisInc/grease/blob/7cc9dc3e3d7e3c595a38b8ec3259a40e7d066730/.github/workflows/docker.yml#L120)). Getting this to work requires tweaking with the `DOCKER_METADATA_ANNOTATIONS_LEVELS` levels in the `docker/metadata-action` steps before it: see [here](https://github.com/GaloisInc/grease/blob/7cc9dc3e3d7e3c595a38b8ec3259a40e7d066730/.github/workflows/docker.yml#L41) and [here](https://github.com/GaloisInc/grease/blob/7cc9dc3e3d7e3c595a38b8ec3259a40e7d066730/.github/workflows/docker.yml#L112) in `grease`'s CI.

Sadly, this part isn't documented in the [standard GitHub Actions template](https://docs.docker.com/build/ci/github-actions/multi-platform/#distribute-build-across-multiple-runners) provided in Docker's documentation. I had to learn this the hard way by figuring out what more experienced people than I do in their multi-platform builds, such as in [this example](https://github.com/zammad/zammad/blob/773f912a3b77690a959fb21f1562510e26447ea5/.github/workflows/docker-release.yaml).

podhrmic:
I decided to follow the `grease` CI more closely, as it already works nicely in the [saw-suite CI](https://github.com/GaloisInc/saw-suite).

RyanGlScott:
This is coming along pretty nicely! That being said, it will be difficult to test if this actually works because you are running this PR from a `crucible` fork, which lacks permissions to actually write to `crucible`'s GitHub container registry. It might be worth trying this from a branch in the main `crucible` repo instead.

podhrmic:
Good point - closing in favor of #1637

PR #1424, opened 2025-06-12
Improve documentation for `Crucible.Backend`

langston-barrett:


crucible/src/Lang/Crucible/Backend/Online.hs
```diff
@@ -4,19 +4,26 @@
...
 --
--- The online backend also allows override definitions access to a
--- persistent SMT solver connection.  This can be useful for some
--- kinds of algorithms that benefit from quickly performing many
--- small solver queries in a tight interaction loop.
+-- The primary intended use-case is to prune unsatisfiable execution
+-- traces during simulation using the execution feature provided by
+-- "Lang.Crucible.Simulator.PathSatisfiability". That execution feature is
+-- parameterized over a function called @considerSatisfiability@, which can be
+-- provided using this module's 'considerSatisfiability'.
```
RyanGlScott: I found it difficult to parse this sentence, partly because it wasn't obvious to me that "a function called `considerSatisfiability`" is distinct from "this module's `considerSatisfiability`". I think it might read a bit more naturally if the sentence were phrased like:

> That execution feature is parameterized by a predicate argument, which can be instantiated using this module's 'considerSatisfiability' function.

langston-barrett: I went with "a function argument" rather than "a predicate argument". Thanks for pointing out that it was confusing and suggesting an improvement!

crucible/src/Lang/Crucible/Backend/Online.hs
```diff
@@ -4,19 +4,26 @@
...
+-- SMT solver (in contrast to "Lang.Crucible.Backend.Simple").
 --
--- The online backend also allows override definitions access to a
--- persistent SMT solver connection.  This can be useful for some
--- kinds of algorithms that benefit from quickly performing many
--- small solver queries in a tight interaction loop.
+-- The primary intended use-case is to prune unsatisfiable execution
+-- traces during simulation using the execution feature provided by
+-- "Lang.Crucible.Simulator.PathSatisfiability". That execution feature is
+-- parameterized over a function argument that can be intantiated  with this
```
RyanGlScott:
```suggestion
-- parameterized over a function argument that can be instantiated with this
```

PR #1426, opened 2025-06-12
Pretty-printer, test-case, and haddocks for `GoalCollector`

langston-barrett:
tbh I'm just trying to grok how this works. Might as well document it for others while I'm here!


crucible-cli/test-data/simulate/assumption-state.out.good
```diff
@@ -0,0 +1,143 @@
...
+         , in main test-data/simulate/assumption-state.cbl:33:5: assumption #3: else branch
+         casmp3@10:b ]
+    else [ The branch in main from test-data/simulate/assumption-state.cbl:18:5 to test-data/simulate/assumption-state.cbl:20:16
+         cb0@3:b
+         , in main test-data/simulate/assumption-state.cbl:21:5: assumption #1: if branch
+         casmp1@5:b ]
+  , in main test-data/simulate/assumption-state.cbl:45:5: assumption #4: end
+  casmp4@13:b
+  , in main test-data/simulate/assumption-state.cbl:10:5: assumption #0
+  casmp0@1:b ]
```
langston-barrett, line 69: I don't yet understand where the assertions are in all this.

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
```diff
@@ -55,16 +59,68 @@ newtype FrameIdentifier = FrameIdentifier Word64
...
+--   part of the state of the simple and online backends.
+--
+--   'GoalCollector' can be somewhat counter-intuitive. The "top"
+--   ('TopCollector') is at the *leaves* when 'GoalCollector' is considered as a
+--   tree (which is a common way to conceptualize recursive algebraic data types
+--   such as this one). Furthermore, the frame identified by the frist argument
+--   of 'CollectorFrame' does not conceptually contain the goals *inside* the
+--   second ('GoalCollector') argument, but rather contains all the assumptions
+--   and goals in whatever 'GoalCollector' *contains* the 'CollectorFrame'
+--   constructor (everything *outside* of the 'CollectorFrame').
```
langston-barrett: I would be interested to know if (1) these facts are counter-intuitive to anyone besides me and (2) if this seems like a good way to clarify them.

crucible-cli/test-data/simulate/assumption-state.out.good
```diff
@@ -0,0 +1,143 @@
...
+
+------------------------------
+else
+----
+Top-level goals:
+[]
+Assumptions:
+  in main test-data/simulate/assumption-state.cbl:10:5: assumption #0
+  casmp0@1:b
+  Frame 8:
```
langston-barrett, line 40: It's a little counter-intuitive that these frame numbers are non-consecutive. The constructor for `AssumptionStack` takes a `NonceGenerator IO t` and `newOnlineBackend` initializes that to some nonce generator from the `ExprBuilder`... Might be better to just create a [`newIONonceGenerator`](https://hackage-content.haskell.org/package/parameterized-utils-2.1.10.0/docs/Data-Parameterized-Nonce.html#v:newIONonceGenerator) for each `AssumptionStack`? They're cheap (it is literally just a pointer to a 64-bit integer), and we only really expect there to be one `AssumptionStack` anyway.

crucible-cli/test-data/simulate/assumption-state.out.good
langston-barrett: Oh wait, yes I do: this `GoalCollector` passes through `gcRemoveObligations`, which clears the goals.

crucible-cli/test-data/simulate/assumption-state.out.good
RyanGlScott: I think that would be reasonable. I don't think we ever compare the fresh values generated by the `AssumptionStack`'s `NonceGenerator` with any other `NonceGenerator`-generated fresh values, so I think it would be fine if there were some overlap between the two.

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
RyanGlScott: I agree that it is a bit counterintuitive. I think you've done a reasonably good job of explaining the nuances of `GoalCollector`.

langston-barrett: Rad!

crucible-cli/test-data/simulate/assumption-state.out.good
langston-barrett: #1427

RyanGlScott (approved):

PR #1428, opened 2025-06-17
debug: Add a command to display the backend state

langston-barrett:
Also, add a method to `Backend` to pretty-print the whole backend state, rather than just the assumptions.

### DIFF
--- a/crucible/src/Lang/Crucible/Backend.hs
+++ b/crucible/src/Lang/Crucible/Backend.hs
@@ -81,6 +81,7 @@ module Lang.Crucible.Backend
   , ppProofObligation
   , backendOptions
   , assertThenAssumeConfigOption
+  , ppAssumptionState
   ) where
 
 import           Control.Exception(Exception(..), throwIO)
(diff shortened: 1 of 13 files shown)


crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -311,6 +313,9 @@ class (IsSymInterface sym, HasSymInterface sym bak) => IsSymBackend sym bak | ba
   resetAssumptionState :: bak -> IO ()
   resetAssumptionState bak = restoreAssumptionState bak PG.emptyGoalCollector
 
+  -- | Render a human-readable representation of the state of the backend
+  debugPrintBackendState :: bak -> IO (PP.Doc ann)
```
RyanGlScott: This method feels a bit off to me, considering that it combines two different things (retrieving the backend state and printing it) into a single method. I would be more inclined to add a `getAssumptionState :: bak -> IO (AssumptionState sym)` method that retrieves the current `AssumptionState`, but without doing any of the snapshot-related stuff that `saveAssumptionState` does. Then you could pretty-print the retrieved state using `ppAssumptionState`.

langston-barrett: Agreed, done.

crucible/src/Lang/Crucible/Backend.hs
```diff
@@ -311,6 +313,9 @@ class (IsSymInterface sym, HasSymInterface sym bak) => IsSymBackend sym bak | ba
   resetAssumptionState :: bak -> IO ()
   resetAssumptionState bak = restoreAssumptionState bak PG.emptyGoalCollector
 
+  -- | Get the state of the backend
+  getBackendState :: bak -> IO (AssumptionState sym)
```
RyanGlScott: Worth expanding these Haddocks to contrast this function's behavior with `saveAssumptionState` (which has the same type signature).

PR #1430, opened 2025-06-17
crucible: Add `gcAddTopLevelAssume`

langston-barrett:
This can be used in conjunction with `{save,restore}AssumptionState` to add assumptions that are in scope for all goals. The intended use is by Macaw's lazy memory model, to dynamically add assumptions that parts of the global memory space are initialized (which must be done in all branches).


langston-barrett:
See discussion here: https://github.com/GaloisInc/crucible/commit/db5b4a43a4c3d4dd2abfdf1937bd2a234aed8ed0

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
```diff
@@ -210,6 +210,23 @@ gcAddGoals g (TopCollector gs) = TopCollector (gs Seq.|> g)
...
+-- | Add an assumption that is in scope for all goals.
+gcAddTopLevelAssume ::
+  Monoid asmp =>
+  asmp ->
+  GoalCollector asmp goal ->
+  GoalCollector asmp goal
+gcAddTopLevelAssume asmp =
+  \case
+    TopCollector gls ->
+      CollectingAssumptions asmp (TopCollector (assuming asmp <$> gls))
```
RyanGlScott, line 229: Worth leaving a comment here about why we duplicate `asmp`.

crucible/src/Lang/Crucible/Backend/ProofGoals.hs
```diff
@@ -210,6 +210,23 @@ gcAddGoals g (TopCollector gs) = TopCollector (gs Seq.|> g)
 gcAddGoals g (CollectingGoals gs gc) = CollectingGoals (gs Seq.|> g) gc
 gcAddGoals g gc = CollectingGoals (Seq.singleton g) gc
 
+-- | Add an assumption that is in scope for all goals.
```
RyanGlScott: Perhaps say something like "... in scope for all goals, even ones that were added in earlier frames."

RyanGlScott (approved):

PR #1435, opened 2025-06-17
`crucible-mir`: Support more `UnsizeVtable` casts

### LINKED ISSUE #1433: Unsized coercion involving `dyn` (but not custom DST) causes `crucible-mir` translation error
Consider this program:

```rs
#![feature(coerce_unsized)]
#![feature(unsize)]

use std::marker::Unsize;
use std::ops::{CoerceUnsized, Deref};

pub struct FakeArc<T: ?Sized>(*const T);

impl<T: ?Sized, U: ?Sized> CoerceUnsized<FakeArc<U>> for FakeArc<T>
where T: Unsize<U> {}

impl<T> FakeArc<T> {
    pub fn new(x: T) -> FakeArc<T> {
        FakeArc(Box::leak(Box::new(x)))
    }
}

impl<T: ?Sized> Deref for FakeArc<T> {
    type Target = T;
    fn deref(&self) -> &T {
        unsafe { &*self.0 }
    }
}

/////

trait Foo {
    fn bar(&self);
}

impl Foo for i32 {
    fn bar(&self) {}
}

/////

#[crux::test]
pub fn crux_test() {
    let fa_i32: FakeArc<i32> = FakeArc::new(0);
    let fa_dyn_foo: FakeArc<dyn Foo> = fa_i32;
    fa_dyn_foo.bar();
}
```

I would expect `crucible-mir` to be able to simulate `crux_test()`, which coerces `fa_i32: FakeArc<i32>` to be of type `FakeArc<dyn Foo>`. Despite this, a translation error occurs when you try to run this with `crux-mir`:

```
test test/4e95c16c::crux_test[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/4e95c16c::crux_test[0] counterexamples ----
[Crux]   test.rs:40:40: 40:46: error: in test/4e95c16c::crux_test[0]
[Crux]   Translation error in test/4e95c16c::crux_test[0]: unimplemented cast: Unsize
[Crux]     ty: TyRawPtr (TyInt B32) Immut
[Crux]     as: TyRawPtr (TyDynamic test/4e95c16c::Foo[0]::_traitab7f1a22c55f3729[0]) Immut

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

There are two problems here, one in `crucible-mir`, and another in `mir-json`:

1. We are coercing from `*const i32` to `*const dyn Foo`, which requires a vtable. Despite this, `mir-json` uses an `Unsize` cast instead of an `UnsizeVtable` cast. This is a limitation in the logic that `mir-json` uses to identify casts that require vtables.
2. Even if this `mir-json` bug were fixed, the type of cast we'd need (an `UnsafeVtable` with `TyRawPtr`s) isn't supported in `crucible-mir` at the moment, so we will need to extend `crucible-mir` to allow this.

(Note that despite the use of `dyn Foo`, `FakeArc<dyn Foo>` is _not_ a custom dynamically sized type, so this is not an instance of https://github.com/GaloisInc/crucible/issues/1399.)

RyanGlScott:
This extends `evalCast'` to support the following unsized cast types:

1. `UnsizeVtable` casts between two structs. This is needed to cast between, say, `Foo<i32>` and `Foo<dyn Bar>`.
2. `UnsizeVtable` casts between two raw pointer types.

This also bumps the `mir-json` submodule to bring in the changes from GaloisInc/mir-json#122. This is needed to ensure that `mir-json` accurately classifies `UnsizeVtable` casts between structs, which is necessary to support the first type of cast listed above.

Fixes #1433.

-----

Marked as a draft pending GaloisInc/mir-json#122 (on which this depends) landing upstream.


crucible-mir/CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ This release supports [version
   removed, as they no longer serve a useful purpose.
 * Added an intrinsic for [`read_volatile`](https://doc.rust-lang.org/std/ptr/fn.read_volatile.html)
   and [`write_volatile`](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html)
+* Support unsized casts involving vtables between struct types and raw pointer
+  types.
```
spernsteiner:
```suggestion
* Support raw-pointer and `CoerceUnsized` casts that introduce vtable metadata.
```
Not sure if this is the best wording, but to me "casts between struct types and raw pointer types" sounds like it's describing an ADT -> rawptr cast.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -957,20 +961,28 @@ evalCast' ck ty1 e ty2  = do
...
         adt2 <- findAdt an2
         case (reprTransparentFieldTy col adt1, reprTransparentFieldTy col adt2) of
-            (Just ty1, Just ty2) -> evalCast' M.Unsize ty1 e ty2
-            (Nothing, Nothing) -> coerceUnsizedNormal adt1 adt2 e
+            (Just ty1, Just ty2) ->
+              -- As of the time of writing, it shouldn't be possible for the
+              -- types to be equal here, because if they were, then `rustc`
+              -- wouldn't generate the unsized coercion in the first place. It
+              -- is conceivable that a future version of `rustc` might do this,
+              -- however, so we leave this check here in place as a safeguard.
```
spernsteiner: Alternate wording:

> The struct is `#[repr(transparent)]`, so it has only one non-ZST field.  This field must be the pointer/ref field being coerced, as rustc currently forbids implementing `CoerceUnsized` if it can't a field to coerce.  Since this field is being coerced, its old and new types should never be equal.  Because `CoerceUnsized` is unstable, a future version of `rustc` might change this behavior, so we leave this check here in place as a safeguard.

Trying to clarify specifically why we expect rustc to reject the coercion.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -981,7 +993,11 @@ evalCast' ck ty1 e ty2  = do
             "coerceUnsized on incompatible types (mismatched fields): " ++ show (an1, an2)
         vals' <- forM (zip3 [0..] (v1 ^. vfields) (v2 ^. vfields)) $ \(i, f1, f2) -> do
             val <- getStructField adt1 i e
-            evalCast' M.Unsize (f1 ^. fty) val (f2 ^. fty)
+            -- As an optimization, we only bother to compute a cast if the types
+            -- are syntactically unequal.
```
spernsteiner: This is not just an optimization - we don't have an `evalCast'` case for `(M.UnsizeVtable vtbl, a, b) | a == b` (and in theory we'd like to get rid of the one for `(M.Unsize, a, b) | a == b` as well), so recursing on matching fields would error.

PR #1437, opened 2025-06-19
crux-mir: Update documentation and create a simple tutorial

podhrmic:
TODO

### DIFF
--- a/crux-mir/README.md
+++ b/crux-mir/README.md
@@ -56,12 +56,47 @@ versions corresponding to the last two `crux-mir` releases.
 [mir-json-readme]: https://github.com/GaloisInc/mir-json#readme
 
 
-## Setup: Build and Install
+## Setup
+
+### Build and Install
 
 Use GHC 9.6, 9.8, or 9.10. From the `crux-mir` directory, run:
 
     $ cabal install exe:crux-mir --overwrite-policy=always
 
+### Binary releases
+
+Both `crux-mir` and `mir-json` comes with nightly builds and stable binary releases. If you want to install the nightly build, do the following:
+
+* Go to the [crux-mir action](https://github.com/GaloisInc/crucible/actions/workflows/crux-mir-build.yml) page
+* Select the last build from `master` and scroll down the page, until you find and download the desired binary release (`crux-mir-macos...` or `crux-mir-ubuntu...`)
+* Unzip the downloaded archive and add it to your path
+* Do the same for `mir-json` by going to the [mir-json action](https://github.com/GaloisInc/mir-json/actions) page
+
+**Galois users only:** We provide a convenient install script for Linux and MacOS - you will need to provide a github access token to the `crucible` and `mir-json` repos:
+
+```
+$ ACCESS_TOKEN=xxx... install-crux.sh
+```
+
+### Library documentation
+
+The documentation for the `crucible` crate can be built locally by following these steps:
+
+* clone https://github.com/GaloisInc/mir-json.git 
+* build `mir-json` as described in the [Building from source](https://github.com/GaloisInc/mir-json?tab=readme-ov-file#building-from-source) section
+* build the docs with:
+  ```
+  rustdoc libs/crucible/lib.rs \
+    --edition 2021 \
+    --crate-name crucible \
+    --extern compiler_builtins=rlibs/libcompiler_builtins.rlib \
+    --extern core=rlibs/libcore.rlib \
+    -L rlibs \
+    --out-dir rustdocs \
+    --crate-type rlib
+  ```
+* open `rustdocs/crucible/index.html` in your browser!
 
 ## Usage
 
@@ -89,7 +124,7 @@ translated libraries:
 In the directory of a Cargo project (such as the [find-first-set
 example](example/ffs)), run the project's symbolic tests:
 
-    $ cargo crux-test
+    $ cargo crux-test --lib
 
 `cargo-crux-test` (part of `mir-json`) will translate the code to MIR, then
 invoke `crux-mir` to symbolically simulate the test cases.
@@ -102,6 +137,54 @@ To compile and test a single Rust program:
 
 (Should print 2.)
 
+### Code coverage
+
+`crux-mir` has a code coverage tool that reports branches where only one side or the other was explored during symbolic execution. Note that due to how `mir-json` works, this will only report issues in functions that are called at least once from a `#[crux::test]` entrypoint. If you have dead code (i.e., functions that are never called by any entrypoint), then `mir-json` will discard the dead functions, which means that they won't be mentioned at all in the coverage output. Code coverage currently does not report the percentage of the coverage, only lists the paths that are not covered.
+
+* Clone this directory with `git clone https://github.com/GaloisInc/crucible.git` as we will need the `report-coverage` utility tool
+* In your crate run:
+  ```
+  $ crux-mir --lib --  --branch-coverage --path-sat --output-directory test-coverage
+  ```
+* In your crate, run the following command, pointing it towards the `report-coverage` folder in `crucible` directory (note that you need to point to a directory in `test-coverage` folder that contains `report_data.js`, and the path is currently rather mangled):
+  ```
+  $ cargo run --manifest-path $PATH_TO_CRUCIBLE_REPO/report-coverage/Cargo.toml -- test-coverage/test/62f2dedb\:\:f\[0\]/report_data.js
+  ```
+* This will report all paths not covered, including ones from the standard library, for example:
+  ```
+  warning: branch condition never has value true
+    ┌─ ./libs/core/src/ub_checks.rs:1:1
+    │
+  1 │ 
+    │ ^
+    │
+    ┌─ ./libs/core/src/num/mod.rs:1:1
+    │
+  1 │ 
+    │ - in this macro invocation
+
+  warning: branch condition never has value true
+    ┌─ ./libs/core/src/ub_checks.rs:1:1
+    │
+  1 │ 
+    │ ^
+
+  warning: branch condition never has value false
+    ┌─ lib.rs:43:13
+    │
+  43 │     assert!(a == b);
+    │             ^^^^^^
+
+  ```
+* To limit the coverage only to the code in your crate, use `--filter` to point the tool to the file you want to analyze. Then you get a more condensed output, for example:
+  ```
+  warning: branch condition never has value false
+    ┌─ test.rs:2:8
+    │
+  2 │     if b {
+    │        ^
+  ```
+
 
 ## Examples
 
@@ -109,6 +192,7 @@ The [example/ffs/](example/ffs) directory in this repository contains a simple
 library with concrete and symbolic tests.  Use `cargo crux-test` to run the
 symbolic test, which proves that an optimized find-first-set function is
 equivalent to a simple reference implementation.
+Have a look at [example/README.md](./example/README.md) to learn more about different ways to use `crux-mir`.
 
 A fork of the curve25519-dalek library with symbolic tests for the `Scalar52`
 type is available [here][dalek-fork].  This is the code that appears in the
@@ -138,7 +222,10 @@ Files that are not yet expected to work correctly begin with `// FAIL: ` and
 a brief comment describing the reason for the expected failure.
 
 
-## Limitations
+## Limitations & How-to
+
+
+### Unsafe code
 
 `crux-mir` does not support reasoning about certain types of unsafe code.  Many
 functions in the standard library rely on unsafe code; we have reimplemented
@@ -147,5 +234,111 @@ unsupported functions still remain.  Test cases that call into unsupported code
 will produce assertion failures with messages like `don't know how to call
 core::intrinsics::transmute`.
 
+
+### References & function pointers
+
 Currently, `crux-mir` also has trouble supporting references and function
 pointers in constant expressions and static initializers.
+
+
+### Symbolic size arrays
+
+`crux-mir` does support vectors (and data structures) with a symbolic capacity if an upper bound on the size is provided. Truly unbounded sizes are not supported. For example, a `Vec` with *symbolic* length but *fixed* capacity can be used as follows:
+
+```Rust
+let mut v = Vec::with_capacity(8);
+for _ in 0 .. usize::symbolic_where("n", |n| n <= 8) {
+    v.push(i32::symbolic("x"));
+}
+```
+
+
+### Don't know how to call...
+
+If you get a translation error and ` Don't know how to call...` message (an example is below), that usually means that your code contains a compiler intrinsic that `crucible-mir` doesn't know how to simulate. Often times, this can be fixed by adding a custom override for the intrinsic. In such case, we encourage you to look through the [existing issues](https://github.com/GaloisInc/crucible/issues) if such missing case was reported already, and if not, open a new issue.
+
+The error will look similar to this:
+
+```
+[Crux]   Translation error in subtle/81045fbd::black_box[0]::_inst2efc261c2cb07b6c[0]: callExp: Don't know how to call core/c7248340::intrinsics[0]::volatile_load[0]::_inst2efc261c2cb07b6c[0]
+```
+
+Similarly, if you see a Haskell error trace in the output, you have likely discovered a bug.
+
+
+### Loop-unrolling
+
+If your code contains loops, the default invocation `cargo crux-mir --lib` might be very slow. If that is the case, add `--path-sat` argument. Then each time around the loop, `crux-mir` will check with the solver whether it should keep unrolling or no. The full command is:
+
+```
+$ cargo crux-test --lib -- --path-sat
+```
+
+
+### Print counterexamples
+
+When your test fails, it is often useful to get concrete counterexample to better understand which input causes the failure. You can use `-m` argument to print the counterexample:
+
+```
+$ cargo crux-test --lib -- -m 
+...
+---- kinds_of_failure/c15b5012::overflow_quicksort[0]::crux_test[0]::midpoint_overflow[0] counterexamples ----
+[Crux] Found counterexample for verification goal
+[Crux]   src/overflow_quicksort.rs:31:27: 31:34: error: in kinds_of_failure/c15b5012::overflow_quicksort[0]::crux_test[0]::midpoint_overflow[0]
+[Crux]   attempt to compute `copy _1 + copy _2`, which would overflow
+Model:
+[{"name": "a","loc": null,"val": "-0x58cf1802","bits": "32"},{"name": "b","loc": null,"val": "0x58cf1803","bits": "32"}]
+```
+
+The model shows the name of variable, as well as its concrete hexadecimal value.
+
+In addition to `-m` (which can be hard to interpret if you have multiple symbolic variables with the same name), you can use `crucible_assert!()` macro. For example `crucible_assert!(x == y, "expected x == y, but got x = {} and y = {}", x, y);` will print some concrete values of x and y. 
+
+### Constraining symbolic values
+
+The code below is equivalent, use either construct as appropriate:
+
+```Rust
+// A
+let x = u8::symbolic_where("x", |x| x < 3);
+
+// B
+let x = u8::symbolic("x");
+crucible_assume!(x < 3);
+
+// C
+let x = u8::symbolic("x");
+if !(x < 3) { crucible_assume_unreachable!(); }
+```
+
+### Where is the source code for the crucible crate?
+
+You are probably looking for [https://github.com/GaloisInc/mir-json/tree/master/libs/crucible](https://github.com/GaloisInc/mir-json/tree/master/libs/crucible) - if you just want to see the crate level docs, have a look at **Library documentation** section above.
+
+
+### Implement symbolic trait for a custom type
+
+To be able to call `MyType::symbolic()` you need to implement the [`Symbolic`](https://github.com/GaloisInc/mir-json/blob/master/libs/crucible/symbolic.rs#L1) trait. For example for a simple `enum` the implementation can be as follows:
+
+```Rust
+#[derive(Copy, Clone, Debug, PartialEq)]
+pub enum Rating {
+    One,
+    Two,
+    Three,
+}
+
+impl Symbolic for Rating {
+    fn symbolic(desc: &'static str) -> Self {
+        let val = u8::symbolic_where(desc, |&x| x <= 3);
+        match val {
+            1 => Rating::One,
+            2 => Rating::Two,
+            3 => Rating::Three,
+            _ => crucible_assume_unreachable!(),
+        }
+    }
+}
+```
+
+In this example `crucible_assume_unreachable!()` is a symbolic equivalent of `unreachable!()` macro.


crux-mir/README.md
```diff
@@ -56,12 +56,47 @@ versions corresponding to the last two `crux-mir` releases.
 [mir-json-readme]: https://github.com/GaloisInc/mir-json#readme
 
 
-## Setup: Build and Install
+## Setup
```
RyanGlScott: Why remove the "Build and Install" part of this section header? I think it's important to include to distinguish it from the "Setup: `mir-json`" section that precedes it.

crux-mir/README.md
```diff
@@ -56,12 +56,47 @@ versions corresponding to the last two `crux-mir` releases.
...
+
+### Build and Install
 
 Use GHC 9.6, 9.8, or 9.10. From the `crux-mir` directory, run:
 
     $ cabal install exe:crux-mir --overwrite-policy=always
 
+### Binary releases
+
+Both `crux-mir` and `mir-json` comes with nightly builds and stable binary releases. If you want to install the nightly build, do the following:
```
RyanGlScott: It might be worth including a link to https://github.com/GaloisInc/crucible/releases for users who want stable binary releases.

crux-mir/README.md
```diff
@@ -56,12 +56,47 @@ versions corresponding to the last two `crux-mir` releases.
...
+
+* Go to the [crux-mir action](https://github.com/GaloisInc/crucible/actions/workflows/crux-mir-build.yml) page
+* Select the last build from `master` and scroll down the page, until you find and download the desired binary release (`crux-mir-macos...` or `crux-mir-ubuntu...`)
+* Unzip the downloaded archive and add it to your path
+* Do the same for `mir-json` by going to the [mir-json action](https://github.com/GaloisInc/mir-json/actions) page
```
RyanGlScott: It might make more sense to put this part in the `Setup: `mir-json` section instead.

crux-mir/README.md
```diff
@@ -56,12 +56,47 @@ versions corresponding to the last two `crux-mir` releases.
...
+
+**Galois users only:** We provide a convenient install script for Linux and MacOS - you will need to provide a github access token to the `crucible` and `mir-json` repos:
+
+```
+$ ACCESS_TOKEN=xxx... install-crux.sh
+```
```
RyanGlScott: I'm not sure what this `install-crux.sh` script is referring to.

crux-mir/README.md
```diff
@@ -56,12 +56,47 @@ versions corresponding to the last two `crux-mir` releases.
...
+  rustdoc libs/crucible/lib.rs \
+    --edition 2021 \
+    --crate-name crucible \
+    --extern compiler_builtins=rlibs/libcompiler_builtins.rlib \
+    --extern core=rlibs/libcore.rlib \
+    -L rlibs \
+    --out-dir rustdocs \
+    --crate-type rlib
+  ```
+* open `rustdocs/crucible/index.html` in your browser!
```
RyanGlScott: I wonder if this should go into the `mir-json` README instead, since nothing about this is specific to `crux-mir`. We could consider linking to that part of the `mir-json` README from the `crux-mir` README if you'd prefer.

crux-mir/README.md
```diff
@@ -102,13 +137,62 @@ To compile and test a single Rust program:
...
 (Should print 2.)
 
+### Code coverage
+
+`crux-mir` has a code coverage tool that reports branches where only one side or the other was explored during symbolic execution. Note that due to how `mir-json` works, this will only report issues in functions that are called at least once from a `#[crux::test]` entrypoint. If you have dead code (i.e., functions that are never called by any entrypoint), then `mir-json` will discard the dead functions, which means that they won't be mentioned at all in the coverage output. Code coverage currently does not report the percentage of the coverage, only lists the paths that are not covered.
+
+* Clone this directory with `git clone https://github.com/GaloisInc/crucible.git` as we will need the `report-coverage` utility tool
+* In your crate run:
+  ```
+  $ crux-mir --lib --  --branch-coverage --path-sat --output-directory test-coverage
```
RyanGlScott: Should this `--lib` flag be here? I think `--lib` is a flag meant to `cargo crux-test`, not `crux-mir` itself.

Alternatively, perhaps you meant to use `cargo crux-test --lib` instead of `crux-mir --lib`.

crux-mir/README.md
```diff
@@ -102,13 +137,62 @@ To compile and test a single Rust program:
...
+  ```
+* In your crate, run the following command, pointing it towards the `report-coverage` folder in `crucible` directory (note that you need to point to a directory in `test-coverage` folder that contains `report_data.js`, and the path is currently rather mangled):
+  ```
+  $ cargo run --manifest-path $PATH_TO_CRUCIBLE_REPO/report-coverage/Cargo.toml -- test-coverage/test/62f2dedb\:\:f\[0\]/report_data.js
+  ```
```
RyanGlScott: Specifically, the reason that the path is mangled is because `crux-mir` generates per-function coverage reports, and functions are uniquely determined by a _disambiguator_ value. The disambiguator is sensitive to things like the contents of the source code and platform-specific information, which is why it is likely to change from invocation to invocation.

I would suggest rephrasing this bullet point like so to make this point clearer:

~~~suggestion
* In your crate, run the following command, pointing it towards the `report-coverage` folder in `crucible` directory, which contains the coverage script. You will need to point to a directory in the `test-coverage` folder that contains a function's coverage report data (`report_data.js`):
  ```
  $ cargo run --manifest-path $PATH_TO_CRUCIBLE_REPO/report-coverage/Cargo.toml -- test-coverage/test/62f2dedb\:\:f\[0\]/report_data.js
  ```

  Note that the `62f2dedb` part of this path will likely be different on your machine due to how `mir-json` works.
~~~

crux-mir/README.md
```diff
@@ -102,13 +137,62 @@ To compile and test a single Rust program:
...
+    │ ^
+
+  warning: branch condition never has value false
+    ┌─ lib.rs:43:13
+    │
+  43 │     assert!(a == b);
+    │             ^^^^^^
+
+  ```
+* To limit the coverage only to the code in your crate, use `--filter` to point the tool to the file you want to analyze. Then you get a more condensed output, for example:
```
RyanGlScott, line 172: Please include a complete invocation of the coverage script using `--filter` to make it clearer how to use this option.

crux-mir/README.md
```diff
@@ -147,5 +234,111 @@ unsupported functions still remain.  Test cases that call into unsupported code
...
+```
+[Crux]   Translation error in subtle/81045fbd::black_box[0]::_inst2efc261c2cb07b6c[0]: callExp: Don't know how to call core/c7248340::intrinsics[0]::volatile_load[0]::_inst2efc261c2cb07b6c[0]
+```
+
+Similarly, if you see a Haskell error trace in the output, you have likely discovered a bug.
+
+
+### Loop-unrolling
+
+If your code contains loops, the default invocation `cargo crux-mir --lib` might be very slow. If that is the case, add `--path-sat` argument. Then each time around the loop, `crux-mir` will check with the solver whether it should keep unrolling or no. The full command is:
```
RyanGlScott:
```suggestion
If your code contains loops, the default invocation `cargo crux-mir --lib` might be very slow. If that is the case, add `--path-sat` argument. Then each time around the loop, `crux-mir` will check with the solver whether it should keep unrolling or not. The full command is:
```

crux-mir/README.md
```diff
@@ -147,5 +234,111 @@ unsupported functions still remain.  Test cases that call into unsupported code
...
+
+```
+$ cargo crux-test --lib -- --path-sat
+```
+
+
+### Print counterexamples
+
+When your test fails, it is often useful to get concrete counterexample to better understand which input causes the failure. You can use `-m` argument to print the counterexample:
```
RyanGlScott:
```suggestion
When your test fails, it is often useful to get a concrete counterexample to better understand which input causes the failure. You can use the `-m` argument to print the counterexample:
```

crux-mir/README.md
```diff
@@ -147,5 +234,111 @@ unsupported functions still remain.  Test cases that call into unsupported code
...
+...
+---- kinds_of_failure/c15b5012::overflow_quicksort[0]::crux_test[0]::midpoint_overflow[0] counterexamples ----
+[Crux] Found counterexample for verification goal
+[Crux]   src/overflow_quicksort.rs:31:27: 31:34: error: in kinds_of_failure/c15b5012::overflow_quicksort[0]::crux_test[0]::midpoint_overflow[0]
+[Crux]   attempt to compute `copy _1 + copy _2`, which would overflow
+Model:
+[{"name": "a","loc": null,"val": "-0x58cf1802","bits": "32"},{"name": "b","loc": null,"val": "0x58cf1803","bits": "32"}]
+```
+
+The model shows the name of variable, as well as its concrete hexadecimal value.
```
RyanGlScott: Note that in general, `val` may not be printed as a hexadecimal value (that output is specific to bitvectors, I believe). Perhaps just say "its concrete value" instead.

crux-mir/README.md
```diff
@@ -147,5 +234,111 @@ unsupported functions still remain.  Test cases that call into unsupported code
...
+
+In addition to `-m` (which can be hard to interpret if you have multiple symbolic variables with the same name), you can use `crucible_assert!()` macro. For example `crucible_assert!(x == y, "expected x == y, but got x = {} and y = {}", x, y);` will print some concrete values of x and y.
```
RyanGlScott:
```suggestion
In addition to `-m` (which can be hard to interpret if you have multiple symbolic variables with the same name), you can use the `crucible_assert!()` macro. For example `crucible_assert!(x == y, "expected x == y, but got x = {} and y = {}", x, y);` will print some concrete values of `x` and `y`. 
```

crux-mir/README.md
```diff
@@ -147,5 +234,111 @@ unsupported functions still remain.  Test cases that call into unsupported code
...
+let x = u8::symbolic("x");
+if !(x < 3) { crucible_assume_unreachable!(); }
+```
+
+### Where is the source code for the crucible crate?
+
+You are probably looking for [https://github.com/GaloisInc/mir-json/tree/master/libs/crucible](https://github.com/GaloisInc/mir-json/tree/master/libs/crucible) - if you just want to see the crate level docs, have a look at **Library documentation** section above.
+
+
+### Implement symbolic trait for a custom type
```
RyanGlScott:
```suggestion
### Implement `Symbolic` trait for a custom type
```

crux-mir/README.md
```diff
@@ -147,5 +234,111 @@ unsupported functions still remain.  Test cases that call into unsupported code
...
+
+To be able to call `MyType::symbolic()` you need to implement the [`Symbolic`](https://github.com/GaloisInc/mir-json/blob/master/libs/crucible/symbolic.rs#L1) trait. For example for a simple `enum` the implementation can be as follows:
```
RyanGlScott:
```suggestion
To be able to call `MyType::symbolic()` you need to implement the [`Symbolic`](https://github.com/GaloisInc/mir-json/blob/master/libs/crucible/symbolic.rs#L1) trait. For example, for a simple `enum` the implementation can be as follows:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
+# crux-mir tutorial
+
+This tutorial is intended for engineers with *some* understanding of formal methods, rather than for formal methods experts. After reading through the tutorial, you should be able to write symbolic tests for your Rust code, and understand the results provided by `crux-mir`. This tutorial is [learning oriented](https://docs.divio.com/documentation-system/), if you need a quick reference for fixing problems, we recommend the [Limitations & How-to](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md) section in the [crux-mir README.md](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md)
+
+Note that while the tool is called `crux-mir`, the invocation is via `cargo crux-test`, mostly for historical reasons. We will be using `crux-mir` and `crux-test` interchangeably.
```
RyanGlScott: I'm not sure that this distinction is accurate. The `crux-mir` tool and `cargo crux-test` do distinct (but related) things. Perhaps it would be better to phrase this as:

```suggestion
The `crux-mir` tool is geared towards compiling and testing a single, standalone Rust file. `mir-json` also offers a `cargo` subcommand, `cargo crux-test`, which generalizes `crux-mir` to allow compiling and testing a complete Cargo project. While we will be using `cargo crux-test` below to compile all of the examples, we will use the terms "`crux-mir`" and "`crux-test`" interchangeably to describe the underlying machinery.
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+## Find First Set
+
+This example implements [Find First Set](https://en.wikipedia.org/wiki/Find_first_set) algorithm, and uses `crux-mir` to prove equivalence between the reference implementation and the fast implementation.
```
RyanGlScott:
```suggestion
This example implements the [Find First Set](https://en.wikipedia.org/wiki/Find_first_set) algorithm and uses `crux-mir` to prove equivalence between the reference implementation and the fast implementation.
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+[Crux]   Proved: 2
+[Crux]   Disproved: 0
+[Crux]   Incomplete: 0
+[Crux]   Unknown: 0
+[Crux] Overall status: Valid.
+```
+
+## Example 1: Failing corner case
+
+**NOTE:** this example was adapted from the [kani tutorial](https://model-checking.github.io/kani/kani-tutorial.html).
```
RyanGlScott: I believe "Kani" is supposed to be capitalized:

```suggestion
**NOTE:** this example was adapted from the [Kani tutorial](https://model-checking.github.io/kani/kani-tutorial.html).
```

(And similarly for all other lowercased occurrences of "kani" in this README.)

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+```
+
+And the output will contain the model value:
+
+```
+Model:
+[{"name": "x","loc": null,"val": "0x3ff","bits": "32"}]
+```
+
+Note that the variable value is printed in hexadecimal format. `x` is the name of the symbolic variable, assigned when creating `u32::symbolic("x")`, and value `0x3ff` is `1023` decimal, which neatly matches the condition guards:
```
RyanGlScott:
```suggestion
Note that the variable value is printed in hexadecimal format. `x` is the name of the symbolic variable, assigned when creating `u32::symbolic("x")`, and the value `0x3ff` is `1023` in decimal, which neatly matches the condition guards:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+In this example, we have a function that requires the input to be smaller than 4096:
+
+```Rust
+fn estimate_size(x: u32) -> u32 {
+    assert!(x < 4096);
+...
+```
+
+In our test suite, we limit the input such that the assertion is not triggered (try running `cargo test`), but when we write our symbolic test, and check the function with *all possible values* we get an error:
```
RyanGlScott:
```suggestion
In our test suite, we limit the input such that the assertion is not triggered (try running `cargo test`), but when we write our symbolic test and check the function with *all possible values*, we get an error:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+```Rust
+#[crux::test]
+fn will_fail() {
+    let x: u32 = u32::symbolic("x");
+    let y = estimate_size(x);
+}
+```
+
+Lets run the `will_fail` test, and because we have more than one test in this example, append `will_fail` at the end of the invocation to filter which tests are run:
```
RyanGlScott:
```suggestion
Let's run the `will_fail` test, and because we have more than one test in this example, append `will_fail` at the end of the invocation to filter which tests are run:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+[Crux]   Total: 1
+[Crux]   Proved: 0
+[Crux]   Disproved: 1
+[Crux]   Incomplete: 0
+[Crux]   Unknown: 0
+[Crux] Overall status: Invalid.
+error: test failed, to rerun pass `--lib`
+```
+
+OK, that looks bad - we gave the `estimate_size()` function a really big `u32` number, and it triggered the assertion (which in turn triggers `panic!` and that is what you see in the error message). Fortunately we can make assumptions about the symbolic values with `crucible_assume!` macro. From the documentation of `crucible_assume!`:
```
RyanGlScott:
```suggestion
OK, that looks bad - we gave the `estimate_size()` function a really big `u32` number, and it triggered the assertion (which in turn triggers `panic!` and that is what you see in the error message). Fortunately we can make assumptions about the symbolic values with the `crucible_assume!` macro. From the documentation of `crucible_assume!`:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+fn check_initialize_prefix() {
+    const LIMIT: usize = 10;
+    let mut buffer: [u8; LIMIT] = [1; LIMIT];
+
+    let length = usize::symbolic("x");
+    initialize_prefix(length, &mut buffer);
+}
+```
+
+Notice we decided to use a relatively small array with only 10 elements. Lets run the test:
```
RyanGlScott:
```suggestion
Notice we decided to use a relatively small array with only 10 elements. Let's run the test:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+```
+$ cargo crux-test --lib --  -m midpoint_overflow
+...
+Model:
+[{"name": "a","loc": null,"val": "0xa730e7fe","bits": "32"},{"name": "b","loc": null,"val": "0x58cf1803","bits": "32"}]
+```
+
+Note that the model representation is a bit hard to read, and might change (see [#1434](https://github.com/GaloisInc/crucible/issues/1434) for details). This shows us two values that when added together overflow `u32`, which is an error. You might get different values (depending on your solver), but they will overflow `u32` when added together.
+
+If you are wondering how to best fix this error, have a look at [Kani tutorial](https://model-checking.github.io/kani/tutorial-kinds-of-failure.html#exercise-classic-overflow-failure) where this example is adapted from.
```
RyanGlScott:
```suggestion
If you are wondering how to best fix this error, have a look at the [Kani tutorial](https://model-checking.github.io/kani/tutorial-kinds-of-failure.html#exercise-classic-overflow-failure) where this example is adapted from.
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+fn bound_check() {
+    const LIMIT: usize = 10;
+    let size = LIMIT;
+    let i = usize::symbolic("i");
+    let array: Vec<u32> = vec![1; size];
+    get_wrapped(i, &array);
+}
+```
+
+Note that while we could create a `Vec` with *symbolic* length but *fixed* capacity, rather than a fixed length vector. It would come at a performance cost, and in this example it does not make a difference. However, have a look at [Limitations & How-to](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md) section in the [crux-mir README.md](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md) for an example how to create a vector with *symbolic* length.
```
RyanGlScott:
```suggestion
Note that while we could create a `Vec` with *symbolic* length but *fixed* capacity, rather than a fixed length vector. It would come at a performance cost, and in this example it does not make a difference. However, have a look at the [Limitations & How-to](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md) section in the [crux-mir README.md](https://github.com/GaloisInc/crucible/blob/master/crux-mir/README.md) for an example of how to create a vector with *symbolic* length.
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+Fortunately we can use `crucible_assert!` macro, which is described as:
+
+> Assert that a condition holds. During symbolic testing, `crux-mir` will search for an assignment to the symbolic variables that violates an assertion.
+> 
+> This macro works just like the standard `assert!`, but currently produces better error messages.
+> 
+> If the error message uses string formatting, `crux-mir` will choose an arbitrary concrete counterexample and use its values when formatting the message. For example: `crucible_assert!(x + y == z, "bad arithmetic: {} + {} != {}", x, y, z);` might print a message like bad arithmetic: `1 + 2 != 4`, assuming `x`, `y`, and `z` are symbolic variables that can take on the values 1, 2, and 4..
+
+We can use `crucible_assert!` in both the crux test code, and in the source code. In this case we will edit the source code, because we want to know what is the value of `idx` before the out of bounds access is attempted:
```
RyanGlScott:
```suggestion
We can use `crucible_assert!` in both the crux test code and in the source code. In this case we will edit the source code, because we want to know what is the value of `idx` before the out of bounds access is attempted:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+        return 0;
+    }
+
+    let idx = i % a.len() + 1;
+    crucible_assert!(idx < a.len(), "idx={}, a.len()={}, i = {}", idx, a.len(), i);
+    return a[idx];
+}
+```
+
+When we run the code we get the assertion print additional debug information:
```
RyanGlScott:
```suggestion
When we run the code, the assertion will print additional debug information:
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+[Crux] Found counterexample for verification goal
+[Crux]   ./libs/crucible/lib.rs:48:13: 55:14 !src/bounds_check.rs:12:5: 12:83: error: in kinds_of_failure/acd766e3::bounds_check[0]::get_wrapped[0]
+[Crux]   MIR assertion at src/bounds_check.rs:12:5:
+[Crux]          idx=10, a.len()=10, i = 089
+...
+```
+
+Seeing that index `i=89` leads to `idx=9+1=10` which is the same as `a.len()` tells us we are off by one! You can use `crucible_assert!` when debugging complex errors in your code. Just don't forget to remove the `crucible` code when you are done debugging, as it would trigger a compilation error when building the code with `cargo build`.
+
+There is also `crucible_assert_unreachable!` which can be used instead of `unreachable!` macro.
```
RyanGlScott:
```suggestion
There is also `crucible_assert_unreachable!` which can be used instead of the `unreachable!` macro.
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+
+## Crux-mir in CI and changing solvers
+
+A fork of the curve25519-dalek library with symbolic tests for the `Scalar52`
+type is available [here](https://github.com/GaloisInc/curve25519-dalek). This is the code that appears in the [`crux-mir` demo video](https://www.youtube.com/watch?v=dCNQFHjgotU). We added a [`prove` CI job](https://github.com/GaloisInc/curve25519-dalek/blob/master/.github/workflows/ci.yml#L59) that shows how to use `crux-mir` in Github CI. Similar approach will work in Gitlab CI.
```
RyanGlScott:
```suggestion
type is available [here](https://github.com/GaloisInc/curve25519-dalek). This is the code that appears in the [`crux-mir` demo video](https://www.youtube.com/watch?v=dCNQFHjgotU). We added a [`prove` CI job](https://github.com/GaloisInc/curve25519-dalek/blob/master/.github/workflows/ci.yml#L59) that shows how to use `crux-mir` in GitHub Actions. Similar approach will work in GitLab CI.
```

crux-mir/example/README.md
```diff
@@ -0,0 +1,449 @@
...
+
+```yml
+...
+ - name: Run proofs
+        run: |
+          cargo crux-test --lib -- -s z3
+...
+```
+
+The `-s z3` flag, which tells `crux-mir` to specifically use `z3` solver (assuming it is installed). In some cases it might be beneficial to use specific solvers, but the details of when are beyond the scope of this tutorial.
```
RyanGlScott:
```suggestion
Note the `-s z3` flag, which tells `crux-mir` to specifically use the `z3` solver (assuming it is installed). In some cases it might be beneficial to use specific solvers, but the details of when are beyond the scope of this tutorial.
```

crux-mir/example/README.md
RyanGlScott, line 1: There is one `example` subdirectory that this `README` _doesn't_ describe, which is the pre-existing `ffs` directory. I feel like we should either mention this example somewhere in the `README` or remove it entirely.

crux-mir/README.md
podhrmic: Good catch, fixed.

crux-mir/example/README.md
podhrmic: Added as `example-1`

crux-mir/README.md
podhrmic: Done

crux-mir/README.md
podhrmic: Updated

crux-mir/README.md
podhrmic: Done

crux-mir/README.md
podhrmic: I removed the script - maybe we can add it in the future if needed

crux-mir/README.md
podhrmic: I reworded the section, please have a look

crux-mir/README.md
podhrmic: Done, see https://github.com/GaloisInc/mir-json/pull/124

crux-mir/README.md
```diff
@@ -28,6 +28,15 @@ You will need to install the following software:
...
 ## Setup: `mir-json`
 
+### Download `mir-json` binaries
+
+To download the latest version of `mir-json`, go to the [mir-json action](https://github.com/GaloisInc/mir-json/actions) page. Select the last build from `master` and scroll down the page, until you find and download the desired binary release (`mir-json-macos...` or `mir-json-ubuntu...`). Unzip the downloaded archive and add it to your path.
+
+The binary release comes with precompiled `rlibs`, so don't forget to set `CRUX_RUST_LIBRARY_PATH` variable pointing to the `rlibs` subfolder of the unzipped archive. Adding `export CRUX_RUST_LIBRARY_PATH=...` to your shell configuration is recommended.
+
+
+### Build `mir-json` from source
```
RyanGlScott: I think it would be helpful to include a brief paragraph right after the "Setup: `mir-json`" section header which states that there are two different ways to set up `mir-json`: either by downloading the binaries from GitHub Actions (the "Download `mir-json` binaries" section) or by building it from source (the "Build `mir-json` from source" section). With the way that the subsections are laid out currently, one might be misled into thinking that you have to both download `mir-json` _and_ build it from source, which could be confusing.

crux-mir/README.md
```diff
@@ -56,7 +65,18 @@ versions corresponding to the last two `crux-mir` releases.
...
+### Download `crux-mir` binaries
+
+`crux-mir` comes with nightly builds and stable [binary releases](https://github.com/GaloisInc/crucible/releases). If you want to install the nightly build, do the following:
+
+* Go to the [crux-mir action](https://github.com/GaloisInc/crucible/actions/workflows/crux-mir-build.yml) page
+* Select the last build from `master` and scroll down the page, until you find and download the desired binary release (`crux-mir-macos...` or `crux-mir-ubuntu...`)
+* Unzip the downloaded archive and add it to your path
+
+
+### Build `crux-mir` from soruce
```
RyanGlScott: Similarly, it would be worth a brief paragraph after the "Setup: `crux-mir`" section header explaining that you can either download it or build it from source.

RyanGlScott (approved):

crux-mir/README.md
podhrmic: Good idea, done!

crux-mir/README.md
podhrmic: Done in 259b511

podhrmic:
@RyanGlScott do you prefer keeping the commit history, or can I squash and merge for a single tutorial-related commit?

RyanGlScott:
Up to you, but I would be in favor of squashing these commits for the sake of making them a bit tidier.

PR #1438, opened 2025-06-19
Better counterexample representation

### LINKED ISSUE #1434: `crux-mir`: incorrect counterexamples
I have the following code (originally from [the kani tutorial](https://model-checking.github.io/kani/tutorial-kinds-of-failure.html):

```Rust
fn find_midpoint(low: u32, high: u32) -> u32 {
    return (low + high) / 2;
}

#[cfg(crux)]
mod crux_test {
    use super::*;
    extern crate crucible;
    use self::crucible::*;

    #[crux::test]
    fn midpoint_overflow() {
        let a = u32::symbolic("a");
        let b = u32::symbolic("b");
        crucible_assert!( (a + b) < u32::MAX, "{} {}", a, b);
        find_midpoint(a, b);
    }
}
```

Running `cargo crux-test -- -s z3 -m --path-sat midpoint` leads to the following output:

```
---- kinds_of_failure/c15b5012::overflow_quicksort[0]::crux_test[0]::midpoint_overflow[0] counterexamples ----
[Crux]   src/overflow_quicksort.rs:31:27: 31:34: error: in kinds_of_failure/c15b5012::overflow_quicksort[0]::crux_test[0]::midpoint_overflow[0]
[Crux]   attempt to compute `copy _1 + copy _2`, which would overflow
Model:
[{"name": "a","loc": null,"val": "-0x58cf1802","bits": "32"},{"name": "b","loc": null,"val": "0x58cf1803","bits": "32"}]
[Crux]   ./libs/crucible/lib.rs:48:13: 55:14 !src/overflow_quicksort.rs:31:9: 31:61: error: in kinds_of_failure/c15b5012::overflow_quicksort[0]::crux_test[0]::midpoint_overflow[0]
[Crux]   MIR assertion at src/overflow_quicksort.rs:31:9:
[Crux]          895 0
Model:
[{"name": "a","loc": null,"val": "-0x80000000","bits": "32"},{"name": "b","loc": null,"val": "0x7fffffff","bits": "32"}]
[Crux]   src/overflow_quicksort.rs:7:12: 7:24: error: in kinds_of_failure/c15b5012::overflow_quicksort[0]::find_midpoint[0]
[Crux]   attempt to compute `copy _1 + copy _2`, which would overflow
Model:
[{"name": "a","loc": null,"val": "-0x58cf1802","bits": "32"},{"name": "b","loc": null,"val": "0x58cf1803","bits": "32"}]

[Crux] Goal status:
[Crux]   Total: 3
[Crux]   Proved: 0
[Crux]   Disproved: 3
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
error: test failed, to rerun pass `--lib`
```

Note that both the asserted values and the printed models seem a little off - instead of `895 0` we should see larger numbers. The hex values seem plausible, but all the numbers should be positive.

podhrmic:
Fixes #1434

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -116,7 +116,12 @@ parsePosition posText =
       | (l,[]):_ <- readDec (Text.unpack line)
       , (c,[]):_ <- readDec (Text.unpack col)
       -> PL.SourcePos fname l c
+    [fname,line,col, _, _]
+      | (l,[]):_ <- readDec (Text.unpack line)
+      , (c,[]):_ <- readDec (Text.unpack col)
+      -> PL.SourcePos fname l c
     _ -> PL.OtherPos posText
+
 
 
 setPosition :: Text.Text -> MirGenerator h s ret ()


podhrmic:
Not sure why the test are failing - probably I will need to re-generate the expected outputs?

RyanGlScott:
Now that the type of `showBVLiteral` (part of `crux`'s public API) has changed, `crux-llvm` now fails to compile. See [this failing CI log](https://github.com/GaloisInc/crucible/actions/runs/15784102678/job/44496588242?pr=1438):

```
src/Crux/LLVM/Compile.hs:328:11: error: [GHC-83865]
Error:     • Couldn't match type: Crux.Model.BVRepr -> String
                     with: [Char]
      Expected: ([Char], [Char],
                 What4.Expr.GroundEval.GroundValue ty -> [Char])
        Actual: ([Char], [Char],
                 bv-sized-1.0.6:Data.BitVector.Sized.Internal.BV w
                 -> Crux.Model.BVRepr -> String)
    • In the expression:
        ("int" ++ show n ++ "_t", "int" ++ show n ++ "_t", showBVLiteral n)
      In a case alternative:
          BaseBVRepr n
            -> ("int" ++ show n ++ "_t", "int" ++ show n ++ "_t", 
                showBVLiteral n)
      In the expression:
        case ty of
          BaseBVRepr n
            -> ("int" ++ show n ++ "_t", "int" ++ show n ++ "_t", 
                showBVLiteral n)
          BaseFloatRepr (FloatingPointPrecisionRepr eb sb)
            | Just Refl <- testEquality eb (knownNat @8),
              Just Refl <- testEquality sb (knownNat @24)
            -> ("float", "float", showFloatLiteral)
          BaseFloatRepr (FloatingPointPrecisionRepr eb sb)
            | Just Refl <- testEquality eb (knownNat @11),
              Just Refl <- testEquality sb (knownNat @53)
            -> ("double", "double", showDoubleLiteral)
          BaseRealRepr -> ("double", "real", (show . toDouble))
          _ -> error ("Type not implemented: " ++ show ty)
    |
328 |           ("int" ++ show n ++ "_t", "int" ++ show n ++ "_t", showBVLiteral n)
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

This should be fixed, and this change to `crux`'s API should be noted in the `crux` changelog.

crux/src/Crux/UI/JS.hs
```diff
@@ -29,7 +34,7 @@ jsLoc x =
 newtype JS = JS { renderJS :: String }
 
 jsList :: [JS] -> JS
-jsList xs = JS $ "[" ++ intercalate "," [ x | JS x <- xs ] ++ "]"
+jsList xs = JS $ prettyPrintJson $ "[" ++ intercalate "," [ x | JS x <- xs ] ++ "]"
```
RyanGlScott: Why was this change made?

crux/src/Crux/Model.hs
```diff
@@ -85,6 +105,17 @@ valsJS ty (Vals xs) =
          , "val"  ~> jsStr (repr (entryValue e))
          , "bits" ~> jsStr (show n)
          ]
+  showEnt'' :: Show b => (a -> BVRepr -> String) -> b -> Entry a -> IO JS
```
RyanGlScott: This really ought to be named something more descriptive than `showEnt''`, especially because this is only used for bitvector values. Perhaps `showBVEnt`?

Moreover, this function only has a single call site, which makes it unnecessary to parameterize it with a higher-order function like `showEnt'` is. I would propose just giving this function a more direct type like `NatRepr w -> Entry (BV w) -> IO JS` and calling `showBVLiteral{,Unsigned,Decimal}` directly in the body of the function.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -113,12 +113,17 @@ parsePosition :: Text.Text -> PL.Position
...
   case Text.split (==':') posText of
     [fname,line,col]
-      | (l,[]):_ <- readDec (Text.unpack line)
-      , (c,[]):_ <- readDec (Text.unpack col)
+      | (l,[]):_ <- readDec (Text.unpack (Text.strip line))
+      , (c,[]):_ <- readDec (Text.unpack (Text.strip col))
+      -> PL.SourcePos fname l c
+    fname : line : col : _rest
+      | (l,[]):_ <- readDec (Text.unpack (Text.strip line))
+      , (c,[]):_ <- readDec (Text.unpack (Text.strip col))
```
RyanGlScott: I can see the appeal behind doing this, but this change actually makes some of the source positions _less_ descriptive, as seen in [this failing CI log](https://github.com/GaloisInc/crucible/actions/runs/15784102682/job/44496585861?pr=1438#step:14:1596):

```
      checked_mul:                 FAIL (2.41s)
        MISMATCH for expected (test/symb_eval/num/checked_mul.good)
                   and actual (test/symb_eval/num/checked_mul.out) output:
            F        |test checked_mul/<DISAMB>::crux_test[0]: returned 44, FAILED
            F        |
            F        |failures:
            F        |
            F        |---- checked_mul/<DISAMB>::crux_test[0] counterexamples ----
            F        |[Crux] Found counterexample for verification goal
            F-expect>|[Crux]   test/symb_eval/num/checked_mul.rs:6:5: 6:12: error: in checked_mul/<DISAMB>::crux_test[0]
            F-actual>|[Crux]   test/symb_eval/num/checked_mul.rs:6:5: error: in checked_mul/<DISAMB>::crux_test[0]
            F        |[Crux]   attempt to compute `const 100_u8 * move _1`, which would overflow
            F        |
            F        |[Crux] Overall status: Invalid.
```

Note that the previous error message (the `F-expect` line) displayed an entire source span (`test/symb_eval/num/checked_mul.rs:6:5: 6:12:`), but after these changes (the `F-actual` line), it only displays the first part of the span (`test/symb_eval/num/checked_mul.rs:6:5:`). This feels like a regression in error message quality.

-----

As far as I understand it, the entire reason for making this change is so that `crux`'s [`jsLoc`](https://github.com/GaloisInc/crucible/blob/b32773c7075edfe9ec0223a43ee6ea518ee51a52/crux/src/Crux/UI/JS.hs#L13-L26) function will successfully be able to parse it. As such, I would propose to leave `crucible-mir`'s `parsePosition` alone for now and instead change `jsLoc` to parse `mir-json`-style source spans into a JS object by only taking the first part of the source span. It is somewhat questionable to drop the last part of the source span, but I'd feel more comfortable doing it in `jsLoc` (which is limited to one very specific `crux` feature) rather than in `parsePosition` (which impacts all `crux-mir` error messages).

RyanGlScott:
> Not sure why the test are failing - probably I will need to re-generate the expected outputs?

I think you may be referring to changes like [this one](https://github.com/GaloisInc/crucible/actions/runs/15784102682/job/44496585861?pr=1438#step:14:1930) in the coverage-related parts of `crux-mir`'s test suite:

```
       coverage_try:                FAIL (2.58s)
        MISMATCH for expected (test/coverage/coverage_try.good)
                   and actual (test/coverage/coverage_try.out) output:
            F-expect>|warning: branch condition never has value 1
            F-expect>|   ┌─ test/coverage/coverage_try.rs:10:17
            F-expect>|   │
            F-expect>|10 │     let y: u8 = y_32.try_into().ok()?;
            F-expect>|   │                 ^^^^^^^^^^^^^^^^^^^^^
            F-expect>|
```

Indeed, the diff shows that after this PR, this test case no longer produces any coverage results at all, which is very unexpected. I suspect that one of your changes accidentally changed the JSON schema of something in a way that the coverage script is sensitive to, although I'm not sure off hand what change would have caused this. Perhaps we should address the code review suggestions above first and then see if the issue persists.

crux/src/Crux/UI/JS.hs
podhrmic: To improve the readability of the printed output - printing a *formatted* JSON is easier to understand, especially if you have more fields showing signed/unsigned/decimal values.

crucible-mir/src/Mir/Trans.hs
podhrmic: Fixed - I modified `jsLoc` as suggested. Maybe not the prettiest code but it seems to function as expected.

crux/src/Crux/Model.hs
podhrmic: Fixed

podhrmic:
The output from `example-2` is now:

```
---- example_1/9ba11644::crux_test[0]::check_estimate_size[0] counterexamples ----
[Crux]   src/lib.rs:16:29: 16:47: error: in example_1/9ba11644::estimate_size[0]
[Crux]   attempted read on the result of an integer-to-pointer cast
Model:
[
    {
        "bits": "32",
        "loc": {
            "col": "58",
            "file": "/Users/mpodhradsky/Documents/Rust/crucible/crux-mir/examples/example-2/libs/crucible/symbolic.rs",
            "line": "24"
        },
        "name": "x",
        "val": "0x3ff",
        "val-decimal": "1023",
        "val-unsigned": "0x3ff"
    }
]
```

which is much easier to read, but the location in the *Model* section is still wrong (or at least not what the user would expect)

crucible-mir/src/Mir/Trans.hs
```diff
@@ -113,12 +113,13 @@ parsePosition :: Text.Text -> PL.Position
...
     [fname,line,col]
-      | (l,[]):_ <- readDec (Text.unpack line)
-      , (c,[]):_ <- readDec (Text.unpack col)
+      | (l,[]):_ <- readDec (Text.unpack (Text.strip line))
+      , (c,[]):_ <- readDec (Text.unpack (Text.strip col))
       -> PL.SourcePos fname l c
     _ -> PL.OtherPos posText
 
 
+
```
RyanGlScott: Please revert these changes—my understanding is that they're no longer necessary, and they only serve to clutter up the diff.

crux/src/Crux/UI/JS.hs
```diff
@@ -3,33 +3,50 @@
...
-           , "col"  ~> jsStr (show c)
-           ]
+    SourcePos fname l c -> parsePos fname l c
+    OtherPos s -> case Text.split (==':') s of
+          fname : line : col : _rest
+            | (l,[]):_ <- readDec (Text.unpack (Text.strip line))
+            , (c,[]):_ <- readDec (Text.unpack (Text.strip col))
+            -> parsePos fname l c
+          _ -> pure Nothing
     _ -> pure Nothing
```
RyanGlScott: Style nit pick: I would prefer using pattern guards here to avoid needing to duplicate the `pure Nothing` fall-through case:

```suggestion
    OtherPos s
      | fname : line : col : _rest <- Text.split (==':') s
      , (l,[]):_ <- readDec (Text.unpack (Text.strip line))
      , (c,[]):_ <- readDec (Text.unpack (Text.strip col)) ->
        parsePos fname l c
    _ -> pure Nothing
```

crux/src/Crux/UI/JS.hs
```diff
@@ -3,33 +3,50 @@
...
-         fabsolute <-
-            if | null fstr -> pure ""
-               | otherwise -> canonicalizePath fstr
-         pure $ Just $ jsObj
-           [ "file" ~> jsStr fabsolute
-           , "line" ~> jsStr (show l)
-           , "col"  ~> jsStr (show c)
-           ]
+    SourcePos fname l c -> parsePos fname l c
+    OtherPos s -> case Text.split (==':') s of
```
RyanGlScott: This `OtherPos` case deserves some comments. In particular, please note the following:

* This case is necessary because of the particular shape of source spans that arise from `mir-json` (e.g., `test/symb_eval/num/checked_mul.rs:6:5: 6:12:`), which will always be represented as `OtherPos`.
* While `crux` doesn't have the machinery to represent the entire source span in its UI framework, we can still achieve partial results by parsing the first location from the source span (e.g., the `test/symb_eval/num/checked_mul.rs:6:5:` bit). This does not show the entire span, but it is still better than nothing.

crux/src/Crux/UI/JS.hs
RyanGlScott: I am leery about this change, as pretty-printing the JSON has the unfortunate side effect of making the generated code have a larger file size. This is one of the reasons why we opt not to generate pretty-printed JSON files in `mir-json`, as doing so would increase the generated code size by several megabytes in the typical case.

My preference would be for users to pretty-print the JSON file themselves using a tool like `jq`.

crux/src/Crux/Model.hs
RyanGlScott: `showEnt''` still needs to be renamed.

crux/src/Crux/Model.hs
```diff
@@ -34,14 +34,25 @@ emptyModelView = ModelView $ MapF.empty
...
 
-
-showBVLiteral :: (1 <= w) => NatRepr w -> BV w -> String
-showBVLiteral w bv =
+showBVLiteralSigned :: (1 <= w) => NatRepr w -> BV w -> String
+showBVLiteralSigned w bv =
     (if x < 0 then "-0x" else "0x") ++ N.showHex i (if natValue w == 64 then "L" else "")
   where
   x = BV.asSigned w bv
   i = abs x
```
RyanGlScott, line 42: Please add a section to the `crux` changelog saying that `showBVLiteral` has been renamed to `showBVLiteralSigned`. (This is a change to the public API, so it should be documented as such.)

crux/src/Crux/Model.hs
podhrmic: My bad, fixed now!

crux/src/Crux/Model.hs
podhrmic: Done.

crux/src/Crux/UI/JS.hs
podhrmic: Done

crucible-mir/src/Mir/Trans.hs
podhrmic: Done

podhrmic:
> My preference would be for users to pretty-print the JSON file themselves using a tool like `jq`.

Should I call `jq` as a system call from `prettyPrintJson` instead of using `encodePretty`? Then we need to make sure users install `jq` so you have another system dependency.

The use case is to display the generated counter examples in the terminal - compare the current state:

```
$ cargo crux-test --lib --  -m midpoint_overflow
...
Model:
[{"name": "a","loc": null,"val": "0xa730e7fe","bits": "32"},{"name": "b","loc": null,"val": "0x58cf1803","bits": "32"}]
```


with the proposed pretty-printed state:

```
$ cargo crux-test --lib --  -m midpoint_overflow
...
Model:
[
    {
        "name": "a",
        "loc": null,
        "val": "0xa730e7fe",
        "bits": "32"
    },
    {
        "name": "b",
        "loc": null,
        "val": "0x58cf1803",
        "bits": "32"
    }
]
```

crux/src/Crux/UI/JS.hs
```diff
@@ -3,33 +3,61 @@
...
-         fabsolute <-
-            if | null fstr -> pure ""
-               | otherwise -> canonicalizePath fstr
-         pure $ Just $ jsObj
-           [ "file" ~> jsStr fabsolute
-           , "line" ~> jsStr (show l)
-           , "col"  ~> jsStr (show c)
-           ]
+    SourcePos fname l c -> parsePos fname l c
+    -- |Attempt to parse `OtherPos` in case it is in fact a code span:
```
RyanGlScott: No need to use Haddock formatting here, since that only works for top-level functions anyway:

```suggestion
    -- Attempt to parse `OtherPos` in case it is in fact a code span:
```

crux/src/Crux/UI/JS.hs
```diff
@@ -3,33 +3,61 @@
...
+    --   * This case is necessary because of the particular shape of source
+    --   spans that arise from `mir-json`
+    --   (e.g., `test/symb_eval/num/checked_mul.rs:6:5: 6:12:`), which will
+    --   always be represented as `OtherPos`.
+    --   * While `crux` doesn't have the machinery to represent the entire
+    --   source span in its UI framework, we can still achieve partial results
+    --   by parsing the first location from the source span
+    --   (e.g., `the test/symb_eval/num/checked_mul.rs:6:5:` bit). This does
+    --   not show the entire span, but it is still better than nothing.
```
RyanGlScott: Slightly more readable:

```suggestion
    -- Attempt to parse `OtherPos` in case it is in fact a code span:
    --
    -- * This case is necessary because of the particular shape of source
    --   spans that arise from `mir-json`
    --   (e.g., `test/symb_eval/num/checked_mul.rs:6:5: 6:12:`), which will
    --   always be represented as `OtherPos`.
    -- * While `crux` doesn't have the machinery to represent the entire
    --   source span in its UI framework, we can still achieve partial results
    --   by parsing the first location from the source span
    --   (e.g., `the test/symb_eval/num/checked_mul.rs:6:5:` bit). This does
    --   not show the entire span, but it is still better than nothing.
```

crux/src/Crux/UI/JS.hs
```diff
@@ -3,33 +3,61 @@
...
 
+import Data.Aeson.Encode.Pretty (encodePretty)
+import Data.Aeson (decode, Value)
+import qualified Data.ByteString.Lazy.Char8 as BL
+
+
 import What4.ProgramLoc
 
+-- |`jsLoc` takes a program location, and optionally returns it
+-- as `JS String`
```
RyanGlScott: Haddock formatting nit picks:

```suggestion
-- | 'jsLoc' takes a program location, and optionally returns it
-- as @JS String@.
```

(In Haddock, `@...@` delimits text that should be rendered in monospace font, and `'...'` is a stronger version of `@...@` that attempts to link to a Haskell identifier with a hyperlink.)

crux/src/Crux/UI/JS.hs
```diff
@@ -3,33 +3,61 @@
...
 import System.Directory( canonicalizePath )
 
+import Data.Aeson.Encode.Pretty (encodePretty)
+import Data.Aeson (decode, Value)
+import qualified Data.ByteString.Lazy.Char8 as BL
+
+
 import What4.ProgramLoc
 
+-- |`jsLoc` takes a program location, and optionally returns it
```
RyanGlScott: "Optionally returns" is a strange way to describe this, as this is a function that can potentially fail (by returning `Nothing`). I would phrase this as:

> `jsLoc` takes a program location and renders it as a JavaScript string. This returns `Nothing` if it is unclear how to render the program location.

RyanGlScott:
> Should I call `jq` as a system call from `prettyPrintJson` instead of using `encodePretty`?

No! I would prefer not pretty-printing this at all. If a user really wants to pretty-print the JSON, they can do so themselves using `jq`.

crux/src/Crux/UI/JS.hs
podhrmic: Done

podhrmic:
@RyanGlScott for size comparison:

* `crux-mir` binary from the latest `master` CI build has `65.3 MB`
* `crux-mir` binary from an earlier `1434-better-counterexamples` CI build (with the prettyPrinter) has `65.5 MB`

Building locally I get binaries that are `94.3 MB` vs `95.5MB` - probably the local build skips some optimizations.

Am I missing something? Because if these numbers are accurate, the JSON pretty printer adds less than 1MB of binary size for a significantly better user experience.

crux/src/Crux/UI/JS.hs
podhrmic: Done

crux/src/Crux/UI/JS.hs
podhrmic: Done. Should have batch committed your suggestions *sigh*

RyanGlScott:
My concern isn't about the size of the `crux-mir` binary, but rather the size of the generated JSON files. To illustrate this point a bit more concretely, here is a MIR JSON file I have lying around on my machine:

```
$ ls -alh ./target/aarch64-apple-darwin/debug/deps/cargo_sandbox-a47b807e6ec52f4e.linked-mir.json
-rw-r--r--@ 1 rscott  staff   4.7M Jun 17 19:42 ./target/aarch64-apple-darwin/debug/deps/cargo_sandbox-a47b807e6ec52f4e.linked-mir.json
```

This compact (i.e., not pretty-printed) JSON file is 4.7 megabytes. If I pretty-print it, the file size jumps to 13 megabytes. Similar considerations would apply here for the generated `translation.json` files.

podhrmic:
I see - we are talking about two different things then! All I care about is pretty printing the JSON in the *console output* that is printed with the `-m` argument (e.g. `cargo crux-test --lib -- -m`). *Not* the generated JSON files. Is `jsLoc` used in both cases? I was under the impression that it is used only for the `-m` output, but perhaps that is not the case?

RyanGlScott:
Yes, the JS code is used for both cases.

> All I care about is pretty printing the JSON in the _console output_ that is printed with the `-m` argument (e.g. `cargo crux-test --lib -- -m`).

I would prefer to solve this problem a different way: namely, we shouldn't be printing raw JSON in human-readable output, as this information is difficult to grok if you aren't a Crux developer. We should instead figure out how to print the model in a prettier way that neatly displays the results.

This is perhaps a larger task than what you set out to do here, however, and I would be happy to defer this to an issue. I do think that adding a dependency on a new library (`aeson-pretty`) is the wrong way to solve this problem.

crux/src/Crux/UI/JS.hs
podhrmic: I reverted 7a39ef0e56f7d2789075d46cc8fbd116a0f19ef5 so that should address your concern!

podhrmic:
Agreed, this one is ready to go then.

RyanGlScott (approved):
Thank you for working on this!

Issue #1222, opened 2024-07-16
`crucible-mir` crashes due to improperly constructing vtable shim types


Torrencem:
I've installed saw and `mir-json` on my system, and I think I have the right version of everything set up.

```
$ saw --version
1.1
$ saw-rustc --version
test build - extract output path - ["rustc", "--version", "--test", "--target", "aarch64-apple-darwin", "--cfg", "crux", "-L", "/Users/sarah.spall/projects/provers/crucible/crux-mir/rlibs", "--sysroot", "/Users/sarah.spall/projects/provers/crucible/crux-mir/rlibs"]
rustc 1.69.0-nightly (5e37043d6 2023-01-22)
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/bin/mir-json-rustc-wrapper.rs:86:27
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

(kind of weird version is panicking but you can see the rustc version is correct in there)

When I try to compile the following:

```
fn add(x: u32, y: u32) -> u32 {
x + y
}
```

it works fine, I'm able to handle it with the following saw script:

```
enable_experimental;

m <- mir_load_module "hello.linked-mir.json";
```

then I run `saw-rustc`, which succeeds

```
$ saw-rustc hello.rs
test build - extract output path - ["rustc", "hello.rs", "--test", "--target", "aarch64-apple-darwin", "--cfg", "crux", "-L", "/Users/sarah.spall/projects/provers/crucible/crux-mir/rlibs", "--sysroot", "/Users/sarah.spall/projects/provers/crucible/crux-mir/rlibs"]
test build - ["rustc", "hello.rs", "--target", "aarch64-apple-darwin", "--cfg", "crux", "-L", "/Users/sarah.spall/projects/provers/crucible/crux-mir/rlibs", "--sysroot", "/Users/sarah.spall/projects/provers/crucible/crux-mir/rlibs", "--cfg", "crux_top_level", "--crate-type", "rlib"]
warning: function `add` is never used
 --> hello.rs:1:4
  |
1 | fn add(x: u32, y: u32) -> u32 {
  |    ^^^
  |
  = note: `#[warn(dead_code)]` on by default

note: Emitting MIR for hello/a2764e5c::add

warning: 1 warning emitted

linking 19 mir files into hello.linked-mir.json
[...]

PR #1444, opened 2025-06-25
`crucible-mir`: Print vtables in `Pretty Collection` instance

### LINKED ISSUE #1443: `crux-mir`: Print vtables with `--print-mir`
Currently, `crux-mir`'s `--print-mir` flag does not print out any vtable definitions in its output. I would find this useful, as there are certain cases where `mir-json` does not eliminate the definition of a function as dead code because it is stored in a vtable.

RyanGlScott:
This causes `crux-mir`'s `--print-mir` flag to print vtable information when displaying a MIR JSON collection. Example output from the program in https://github.com/GaloisInc/crucible/issues/1222#issuecomment-2233313541:

```
VTABLEs
vtable core/c7248340::ops[0]::function[0]::Fn[0]::_vtblefc1a1550237cdbd[0] {
   ( core/c7248340::ops[0]::function[0]::FnOnce[0]::call_once[0]::_callonce579d6f58bdd54617[0]
   , core/c7248340::ops[0]::function[0]::FnOnce[0]::call_once[0] );
   ( test/ab048184::bar[0]::{closure#0}[0]::_inst485a3c886f64d386[0]
   , test/ab048184::bar[0]::{closure#0}[0] );
   ( test/ab048184::bar[0]::{closure#0}[0]::_inst485a3c886f64d386[0]
   , test/ab048184::bar[0]::{closure#0}[0] );
}
```

Fixes #1443.

### DIFF
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -310,6 +310,15 @@ instance Pretty BinOp where
       Unchecked op' -> pretty op' <> pretty "!"
       WithOverflow op' -> pretty op' <> pretty "?"
 
+instance Pretty VtableItem where
+  pretty (VtableItem fn def) = tupled [pretty fn, pretty def] <> semi
+
+instance Pretty Vtable where
+  pretty (Vtable name items) =
+    vcat [pretty "vtable" <+> pretty name <+> lbrace ,
+          indent 3 (vcat (map pretty items)),
+          rbrace]
+
 instance Pretty CastKind where
     pretty = viaShow
 
@@ -415,6 +424,8 @@ instance Pretty Collection where
           map pretty (Map.elems (col^.adts)) ++
           [pretty "TRAITs"] ++
           map pretty (Map.elems (col^.traits)) ++
+          [pretty "VTABLEs"] ++
+          map pretty (Map.elems (col^.vtables)) ++
           [pretty "INTRINSICSs"] ++
           map pretty (Map.elems (col^.intrinsics)) ++
           [pretty "STATICS"] ++
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -311,7 +311,11 @@ instance Pretty BinOp where
       WithOverflow op' -> pretty op' <> pretty "?"
 
 instance Pretty VtableItem where
-  pretty (VtableItem fn def) = tupled [pretty fn, pretty def] <> semi
+  pretty (VtableItem fn def) =
+    vcat
+      [ pretty "(" <+> pretty fn
+      , pretty "," <+> pretty def <+> pretty ");"
+      ]
 
 instance Pretty Vtable where
   pretty (Vtable name items) =
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -313,8 +313,8 @@ instance Pretty BinOp where
 instance Pretty VtableItem where
   pretty (VtableItem fn def) =
     vcat
-      [ pretty "(" <+> pretty fn
-      , pretty "," <+> pretty def <+> pretty ");"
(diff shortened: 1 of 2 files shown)


crucible-mir/src/Mir/PP.hs
```diff
@@ -310,6 +310,15 @@ instance Pretty BinOp where
       Unchecked op' -> pretty op' <> pretty "!"
       WithOverflow op' -> pretty op' <> pretty "?"
 
+instance Pretty VtableItem where
+  pretty (VtableItem fn def) = tupled [pretty fn, pretty def] <> semi
```
RyanGlScott: Suggestions welcome on how to more neatly print this. I couldn't easily find reference output for how Rust vtables are conventionally displayed, so I more-or-less improvised things here.

samcowger: I'm comfortable with how this looks, though I may not be the best person to judge.

I do think the in-tuple newlines in particular are nice to have, since they make it easier to see a short difference between two elements that share a long prefix, as in your example. Will `tupled` always produce newlines like that, do you know?

RyanGlScott: `tupled` doesn't always insert newlines, no. It attempts to lay out all the elements on one line, but it will insert a line break if it doesn't fit the width the page (usually 80 characters when using `prettyprinter`'s default settings). I could try to define this in terms of `vcat` instead to ensure that the two `DefId`s in each entry are always stacked on top of each other vertically, however.

crucible-mir/src/Mir/PP.hs
```diff
@@ -310,6 +310,19 @@ instance Pretty BinOp where
       Unchecked op' -> pretty op' <> pretty "!"
       WithOverflow op' -> pretty op' <> pretty "?"
 
+instance Pretty VtableItem where
+  pretty (VtableItem fn def) =
+    vcat
+      [ pretty "(" <+> pretty fn
+      , pretty "," <+> pretty def <+> pretty ");"
+      ]
```
spernsteiner: I would flip these around, and maybe change the syntax to make it clearer that this is a key-value mapping.  `def` is the key and `fn` is the value, at least based on my interpretation of the `VtableItem` type:
```Haskell
data VtableItem = VtableItem
    { _vtFn :: DefId        -- ^ ID of the implementation that should be stored in the vtable
    , _vtDef :: DefId       -- ^ ID of the item definition in the trait
    }
```

spernsteiner:
```
VTABLEs
vtable core/c7248340::ops[0]::function[0]::Fn[0]::_vtblefc1a1550237cdbd[0] {
   ( core/c7248340::ops[0]::function[0]::FnOnce[0]::call_once[0]::_callonce579d6f58bdd54617[0]
   , core/c7248340::ops[0]::function[0]::FnOnce[0]::call_once[0] );
   ( test/ab048184::bar[0]::{closure#0}[0]::_inst485a3c886f64d386[0]
   , test/ab048184::bar[0]::{closure#0}[0] );
   ( test/ab048184::bar[0]::{closure#0}[0]::_inst485a3c886f64d386[0]
   , test/ab048184::bar[0]::{closure#0}[0] );
}
```

That's some very strange behavior from mir-json.  The key `test::bar::{closure#0}` appears twice (fortunately with the same value both times).  But `test::bar::{closure#0}` isn't a member of the `Fn` trait, so I wouldn't expect it to appear here at all.

RyanGlScott:
> That's some very strange behavior from mir-json.

And yet, that's what it tells me... should I open a `mir-json` issue about this?

spernsteiner:
Opened GaloisInc/mir-json#127

crucible-mir/src/Mir/PP.hs
RyanGlScott: I've flipped these around and replaced the comma with `=>` to make it clearer that this is a key-value mapping.

spernsteiner (approved):

PR #1445, opened 2025-06-25
crucible-mir: support custom dynamically-sized types

### LINKED ISSUE #1399: `crucible-mir`: Support custom dynamically sized types
We should add support for custom dynamically sized types, which @spernsteiner describes as:

`struct S<T: ?Sized> { x: i32, y: T }`, then cast `&S<X>` to `&S<dyn Trait>` or `&S<[Y; N]>` to `&S<[Y]>`

samcowger:
Fixes #1399.

This implements three main forms of DST interaction:
- Introduction via reference/pointer cast, as in e.g. `&S<Concrete> as &S<dyn Trait>`
- Dereferencing
- Field access, for both normally-sized fields and unsized/dynamically-sized fields

This is in draft form for several reasons:
- Two of the tests I've added, `dyn/custom_dst.rs` and `dyn/arc.rs`, currently fail, because of the `mir-json` limitation mentioned in https://github.com/GaloisInc/crucible/issues/1399#issuecomment-3002103350. I'll make the necessary `mir-json` changes and update its submodule, which I hope will let them pass.
- I'm seeing strange behavior with two existing tests, `cell/slice.rs` and `cell/slice_from_mut.rs`, both of which are expected to fail. The former still fails, but I'd expect my changes to make it pass. The latter passes, but I don't yet understand why my changes impacted it.

I'll be looking into those issues, but in the meantime, I think this is largely ready to be reviewed.


crucible-mir/src/Mir/Generator.hs
```diff
@@ -657,6 +657,20 @@ subfieldRef ::
   MirGenerator h s ret (R.Expr MIR s MirReferenceType)
 subfieldRef ctx ref idx = G.extensionStmt (MirSubfieldRef ctx ref idx)
 
+-- | Extend the reference path encapsulated in the provided `Expr MIR s
+-- MirReferenceType` (i.e. `ref`), which should terminate in a struct, to
```
RyanGlScott: Haddock nit pick: surrounding text with backticks (or single quotes) in Haddock (`` `...` ``) is meant for hyperlinking an individual identifier. I think you want to use `@...@` here instead, which simply renders the text in monospace font:

```suggestion
-- | Extend the reference path encapsulated in the provided @Expr MIR s
-- MirReferenceType@ (i.e. @ref@), which should terminate in a struct, to
```

And similarly for other uses of backticks in this PR.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -861,6 +861,13 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
      !(f MirReferenceType) ->
      !(Index ctx tp) ->
      MirStmt f MirReferenceType
+  -- | Like `MirSubfieldRef`, but for fields with statically-unknown types, such
+  -- as trait objects.
+  MirSubfieldRef_Untyped ::
+     !(f MirReferenceType) ->
+     !Int ->
+     !(Maybe (Some TypeRepr)) ->
```
RyanGlScott, line 872: Instead of using `Some TypeRepr` here, why not use `TypeRepr tp`, where `tp` is existentially quantified? Better yet, it might help to give `tp` a more descriptive name that indicates what its purpose is in `MirSubfieldRef_Untyped`.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -861,6 +861,13 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
      !(f MirReferenceType) ->
      !(Index ctx tp) ->
      MirStmt f MirReferenceType
+  -- | Like `MirSubfieldRef`, but for fields with statically-unknown types, such
+  -- as trait objects.
```
RyanGlScott: Please comment on what the `Int` and `TypeRepr` fields represent.

crucible-mir/src/Mir/Generator.hs
```diff
@@ -657,6 +657,20 @@ subfieldRef ::
...
+-- terminate at the `fieldNum`th field of that struct.
```
RyanGlScott: Also mention what `expectedTy`'s role is here.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -944,6 +959,67 @@ evalCast' ck ty1 e ty2  = do
       _ -> mirFail $ "unimplemented cast: " ++ (show ck) ++
         "\n  ty: " ++ (show ty1) ++ "\n  as: " ++ (show ty2)
   where
+    -- | Perform an unsizing cast from a reference to a struct containing an
```
RyanGlScott: Did you mean:

```suggestion
    -- | Perform an unsized cast from a reference to a struct containing an
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -659,6 +722,38 @@ checkFieldKind True tpr tpr' desc = do
...
+-- per `adtSizedness`.
+tySizedness :: HasCallStack => M.Collection -> M.Ty -> Sizedness
+tySizedness col ty =
+  case ty of
+    M.TyDynamic _ -> Unsized
+    M.TySlice _ -> Unsized
+    M.TyStr -> Unsized
+    M.TyAdt adtName _ _ ->
+      case col ^? M.adts . ix adtName of
+        Nothing -> error $ "unknown ADT: "<>show adtName
```
RyanGlScott: Perhaps:

```suggestion
        Nothing -> error $ "tySizedness: unknown ADT: "<>show adtName
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -669,41 +764,77 @@ structInfo adt i = do
...
+            mirFail "encountered unsized field in non-last position"
+        Unsized ->
+          case fldTy of
+            M.TySlice innerTy -> do
+              Some innerRepr <- tyToReprM innerTy
+              pure $ UnsizedSliceField innerRepr
+            M.TyStr -> do
+              Some innerRepr <- tyToReprM (M.TyUint M.B8)
+              pure $ UnsizedSliceField innerRepr
+            _ -> pure UnsizedNonSliceField
```
RyanGlScott: Style nit pick: use an `otherwise` guard here so that it's more obvious at a glance that this is all part of the `Unsized` case:

```suggestion
        Unsized
          | i /= length (var ^. M.vfields) - 1 ->
            mirFail "encountered unsized field in non-last position"
          | otherwise ->
            case fldTy of
              M.TySlice innerTy -> do
                Some innerRepr <- tyToReprM innerTy
                pure $ UnsizedSliceField innerRepr
              M.TyStr -> do
                Some innerRepr <- tyToReprM (M.TyUint M.B8)
                pure $ UnsizedSliceField innerRepr
              _ -> pure UnsizedNonSliceField
```

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Doesn't that require you to choose a `tp` even when the `Maybe` is `Nothing`?

crucible-mir/src/Mir/Trans.hs
```diff
@@ -944,6 +959,67 @@ evalCast' ck ty1 e ty2  = do
...
+      -- Sanity-check that we are indeed casting to a type with a slice in the
+      -- expected position
+      let ref = MirExp MirSliceRepr (mkSlice adtRef lenExpr)
+      case findLastFieldRec col an2 of
+        Nothing ->
+          mirFail $ "unsizedAdtSlice: unable to determine last field of "<>show an2
+        Just field ->
+          case field ^. M.fty of
+            M.TySlice _elemTy -> pure ref
+            M.TyStr -> pure ref
```
spernsteiner: No types can `Unsize` to `str`.  See the docs for the [`Unsize` trait](https://doc.rust-lang.org/stable/core/marker/trait.Unsize.html) - there are no cases that implement `Unsize<str>`.  The only way to get `&str` is through library functions that unsize to `&[u8]` and then transmute `&[u8]` to `&str`.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1315,7 +1395,59 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
 
-evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) (M.PField idx _mirTy) = do
+    -- For (dynamically-sized) ADTs wrapping slices, with MirPlace input of the
+    -- form:
+    --
+    --   MirPlace (*S<[T]>) <expr: **S<[T; len]> NoMeta
+    --
+    -- We produce output of the form:
+    --
+    --   MirPlace S<[T]> <expr: *S<[T; len]> (SliceMeta len)
```
spernsteiner: I think the input expr is actually `<expr: **S<[T]>>` (consistent with the normal `NoMeta` case, which looks like `MirPlace U <expr: *U> NoMeta` for some `U`).  From that pointer, we load a value of type `*S<[T]>` (`MirSliceRepr`), which is a pair of `<expr: *S<[T; len]>>` and `len`, giving us the two parts that appear in the output `MirPlace`.

spernsteiner: Looking at the output `MirPlace` here, it seems like we're accumulating more and more cases where the `MirPlace`'s `TypeRepr` field doesn't correspond to the actual pointee type.  I wonder if we should try to get rid of it at some point.  Historically we needed a `TypeRepr tp` to go with the `MirReferenceType tp` for the pointer field, but nowadays the pointer is just `MirReferenceType`.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1315,7 +1395,59 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
+        return $ MirPlace adtRepr adtPtr (SliceMeta sliceLen)
+
+    -- For (dynamically-sized) ADTs wrapping trait objects, with MirPlace input
+    -- of the form:
+    --
+    --   MirPlace (*S<dyn Trait>) <expr: **S<Concrete> NoMeta
+    --
+    -- We produce output of the form:
+    --
+    --   MirPlace S<dyn Trait> <expr: *S<Concrete> (DynMeta vtable)
```
spernsteiner: Input is `<expr: **S<dyn Trait>>`

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1315,7 +1395,59 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
+    --   MirPlace S<[T]> <expr: *S<[T; len]> (SliceMeta len)
+    --
+    -- Where `len` is the wrapped slice's length.
+    doSliceAdt :: M.Ty -> R.Expr MIR s MirReferenceType -> MirGenerator h s ret (MirPlace s)
+    doSliceAdt adtTy ref' =
+      do
+        -- Normally we'd use `tyToReprM adtTy` to get the ADT's representation,
+        -- but that would involve computing its slice field's representation,
+        -- which should error
+        let adtRepr = R.exprType ref'
```
spernsteiner: If `R.exprType` gives the `CrucibleType` of an `R.Expr`, then this will always return `MirReferenceType`, which is definitely not the repr of the ADT.  But (as mentioned above) the `MirPlace` `TypeRepr` doesn't make a lot of sense here, so maybe we can get away with using `AnyType` or something as a placeholder.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: I tried this at first, but ran into some errors in the `TestEqualityFC` and `OrdFC` instances for `MirStmt` that I had trouble understanding. I've gone back and tried stubbing out those instances, though, and I don't see errors at calls that provide `Nothing`, so I could probably make that change, if I could work out those instance definitions.

Is one approach considered more idiomatic than the other when working with existentially-quantified types in this way?

RyanGlScott: I don't have a strong opinion on which is used. I only suggest avoiding `Some` since it's one extra layer of indirection, and I don't think it's necessary in this case.

crucible-mir/src/Mir/Generator.hs
samcowger: I did, thanks!

crucible-mir/src/Mir/Trans.hs
samcowger: Oh, yeah, I guess that's pretty manifestly incorrect. I'm a bit surprised that this doesn't seem to matter - my tests pass with this, or `AnyRepr`, or even `undefined`. Where/under what circumstances would that `TypeRepr` normally be examined?

spernsteiner: It's mainly used for load and store operations, which are only possible on sized types.

samcowger:
With the changes from https://github.com/GaloisInc/mir-json/pull/133, I now see the `dyn/custom_dst.rs` test pass. The also-previously-failing `dyn/arc.rs`, however, seems to enter an infinite loop at simulation time. If I run `crux-mir` on that file with `-d 5`, I see a lot of output, culminating in this stanza, which repeats itself:
```
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $1 = newMirRef (Struct Any (MirReference )) % internal
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $2 = newMirRef Unit % internal
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $3 = emptyApp() % internal
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $4 = writeMirRef $2 <- $3 % internal
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $5 = writeMirRef $1 <- $0 % internal
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  jump %0($1, $2) % internal
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $2 = readMirRef $0 % ./libs/core/src/ptr/mod.rs:523:1: 523:56
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $3 = handleLit(core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]) % ./libs/core/src/ptr/mod.rs:523:1: 523:56
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $4 = getStruct($2, 0, IntrinsicRepr MirReference []) % ./libs/core/src/ptr/mod.rs:523:1: 523:56
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $5 = getStruct($2, 1, AnyRepr) % ./libs/core/src/ptr/mod.rs:523:1: 523:56
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $6 = mkStruct([IntrinsicRepr MirReference [], AnyRepr], [$4, $5]) % ./libs/core/src/ptr/mod.rs:523:1: 523:56
core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]:
  $7 = call $3($6); % ./libs/core/src/ptr/mod.rs:523:1: 523:56
```

I'm not totally sure how to read/interpret this. One thought I had was that this was somehow entering the unconditionally-recursive implementation of [`core::ptr::drop_in_place`](https://github.com/GaloisInc/mir-json/blob/0de09fdefee7b69d05467b525b6301c81d63f4a8/libs/core/src/ptr/mod.rs#L523-L529) - but modifying `drop_in_place` to have an empty body didn't fix the problem.

Another thought is that during translation I'm somehow creating a cyclic reference structure that I then attempt to drop. Even then, though, I'd think I'd see more computation being repeated, because the `Drop` implementation for `Arc` that calls `drop_in_place` does a bit of work before calling it. I've also done some manual tracing of the code I've written to see how values are constructed in the lead-up to the `drop_in_place` call, but nothing seems obviously malformed.

My next move is probably to try to break down the computation involved in `Drop`ping `Arc`s, to hopefully get more clarity on what sort of value is being passed to `drop_in_place`. I'd welcome thoughts/suggestions about this or other approaches!

spernsteiner:
Try looking up `core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]` in the linked JSON - it should (eventually) map to an instance definition that gives the generic arguments (along with the original DefId, which should be `core/c7248340::ptr[0]::drop_in_place[0]`), and those will include what type it's trying to drop

samcowger:
> Try looking up `core/c7248340::ptr[0]::drop_in_place[0]::_drop1838a092de795f25[0]` in the linked JSON - it should (eventually) map to an instance definition that gives the generic arguments (along with the original DefId, which should be `core/c7248340::ptr[0]::drop_in_place[0]`), and those will include what type it's trying to drop

Ah, good idea. As expected/hoped, its parameter is a `*mut dyn Foo`.

Further, if I look at the MIR JSON definition for that version of `drop_in_place`, I see that one of its basic blocks has a `Drop` terminator, which says that the `*mut dyn Foo` value's drop function is the very same version of `drop_in_place`. I don't think I see this pattern elsewhere in the generated MIR - is that the proper way to specify a drop in MIR, or could that be the source of the problem?

spernsteiner:
It looks like we're relying on rustc to generate proper MIR for `drop_in_place` implementations, which it does in most cases.  For example, in `test/conc_eval/vec/drop.rs`, it generates a `drop_in_place::<[T]>` that iterates over the slice and drops each `T`.  But it seems like it generates the wrong thing for `drop_in_place::<dyn Trait>` - maybe it's relying on the backend to do the lowering.  What this code is supposed to do is get the function pointer for the drop glue from the `*mut dyn Trait` vtable and pass the data pointer to it.  Maybe we need to implement that logic ourselves either in mir-json or in crucible-mir.

samcowger:
I see, that's somewhat unfortunate. Two questions about that:
- Would it be appropriate/possible for `mir-json` to include a pointer to the appropriate destructor in the vtable it emits for a given trait object? Or does it lack access to the concrete type of a trait object while constructing its vtable?
- Does `crucible-mir` already do something like this - finding/substituting and executing drop glue - for other cases, or would that functionality need to be built from whole cloth, do you know?

spernsteiner:
>Would it be appropriate/possible for mir-json to include a pointer to the appropriate destructor in the vtable it emits for a given trait object? Or does it lack access to the concrete type of a trait object while constructing its vtable?

Yes, I think that's reasonable.  In actual rustc, the vtable contains a pointer to the `drop_in_place` function for the concrete type (and a couple other things) in addition to the function pointers for the user-visible methods.  mir-json's vtable generation has access to both the trait and the type.

>Does crucible-mir already do something like this - finding/substituting and executing drop glue - for other cases, or would that functionality need to be built from whole cloth, do you know?

IIRC mir-json's output for `TerminatorKind::Drop` includes the name of the `drop_in_place` function for the value being dropped, and crucible-mir basically just calls that function.  For code like `Arc::drop` that calls a `drop_in_place` function directly, mir-json outputs the MIR that rustc internally generates for each instantiation of `drop_in_place`.  This seems to work for just about everything other than `dyn Trait`.  (I tried looking for the code in rustc that actually generates these shims to get a better idea about what we should expect to work, but I couldn't find it.)

I think what we should do here is have mir-json include `drop_in_place` in the vtable, and add a `TransCustom` override for `drop_in_place::<dyn Trait>` that retrieves that function pointer from vtable and calls it.

samcowger:
Progress continues on this. I've just opened https://github.com/GaloisInc/mir-json/pull/138 to accommodate some strange Rust behavior with regard to drop glue for trait objects, which seems to be the cause of the apparently-recursive drop behavior I saw on my `dyn/arc.rs` test. This will allow me to write an override for `drop_in_place`, specific to trait objects, that fetches and calls the right drop function (which https://github.com/GaloisInc/mir-json/pull/138 will cause `mir-json` to include in the traits and vtables it emits).

crucible-mir/src/Mir/Trans.hs
samcowger: Noted. I've switched to `AnyRepr` and have mentioned your point in the comment.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Ryan and I spoke offline, and figured that the benefit of eliminating `Some` was roughly balanced by the cost of more complicated `{Ord,TestEquality}FC` instance implementations, so I'll leave this as-is.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1266,6 +1396,10 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
     doRef (M.TySlice ty') | MirSliceRepr <- tpr = doSlice ty' ref
     doRef M.TyStr | MirSliceRepr <- tpr = doSlice (M.TyUint M.B8) ref
     doRef (M.TyDynamic _) | DynRefRepr <- tpr = doDyn ref
+    doRef adtTy@(M.TyAdt _ _ _) | MirSliceRepr <- tpr = doSliceAdt adtTy ref
+    doRef adtTy@(M.TyAdt _ _ _) | DynRefRepr <- tpr = doDynAdt adtTy ref
+    -- TODO: do we need to go to the trouble here of figuring out if this ADT is
+    -- a DST, if it didn't come with a DST-indicative TypeRepr?
```
samcowger: Opinions welcome. It wouldn't be too hard to check, but I think that such a thing ought to be an error, and all that checking would get us would be a very slightly more useful error message than `doRef`'s fallback case.

RyanGlScott: How likely is this to happen? If it's unlikely, I'm inclined to not bother and just use the fallback error message.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -624,6 +630,49 @@ null_ptr_impl what substs = case substs of
...
+            Just $ CustomOp $ \argTys [op@(MirExp selfTy selfExpr)] -> do
+                col <- use $ cs . collection
+                let argTys = Ctx.empty
+                let argExprs = Ctx.empty
+                let retTy = C.UnitRepr
+                -- TODO: is it appropriate to check here that the vtable has at
+                -- least one method? `doVirtCall` (via `mkVirtCall`) will
+                -- indirectly perform this check, but we could emit a more
+                -- useful error message here, or perhaps avoid overriding at
+                -- all?
```
samcowger: Opinions welcome. The `box_send.rs` test in this PR is an example of code that would trigger this error and might benefit from an explicit vtable size check here. At the moment, running `crux-mir` on it yields output like:
```
failed to generate virtual-call shim for method 0 of trait trait/0::empty[0] method index out of range for vtable: method = 0 ; size = 0
  error, called at src/Mir/Trans.hs:2698:17 in crucible-mir-0.4.0.0.99-inplace:Mir.Trans
  mkVirtCall, called at src/Mir/Trans.hs:2753:23 in crucible-mir-0.4.0.0.99-inplace:Mir.Trans
  doVirtCall, called at src/Mir/TransCustom.hs:654:21 in crucible-mir-0.4.0.0.99-inplace:Mir.TransCustom
  ...
```

RyanGlScott: If it's cheap to implement, then I think it would be worthwhile to improve the error message here. The current one says that the method index is out of range, but doesn't really indicate _why_, which might be helpful for finding the relevant `mir-json` limitation.

samcowger: I suspect it will be cheap to implement - I'll give it a go.

samcowger:
While I address the above comment, I'm going to go ahead and promote this from draft status, because I think it's otherwise complete, and the tests I've added now pass (or, in one case, fail) as expected.

crucible-mir/src/Mir/Trans.hs
samcowger: I don't think it's particularly likely, so I share your inclination.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Well, I'd hoped that a stanza like this would cause the "no drop method available" message to show up prominently in the output, but it ends up subsumed by another error, and is only visible at higher debug levels:
```hs
() <- case col ^. traits . at traitName of
    Nothing -> mirFail $ "undefined trait: "<>show traitName
    Just trait ->
        case trait ^. traitItems of
            [] -> mirFail $ "no drop method available in trait "<>show traitName
            (_:_) -> pure ()
```
It does avoid a call to `error` from `mkVirtCall`, though, which allows translation to continue and defers failure until simulation time. Is that better than the current state of things, do you think?

RyanGlScott: I would definitely prefer calling `mirFail` over `error`, so I think that this is a slight improvement.

crucible-mir/src/Mir/Trans.hs
spernsteiner: I agree this is fine to leave as-is.  It may throw an error from `tyToReprM` below in the case of custom DSTs with a `extern type` (`TyForeign`) tail, but extern types are currently unsupported in lots of other parts of crux-mir too.

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: I agree `mirFail` is a better choice than `error` in basically all cases.

I was curious how the error was getting superseded, and it looks like mir-json is omitting the vtable for `Box<i32>` to `Box<dyn Send>`:
```
[Crux]   test/conc_eval/dyn/box_send.rs:10:28: 10:44: error: in box_send/8d578f4a::crux_test[0]
[Crux]   Translation error in box_send/8d578f4a::crux_test[0]: unimplemented cast: Unsize
[Crux]     ty: TyRawPtr (TyUint USize) Immut
[Crux]     as: TyRawPtr (TyDynamic trait/0::empty[0]) Immut
```

There's a [TODO comment](https://github.com/GaloisInc/mir-json/blob/dbdc8b75cffeea85b48738712a97f338c1151da0/src/analyz/mod.rs#L163) about this in mir-json.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1250,6 +1262,55 @@ subfieldMirRefLeaf ctx ref idx =
...
+            Some fieldIdx <-
+              case Ctx.intIndex fieldNum (Ctx.size structCtx) of
+                Just someIdx -> pure someIdx
+                Nothing ->
+                  bail $ unwords $
+                    [ "out-of-bounds field access: field"
+                    , show fieldNum
+                    , "of struct"
+                    , show structCtx
+                    ]
```
spernsteiner:
```suggestion
                Nothing -> bail $ unwords $ [ "out-of-bounds field access: field"
                    , show fieldNum , "of struct" , show structCtx ]
```
I normally flow error messages like this into a smaller number of lines so they don't take up too much vertical space and distract from the important logic.  (There are several other cases like this elsewhere in the PR.)

crucible-mir/src/Mir/Trans.hs
```diff
@@ -944,6 +960,120 @@ evalCast' ck ty1 e ty2  = do
...
+      col <- use $ cs . collection
+      eventualTraitObject <- case findLastFieldRec col castSource of
+        Nothing ->
+          mirFail $ "unsizedAdtDyn: unable to determine last field of "<>show castSource
+        Just field ->
+          pure (field ^. M.fty)
+      case tyToFieldRepr col eventualTraitObject of
+        Some (FieldRepr (FkMaybe _)) ->
+          mirFail "error: unsizeAdtDyn with non-initializable unsized field"
+        Some (FieldRepr (FkInit _)) ->
```
spernsteiner, line 1080: Please add a comment explaining why we distinguish `FkMaybe` from `FkInit` here.  From previous discussion I think the answer is that `FkMaybe` in the last field would require an additional step in the ref path when doing field projections, but we don't have a good way to detect whether or not that step is needed, and anyway we expect `FkMaybe` to be pretty rare these days.

crucible-mir/src/Mir/Trans.hs
samcowger: Alright, I've removed the TODO.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Yes, thanks for writing that down here. Do you think it would be worthwhile to include a case in `evalCast'` for `Unsize` casts from `&T`/`*T` to `&dyn`/`*dyn`, to give a more informative error message than this?

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1315,7 +1447,63 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
 
-evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) (M.PField idx _mirTy) = do
+    -- For (dynamically-sized) ADTs wrapping slices, with MirPlace input of the
+    -- form:
+    --
+    --   MirPlace (*S<[T]>) <expr: **S<[T]> NoMeta
+    --
+    -- We produce output of the form:
+    --
+    --   MirPlace S<[T]> <expr: *S<[T; len]> (SliceMeta len)
```
spernsteiner:
```suggestion
    --   MirPlace AnyRepr <expr: *S<[T; len]> (SliceMeta len)
```
To match the behavior of the implementation below

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1315,7 +1447,63 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
...
+        return $ MirPlace adtRepr adtPtr (SliceMeta sliceLen)
+
+    -- For (dynamically-sized) ADTs wrapping trait objects, with MirPlace input
+    -- of the form:
+    --
+    --   MirPlace (*S<dyn Trait>) <expr: **S<dyn Trait> NoMeta
+    --
+    -- We produce output of the form:
+    --
+    --   MirPlace S<dyn Trait> <expr: *S<Concrete> (DynMeta vtable)
```
spernsteiner:
```suggestion
    --   MirPlace AnyRepr <expr: *S<Concrete> (DynMeta vtable)
```

crucible-mir/src/Mir/TransCustom.hs
samcowger: I've added such a case.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -624,6 +630,54 @@ null_ptr_impl what substs = case substs of
...
+                col <- use $ cs . collection
+                let argTys = Ctx.empty
+                let argExprs = Ctx.empty
+                let retTy = C.UnitRepr
+
+                -- We expect @mir-json@ to have placed this trait object's drop
+                -- method at index 0, unless the trait object lacks a principal
+                -- trait. Check here whether the trait has any methods, to emit
+                -- a more relevant error message than `doVirtCall` would emit on
+                -- its own.
```
spernsteiner: Would be good to note that we consider the current mir-json behavior for `dyn Send` to be a bug/limitation - mir-json should emit a vtable with only a drop and no other methods, but that would take some work, so we haven't done it yet.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -624,6 +630,54 @@ null_ptr_impl what substs = case substs of
...
+                        col
+                        traitName
+                        dropMethodIndex
+                        selfTy
+                        selfExpr
+                        argTys
+                        argExprs
+                        retTy
+                pure (MirExp retTy callExpr)
+        Substs [_] -> Nothing
```
spernsteiner: Should note: for all other types, we return no `CustomOp`, so the `drop_in_place` implementation provided by rustc will be used instead.

crux-mir/test/conc_eval/box/dyn.rs
```diff
@@ -0,0 +1 @@
+../dyn/box.rs
```
spernsteiner: Is this a symlink?  Why run the same test twice?

spernsteiner (approved):
Looks good - only some minor notes on style / documentation.

Please also check whether you can remove the `mem::forget` call in the `fn_dyn_box_mut.rs` test case now that drop works for `Box<dyn Trait>`.

samcowger:
> Please also check whether you can remove the `mem::forget` call in the `fn_dyn_box_mut.rs` test case now that drop works for `Box<dyn Trait>`.

I've done so in 538c4b6a4c793e74d0572688955120f2b7b7c6cb, and the test still passes - see e.g. [the logs](https://github.com/GaloisInc/crucible/actions/runs/16154426590/job/45593479241#step:14:888) of the CI run two commits later for f3d89b630a0e2947de2a8f58389a1f810c2ddf06.

crux-mir/test/conc_eval/box/dyn.rs
samcowger: It is a symlink. I couldn't decide whether a test for `Box<dyn Trait>` belonged in `box/` or `dyn/`, so I split the difference and put it in both. I was thinking the boost in discoverability would outweigh the cost of duplication, but I can pick a single spot for it if you think that's more appropriate.

RyanGlScott: I don't think the subdirectory names are intended to be that prescriptive about containing every test case related to a particular language feature. I think it's fine to a single `dyn/box.rs` test case, especially since if you'd want to find this test case, you're probably going to search for `Box<dyn` anyway (and then it won't particularly matter where it lives).

samcowger: Alright, I've removed it.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Done in 01ec64b.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Done in 5d54bef.

crucible-mir/src/Mir/Trans.hs
samcowger: Done in 41485a6.

crucible-mir/src/Mir/Trans.hs
samcowger: Done in 41485a6.

crucible-mir/src/Mir/Trans.hs
samcowger: Done in 9f7fe0f, by pointing to an explanation in this function's caller's docstring.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Fair enough - I've found some opportunities to economize in 3b0de1e.

RyanGlScott (approved):

PR #1449, opened 2025-07-02
crucible-mir: allow `TyRawPtr` receivers in `transVtableShim`

samcowger:
The need for this change arose when trying to test `Box`es containing trait objects, as part of implementing support for custom DSTs in #1445.


samcowger:
> It might be worth factoring out the code shared in common between the `TyRef` and `TyRawPtr` cases, since it's somewhat involved. Other than that, LGTM.

Done!

Issue #1406, opened 2025-05-23
`llvm`: Finish migration of string-manipulating functions to their own module


langston-barrett:
https://github.com/GaloisInc/crucible/pull/1405 began a process of migrating some functions that manipulate C-style null-terminated strings out of the giant `MemModel` module into a new module. In particular, it began by re-exporting `Lang.Crucible.LLVM.MemModel.{loadString,loadMaybeString,strLen}` at `Lang.Crucible.LLVM.MemModel.Strings.{loadString,loadMaybeString,strLen}`, and noting that these functions should now be imported using the latter paths.

The next step is to officially deprecate the exports from `MemModel`. Such deprecation pragmas were provided in the linked PR, but they are currently commented out because they only became available with GHC 9.8. Hence, this is blocked on #1308.

Once the exports are deprecated, we should wait a while (one or two releases), then migrate the implementations of these functions to `Strings` and remove the exports from `MemModel`.

PR #1450, opened 2025-07-02
llvm: API for loading symbolic null-terminated strings from memory

### LINKED ISSUE #1061: `crucible-llvm`: Support loading mostly symbolic strings

langston-barrett:
Fixes #1061. It would be good to go further and provide a similarly flexible API for string-length calculations. I'll leave this for a follow-up.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
@@ -120,7 +120,9 @@ ptrToBv8 bak bytePtr = do
   let err = LCS.AssertFailureSimError "Found pointer instead of byte when loading string" ""
   Partial.ptrToBv bak err bytePtr
 
--- Currently unused, but analogous with 'Lang.Crucible.LLVM.MemModel.loadString'
+-- Currently unused, but analogous with
+-- 'Lang.Crucible.LLVM.MemModel.loadString'. In fact, it would be good to define
+-- that function in terms of this one. However, this is blocked on TODO(#1406).
 _fullyConcreteNullTerminatedString ::
   GHC.HasCallStack =>
   LCB.IsSymBackend sym bak =>


crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax/Overrides/String.hs
```diff
@@ -96,7 +92,6 @@ readCStringOverride memVar =
 --
 -- * The loaded string must be concrete. If a symbolic character is encountered
 --   while loading, this function will generate an assertion failure.
---   (TODO(#1061))
```
langston-barrett, line 99: I just removed these TODOs, because it's not actually clear to me how we would UTF-8-decode symbolic bytes.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -3,27 +3,261 @@
...
+
+-- | Load a null-terminated string (with a concrete null terminator) from memory.
+--
+-- The string must contain a concrete null terminator. If a maximum number of
+-- characters is provided, no more than that number of characters will be read.
+-- In either case, 'loadConcretelyNullTerminatedString' will stop reading if it
+-- encounters a (concretely) null terminator.
+--
+-- Note that the loaded string may actually be smaller than the returned list if
+-- any of the symbolic bytes may be 0.
```
RyanGlScott:
```suggestion
-- Note that the loaded string may actually be smaller than the returned list if
-- any of the symbolic bytes are equal to 0.
```

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -3,27 +3,261 @@
...
+
+-- | Load a null-terminated string from memory.
+--
+-- Consults an SMT solver to check if the loaded byte is known to be null (0).
+-- If a maximum number of characters is provided, no more than that number
+-- of charcters will be read. In either case, 'loadSymbolicString' will stop
+-- reading if it encounters a null terminator.
+--
+-- Note that the loaded string may actually be smaller than the returned list if
+-- any of the symbolic bytes may be 0.
```
RyanGlScott:
```suggestion
-- any of the symbolic bytes are equal to 0.
```

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -3,27 +3,261 @@
...
+  Mem.MemImpl sym ->
+  Mem.LLVMPtr sym wptr ->
+  -- | Maximum number of characters to read
+  Maybe Int ->
+  IO [WI.SymBV sym 8]
+loadConcretelyNullTerminatedString bak mem ptr limit =
+  case limit of
+    Nothing -> loadBytes bak mem id ptr concretelyNullTerminatedString
+    Just l ->
+      let byteChecker = withMaxChars l (pure . ($ [])) concretelyNullTerminatedString in
```
RyanGlScott: Style nit pick: the expression `pure . ($ [])` is a bit too point-free for my liking. I'd prefer eta-expanding it to `\f -> pure (f [])` to make it a bit clearer what is being applied where.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -3,27 +3,261 @@
...
+  LCB.IsSymBackend sym bak =>
+  ByteChecker IO sym bak ([Word8] -> [Word8]) [Word8]
+_fullyConcreteNullTerminatedString =
+  ByteChecker $ \bak acc bytePtr -> do
+    byte <- ptrToBv8 bak bytePtr
+    case BV.asUnsigned <$> WI.asBV byte of
+      Just 0 -> pure (Break (acc []))
+      Just c -> do
+        let c' = toEnum @Word8 (fromInteger c)
+        pure (Continue (acc . (c':)))
```
RyanGlScott: Similar comments here (and in related functions) about `acc . (c':)` being a bit too pointfree for my liking.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -3,27 +3,261 @@
...
+loadConcretelyNullTerminatedString bak mem ptr limit =
+  case limit of
+    Nothing -> loadBytes bak mem id ptr concretelyNullTerminatedString
+    Just l ->
+      let byteChecker = withMaxChars l (pure . ($ [])) concretelyNullTerminatedString in
+      loadBytes bak mem (id, 0) ptr byteChecker
+
+-- | Load a null-terminated string from memory.
+--
+-- Consults an SMT solver to check if the loaded byte is known to be null (0).
```
RyanGlScott:
```suggestion
-- Consults an SMT solver to check any of the loaded bytes are known to be null (0).
```

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -3,27 +3,261 @@
...
+data ControlFlow a b
+  = Continue a
+  | Break b
+  deriving Functor
+
+instance Bifunctor ControlFlow where
+  bimap l r =
+    \case
+      Continue x -> Continue (l x)
+      Break x -> Break (r x)
```
RyanGlScott, line 59: Given that `ControlFlow` isn't exported, is there any reason to define this instance? As far as I can tell, nothing is making use of it.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -3,27 +3,261 @@
...
+  LCB.IsSymBackend sym bak =>
+  bak ->
+  Mem.LLVMPtr sym 8 ->
+  IO (WI.SymBV sym 8)
+ptrToBv8 bak bytePtr = do
+  let err = LCS.AssertFailureSimError "Found pointer instead of byte when loading string" ""
+  Partial.ptrToBv bak err bytePtr
+
+-- Currently unused, but analogous with 'Lang.Crucible.LLVM.MemModel.loadString'
+_fullyConcreteNullTerminatedString ::
```
RyanGlScott: Any reason not to define `loadString` in terms of `_fullyConcreteNullTerminatedString`, now that we have it?

RyanGlScott (commented):
Please mention these new functions in the `crucible-llvm` changelog.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
langston-barrett: `first` is used below, added a comment mentioning it

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
langston-barrett: Because the module dependency goes the other way - `Strings` imports `MemModel`. I could add this as a TODO referencing #1406?

RyanGlScott: Yes, let's add a TODO mentioning #1406.

PR #1452, opened 2025-07-03
`crucible-mir`: Generalize `rotate_{left,right}` overrides to work on many different bit widths

### LINKED ISSUE #1451: `crucible-mir`: Custom translation for `rotate_{left,right}` only works for `u32`
`crucible-mir` has custom overrides for the [`rotate_left`](https://doc.rust-lang.org/std/intrinsics/fn.rotate_left.html) and [`rotate_right`](https://doc.rust-lang.org/std/intrinsics/fn.rotate_right.html) intrinsics. These intrinsics can work over a variety of different integer types, and there are `rotate_{left,right}` variants for `u8`, `u16`, `u32`, `u64`, and `u128`. (Similarly for the signed integer types `i*`.) Despite this, `crucible-mir`'s custom overrides only work for `u32` in practice. For instance, `crux-mir` has no issues simulating the following program, which uses `u32::rotate_left`:

```
// test.rs
#[crux::test]
pub fn f() {
    let n: u32 = 3;
    let expected = 12;
    let actual = n.rotate_left(2);
    assert!(expected == actual);
}
```

If you change `n: u32` to `n: u64`, however, then `crux-mir` fails to translate it:

```
test test/2c090053::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/2c090053::f[0] counterexamples ----
[Crux]   ./libs/core/src/num/uint_macros.rs:261:20: 261:52 !./libs/core/src/num/mod.rs:1121:5: 1139:6: error: in core/c7248340::num[0]::{impl#9}[0]::rotate_left[0]
[Crux]   Translation error in core/c7248340::num[0]::{impl#9}[0]::rotate_left[0]: invalid arguments for rotate_left

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

The culprit can be seen in the implementations of the `rotate_{left,right}` overrides:

https://github.com/GaloisInc/crucible/blob/f9cc8a48c8636facc190d49191f78f3634f1bb36/crucible-mir/src/Mir/TransCustom.hs#L929-L943

In particular, note the requirement that the first argument's bit-width must be equal to the second argument's bit-width. This is a problem, because the Rust type signatures for these intrinsics look like:

```rs
pub const fn rotate_{left,right}<T>(x: T, shift: u32) -> T
```

Note that the second argument is always 32 bits wide, even though the first argument's bit-width is allowed to differ. As a result, the `crucible-mir` overrides will work at 32 bits, but not at any other widths.

We should generalize these overrides to work at other widths. We can take inspiration from how `what4` implements a similar operation [here](https://github.com/GaloisInc/what4/blob/cb2cdfc2ebc17f2120502cedd070b7d69e095cb0/what4/src/What4/SWord.hs#L693-L716). `what4`'s low-level rotation operations require the bit-widths of the arguments to be the same, but it also provides an abstraction on top of it which zero-extends or truncates the second argument as appropriate to be the same bit-width as the first argument.

RyanGlScott:
Fixes #1451.

### DIFF
--- a/crux-mir/test/conc_eval/intTest/test1451.rs
+++ b/crux-mir/test/conc_eval/intTest/test1451.rs
@@ -1,0 +1,88 @@
+// A regression test for #1451. This tests the rotate_{left,right} intrinsics
+// at various, commonly-encountered bit widths to ensure that translation
+// succeeds and behaves as expected.
+
+macro_rules! rol_test {
+    ($name:ident, $ty:ty) => {
+        pub fn $name() {
+            let n: $ty = 3;
+            let expected = 12;
+            let actual = n.rotate_left(2);
+            assert!(expected == actual);
+        }
+    }
+}
+
+macro_rules! ror_test {
+    ($name:ident, $ty:ty) => {
+        pub fn $name() {
+            let n: $ty = 12;
+            let expected = 3;
+            let actual = n.rotate_right(2);
+            assert!(expected == actual);
+        }
+    }
+}
+
+rol_test!(i8_rol_test, i8);
+rol_test!(i16_rol_test, i16);
+rol_test!(i32_rol_test, i32);
+rol_test!(i64_rol_test, i64);
+rol_test!(i128_rol_test, i128);
+rol_test!(isize_rol_test, isize);
+
+rol_test!(i8_ror_test, i8);
+rol_test!(i16_ror_test, i16);
+rol_test!(i32_ror_test, i32);
+rol_test!(i64_ror_test, i64);
+rol_test!(i128_ror_test, i128);
+rol_test!(isize_ror_test, isize);
+
+ror_test!(u8_rol_test, u8);
+ror_test!(u16_rol_test, u16);
+ror_test!(u32_rol_test, u32);
+ror_test!(u64_rol_test, u64);
+ror_test!(u128_rol_test, u128);
+ror_test!(usize_rol_test, usize);
+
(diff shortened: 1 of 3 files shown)


crux-mir/test/conc_eval/intTest/test1451.rs
```diff
@@ -0,0 +1,88 @@
...
+rol_test!(i64_ror_test, i64);
+rol_test!(i128_ror_test, i128);
+rol_test!(isize_ror_test, isize);
+
+ror_test!(u8_rol_test, u8);
+ror_test!(u16_rol_test, u16);
+ror_test!(u32_rol_test, u32);
+ror_test!(u64_rol_test, u64);
+ror_test!(u128_rol_test, u128);
+ror_test!(usize_rol_test, usize);
```
samcowger, line 46: A bit late, but: shouldn't the top 6 macros be `ror_test!` rather than `rol_test!` (and vice versa for the bottom 6)?

RyanGlScott: Oops, quite right. I'll address this in a follow-up PR.

RyanGlScott: https://github.com/GaloisInc/crucible/pull/1454

PR #1457, opened 2025-07-07
Don't throw exceptions if we don't understand the output of `java`

yav:
Modify the functions that interact with an external `java` process to return `Nothing` instead of throwing exceptions.
This is helpful to fix https://github.com/GaloisInc/saw-script/issues/1709#issuecomment-1189188736

### DIFF
--- a/crucible-jvm/src/Lang/JVM/JavaTools.hs
+++ b/crucible-jvm/src/Lang/JVM/JavaTools.hs
@@ -16,6 +16,8 @@ module Lang.JVM.JavaTools
 import Control.Exception(handle,SomeException(..))
 import Data.List.Extra (dropPrefix, firstJust, stripInfix)
 import Data.Maybe
+import Data.Text(Text)
+import qualified Data.Text as Text
 import System.Directory
 import Text.Read(readMaybe)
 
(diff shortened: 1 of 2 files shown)


crucible-jvm/src/Lang/JVM/JavaTools.hs
```diff
@@ -75,12 +78,12 @@ findJavaProperty javaPath propertyNeedle = do
 
 -- | @'findJavaMajorVersion' javaPath@ will consult @javaPath@ to find the
 -- major version number corresponding to that Java release.
-findJavaMajorVersion :: FilePath -> IO Int
+findJavaMajorVersion :: FilePath -> IO (Maybe Int)
```
RyanGlScott: Rather than returning `Maybe Int`, I think we should return something akin to `Either Text Int`. That way, `crucible-jvm` can report what went wrong in the event that `findJavaMajorVersion` fails.

crucible-jvm/tool/Main.hs
```diff
@@ -124,14 +124,18 @@ processJVMOptions opts@JVMOptions{javaBinDirs} = do
...
+        Just v
+          | v >= 9 ->
+            putStrLn (unlines
+              [ "WARNING: crucible-jvm's support for JDK 9 or later is experimental."
+              , "See https://github.com/GaloisInc/crucible/issues/641."
+              ])
+            >> pure opts
+          | otherwise -> addRTJar javaPath opts
+        Nothing ->
+          putStrLn "WARNING: failed to determine JDK version." >> pure opts
```
RyanGlScott: I think this should fail instead of silently succeeding. If we can't determine the Java version, then we don't have a realistic chance of things working as expected.

crucible-jvm/src/Lang/JVM/JavaTools.hs
yav: We don't really know what went wrong beyond that we couldn't get the version (i.e. the error is always "parse error").  I guess I could return the output of the command, and we could show that to the user, although I wonder how useful it is unless we show them how run the command as well...

RyanGlScott: Ah, sorry, to be clear: I think we should also change the return type of `findJavaProperty` to return `Either Text` so that we can display the output of the command.

PR #1459, opened 2025-07-09
crux-mir: perform tests in alphabetical order by test directory/filename

samcowger:
This should:
- Make the test suite a hair more robust, by eliminating some needless nondeterminism, and
- Offer a better experience when scrolling through e.g. CI output to find results for particular tests, since test results will appear in a standard, stable order.

### DIFF
--- a/crucible-debug/test/Test.hs
+++ b/crucible-debug/test/Test.hs
@@ -86,7 +86,10 @@ mkTest dir path =
 
 loadTests :: FilePath -> IO Tasty.TestTree
 loadTests dir = do
-  files <- Dir.listDirectory dir
+  -- This `List.sort` is not technically necessary, it just ensures that test
+  -- cases will be performed in a stable ordering, since `Dir.listDirectory`
+  -- doesn't guarantee such an ordering.
+  files <- List.sort <$> Dir.listDirectory dir
   let dbgScripts = List.filter (".txt" `List.isSuffixOf`) files
   let tests = map (uncurry mkTest) (map (dir,) dbgScripts)
   pure (Tasty.testGroup dir tests)


crux-mir/test/Test.hs
```diff
@@ -284,6 +284,6 @@ testDir oracleTest dir = do
       gen f = doesDirectoryExist (dir </> f) >>= \case
         False -> return Nothing
         True -> Just <$> testDir oracleTest (dir </> f)
-  fs <- listDirectory dir
+  fs <- sort <$> listDirectory dir
```
RyanGlScott, line 290: Perhaps worth a comment that the `sort` is technically unnecessary, but it ensures a stable ordering of test cases regardless of the underlying order in which the OS lists the files.

RyanGlScott (commented):
Thanks! This is something that has bugged me for a while.

While we're at it, I noticed that the `crucible-debug` test suite also suffers from the same issue when enumerating its test cases:

https://github.com/GaloisInc/crucible/blob/5a46059d2084eb0c591735f1cd4830df89b55b81/crucible-debug/test/Test.hs#L88-L89

Could you add a `sort` here as well?

crux-mir/test/Test.hs
samcowger: Sure, done.

samcowger:
> While we're at it, I noticed that the `crucible-debug` test suite also suffers from the same issue when enumerating its test cases:
> 
> https://github.com/GaloisInc/crucible/blob/5a46059d2084eb0c591735f1cd4830df89b55b81/crucible-debug/test/Test.hs#L88-L89
> 
> Could you add a `sort` here as well?

Done in 3441f67.

PR #1460, opened 2025-07-09
llvm: Export more "low-level" string-loading functionality

langston-barrett:
This is necessary for effectively loading strings via the Macaw memory model, which uses `doReadMemModel` rather than `doLoad`. We should add a module like `Strings` to macaw-symbolic that wraps these low-level APIs much the same way `Strings` does. I might do this (and use it in GREASE) before un-drafting this PR, but feel free to review in the meantime.


langston-barrett:
I've plumbed this through Macaw and GREASE and it appears to work well with the lazy Macaw memory model.

RyanGlScott (approved):

PR #1461, opened 2025-07-10
MIR: recoverably fail on unsupported type

glguy:
The updates tyToRepr to report a recoverable error when we encounter a type we don't support.

This allows us to emit a symbolic-execution runtime assertion failure for code that uses types we don't support. Treating these as soft failures allow us to verify other aspects of the program that don't execute the unsupported portion both through assumption and through overrides.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -229,7 +229,7 @@ transConstVal ty@(M.TyAdt aname _ _) tpr (ConstStruct fields) = do
         Nothing -> do
             let fieldDefs = adt ^. adtvariants . ix 0 . vfields
             let fieldTys = map (\f -> f ^. fty) fieldDefs
-            exps <- zipWithM (\val ty -> transConstVal ty (tyToRepr col ty) val) fields fieldTys
+            exps <- zipWithM (\val ty -> tyToReprM ty >>= \rpr -> transConstVal ty rpr val) fields fieldTys
             buildStruct adt exps
 
 transConstVal ty@(M.TyAdt aname _ _) tpr (ConstEnum variant fields) = do
@@ -241,7 +241,7 @@ transConstVal ty@(M.TyAdt aname _ _) tpr (ConstEnum variant fields) = do
         Nothing -> do
             let fieldDefs = adt ^. adtvariants . ix variant . vfields
             let fieldTys = map (\f -> f ^. fty) fieldDefs
-            exps <- zipWithM (\val ty -> transConstVal ty (tyToRepr col ty) val) fields fieldTys
+            exps <- zipWithM (\val ty -> tyToReprM ty >>= \rpr -> transConstVal ty rpr val) fields fieldTys
             buildEnum adt variant exps
 transConstVal ty (Some MirReferenceRepr) init = do
     let pointeeTy = M.typeOfProj M.Deref ty
@@ -285,7 +285,9 @@ transConstTuple tys tprs vals = do
                      _ ->
                        mirFail $ "transConstTuple: expected tuple field to have MaybeType, but got " ++ show tpr)
                tys (toListFC Some tprs) vals
-    return $ buildTupleMaybe col tys $ map Just vals'
+    case buildTupleMaybe col tys $ map Just vals' of
+      Left err -> mirFail ("transConstTuple: " ++ err)
+      Right rpr -> return rpr
 
 -- Translate a struct or enum marked with repr(transparent).
 transTransparentVal ::
@@ -475,7 +477,9 @@ transBinOp bop op1 op2 = do
         WithOverflow bop' -> do
             (res, overflow) <- evalBinOp bop' mat me1 me2
             col <- use $ cs . collection
-            return $ buildTupleMaybe col [error "not needed", TyBool] [Just res, Just $ MirExp (C.BoolRepr) overflow]
+            case buildTupleMaybe col [error "not needed", TyBool] [Just res, Just $ MirExp (C.BoolRepr) overflow] of
+              Left err -> mirFail ("transBinOp: " ++ err)
+              Right rpr -> return rpr
         _ -> fst <$> evalBinOp bop mat me1 me2
 
 -- Evaluate a binop, returning both the result and an overflow flag.
@@ -1075,9 +1079,10 @@ evalCast' ck ty1 e ty2  = do
       -- prohibiting them here is described in `unsizeAdtDyn`, above, and
       -- `Mir.TransTy.structFieldRef`
       case tyToFieldRepr col eventualTraitObject of
-        Some (FieldRepr (FkMaybe _)) ->
+        Left err -> mirFail ("unsizeAdtDynNormal: " ++ err)
+        Right (Some (FieldRepr (FkMaybe _))) ->
           mirFail "error: unsizeAdtDyn with non-initializable unsized field"
-        Some (FieldRepr (FkInit _)) ->
+        Right (Some (FieldRepr (FkInit _))) ->
           case findLastFieldRec col castTarget of
             Nothing ->
               mirFail $ "unsizedAdtDyn: unable to determine last field of "<>show castTarget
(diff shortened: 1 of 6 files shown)


RyanGlScott (commented):
Looks reasonable at a first glance, although we should advertise the relevant API changes in the `crucible-mir` changelog.

RyanGlScott:
Also, I think we should add a `crux-mir` test case that demonstrates this new feature. Something like this:

```rs
#[repr(C)]
union MyUnion {
    f1: u32,
    f2: f32,
}

fn f() -> u32 {
    let u = MyUnion { f1: 1 };
    unsafe { u.f1 }
}

fn g(x: bool) -> u32 {
    if x {
        f()
    } else {
        2
    }
}

#[crux::test]
pub fn h() -> u32 {
    g(false)
}
```

This fails with a translation error on `master` but succeeds with this branch.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2391,15 +2402,17 @@ mkHandleMap :: (HasCallStack) => Collection -> FH.HandleAllocator -> IO HandleMa
...
-       in
-          tyListToCtx col targs $ \argctx -> do
-          tyToReprCont col (fsig^.fsreturn_ty) $ \retrepr -> do
-             h <- FH.mkHandle' halloc handleName argctx retrepr
-             return $ MirHandle fname fsig h
+    mkHandle (M.Fn fname _fargs fsig _fbody)  = do
+      let targs = abiFnArgs fsig
+          handleName = FN.functionNameFromText (M.idText fname)
+      Some argctx <- case tyListToCtx col targs (Right . Some) of
+         Left err -> fail ("mkHandle: " ++ err)
```
samcowger, line 2409: Is the use of `fail` here, rather than `error`, intentional?

crux-mir/test/symb_eval/resilience/ignore_unused_union.rs
RyanGlScott, line 1: Can you put this test case in `conc_eval` instead of `symb_eval`? `symb_eval` is generally meant for test cases that [create symbolc values and make assertions about them](https://github.com/GaloisInc/crucible/blob/96f7e8f92f17b2938bddc8606ca5ef5c17e6bf1f/crux-mir/README.md#L105-L106).

As part of this change, you won't need the `.good` file anymore, as the test suite compares the output of simulation against the actual compiled Rust code. To make this work, you'll have to set up a bit of additional scaffolding to allow the code to compile with and without `crux-mir`:

```rs
#[cfg_attr(crux, crux::test)]
pub fn h() -> u32 {
    g(false)
}

pub fn main() {
}
```

See the other test cases in `conc_eval` for inspiration.

RyanGlScott (approved):
LGTM, aside from one suggestion.

crucible-mir/src/Mir/Trans.hs
glguy: Yes, `fail` fails in the sequencing of IO actions. `error` is the failure method of last resort where you making a pure function and you didn't leave yourself any other way to get out.

samcowger:
> Also, I think we should add a `crux-mir` test case that demonstrates this new feature. Something like this:
> 
> ```rust
> #[repr(C)]
> union MyUnion {
>     f1: u32,
>     f2: f32,
> }
> 
> fn f() -> u32 {
>     let u = MyUnion { f1: 1 };
>     unsafe { u.f1 }
> }
> 
> fn g(x: bool) -> u32 {
>     if x {
>         f()
>     } else {
>         2
>     }
> }
> 
> #[crux::test]
> pub fn h() -> u32 {
>     g(false)
> }
> ```
> 
> This fails with a translation error on `master` but succeeds with this branch.

I pulled this and played around with it a bit, and I found that inlining `f()` into `g` causes this test to fail with a translation error. Is that expected?

RyanGlScott:
> Is that expected?

Unfortunately, yes. This is perhaps easiest to see if you run `crux-mir` with the `--print-mir` flag. That is, you if you have this program:

```rs
#[repr(C)]
union MyUnion {
    f1: u32,
    f2: f32,
}

fn g(x: bool) -> u32 {
    if x {
	    let u = MyUnion { f1: 1 };
	    unsafe { u.f1 }
    } else {
        2
    }
}

#[crux::test]
pub fn h() -> u32 {
    g(false)
}
```

And run it like so:

```
$ cabal -v0 run exe:crux-mir -- test.rs --print-mir
FNs
fn test/785661fb::g[0](_1 : bool) -> u32 {
   let mut _0 : u32;
   let const _2 : test/785661fb::MyUnion[0]<>;
   bb0: {
      switchint _1 :bool [0] -> [bb2, bb1]
   }
   bb1: {
      _2 = AdtAg(test/785661fb::MyUnion[0]::_adtc5a5405abab57fd5[0], 0, [1], 0);
      _0 = use(_2.0);
      goto(bb3);
   }
   bb2: {
      _0 = use(2);
      goto(bb3);
   }
   bb3: {
      return;
   }
}
fn test/785661fb::h[0]() -> u32 {
   let mut _0 : u32;
   bb0: {
      call(test/785661fb::g[0](False), (_0, bb1))
   }
   bb1: {
      return;
   }
}
ADTs
 Union test/785661fb::MyUnion[0]::_adtc5a5405abab57fd5[0][test/785661fb::MyUnion[0] <>]( Variant( test/785661fb::MyUnion[0]
, Relative(0)
, [ Field(test/785661fb::MyUnion[0]::f1[0], u32)
  , Field(test/785661fb::MyUnion[0]::f2[0], f32) ]
,
,
, True ) )
TRAITs
VTABLEs
INTRINSICSs
test/785661fb::g[0] = IkItem test/785661fb::g[0]<<>>
test/785661fb::h[0] = IkItem test/785661fb::h[0]<<>>
STATICS
```

The problem is that `crucible-mir` translates the temporary variable declaration `let const _2 : test/785661fb::MyUnion[0]<>;` _before_ it translates the rest of `g`'s function body. (We must do this in general, although in this particular example, `_2` is only used in one branch of an `if` statement.) As a result, when `crucible-mir` translates the type of `_2`, it generates a failing statement when it encounters `MyUnion`, and is immediately what gets simulated when you invoke `g`.

glguy:
> > Also, I think we should add a `crux-mir` test case that demonstrates this new feature. Something like this:
> > ```rust
> > #[repr(C)]
> > union MyUnion {
> >     f1: u32,
> >     f2: f32,
> > }
> > 
> > fn f() -> u32 {
> >     let u = MyUnion { f1: 1 };
> >     unsafe { u.f1 }
> > }
> > 
> > fn g(x: bool) -> u32 {
> >     if x {
> >         f()
> >     } else {
> >         2
> >     }
> > }
> > 
> > #[crux::test]
> > pub fn h() -> u32 {
> >     g(false)
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This fails with a translation error on `master` but succeeds with this branch.
> 
> I pulled this and played around with it a bit, and I found that inlining `f()` into `g` causes this test to fail with a translation error. Is that expected?

It's undesirable but probably unavoidable.

Looking at the generated MIR I can see that the union const `_2` is lifted up to the top of the functions. I won't be surprised to find that we have to settle for using function-boundaries as the place to hide unsupported features.

```
fn sm/e4a6144a::g[0](_1 : bool) -> u32 {
   let mut _0 : u32;
   let const _2 : sm/e4a6144a::MyUnion[0]<>;
   bb0: {
      switchint _1 :bool [0] -> [bb2, bb1]
   }
   bb1: {
      _2 = AdtAg(sm/e4a6144a::MyUnion[0]::_adtc5a5405abab57fd5[0], 0, [1], 0);
      _0 = use(_2.0);
      goto(bb3);
   }
   bb2: {
      _0 = use(2);
      goto(bb3);
   }
   bb3: {
      return;
   }
}
```

samcowger:
I see, thanks. That makes sense, though it's definitely unfortunate.

crucible-mir/src/Mir/Trans.hs
samcowger: Gotcha. Part of my motivation for asking is because `error` prints call stacks by default, while `fail` (at `IO`, at least) seems not to. Given that you're already in a `HasCallStack` context at a lot of these `fail` sites, could you append call stack information to the error messages you're creating?

PR #1462, opened 2025-07-10
Mir personality support

yav:
This PR adds support for custom user state to concurrent crux, and crux-mix.   This is in preparation for some changes to cryx-mir-comp, to add support for working with uninterpreted functions.

### DIFF
--- a/crucible-concurrency/src/Cruces/CrucesMain.hs
+++ b/crucible-concurrency/src/Cruces/CrucesMain.hs
@@ -129,7 +129,7 @@ run (cruxOpts, opts) =
                   (IsSymBackend s bak, IsExprBuilder s) =>
                   bak ->
                   IO ( FnVal s Ctx.EmptyCtx C.UnitType
-                     , ExplorePrimitives (ThreadExec() DPOR s () C.UnitType) s ()
+                     , ExplorePrimitives (ThreadExec () DPOR s () C.UnitType) s ()
                      , [Some GlobalVar]
                      , FunctionBindings (ThreadExec () DPOR s () C.UnitType) s ()
                      )
--- a/crucible-concurrency/src/Cruces/CrucesMain.hs
+++ b/crucible-concurrency/src/Cruces/CrucesMain.hs
@@ -129,9 +129,9 @@ run (cruxOpts, opts) =
                   (IsSymBackend s bak, IsExprBuilder s) =>
                   bak ->
                   IO ( FnVal s Ctx.EmptyCtx C.UnitType
-                     , ExplorePrimitives (ThreadExec DPOR s () C.UnitType) s ()
+                     , ExplorePrimitives (ThreadExec() DPOR s () C.UnitType) s ()
                      , [Some GlobalVar]
-                     , FunctionBindings (ThreadExec DPOR s () C.UnitType) s ()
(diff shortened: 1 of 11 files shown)


crucible-concurrency/src/Cruces/CrucesMain.hs
```diff
@@ -129,9 +129,9 @@ run (cruxOpts, opts) =
                   (IsSymBackend s bak, IsExprBuilder s) =>
                   bak ->
                   IO ( FnVal s Ctx.EmptyCtx C.UnitType
-                     , ExplorePrimitives (ThreadExec DPOR s () C.UnitType) s ()
+                     , ExplorePrimitives (ThreadExec() DPOR s () C.UnitType) s ()
```
RyanGlScott: Minor style nit pick:

```suggestion
                     , ExplorePrimitives (ThreadExec () DPOR s () C.UnitType) s ()
```

crux-mir/src/Mir/Language.hs
```diff
@@ -156,25 +163,51 @@ withMirLogging computation =
...
+instance MirPersonality (CruxMir s sym) where
+  type MirState (CruxMir s sym) = s 
+  mirStateField = lens cruxMirState (const CruxMir)
+
+-- | This is used when we execute with concurrency
+instance MirPersonality p => MirPersonality (Explore.Exploration p alg ext ret sym) where
+  type MirState (Explore.Exploration p alg ext ret sym) = MirState p
+  mirStateField = Explore.personality . mirStateField
+
+mirState :: HasMirState (p sym) s => Lens' (C.SimState (p sym) sym ext r f a) s
```
RyanGlScott: Why not generalize this to:

```suggestion
mirState :: HasMirState p s => Lens' (C.SimState p sym ext r f a) s
```

crux-mir/src/Mir/Language.hs
```diff
@@ -156,25 +163,51 @@ withMirLogging computation =
...
+        $ runTestsWithExtraOverrides customState bindExtra
+
+type HasMirState p s = (MirPersonality p, MirState p ~ s)
+
+-- | Provides access to some custom state
+class MirPersonality p where
+  type MirState p
+  mirStateField :: Lens' p (MirState p)
+
+newtype CruxMir s sym = CruxMir { cruxMirState :: s }
```
RyanGlScott: What is the purpose of quantifying over `sym` here?

crux-mir/src/Mir/Language.hs
yav: Yeah, there something somewhere in Crux that seems to require the personalities to be perameterized by the `sym`, which is why I wrote things like that.   It looks pretty weird to me too, so I'll see if I can figure out if this is actually needed or not, and maybe put a note somewhere if it is needed.

yav: I just pushed a change that generalizes this in Crux

RyanGlScott: Thanks! It's probably worth mentioning this change in the `crux` changelog. (Case in point: CI reveals that change broke some other code in `crux-llvm`.)

yav:
I ended up doing this in a different way, so I'll make a new PR shortly

PR #1465, opened 2025-07-11
llvm: Assume previous bytes were non-null when loading strings

### LINKED ISSUE #1463: llvm: String-handling functions are imprecise
The string-handling functions in [`Lang.Crucible.LLVM.MemModel.Strings`](https://github.com/GaloisInc/crucible/blob/96f7e8f92f17b2938bddc8606ca5ef5c17e6bf1f/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs) are *imprecise* in the sense that they can cause assertion failures where none is warranted.

These functions load bytes one at a time until a provably null (most often, concretely null) terminator can be found. Loading each byte results in a safety assertion that the load succeeds. However, if the string contains a symbolic byte that *might* be equal to zero, then the assertions about loading subsequent bytes should only be made *assuming* the earlier byte was nonzero.

This reasoning is applied in [`Lang.Crucible.LLVM.MemModel.strLen`](https://github.com/GaloisInc/crucible/blob/96f7e8f92f17b2938bddc8606ca5ef5c17e6bf1f/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs), which correctly checks if the current path condition implies the safety predicate:

https://github.com/GaloisInc/crucible/blob/96f7e8f92f17b2938bddc8606ca5ef5c17e6bf1f/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L1146-L1154

langston-barrett:
Fixes #1463, though I'm not super confident in it... it looks pretty different from the existing logic in `strLen`, but I think it should generate the same proof goals?

Also seems a bit tricky to make a test-case for, the safety assertion for loading a particular byte would have to depend integrally on the non-zero-ness of a previous byte...

### DIFF
--- a/crucible-llvm-cli/test-data/T1463-read-bytes.cbl
+++ b/crucible-llvm-cli/test-data/T1463-read-bytes.cbl
@@ -1,0 +1,34 @@
+; A regression test for #1463, wherein some string-handling functions generated
+; imprecise assertions that could spurriously fail because the safety assertions
+; for loads of later bytes in strings didn't have access to the assumption that
+; earlier bytes were nonzero.
+;
+; If `read-bytes` were subject to that bug, then this particular test-case
+; would cause an assertion failure. In the case that the fresh byte is zero,
+; then the first byte of the string is zero and `read-bytes` should immediately
+; stop reading without generating further assertions. Otherwise, the byte would
+; have been written to both the first and second positions of the string, so an
+; uninitialized read should be impossible.
+
+(declare @read-bytes ((x Pointer)) (Vector (Bitvector 8)))
+
+(defun @main () Unit
+  (start start:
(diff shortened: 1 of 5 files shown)


RyanGlScott:
This seems plausible, but this code deserves some commentary. What is an example where this assumption is critical to making it work? (If it's difficult to distill into a test case, that's fine—I'm just not sure of where this would come up.)

langston-barrett:
I made tests for `strlen` (showing that it is not subject to this bug) and `read-bytes` (showing that it is) in a first commit, and fixed them in the second commit (bringing their results in line with one another). Also added a comment referring to the test-case and issue.

RyanGlScott (approved):
Wonderful, thanks!

PR #1468, opened 2025-07-11
llvm: APIs and overrides for `strnlen`, `strcpy`

langston-barrett:
Also some minor renaming of existing string-loading functionality for consistency. To test this, I'll create an override and test-case for `strnlen`.

### DIFF
--- a/crucible-llvm-cli/test-data/strcpy.cbl
+++ b/crucible-llvm-cli/test-data/strcpy.cbl
@@ -1,4 +1,6 @@
 (declare @strcpy ((dst (Ptr 64)) (src (Ptr 64))) (Ptr 64))
+(declare @read-c-string ((x Pointer)) (String Unicode))
+(declare @write-c-string ((dst Pointer) (src (String Unicode))) Unit)
 
 (defun @main () Unit
   (start start:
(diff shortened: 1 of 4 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -15,29 +15,46 @@
...
+  -- * Loading strings
+  , Mem.loadString
   , Mem.loadMaybeString
-  , Mem.strLen
   , loadConcretelyNullTerminatedString
-  , loadSymbolicString
-  , storeString
+  , loadProvablyNullTerminatedString
+  -- * String length
+  , Mem.strLen
```
RyanGlScott, line 25: Only somewhat related to the changes here, but are there plans to define `Mem.strLen` in terms of the new `ByteChecker` machinery?

langston-barrett: Yes, that will become possible when `strLen` is moved to `Strings`, see https://github.com/GaloisInc/crucible/issues/1406 and specifically https://github.com/GaloisInc/crucible/blob/dcdb1ddd01bb3eb5f4c2a1689a0a46542ac6ab04/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L95

langston-barrett:
> Perhaps it would be worth documenting these restrictions in crucible-llvm/doc/limitations.md?

Good call, done!

PR #1469, opened 2025-07-11
crucible-mir: add debug helpers

### LINKED ISSUE #1311: `crucible-mir`: Add override for `println!` macro (or a `println!`-like function) for debugging
`crucible-llvm` features a [`printf` override](https://github.com/GaloisInc/crucible/blob/50d1550834e9fc0b0c0a898aa6d699aa53be4c89/crucible-llvm/src/Lang/Crucible/LLVM/Printf.hs) that goes as far as actually printing its contents to `stderr` (it will print `?` characters for symbolic contents). This is tremendously useful for debugging purposes, as it allows `crux-llvm` users to see what values are passed around at simulation time.

This issue exists to explore whether we could have a similar capability at the `crucible-mir` level. The ideal vision would be to override the `println!` macro such that it behaves analogously to `crucible-llvm`'s `printf` override. This is much easier said than done, however, as the `println!` macro usually expands to fairly difficult-for-Crucible-to-simulate code. For instance, given this seemingly innocent-looking Rust code:

```rs
pub fn main() {
    println!("hello world");
}
```

`rustc` will turn it into the following MIR:

```
fn test/13e15d23::main[0]() -> () {
   let mut _0 : ();
   let const _1 : ();
   let mut _2 : core/092bc89a::fmt[0]::Arguments[0]<lifetime>;
   let mut _3 : &[&str];
   let mut _4 : &[&str; 1];
   let const _5 : &[&str; 1];
   let mut _6 : &[core/092bc89a::fmt[0]::ArgumentV1[0]<lifetime>];
   let mut _7 : &[core/092bc89a::fmt[0]::ArgumentV1[0]<lifetime>; 0];
   let const _8 : &[core/092bc89a::fmt[0]::ArgumentV1[0]<lifetime>; 0];
   let mut _9 : &[core/092bc89a::fmt[0]::ArgumentV1[0]<lifetime>; 0];
   let mut _10 : &[&str; 1];
   bb0: {
      _10 = use(&test/13e15d23580eea4b::{{alloc}}[1]);
      _5 = use(_10);
      _4 = use(_5);
      _3 = Cast(Unsize, _4, &[&str]);
      _9 = use(&test/13e15d23580eea4b::{{alloc}}[2]);
      _8 = use(_9);
      _7 = use(_8);
      _6 = Cast(Unsize, _7, &[core/092bc89a::fmt[0]::ArgumentV1[0]<lifetime>]);
      call(_2 = <ZST>(_3, _6), bb1)
   }
   bb1: {
      call(_1 = <ZST>(_2), bb2)
   }
   bb2: {
      return;
   }
}
```

Currently, `crucible-mir` chokes on this code:

```
Symbolic execution failed.
  internal: error: in core/092bc89a::mem[0]::needs_drop[0]::_instaddce72e1232152c[0]
  Translation error in core/092bc89a::mem[0]::needs_drop[0]::_instaddce72e1232152c[0]: callExp: Don't know how to call core/092bc89a::intrinsics[0]::{extern#0}[0]::needs_drop[0]::_instaddce72e1232152c[0]
```

We would need to dig more into the details of how `println!` works in order to ensure that `crucible-mir` is able to cope with everything that it can do.

As an alternative, it might be worth adding a family of no-op functions that do nothing at Rust execution time, but are overriden by Crucible to print things. I'm imagining:

* `crucible_print_i8`
* `crucible_print_i16`
* `crucible_print_u8`
* `crucible_print_u16`
* `crucible_print_str`

And so on for each primitive Rust type. This would be much simpler than having to support `println!` and all of its various quirks.

spernsteiner:
This adds some debugging functions for inspecting parts of crux-mir's internal representation, based on the ones @glguy and I used to investigate GaloisInc/saw-script#2477.  The main function is `dump_any`, which takes any Rust value and prints its representation as a Crucible `RegValue`.  It also adds `print_str`, which prints a string during symbolic execution; this is mainly for use with `MethodSpec::pretty_print` which returns its result as `&str`.

Fixes #1311

### DIFF
--- a/crux-mir/src/Mir/Overrides.hs
+++ b/crux-mir/src/Mir/Overrides.hs
@@ -10,6 +10,7 @@
 {-# Language TypeApplications #-}
 {-# Language PartialTypeSignatures #-}
 {-# Language FlexibleContexts #-}
+{-# Language AllowAmbiguousTypes #-}
 
 module Mir.Overrides (bindFn, getString) where
 
@@ -33,8 +34,11 @@ import qualified Data.Vector as V
 
 import System.IO (hPutStrLn)
 
+import qualified Prettyprinter as PP
+
 import Data.Parameterized.Context (pattern Empty, pattern (:>))
 import qualified Data.Parameterized.Context as Ctx
+import qualified Data.Parameterized.Map as MapF
 import Data.Parameterized.NatRepr
 
 import What4.Config( getOpt, setOpt, getOptionSetting )
@@ -55,6 +59,7 @@ import Lang.Crucible.Backend.Online
 import Lang.Crucible.CFG.Core (CFG, cfgArgTypes, cfgHandle, cfgReturnType)
 import Lang.Crucible.FunctionHandle
 import Lang.Crucible.Panic
+import Lang.Crucible.Pretty
 import Lang.Crucible.Simulator.ExecutionTree
 import Lang.Crucible.Simulator.GlobalState
 import Lang.Crucible.Simulator.OverrideSim
@@ -462,7 +467,7 @@ bindFn symOnline cs name cfg
         str <- getString strRef >>= \x -> case x of
             Just x -> return x
             Nothing -> fail "dump_rv: desc string must be concrete"
-        liftIO $ outputLn $ Text.unpack str ++ " = " ++ showRV tpr expr
+        liftIO $ outputLn $ Text.unpack str ++ " = " ++ showRV @sym tpr expr
 
   where
     explodedName = textIdKey name
@@ -476,24 +481,6 @@ bindFn symOnline cs name cfg
         Just (edidInit, edidLast) ->
           pfxInit == edidInit &&
           "_inst" `Text.isPrefixOf` edidLast
-
-    showRV :: TypeRepr tp' -> RegValue sym tp' -> String
-    showRV (BVRepr _) x = show (printSymExpr x)
-    showRV AnyRepr (AnyValue tpr' rv') = "(AnyValue " ++ show tpr' ++ " " ++ showRV tpr' rv' ++ ")"
-    showRV UnitRepr () = "Unit"
-    showRV (MaybeRepr tpr') Unassigned = "(Maybe Unassigned)"
-    showRV (MaybeRepr tpr') (PE cond rv') = "(Maybe " ++ showRV tpr' rv' ++ " " ++ show (printSymExpr cond) ++ ")"
-    showRV (VectorRepr tpr') v = "(Vector " ++ show tpr' ++ " " ++
-      unwords (map (showRV tpr') (V.toList v)) ++ ")"
-    showRV (StructRepr ctx) rvs = "(Struct" ++ showStruct ctx rvs ++ ")"
-    showRV MirReferenceRepr ref = "(Ref " ++ show ref ++ ")"
-    -- Add more cases as needed
-    showRV tpr _ = "[unsupported: " ++ show tpr ++ "]"
-
-    showStruct :: CtxRepr ctx -> Ctx.Assignment (RegValue' sym) ctx -> String
-    showStruct Ctx.Empty Ctx.Empty = ""
-    showStruct (ctx Ctx.:> tpr) (rvs Ctx.:> RV rv) =
-      showStruct ctx rvs ++ " " ++ showRV tpr rv
 
 bindFn _symOnline _cs fn cfg =
   ovrWithBackend $ \bak ->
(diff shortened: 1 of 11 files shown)


crux-mir/src/Mir/Overrides.hs
```diff
@@ -430,16 +431,39 @@ bindFn symOnline cs name cfg
...
+
+  | hasInstPrefix ["crucible", "dump_rv"] explodedName
+  , Empty :> MirSliceRepr :> tpr <- cfgArgTypes cfg
+  , UnitRepr <- cfgReturnType cfg
+  = bindFnHandle (cfgHandle cfg) $ UseOverride $
+    mkOverride' "dump_rv" UnitRepr $ do
+        RegMap (Empty :> RegEntry _ strRef :> RegEntry _ expr) <- getOverrideArgs
+        str <- getString strRef >>= \x -> case x of
+            Just x -> return x
+            Nothing -> fail $ "dump_rv: desc string must be concrete"
```
RyanGlScott: Redundant `$`:

```suggestion
            Nothing -> fail "dump_rv: desc string must be concrete"
```

crux-mir/src/Mir/Overrides.hs
```diff
@@ -430,16 +431,39 @@ bindFn symOnline cs name cfg
...
 
+  | ["crucible", "print_str"] == explodedName
+  , Empty :> MirSliceRepr <- cfgArgTypes cfg
+  , UnitRepr <- cfgReturnType cfg
+  = bindFnHandle (cfgHandle cfg) $ UseOverride $
+    mkOverride' "print_str" UnitRepr $ do
+        RegMap (Empty :> RegEntry _ strRef) <- getOverrideArgs
+        str <- getString strRef >>= \x -> case x of
+            Just x -> return x
+            Nothing -> fail $ "print_str: desc string must be concrete"
```
RyanGlScott: Similarly:

```suggestion
            Nothing -> fail "print_str: desc string must be concrete"
```

crux-mir/src/Mir/Overrides.hs
```diff
@@ -453,6 +477,24 @@ bindFn symOnline cs name cfg
           pfxInit == edidInit &&
           "_inst" `Text.isPrefixOf` edidLast
 
+    showRV :: TypeRepr tp' -> RegValue sym tp' -> String
```
RyanGlScott: I wonder if this should return a `Doc` instead of a `String`, and then have call sites convert the `Doc` to a `String` at the outermost level by calling `show` on it. In particular, I think this pretty-printing logic would benefit from having the `prettyprinter` library intelligently lay out the text in a way that overflows lines based on the width of the page so as to avoid overly long lines. (`printSymExpr` in particular is susceptible to this when very symbolic expressions are involved.)

crux-mir/src/Mir/Overrides.hs
```diff
@@ -453,6 +477,24 @@ bindFn symOnline cs name cfg
...
+    showRV (BVRepr _) x = show (printSymExpr x)
+    showRV AnyRepr (AnyValue tpr' rv') = "(AnyValue " ++ show tpr' ++ " " ++ showRV tpr' rv' ++ ")"
+    showRV UnitRepr () = "Unit"
+    showRV (MaybeRepr tpr') Unassigned = "(Maybe Unassigned)"
```
RyanGlScott: For consistency with the `PE` case, should this print the `TypeRepr`?

RyanGlScott (commented):
Awesome! `print_str` in particular scratches the itch I had in https://github.com/GaloisInc/crucible/issues/1311, and I would even be willing to go as far as to say that this PR fixes https://github.com/GaloisInc/crucible/issues/1311.

crux-mir/src/Mir/Overrides.hs
sauclovian-g: Yes please, though I suspect it's a can of worms and maybe shouldn't go in this patch.

langston-barrett: This looks a lot like [`ppRegVal`](https://github.com/GaloisInc/crucible/blob/04a125618d986472ec0bde03b1c2d96678bac0ea/crucible-debug/src/Lang/Crucible/Pretty.hs#L61)

spernsteiner: Reworked this to use `ppRegVal` (thanks for the tip!)

crux-mir/src/Mir/Overrides.hs
spernsteiner: I went the other direction when adding `MaybeRepr` support to `ppRegVal` and made neither case print the `TypeRepr`.  Most of the other `ppRegVal` cases don't print any `TypeRepr`s, so I figured it was more consistent to leave it out.

PR #1470, opened 2025-07-14
crux-mir: add tests for Arc<dyn Fn> and Rc<dyn Fn>

### LINKED ISSUE #1399: `crucible-mir`: Support custom dynamically sized types

spernsteiner:
We fixed #1399 with the goal of making `Arc<dyn Fn>` work, but never added a test for `Arc<dyn Fn>` specifically (though we do have separate tests for `&dyn Fn`, `Box<dyn Fn>`, and `Arc<dyn Trait>`).  This branch adds one.

This branch also adds a test for `Rc<dyn Fn>` and an even more basic test that uses `Rc<i32>`, both of which currently fail due to the implementation of `Rc` using unsupported features (unrelated to the features we needed for `Arc<dyn Fn>`).

### DIFF
--- a/crux-mir/test/conc_eval/clos/fn_dyn_arc.rs
+++ b/crux-mir/test/conc_eval/clos/fn_dyn_arc.rs
@@ -1,0 +1,15 @@
+use std::sync::Arc;
+
+fn call_closure_arc(mut f: Arc<dyn Fn(i32, i32) -> i32>) -> i32 {
+    let r = f(1, 2);
+    r
+}
+
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> i32 {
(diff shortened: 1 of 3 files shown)


RyanGlScott (approved):
It might be worth opening follow-up issues about the `Rc`-related functions we'd need to patch or override in order to make the `Rc` test cases work.

spernsteiner:
Opened GaloisInc/mir-json#143 for `Rc` support

Issue #1108, opened 2023-09-11
`crucible-mir` translation is sensitive to `static` name ordering


RyanGlScott:
`crux-mir` is successfully able to verify this program:

```rs
static A: Option<u32> = Some(42);
static B: Option<u32> = A;

#[crux::test]
pub fn f() -> u32 {
    B.unwrap_or(27)
}
```
```
$ cabal run exe:crux-mir -- test.rs 
test test/f225316c::f[0]: returned 42, ok

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] All goals discharged through internal simplification.
[Crux] Overall status: Valid.
```

So far, so good. What happens if we rename the `static` values, however? This program should be equivalent to the first one, as we have simply swapped the names for `A` and `B`:

```rs
static B: Option<u32> = Some(42);
static A: Option<u32> = B;

#[crux::test]
pub fn f() -> u32 {
    A.unwrap_or(27)
}
```

But `crux-mir` fails to run this version of the program:

```
$ cabal run exe:crux-mir -- test.rs 
test test/efc0f956::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/efc0f956::f[0] counterexamples ----
[Crux] Found counterexample for verification goal
[Crux]   test.rs:2:25: 2:26: error: in test/efc0f956::A[0]
[Crux]   Attempted to read uninitialized reference cell
[Crux] Found counterexample for verification goal
[Crux]   test.rs:2:25: 2:26: error: in test/efc0f956::A[0]
[Crux]   attempted to read empty mux tree

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] Goal status:
[Crux]   Total: 2
[Crux]   Proved: 0
[Crux]   Disproved: 2
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Invalid.
```

Here is what I think is going on:

1. When translating the `static` values (using the [`transStatics`](https://github.com/GaloisInc/crucible/blob/602cb8f04ec5f42e76a2ba0833f7e91a95d7109c/crucible-mir/src/Mir/Trans.hs#L2254-L2305) function), `crucible-mir` will translate each of the statics that appear in the `statics :: Map DefId Static` field of a `Collection`.
[...]

PR #1471, opened 2025-07-14
`crucible-mir`: Translate "constant" statics before non-constant ones

RyanGlScott:
`mir-json` emits two types of static item definitions: ones with constant initializer expressions, and ones without them. It is possible to write a program where a non-constant static item's definition depends on a constant static item, and in order for `crucible-mir` to handle such a program, it is important that `crucible-mir` translate the constant static item _before_ the non-constant static item that uses it.

This is difficult to arrange for in the general case (see https://github.com/GaloisInc/crucible/issues/1108), but in the common case, we can make this much more likely to work by always translating every constant static item before every non-constant static item. This provides a partial workaround (but not a full fix) for https://github.com/GaloisInc/crucible/issues/1108.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -28,6 +28,9 @@ This release supports [version
 * Type translation functions like `tyToRepr` now fail gracefully
   so that failed translations can be handled by upstream tooling
   instead of failing using `error`
+* Fix a bug in which static items with non-constant initializer expressions that
+  depend on static items with constant initializer expressions would fail to
+  simulate correctly.
 
 # 0.4 -- 2025-03-21
 
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -3002,7 +3002,14 @@ transCollection col halloc = do
                 }
 
 -- | Produce a crucible CFG that initializes the global variables for the static
--- part of the crate
+-- part of the crate.
+--
+-- Note that while static items' definitions can depend on other static items,
+-- this function does not currently translate static items in a way that
+-- respects dependency order in general. (See #1108.) We do make a limited
+-- effort to minimize the chances of static items being translated in the wrong
+-- order, however. See the comments near @constValStatics@/@nonConstValStatics@
+-- below.
 transStatics ::
   (?debug::Int,?customOps::CustomOpMap,?assertFalseOnError::Bool) =>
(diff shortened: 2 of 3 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -3042,10 +3042,28 @@ transStatics colState halloc = do
   -- TODO: make the name of the static initialization function configurable
   let initName = FN.functionNameFromText "static_initializer"
   initHandle <- FH.mkHandle' halloc initName Ctx.empty C.UnitRepr
+  let allStatics :: [Static]
+      allStatics = map snd (Map.toList (colState^.collection.statics))
+  -- Partition the static items into those with constant initializer expressions
+  -- (constValStatics) and those non-constant initializer expressions
```
```suggestion
  -- (constValStatics) and those with non-constant initializer expressions
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -3042,10 +3042,28 @@ transStatics colState halloc = do
   -- TODO: make the name of the static initialization function configurable
   let initName = FN.functionNameFromText "static_initializer"
   initHandle <- FH.mkHandle' halloc initName Ctx.empty C.UnitRepr
+  let allStatics :: [Static]
+      allStatics = map snd (Map.toList (colState^.collection.statics))
```
samcowger: Perhaps:
```suggestion
      allStatics = Map.elems (colState^.collection.statics)
```

crucible-mir/CHANGELOG.md
```diff
@@ -28,6 +28,9 @@ This release supports [version
 * Type translation functions like `tyToRepr` now fail gracefully
   so that failed translations can be handled by upstream tooling
   instead of failing using `error`
+* Fix a bug in which static items with non-constant initializer expressions that
+  depend on static items with constant initializer expressions would fail to
+  translate.
```
spernsteiner: Small nitpick - they would translate fine (no `error` or `mirFail`), but the `static_initializer` function (which runs automatically before the test) would fail at run time due to reading from a pointer to uninitialized memory.

spernsteiner (approved):

crucible-mir/CHANGELOG.md
RyanGlScott: I changed "fail to translate" to "fail to simulate correctly".

RyanGlScott:
I've expanded `transStatics`' docstring to mention its current shortcomings with respect to #1108.

PR #1472, opened 2025-07-15
crucible-mir: adjust `cloneShimTuple`'s representation of unit expressions

samcowger:
Before, it would represent units as empty struct-like expressions (using `StructRepr`s). This leads to problems in cases like this PR's `clone_unit.rs` test, whose `my_clone` function yields this MIR:

```
fn clone_unit/43042312::my_clone[0]::_inst581cc2350a9a65a8[0](_1 : &()) -> () {
   let mut _0 : ();
   bb0: {
      call( core/7bc31080::clone[0]::Clone[0]::clone[0]::_shim581cc2350a9a65a8[0]( _1 )
      , (_0, bb1) )
   }
   bb1: {
      return;
   }
}
```

`bb0`'s assignment of the clone shim's output to `_0` would trigger a translation error, complaining about this `StructRepr`/`UnitRepr` mismatch:

```
Translation error in <...my_clone...>: ill-typed assignment of StructRepr [] to UnitRepr (TyTuple []) LBase (Var {_varname = "_0", _varmut = Mut, _varty = TyTuple [], _varIsZST = True})
```

### DIFF
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -1956,6 +1956,12 @@ cloneShimDef ty parts = CustomOp $ \_ _ -> mirFail $ "cloneShimDef not implement
 
 -- | Create an 'IkCloneShim' implementation for a tuple or closure type.
 cloneShimTuple :: [Ty] -> [M.DefId] -> CustomOp
+cloneShimTuple [] [] = CustomMirOp $ \_ops ->
+    -- We're cloning unit, i.e. `()`. Without this case, deferring to
+    -- `buildTupleMaybeM` to construct `()` results in a struct-like expression
+    -- with an empty `StructRepr`. In this particular case, that's wrong - what
+    -- we need to produce is a `UnitRepr` expression.
+    pure $ MirExp C.UnitRepr $ R.App E.EmptyApp
 cloneShimTuple tys parts = CustomMirOp $ \ops -> do
     when (length tys /= length parts) $ mirFail "cloneShimTuple: expected tys and parts to match"
     lv <- case ops of
--- a/crux-mir/test/conc_eval/tuple/clone_unit.rs
+++ b/crux-mir/test/conc_eval/tuple/clone_unit.rs
@@ -1,0 +1,19 @@
+//! Test that the clone shim generated for unit values is appropriately typed.
+//! All we really want to do is clone `()` - we only use the generalized
+//! `my_clone`, rather than e.g. `().clone()`, to ensure rustc doesn't optimize
+//! out the faulty clone shim.
+
+fn my_clone<T: Clone>(x: &T) -> T {
+    x.clone()
+}
+
+#[cfg_attr(crux, crux::test)]
+pub fn f() {
+    let x = ();
+    let y = my_clone(&x);
+    assert!(x == y);
+}
+
+fn main() {
+    println!("{:?}", f());
+}


crux-mir/test/conc_eval/tuple/clone_unit.rs
```diff
@@ -0,0 +1,23 @@
+//! Test that the clone shim generated for unit values is appropriately typed.
+//! What we care about doing is cloning `()` - this `MyClone` trait is only used
+//! as a bit of indirection to prevent rustc from optimizing out a simpler
+//! expression like `().clone()`.
+
+trait MyClone: Clone {}
+
+impl MyClone for () {}
```
spernsteiner: This seems unused - `my_clone` requires `T: Clone`, not `T: MyClone`

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1956,6 +1956,12 @@ cloneShimDef ty parts = CustomOp $ \_ _ -> mirFail $ "cloneShimDef not implement
 
 -- | Create an 'IkCloneShim' implementation for a tuple or closure type.
 cloneShimTuple :: [Ty] -> [M.DefId] -> CustomOp
+cloneShimTuple [] [] = CustomMirOp $ \_ops ->
+    -- We're cloning unit, i.e. `()`. Without this case, deferring to
+    -- `buildTupleMaybeM` to construct `()` results in a struct-like expression
+    -- with an empty `StructRepr`. That's not wrong in general, but it's wrong
+    -- for us here - what we need to produce is a `UnitRepr` expression.
```
spernsteiner: It is wrong in general - everything in crux-mir expects `TyTuple []` to be represented as `UnitRepr`, not `StructRepr []`.

crux-mir/test/conc_eval/tuple/clone_unit.rs
samcowger: Whoops, you're right. I was experimenting with how much I could minimize indirection, and I guess I minimized it more than I thought. I'll remove the trait and update the comment.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Hmm, I see. If that's the case, then wouldn't we want to discriminate on empty/nonempty tuple construction in `buildTupleMaybe`? When we discussed this yesterday, I thought you said that specifying it there might disrupt other parts of crucible, because they might expect empty tuples to use an empty `StructRepr`, but now I'm wondering if I misunderstood.

samcowger:
> Please double-check the `--print-mir` output to make sure the new test is actually using a clone shim for `<()>::clone`. (It would be nice to automate that check, but currently I don't think we have a convenient way to do that.) Otherwise, this PR looks good.

It is definitely using the shim, notwithstanding the spurious `MyClone` trait. I copied the MIR in the PR description/commit straight from the output of `--print-mir` on this branch, and I copied the error message from trying and failing to run `crux-mir` on this test on `master`.

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: I think `buildTupleMaybe` is also used for closures, which do use `StructRepr []` for the empty case.

RyanGlScott: Tricky! Perhaps we should instead say "In this particular case, that's wrong - what we need to produce [...]".

RyanGlScott (approved):

crucible-mir/src/Mir/TransCustom.hs
samcowger: Ah, thanks for clarifying; I've updated the wording.

PR #1474, opened 2025-07-16
`crucible-mir`: implement `fnPtrShimDef` for `TyFnPtr`

samcowger:
This allows calling function _pointers_ via `Fn` call shims. This is distinct from calling function _handles_ via those same shims - see the included test for an illustration of this distinction.

### DIFF
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -1907,12 +1907,12 @@ fnPtrShimDef fnTy = case fnTy of
                     Copy lv -> return lv
                     Move lv -> return lv
                     _ -> mirFail $ "unsupported argument tuple operand " ++ show argTuple ++
-                        " for fnptr shim of " ++ show defId
+                        " for fnptr shim of type " ++ show fnTy
                 let argOps = zipWith (\ty i -> Move $ LProj argBase (PField i ty)) argTys [0..]
                 callExp defId argOps
             ty -> mirFail $ "unexpected argument tuple type " ++ show ty ++
-                " for fnptr shim of " ++ show defId
-        _ -> mirFail $ "unexpected arguments " ++ show ops ++ " for fnptr shim of " ++ show defId
+                " for fnptr shim of type " ++ show fnTy
+        _ -> mirFail $ "unexpected arguments " ++ show ops ++ " for fnptr shim of type " ++ show fnTy
     _ -> CustomOp $ \_ _ -> mirFail $ "fnPtrShimDef not implemented for " ++ show fnTy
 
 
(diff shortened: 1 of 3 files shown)


crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1897,22 +1897,39 @@ ctpop = (["core", "intrinsics", "ctpop"],
...
+        callExp defId argOps
+    TyFnPtr sig -> CustomMirOp $ \ops -> do
+        (fnPtr, argOps) <- untupleArgOps ops
+        fnPtrExp <- evalOperand fnPtr
+        Some argTys <- Ctx.fromList <$> mapM tyToReprM (sig ^. fsarg_tys)
+        Some retTy <- tyToReprM (sig ^. fsreturn_ty)
+        let fnRepr = C.FunctionHandleRepr argTys retTy
+        fnExp <- case fnPtrExp of
+            MirExp MirReferenceRepr fnRef -> MirExp fnRepr <$> readMirRef fnRepr fnRef
+            _ -> mirFail $
```
samcowger: I would appreciate reviewer confirmation that this does not need to account for `FunctionHandleRepr`-type function expressions (though if it does, the fix ought to be simple)

spernsteiner: The shim might take the `FnPtr` by value in the `FnOnce` case - try extending the test to cover `F: FnOnce` in addition to `F: Fn` and see if you get an error

samcowger: Indeed I do! I'll go ahead and allow the `FunctionHandleRepr` case.

crux-mir/test/conc_eval/fnptr/pass.rs
```diff
@@ -0,0 +1,21 @@
...
+    f()
+}
+
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> i32 {
+    let f: fn() -> i32 = make_i32;
+    // By passing `f`, a function pointer, we're ensuring that `call_closure`
+    // can interact with function _pointer_ shims in particular. It does not
+    // suffice to pass `make_i32` directly, as this resolves to a function
+    // handle, rather than a function pointer.
```
spernsteiner:
```suggestion
    // can interact with function _pointer_ shims in particular. It does not
    // suffice to pass `make_i32` directly, as this resolves to a function
    // item type, rather than a function pointer.
```
"Function item type" is the name the Rust Reference uses for `TyKind::FnDef`: https://doc.rust-lang.org/reference/types/function-item.html

samcowger: Ah, thanks!

PR #1478, opened 2025-07-22
crucible-mir: override `core::array::from_ref::crucible_array_from_ref_hook`

samcowger:
Without this override, crux-mir fails to simulate `core::array::from_ref` (and `core::slice::from_ref`, a notable caller of `array::from_ref`) properly, and errors out with a message like:

```
[Crux]   crux-mir/test/conc_eval/array/from_ref.rs:8:5: 8:17: error: in from_ref/a38569f9::crux_test[0]
[Crux]   subindex requires a reference to IntrinsicRepr MirVector [BVRepr 32], but got a reference to BVRepr 32
[Crux]   crux-mir/test/conc_eval/array/from_ref.rs:8:5: 8:17: error: in from_ref/a38569f9::crux_test[0]
```

To handle this, we take a similar approach to the one taken in `core::slice::as_array::crucible_array_from_slice_hook`: we define and override a hook function in `from_ref` to avoid simulating its non-`crucible`-compatible cast.

`core::array::from_ref::crucible_array_from_ref_hook` was defined in https://github.com/GaloisInc/mir-json/pull/147.

### DIFF
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -126,6 +126,7 @@ customOpDefs = Map.fromList $ [
                          , mem_bswap
                          , mem_crucible_identity_transmute
                          , array_from_slice
+                         , array_from_ref
 
                          , vector_new
                          , vector_replicate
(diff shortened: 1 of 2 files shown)


crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1220,6 +1221,24 @@ array_from_slice = (["core","slice", "{impl}", "as_array", "crucible_array_from_
...
 
+array_from_ref ::  (ExplodedDefId, CustomRHS)
+array_from_ref = (["core", "array", "from_ref", "crucible_array_from_ref_hook"],
+    \substs -> Just $ CustomOpNamed $ \fnName ops -> do
+        fn <- findFn fnName
+        case (fn ^. fsig . fsreturn_ty, ops) of
+            (TyRef (TyArray elemTy 1) Immut, [MirExp MirReferenceRepr elemRef]) -> do
+                -- TODO: Like `array_from_slice`, above, this would be more
+                -- correctly implemented as a type cast, so that the input and
+                -- output are aliases.
```
samcowger, line 1232: As best I could tell, `array_from_slice`'s rationale applies here, but please let me know if that's not true, and/or if we can implement this in a way that's more faithful to the proper aliasing semantics.

qsctr: I am not a Crucible expert, but this looks reasonable to me. I am not sure I fully understand the comment in `array_from_slice` though. It seems to me that the output of that function is also a `MirVector` and not a plain crucible `Vector`.

qsctr (approved):

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: > As best I could tell, `array_from_slice`'s rationale applies here

That's correct (unfortunately).  A limitation of the memory model is that it expects array pointers like `&[T; 1]` to point to an actual `MirVectorRepr`.  This is unlike slice pointers `&[T]`, where the data pointer can point to any `T`, even one that's not in an array.

In the long term I think we should move to a "flat allocation" model, where each allocation is a sequence of primitive values with no internal structure.  That would fix this issue (i.e. make `array_from_slice` and `array_from_ref` properly match Rust semantics), remove the need for typed allocations (letting us undo some library patches), remove the need for special handling of `#[repr(transparent)]`, and properly support `slice::from_ref(&x).iter()`, unsafe `ptr.byte_add(n)`, and some uses of unions.

spernsteiner (approved):

PR #1480, opened 2025-07-28
crucible-mir: override `crucible_slice_from_{ref,mut}_hook`

samcowger:
This approach is akin to the `core::array::from_ref` override in https://github.com/GaloisInc/crucible/pull/1478, but because it operates on slices, it avoids the aliasing issue in that override.

### DIFF
--- a/crux-mir/test/conc_eval/slice/from_ref.rs
+++ b/crux-mir/test/conc_eval/slice/from_ref.rs
@@ -1,0 +1,23 @@
+use std::cell::Cell;
+
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> i32 {
+    // We want to check that our override of `crucible_slice_from_ref_hook`
+    // properly aliases its input and output. We do so by making the conversion
+    // source mutable via `Cell`, and checking that writes to that source can be
+    // observed by reading the conversion target, and vice versa.
+    let foo: Cell<i32> = Cell::new(42);
+    let foo_slice: &[Cell<i32>] = core::slice::from_ref(&foo);
+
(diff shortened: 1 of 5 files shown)


crux-mir/test/conc_eval/slice/from_ref.rs
```diff
@@ -0,0 +1,19 @@
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> i32 {
+    let foo: i32 = 42;
+    let foo_slice: &[i32] = core::slice::from_ref(&foo);
+
+    // We want to check that our override of `crucible_slice_from_ref_hook`
+    // properly aliases its input and output. We do so by converting `foo_slice`
+    // into a mutable reference, and checking to see whether writes to it can be
+    // observed by reading from `foo`.
+    let foo_mut: &mut i32 = unsafe { foo_slice.as_ptr().cast_mut().as_mut().unwrap() };
```
spernsteiner: Converting `&T` to `&mut T` is undefined behavior.  You can check the aliasing by using `Cell<i32>` instead, which doesn't require creating `&mut T`.

samcowger: Done.

samcowger:
> Can you please add another test that tries to iterate over the result of `slice::from_ref`? I'd expect this to fail - it will try to do arithmetic on the data pointer to get the end pointer for the iterator, but pointer arithmetic isn't allowed because the data pointer doesn't point into an array. Maybe also test `&foo_slice[1..]`, which should fail for a similar reason.

Done in a8c6be0.

PR #1481, opened 2025-07-29
`crucible-mir`: test `[u8; N]` allocation, part 2

samcowger:
Like #1479 (b9ecbb0), this pulls in `mir-json` changes that fix issues with `[u8; N]` allocation - in particular, to support `Vec<[u8; N]>`s - and tests them. These changes come from https://github.com/GaloisInc/mir-json/pull/150.


crux-mir/test/conc_eval/vec/of_u8_array.rs
```diff
@@ -1,19 +1,28 @@
...
+fn test_macro() -> [u8; 4] {
+    let mut v: Vec<[u8; 4]> = alloc::vec!([5, 6, 7, 8]);
+    v[0]
+}
+
 fn main() {
-    println!("{:?}", test());
+    println!("{:?}", test_push());
+    println!("{:?}", test_macro());
 }
```
spernsteiner: Does this actually work with `cabal test`?  I don't think we have any other `conc_eval` tests with multiple `#[crux::test]` functions in the same file.

spernsteiner (approved):
Looks good, assuming there's no issue with multiple tests in one file

crux-mir/test/conc_eval/vec/of_u8_array.rs
samcowger: Ah, I didn't even think of that. I was just running `crux-mir` directly on this file.

Local experimentation suggests that this approach is _almost_ compatible with `cabal test`. The test harness seems to have no issues comparing output for multiple tests, but `crux-mir` seems to run a file's tests in alphabetical order by function name: https://github.com/GaloisInc/crucible/blob/5c9211099ec54bd34c179877b7d639d264767374/crux-mir/src/Mir/Language.hs#L290

Seems like that was done in 3f6510c0a0dbb717b56df2f04be9e04e11316ad5 as part of better mimicking `cargo` output. Would you prefer to maintain that behavior and split this up into multiple files, or do you think it'd be acceptable to remove that `List.sort` call?

spernsteiner: Unsorted output seems like it would be harder to read, so I think it would be better to split the test case into two.

RyanGlScott (approved):

PR #1487, opened 2025-08-06
`crucible-mir`: Add type layout info and fix `align_of`/`size_of` implementation

qsctr:
Incorporates the layout information added in GaloisInc/mir-json#152 by storing it in a new field `layouts :: Map Ty (Maybe Layout)` in `Collection`.

Also fixes the implementation of the `align_of` and `size_of` nullops and intrinsics to use the actual layout information instead of placeholder values.

This bumps the required `mir-json` schema version to 3 for `crucible-mir` and `crux-mir`.

### DIFF
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -279,9 +279,13 @@ data Collection = Collection {
     _statics   :: !(Map DefId Static),
     _vtables   :: !(Map VtableName Vtable),
     _intrinsics :: !(Map IntrinsicName Intrinsic),
+    -- | This corresponds to the @tys@ table from @mir-json@ and is cleared
+    -- after uninterning. The @TyName -> Ty@ mappings are used for uninterning
+    -- the types in the rest of the 'Collection', and the uninterned @(Ty, Maybe
+    -- Layout)@ pairs are saved into '_layouts'.
     _namedTys  :: !(Map TyName (Ty, Maybe Layout)),
     -- | Layouts for known types. If the value is 'Nothing' then the type is
-    -- unsized.
+    -- unsized. This is not populated until uninterning.
     _layouts   :: !(Map Ty (Maybe Layout)),
     -- | Map the original 'DefId's for lang items to their custom, @$lang@-based
     -- 'DefId's (e.g., map @core::option::Option@ to @$lang/Option@).
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -118,7 +118,21 @@ data Ty =
       | TyInterned TyName
       deriving (Eq, Ord, Show, Generic)
 
-data NamedTy = NamedTy { _ntName :: Text, _ntTy :: Ty }
+data NamedTy = NamedTy
+  { _ntName :: Text
+  , _ntTy :: Ty
+    -- | If 'Nothing' then the type is unsized.
+  , _ntLayout :: Maybe Layout
+  }
+  deriving (Eq, Ord, Show, Generic)
+
+-- | Layout of a type.
+data Layout = Layout
+  { -- | ABI alignment in bytes
+    _layAlign :: Word64
+    -- | Size in bytes
+  , _laySize :: Word64
+  }
   deriving (Eq, Ord, Show, Generic)
 
 data LangItem = LangItem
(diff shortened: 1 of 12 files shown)


crux-mir/test/conc_eval/mem/align_of.rs
```diff
@@ -0,0 +1,8 @@
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> usize {
+    align_of::<u8>()
```
spernsteiner: Might be more interesting to test with a type whose alignment isn't 1.  Or maybe have the test return `[usize; 4]` and check the alignments of several different types.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -684,17 +684,23 @@ evalBinOp bop mat me1 me2 =
...
-    return $ MirExp UsizeRepr $ R.App $ usizeLit 1
-transNullaryOp M.UbChecks _ = do
-    -- Disable undefined behavior checks.
-    -- TODO: re-enable this later, and fix the tests that break
-    -- (see https://github.com/GaloisInc/mir-json/issues/107)
-    return $ MirExp C.BoolRepr $ R.App $ E.BoolLit False
+transNullaryOp nop ty =
+  case nop of
+    M.AlignOf -> lookupLayout layAlign
+    M.SizeOf -> lookupLayout laySize
```
spernsteiner: There are also overrides for `size_of` and `align_of` in `TransCustom.hs`, which should be updated similarly.  I'm not sure whether the overrides or the `NullOp`s actually get used.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -684,17 +684,23 @@ evalBinOp bop mat me1 me2 =
...
+    M.UbChecks -> do
+      -- Disable undefined behavior checks.
+      -- TODO: re-enable this later, and fix the tests that break
+      -- (see https://github.com/GaloisInc/mir-json/issues/107)
+      return $ MirExp C.BoolRepr $ R.App $ E.BoolLit False
+  where
+    lookupLayout layField = do
```
RyanGlScott: Worth giving this a type signature, I think, since it took me a bit to realize that the argument is a lens.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -684,17 +685,24 @@ evalBinOp bop mat me1 me2 =
...
+      return $ MirExp C.BoolRepr $ R.App $ E.BoolLit False
+  where
+    lookupLayout :: Getter Layout Word64 -> MirGenerator h s ret (MirExp s)
+    lookupLayout layFieldLens = do
+      lays <- use (cs . collection . layouts)
+      case Map.lookup ty lays of
+        Just (Just lay) -> return $
+          MirExp UsizeRepr $ R.App $ usizeLit $ toInteger $ lay ^. layFieldLens
+        Just Nothing -> mirFail $ fmt nop ++ " on unsized type " ++ fmt ty
+        Nothing -> mirFail $ fmt nop ++ ": no layout info for " ++ fmt ty
```
RyanGlScott: This lookup-related code is nearly copy-pasted between here and in the two custom overrides in `TransCustom`. Worth factoring out this code, I'd say.

crucible-mir/src/Mir/Mir.hs
```diff
@@ -265,7 +279,10 @@ data Collection = Collection {
     _statics   :: !(Map DefId Static),
     _vtables   :: !(Map VtableName Vtable),
     _intrinsics :: !(Map IntrinsicName Intrinsic),
-    _namedTys  :: !(Map TyName Ty),
+    _namedTys  :: !(Map TyName (Ty, Maybe Layout)),
+    -- | Layouts for known types. If the value is 'Nothing' then the type is
+    -- unsized.
+    _layouts   :: !(Map Ty (Maybe Layout)),
```
RyanGlScott: The Haddocks should state that there is a precondition that the entries in this `Map` should precisely correspond to the pairs in the range of `_namedTys` (and that this `Map` only exists for caching purposes).

RyanGlScott (approved):

crucible-mir/src/Mir/Mir.hs
qsctr: `_namedTys` gets deleted in the uninterning step right after parsing, so this is the only place the `Layout`s will remain. I'll add a haddock for `_namedTys`.

PR #1489, opened 2025-08-08
crucible: New API for `abortedGlobals`, `exec{State,Result}Globals`

### LINKED ISSUE #1488: crucible: Fix panic in `abortedResultGlobals`
The `abortedGlobals` function can panic, because it calls `globalMuxFun` when its preconditions are not met.

https://github.com/GaloisInc/crucible/blob/db67d851256a55cc3a0546f145ff1ad421221463/crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs#L443-L456

See this GREASE commit: https://github.com/GaloisInc/grease/commit/1c68496645941c5aac9d98877e7db17292437895

The best fix is probably letting API consumers decide how to handle muxes. For the GREASE use-case, for instance, we have a model from the solver (`GroundEvalFn`) that we can use to pick between the branches. Other use-cases might just want to return a partial result (e.g., `Maybe`). We could accomplish this by adding a callback parameter to `abortedGlobals` that takes a `PredSym` and two global states, and returns a global state in some `Applicative`.

I don't have a standalone test case.

langston-barrett:
These functions previously internally called `globalMuxFun` when its preconditions weren't met, resulting in a panic. Instead, let callers decide how to handle muxes.

This is not breaking, because these functions haven't been through a release cycle.

Fixes #1488.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
+++ b/crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
@@ -169,7 +169,7 @@ import           Lang.Crucible.CFG.Extension (StmtExtension, ExprExtension)
 import           Lang.Crucible.FunctionHandle (FnHandleMap, HandleAllocator, mkHandle')
 import           Lang.Crucible.Simulator.CallFrame
 import           Lang.Crucible.Simulator.Evaluation (EvalAppFunc)
-import           Lang.Crucible.Simulator.GlobalState (SymGlobalState, globalMuxFn)
+import           Lang.Crucible.Simulator.GlobalState (SymGlobalState)
 import           Lang.Crucible.Simulator.Intrinsics (IntrinsicTypes)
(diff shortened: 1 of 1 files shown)


crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
```diff
@@ -441,47 +441,50 @@ execStateSimState = \case
   InitialState _ _ _ _ _         -> Nothing
 
 abortedGlobals ::
-  IsSymInterface sym =>
-  sym ->
-  IntrinsicTypes sym ->
+  Monad f =>
+  -- | How to handle 'AbortedBranch'
```
RyanGlScott: It might be worth listing some common examples of things one can do here (e.g., `globalMuxFn` or returning a `Maybe`).

RyanGlScott (approved):

PR #1490, opened 2025-08-11
crucible: Make `CrucibleType` into `type data`

langston-barrett:
`-XTypeData` enables the declaration of type-level datatypes without the need for `-XDataKinds`-style "promotion" from the value level to the type level. `CrucibleType` is only ever used at the type level, so it is a prime candidate for a declaration as `type data`. In fact, it was one of the use-cases for the GHC proposal introducing the extension! Practically speaking, this lets us reduce a bunch of duplication and boilerplate.

This functionality was introduced in GHC 9.6, so using it explicitly drops support for GHC 9.4. This is in line with our versioning policy as described in `doc/dev.md`, which states that we aim to support the three most recent versions of GHC at a time. We currently support GHC 9.6-9.10.

See:
https://github.com/ghc-proposals/ghc-proposals/blob/3b4e702f217884cf1c2d9cfbd9cb22ad642b22b0/proposals/0106-type-data.rst https://downloads.haskell.org/ghc/9.6.1/docs/users_guide/9.6.1-notes.html

### DIFF
--- a/crucible/src/Lang/Crucible/Types.hs
+++ b/crucible/src/Lang/Crucible/Types.hs
@@ -39,46 +39,29 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TypeData #-}
 {-# LANGUAGE TypeFamilies #-}
 {-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 {-# LANGUAGE ViewPatterns #-}
 module Lang.Crucible.Types
   ( -- * CrucibleType data kind
-    type CrucibleType
+    type CrucibleType(..)
     -- ** Constructors for kind CrucibleType
-  , AnyType
-  , UnitType
   , BoolType
-  , NatType
   , IntegerType
   , RealValType
   , SymbolicStructType
   , ComplexRealType
   , BVType
-  , FloatType
   , IEEEFloatType
-  , CharType
   , StringType
-  , FunctionHandleType
-  , MaybeType
-  , RecursiveType
-  , IntrinsicType
-  , VectorType
-  , SequenceType
-  , StructType
-  , VariantType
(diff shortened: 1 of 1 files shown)


RyanGlScott:
This looks promising. I am slightly hesitant about pulling the trigger here, however. Not because of backwards compatibility with old versions of GHC, but rather because I know of [at least one `TypeData` bug](https://gitlab.haskell.org/ghc/ghc/-/issues/25739) that can cause Haddock to crash when rendering the documentation for modules that re-export definitions that leverage `TypeData`. It's not entirely clear to me under what circumstances this bug triggers, however—I note that `crucible`'s Haddocks still build without issue even though `Lang.Crucible.CFG.Core` re-exports `Lang.Crucible.Types`...

langston-barrett:
Gotcha. I see that the bug is fixed, though only as of July 7. The fix is likely not on the most recent GHC patch releases, nor do I see any information in the GHC thread about backporting.

What do you think the best way forward is? We could (1) merge as this doesn't seem to be affecting us right now, (2) test with more repos, e.g., Macaw (though apparently [only Crucible re-exports this module](https://github.com/search?q=org%3AGaloisInc+%2Fmodule+Lang%5C.Crucible%5C.Types%2F&type=code)), (3) wait until our newest supported GHC has a patch for the bug, (4) wait until all of our supported GHCs have the fix, or (5) something else?

RyanGlScott:
The [merge request which fixed the GHC issue](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/13993) has been marked to be backported to several GHC release series. (Thus far, it has been backported to GHC 9.10.3, which is slated to be released [later this month](https://mail.haskell.org/pipermail/ghc-devs/2025-August/022088.html).

Personally, I am inclined to wait until all supported GHCs have this fix, but if you are keen on landing this, then I won't object.

langston-barrett:
Update: the fix is in 9.{10,12}.3, but has not been backported to GHC 9.6 nor 9.8. That may change, but if it doesn't we can merge this after they both slip out of our support window, after #1693.

PR #1491, opened 2025-08-11
`crucible-mir`: support `Subslice` projections on slices

samcowger:
See the included tests for examples of Rust syntax that results in these projections appearing in MIR. Although these projections are also valid on arrays - see e.g. the included xfail test, which Rust can comprehend but Crucible cannot - Crucible's representation of arrays hinders us from supporting them. See comments in code for details.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1628,21 +1628,7 @@ evalPlaceProj ty (MirPlace tpr ref meta) (M.Subslice fromIndex toIndex fromEnd) 
           newMeta = SliceMeta newLen
       in MirPlace elemTpr <$> mirRef_offset headRef firstIndex <*> pure newMeta
 
-    -- TODO: Crucible's representation of `MirReference`s to arrays is not
-    -- expressive enough to implement this case properly. A Crucible reference
-    -- to an array is a reference to the entire array, and cannot have
-    -- `mirRef_offset` applied to it - unlike slices, a Crucible reference to
-    -- which is represented as a reference to its first element.
-
-    -- A couple possible ways past this:
-    -- - Make Crucible array references actually point to the first element of
-    --   the array, mimicking Crucible's approach to slices and Rust itself's
-    --   approach to arrays. Then, they'd be amenable to `mirRef_offset`.
-    -- - Implement a new reference path extension, something like
-    --   "MirRef_Subarray". Such a path extension would behave somewhat like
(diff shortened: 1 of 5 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -1618,6 +1618,46 @@ evalPlaceProj _ pl (M.Downcast _idx) = return pl
...
+    --   approach to arrays. Then, they'd be amenable to `mirRef_offset`.
+    -- - Implement a new reference path extension, something like
+    --   "MirRef_Subarray". Such a path extension would behave somewhat like
+    --   `MirRef_Index`, but instead of offering read/write access to a single
+    --   element of the parent array, it would offer read/write access to a
+    --   subarray of elements of the parent array.
+    --
+    -- NB: after implementing this, update `Mir.Mir.typeOfProj`, which currently
+    -- declares that `Subslice` unconditionally yields a slice. This is
+    -- incorrect; `Subslice` applied to an array will yield an array.
```
RyanGlScott: I think it would be helpful to turn all of this context into a `crucible` issue, then link back to the issue number here.

RyanGlScott (approved):
Please mention support for simulating slice subslices in the `crucible-mir` changelog.

crux-mir/test/conc_eval/slice/subslice.rs
```diff
@@ -0,0 +1,37 @@
+#[cfg_attr(crux, crux::test)]
+fn crux_test() {
+    let mut xs: [u32; 6] = [1, 2, 3, 4, 5, 6];
+    let ys: &mut [u32] = &mut xs;
+
+    let [_, rest @ ..] = ys else { panic!() };
+    assert_eq!(*rest, [2, 3, 4, 5, 6]);
```
spernsteiner, line 7: Does `slice/subslice.rs` exercise the non-`fromEnd` case, or does that only apply to arrays?

spernsteiner (approved):

crux-mir/test/conc_eval/slice/subslice.rs
samcowger: It looks like all the examples in this test get represented with `fromEnd == True`, and all the examples in the array test get represented with `fromEnd == False`.

crucible-mir/src/Mir/Trans.hs
samcowger: Done - see #1494.

PR #1492, opened 2025-08-11
ci: manually `cabal build` packages

samcowger:
It turns out `cabal test` will not build a package unless it contains a test suite - so if a package lacks a test suite and fails to build, CI would not catch such a failure. Unfortunately, this seems to have happened for `crucible-mir-debug` (introduced in https://github.com/GaloisInc/crucible/pull/1486). This addition ought to prevent this in the future.

### DIFF
--- a/.github/workflows/crux-llvm-build.yml
+++ b/.github/workflows/crux-llvm-build.yml
@@ -180,9 +180,6 @@ jobs:
         run: .github/ci.sh test crucible-llvm-cli
         if: runner.os == 'Linux'
 
-      - name: Test crucible-llvm-debug
-        run: .github/ci.sh test crucible-llvm-debug
-
       - name: Test crux-llvm
         run: .github/ci.sh test crux-llvm
         if: runner.os == 'Linux' || runner.os == 'macOS'
--- a/.github/workflows/crux-mir-build.yml
+++ b/.github/workflows/crux-mir-build.yml
@@ -165,7 +165,6 @@ jobs:
 
       - run: .github/ci.sh test crucible-mir-syntax
       - run: .github/ci.sh test crucible-mir-cli
-      - run: .github/ci.sh test crucible-mir-debug
       - run: .github/ci.sh test crux-mir
 
       - uses: actions/cache/save@v4
--- a/.github/workflows/crux-mir-build.yml
+++ b/.github/workflows/crux-mir-build.yml
@@ -146,7 +146,18 @@ jobs:
       - name: Generate source distributions
         run: cabal sdist crucible-syntax crucible-concurrency crucible-mir crux-mir
 
-      - run: cabal build exe:crux-mir
+      - run: >
+          cabal build
+          crucible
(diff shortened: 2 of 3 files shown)


RyanGlScott:
I like the spirit of this PR, although I disagree with the particulars of the fix. To me, if you call `test` on a particular `cabal` target, then it really ought to just call `cabal test`, full stop. If you want to `cabal build` something, then I think we should declare that explicitly.

On a related note, the problem is that we do the following:

https://github.com/GaloisInc/crucible/blob/1689cf98297378aa019c0e115591f8a8a6c823d6/.github/workflows/crux-llvm-build.yml#L183-L184

https://github.com/GaloisInc/crucible/blob/1689cf98297378aa019c0e115591f8a8a6c823d6/.github/workflows/crux-mir-build.yml#L157

Neither of these packages have a test suite, so calling `cabal test` on them will be a no-op, which means that absolutely no code will be built from either package. Eek! We really ought to call `cabal build exe:crucible-{llvm,mir}-debug` instead. We should also make sure that we aren't accidentally skipping over the executable components of any other packages.

samcowger:
Yeah, I guess this PR reflects a bit of bias. I personally feel that saying `test <pkg>` (regardless of the language or tool implementing `test`) _should_ entail building it, and I wanted to eliminate surprise when that didn't happen (especially since it seems to have happened a few times now, to the tools' detriment). I recognize that that opinion isn't universal, though, so I'm amenable to a different approach.

To be clear, it sounds like you'd prefer something like adding an explicit `cabal build <pkg>` (or perhaps adding and using a new `.github/ci.sh build <pkg>` command) before each occurrence of `.github/ci.sh test <pkg>`?

langston-barrett:
FWIW, from a build time perspective it is ideal to build all of the packages in a CI workflow together in one invocation to `cabal build`, as is done in the `crux-llvm` workflow:

https://github.com/GaloisInc/crucible/blob/1689cf98297378aa019c0e115591f8a8a6c823d6/.github/workflows/crux-llvm-build.yml#L127-L137

with `--semaphore`, this allows Cabal and GHC to make optimal use of the parallelism available on the runner.

RyanGlScott:
Yes, I would prefer explicitly calling `cabal build` before we call `cabal test`. I like @langston-barrett's suggstion of `cabal build`ing all of the relevant packages en masse before doing any testing. I think something like this ought to work:

```
cabal build
crucible
crucible-cli
crucible-concurrency
crucible-debug
crucible-mir
crucible-mir-cli
crucible-mir-debug
crucible-mir-syntax
crucible-syntax
crux-mir
```

Once we do that, then we should be able to get rid of the `cabal test crucible-{llvm,mir}-debug` steps, which don't do anything.

RyanGlScott:
Separately, I've pushed a commit which should address the compilation error that I accidentally introduced in https://github.com/GaloisInc/crucible/pull/1486.

langston-barrett (approved):

PR #1503, opened 2025-08-21
crux-mir: add MirAggregateType and use it for TyTuple

spernsteiner:
This branch adds an initial implementation of `MirAggregateType`, as described in #1498.  See the comments on `MirAggregate` in `Mir/Intrinsics.hs` for details.  This also updates the translation to use `MirAggregateRepr` for `TyTuple` and `TyClosure`.

Since mir-json doesn't export layout information yet, for now crucible-mir always stores element `i` of the tuple in bytes `i .. i + 1`.  Also, flattening of nested `MirAggregate`s is not yet implemented: a type like `(u8, (u8, u8))` is represented as a `MirAggregate` that contains a bitvector and another `MirAggregate`.

This is ready for review, but I don't plan to merge it until I have a matching saw-script PR ready to go, so the two repos don't get out of sync.

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -727,7 +727,7 @@ muxMirAggregate sym itefns c (MirAggregate sz1 m1) (MirAggregate sz2 m2) = do
   return $ MirAggregate sz1 m'
   where
     muxEntries off e1 e2 = muxMirAggregateEntry sym itefns off' c e1 e2
-      where off' = (fromIntegral :: Int -> Word) off
+      where off' = fromIntegral off
 
 -- | Return the @(offset, regValue)@ pair for each entry whose type is @tpr@.
 -- When performing a typed access, these are all the entries that the access


RyanGlScott:
> Since mir-json doesn't export layout information yet

Isn't this the case as of https://github.com/GaloisInc/mir-json/pull/152?

crucible-debug/src/Lang/Crucible/Pretty.hs
```diff
@@ -106,8 +106,10 @@ ppRegVal iFns tp (C.RV v) =
     C.FunctionHandleRepr args ret -> ppFnVal args ret v
     C.MaybeRepr tpr -> case v of
       W4.Unassigned -> "Nothing"
-      W4.PE cond v' ->
-        "Just" PP.<+> ppRegVal iFns tpr (C.RV v') PP.<+> "if" PP.<+> W4.printSymExpr cond
+      W4.PE cond v'
+        | Just True <- W4.asConstantPred cond -> "Just" PP.<+> ppRegVal iFns tpr (C.RV v')
+        | otherwise ->
+          "Just" PP.<+> ppRegVal iFns tpr (C.RV v') PP.<+> "if" PP.<+> W4.printSymExpr cond
```
RyanGlScott: To avoid duplicating some code, how about writing this like:

```hs
      W4.PE cond v' ->
          let ppCond | Just True <- W4.asConstantPred cond ->
                       "if" PP.<+> W4.printSymExpr cond
                     | otherwise ->
                       PP.empty in
          "Just" PP.<+> ppRegVal iFns tpr (C.RV v') PP.<+> ppCond
```

spernsteiner:
> > Since mir-json doesn't export layout information yet
> 
> Isn't this the case as of [GaloisInc/mir-json#152](https://github.com/GaloisInc/mir-json/pull/152)?

That only outputs size and alignment info, not the offsets of the fields.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -588,6 +590,108 @@ muxRef sym iTypes c (MirReferenceMux mt1) (MirReferenceMux mt2) =
...
+  sym ->
+  IntrinsicTypes sym ->
+  Word ->
+  Pred sym ->
+  Maybe (MirAggregateEntry sym) ->
+  Maybe (MirAggregateEntry sym) ->
+  IO (MirAggregateEntry sym)
+muxMirAggregateEntry sym itefns offset c mEntry1 mEntry2 =
+  case (mEntry1, mEntry2) of
+    (Nothing, Nothing) -> error "impossible"
```
RyanGlScott: I think this should `panic` instead.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -588,6 +590,108 @@ muxRef sym iTypes c (MirReferenceMux mt1) (MirReferenceMux mt2) =
...
+    fail $ "bad MirAggregate merge: size " ++ show sz1 ++ " != " ++ show sz2
+  m' <- sequence $ IntMap.mergeWithKey
+    (\offset entry1 entry2 -> Just $ muxEntries offset (Just entry1) (Just entry2))
+    (IntMap.mapWithKey $ \offset entry1 -> muxEntries offset (Just entry1) Nothing)
+    (IntMap.mapWithKey $ \offset entry2 -> muxEntries offset Nothing (Just entry2))
+    m1 m2
+  return $ MirAggregate sz1 m'
+  where
+    muxEntries off e1 e2 = muxMirAggregateEntry sym itefns off' c e1 e2
+      where off' = (fromIntegral :: Int -> Word) off
```
RyanGlScott: Style nit pick: I'd prefer to use type applications here instead:

```hs
      where off' = fromIntegral @Int @Word off
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -690,6 +707,162 @@ muxMirAggregate sym itefns c (MirAggregate sz1 m1) (MirAggregate sz2 m2) = do
     muxEntries off e1 e2 = muxMirAggregateEntry sym itefns off' c e1 e2
       where off' = (fromIntegral :: Int -> Word) off
 
+-- | Return the `(offset, regValue)` pair for each entry whose type is `tpr`.
```
RyanGlScott: Haddock nit pick: avoid trying to hyperlink things that aren't identifiers:

```hs
-- | Return the @(offset, regValue)@ pair for each entry whose type is @tpr@.
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -690,6 +707,162 @@ muxMirAggregate sym itefns c (MirAggregate sz1 m1) (MirAggregate sz2 m2) = do
...
+    (\(o, MirAggregateEntry _ tpr' rv) ->
+      case testEquality tpr tpr' of
+        Just Refl -> Just (fromIntegral o, rv)
+        Nothing -> Nothing)
+    (IntMap.toAscList m)
+
+wordLit :: IsSymInterface sym => sym -> Word -> IO (RegValue sym UsizeType)
+wordLit sym o = bvLit sym knownNat (BV.mkBV knownNat (fromIntegral o))
+
+-- | Lift `iteFn` from type `tp` to type `MaybeType tp`.
```
RyanGlScott: Similarly:

```hs
-- | Lift @iteFn@ from type @tp@ to type @MaybeType tp@.
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -690,6 +707,162 @@ muxMirAggregate sym itefns c (MirAggregate sz1 m1) (MirAggregate sz2 m2) = do
...
+  Pred sym ->
+  RegValue sym (MaybeType tp) ->
+  RegValue sym (MaybeType tp) ->
+  IO (RegValue sym (MaybeType tp))
+liftIteFnMaybe sym _tpr iteFn c x y =
+  mergePartial sym (\c' x' y' -> lift $ iteFn c' x' y') c x y
+
+readMirAggregateWithSymOffset ::
+  forall sym bak tp.
+  (IsSymBackend sym bak) =>
```
RyanGlScott: Redundant parentheses:

```hs
  IsSymBackend sym bak =>
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1247,11 +1250,24 @@ mkTraitObject traitName vtableName e = do
...
                 "of trait", show traitName, ":", show vtableTy, "!=", show vtableTy']
 
-    return $ buildTuple
+    return $ buildStructExp
         [ e
         , packAny vtable
         ]
 
+    where
+        buildStructExp :: [MirExp s] -> MirExp s
```
RyanGlScott: Unlike the similarly named `buildStruct'` function, this does _not_ wrap each field in a `MaybeRepr`. Is it worth leaving a comment about why that is not done here?

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -573,33 +572,43 @@ buildArrayLit trep exps = do
...
 
-accessAggregateMaybe (MirExp ty a) b = mirFail $ "invalid access of " ++ show ty ++ " at field (maybe) " ++ (show b)
+getTupleElemTyped :: HasCallStack => MirExp s -> Int -> M.Ty -> MirGenerator h s ret (MirExp s)
+getTupleElemTyped (MirExp tpr ag) i ty = do
+    col <- use $ cs . collection
+    case isZeroSized col ty of
+        False -> do
+            let tpr' = MirAggregateRepr
+            Refl <- testEqualityOrFail tpr tpr' $
+                "getTupleElem: expected tuple to be " ++ show tpr' ++ ", but got " ++ show tpr
```
RyanGlScott: s/getTupleElem/getTupleElemTyped/

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -573,33 +572,43 @@ buildArrayLit trep exps = do
...
+            Some tpr <- tyToReprM ty
+            MirExp tpr <$> mirAggregate_get (fromIntegral i) 1 tpr ag
+        True -> do
+            mVal <- initialValue ty
+            case mVal of
+                Just x -> return x
+                Nothing -> error "zero-sized type with no initialValue?"
```
RyanGlScott: Should this `panic` instead of `error`ing?

spernsteiner: Is there ever a reason to use `error` over `panic`?  I wasn't thinking about it when I wrote this code, but isn't the general rule to always use `panic`, which gives better error messages?

RyanGlScott: We generally stick to the convention that `panic` is reserved for code paths that are not intended to be reachable unless there is a bug somewhere in the implementation of the code. We use other forms of error reporting (`error`, `fail`, etc.) for error paths that end users might see.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Fixed comments in eb52d390aa9d79a5aadf6be34212c95b74fa9cae

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Fixed comments in eb52d390aa9d79a5aadf6be34212c95b74fa9cae

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Switched `error` to `panic` in 6014ebf6a6cac4ad43fdcadab14a4e0136b6f403

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Switched `error` to `mirFail` in 6014ebf6a6cac4ad43fdcadab14a4e0136b6f403

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: I think the type ascription is clearer because the reader doesn't need to know which type parameter is which

RyanGlScott: OK. If you go that route, I would personally prefer to just give `off'` a top-level type signature (e.g., `off' :: Word; off' = ...`), since that adds less visual clutter in the expression part. I don't feel strongly about this, however, so I'll leave this to your judgment.

crucible-debug/src/Lang/Crucible/Pretty.hs
spernsteiner: Deduplicated in da984da77b4491d68651920ecbd2b27b9c9d8250

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Cleaned up in 1b2f3f5d2057ff19b64a058133bfe55ae4bc0c69

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Fixed in 1b2f3f5d2057ff19b64a058133bfe55ae4bc0c69

crucible-mir/src/Mir/Trans.hs
spernsteiner: Renamed `buildStructExp` and added a comment to avoid confusion with `TransTy.buildStruct`: de3ff7218f54e9f04d94088d095fe63c760aca0a

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: I removed the ascription entirely (in f05829f811e920e6873d055b2ecad0d8faf0f6fc). It was necessary for type inference at some point, but I guess not any more.

spernsteiner:
GaloisInc/saw-script#2567 is ready, so I'll merge this now and then merge that one.

PR #1506, opened 2025-08-26
crux-mir: Add test for #[derive(Symbolic)]

glguy:


RyanGlScott (approved):
LGTM, assuming that the CI passes.

RyanGlScott (commented):
Actually, I do see one thing that needs to be done: make sure to bump the `mir-json` submodule to bring in the necessary changes!

PR #1507, opened 2025-08-26
Add RME as an offline solver for Crux

glguy:


dependencies/mir-json
RyanGlScott, line 1: Why does `mir-json` need to be bumped here?

crux/src/Crux/Config/Solver.hs
```diff
@@ -113,24 +114,26 @@ asOnlyOfflineSolver s =
   case s of
     "dreal" -> pure DReal
     "boolector" -> pure Boolector
+    "rme" -> pure RME
```
RyanGlScott, line 117: Don't forget to update the Haddocks of this function to mention RME.

RyanGlScott (commented):
Thanks for doing this!

dependencies/mir-json
glguy: *fixed*

RyanGlScott (approved):

PR #1511, opened 2025-08-28
`crux-mir`: Add test case for formatting `{u,i}128` values

### LINKED ISSUE #1509: `crucible-mir`: Translation error when displaying a `u128`/`i128` value
If you try to display a value of type `u128` or `i128`, such as in this example:

```rs
// test.rs
extern crate crucible;
use crucible::*;

#[crux::test]
pub fn f() {
    let x: u128 = 42;
    crucible_assert!(false, "x is {:?}", x);
}
```

Then `crucible-mir` will fail to translate it:

```
test test/b83b98b4::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/b83b98b4::f[0] counterexamples ----
[Crux]   ./libs/core/src/fmt/num.rs:569:17: 569:28: error: in core/fae87615::fmt[0]::num[0]::parse_u64_into[0]::_inst07f20e5c9e24637f[0]
[Crux]   Translation error in core/fae87615::fmt[0]::num[0]::parse_u64_into[0]::_inst07f20e5c9e24637f[0]: unimplemented cast: Misc
[Crux]     ty: TyFloat F64
[Crux]     as: TyUint B64

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

On the other hand, displaying other primitive integer types (e.g., `u8`, `u16`, `u32`, or `u64`) does not cause a translation error. This is ultimately because the `Display` impls for `u128` and `i128` use very different code paths than the impls for other primitive integer types do. The specific line of code that `crucible-mir` falls over on can be found [here](https://github.com/GaloisInc/mir-json/blob/7b14ef37191f81b47ba4849b9cdfe61c4f8e8f00/libs/core/src/fmt/num.rs#L569):

```rs
        if n >= 1e16 as u64 {
```

This casts `1e16` from an `f64` to a `u64`, which `crucible-mir` currently does not support. It is likely that we could implement support for such a cast, but even if we do, there might be other tricky parts of this impl lying in wait (e.g., a call to the `write_bytes` intrinsic [here](https://github.com/GaloisInc/mir-json/blob/7b14ef37191f81b47ba4849b9cdfe61c4f8e8f00/libs/core/src/fmt/num.rs#L684-L688)). (See also https://github.com/GaloisInc/crucible/issues/1510.)

For now, a crude workaround is to convert the `u128` value into an array of bytes, e.g.,

```rs
// test.rs
extern crate crucible;
use crucible::*;

#[crux::test]
pub fn f() {
    let x: u128 = 42;
    crucible_assert!(false, "x is {:?}", u128::to_be_bytes(x));
}
```
```
test test/a25b48a9::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/a25b48a9::f[0] counterexamples ----
[Crux]   ./libs/crucible/lib.rs:48:13: 55:14 !test.rs:8:5: 8:63: error: in test/a25b48a9::f[0]
[Crux]   MIR assertion at test.rs:8:5:
[Crux]   	x is [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 42]

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

RyanGlScott:
This bumps the `mir-json` submodule to bring in the changes from https://github.com/GaloisInc/mir-json/pull/160 and adds a test case.

Fixes #1509.


crux-mir/test/conc_eval/str/format_u128_i128.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg_attr(crux, crux::test)]
+fn crux_test() {
+    let s = format!("{:?} {:?}", 42u128, -42i128);
+    assert!(&s == "42 -42");
```
spernsteiner: Looking at `showRegEntry` in `Language.hs`, I think you're supposed to be able to return `&'static str` (or `[&'static str; 2]` for multiple results) and let the test harness compare to the concrete implementation.  You can use `String::leak(format!("..."))` to get the `'static` lifetime it needs.

crux-mir/test/conc_eval/str/format_u128_i128.rs
```diff
@@ -0,0 +1,9 @@
+#[cfg_attr(crux, crux::test)]
+fn crux_test() {
+    let s = format!("{:?} {:?}", 42u128, -42i128);
```
spernsteiner: Please also try it with something that doesn't fit in a u64

crux-mir/test/conc_eval/str/format_u128_i128.rs
RyanGlScott: Although `TyStr` (i.e., `str`) is supported, it doesn't appear that `&str` is supported. When I changed the implementation to this:

```rs
#[cfg_attr(crux, crux::test)]
fn crux_test() -> &'static str {
    String::leak(format!("{:?} {:?}", 42u128, -42i128))
}

pub fn main() {
}
```

Then `crux-mir` complains thusly:

```
$ cabal -v0 run crux-mir-test -- -p format_u128_i128
crux-mir
  crux concrete

      str
        format_u128_i128: FAIL (0.90s)
          Compiling and running oracle program                       (0.43s)
          Oracle output: "42 -42"                                    (0.47s)
          Crux output: I don't know how to print result of type str

            test/Test.hs:124:
            crux doesn't match oracle

1 out of 1 tests failed (0.90s)
```

As such, I'd prefer to keep the original approach of using `assert`, if you don't mind. That should still ensure that both the simulator and the concrete implementation format the string in the same way, correct?

If you'd like, I could open a separate issue about the test harness not being able to render `&str` values.

crux-mir/test/conc_eval/str/format_u128_i128.rs
RyanGlScott: I'd updated the test case to test printing the values `85000000000000000000000000000000000000u128` and `-85000000000000000000000000000000000000i128`, which do not fit in `{u,i}64`.

spernsteiner (approved):

crux-mir/test/conc_eval/str/format_u128_i128.rs
RyanGlScott: A bit late, but I've opened https://github.com/GaloisInc/crucible/issues/1519 to track being able to return `&str`s in `crux::test`s.

PR #1514, opened 2025-08-29
More detailed symbolic coverage

yav:
This MR adds support for more detailed symbolic coverage statistics.  It require `mir-json` version 4.

### DIFF
--- a/crux-mir/README.md
+++ b/crux-mir/README.md
@@ -49,7 +49,7 @@ Next, navigate to the `crucible/dependencies/mir-json` directory. Install
 [the `mir-json` README][mir-json-readme].
 
 Currently, `crux-mir` supports [version
-3](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#3) of
+4](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#4) of
 `mir-json`'s schema. Note that the schema versions produced by `mir-json` can
 change over time as dictated by internal requirements and upstream changes. To
 help smooth this over:
@@ -135,7 +135,7 @@ To compile and test a single Rust program:
 * Clone this directory with `git clone https://github.com/GaloisInc/crucible.git` as we will need the `report-coverage` utility tool
 * In your crate run:
   ```
-  $ cargo crux-test --lib --  --branch-coverage --path-sat --output-directory test-coverage
+  $ EXPORT_ALL=true cargo crux-test --lib --  --branch-coverage --path-sat --output-directory test-coverage
   ```
 * In your crate, run the following command, pointing it towards the `report-coverage` folder in `crucible` directory, which contains the coverage script. You will need to point to a directory in the `test-coverage` folder that contains a function's coverage report data (`report_data.js`):
   ```
@@ -143,6 +143,15 @@ To compile and test a single Rust program:
   ```
 
   Note that the `62f2dedb` part of this path will likely be different on your machine due to how `mir-json` works. And `$YOUR_CRATE_NAME` is the `name` specified in the crate's `Cargo.toml` file.
+
+  It is also possible to run `report-coverage` with multiple input files, in which case you will
+  get the combined coverge from all tests.  For example:
+  ```
+  find ./test-coverage -name 'report_data.js' | xargs cargo run --manifest-path $COV_REP --
+  ```
+  where `COV_REP` is the path to `report-coverage` as above.
+
+  XXX: UPDATE OUTPUT
 * This will report all paths not covered, including ones from the standard library, for example:
   ```
   warning: branch condition never has value true
--- a/crux-mir/report-coverage/src/main.rs
+++ b/crux-mir/report-coverage/src/main.rs
@@ -312,54 +312,62 @@ fn parse_branch(json: &Value) -> Result<BranchTrans, String> {
     }
 }
 
+/// Info about the coverage of a funcion
+#[derive(Default)]
+struct FnCoverage<'a> {
+    /// Did we call this function at all
+    fn_called: bool,
 
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
-struct CoverageKey<'a> {
-    fn_id: &'a str,
-    branch_span: &'a str,
+    /// Information about the branches in the functin
+    branch_cov: HashMap<&'a str, BranchCoverage>,
 }
+
+impl<'a> FnCoverage<'a> {
+
+    pub fn branch(&mut self, span: &'a str) -> &mut BranchCoverage {
+        self.branch_cov.entry(span).or_insert_with(||BranchCoverage::default())
+    }
+
+    pub fn iter_sorted<'b>(&'b self) -> impl Iterator<Item = (&'a str, &'b BranchCoverage)> + 'b {
+        let mut keys = self.branch_cov.keys().collect::<Vec<_>>();
+        keys.sort();
+        keys.into_iter().map(move |k| (*k, self.branch_cov.get(*k).unwrap()))
+    }
+}
+
 
 struct Coverage<'a> {
     /// Whether to merge multiple monomorphized instances of a function when computing coverage.
     /// If set, then an exit is considered covered if it is covered in any monomorphization of the
     /// source function; otherwise, every exit must be covered in every monomorphization.
     merge_functions: bool,
-    branches: HashMap<CoverageKey<'a>, BranchCoverage>,
+
+    // Information about the coverage a function
+    fun_cov: HashMap<&'a str, FnCoverage<'a>>
 }
 
 impl<'a> Coverage<'a> {
     pub fn new(merge_functions: bool) -> Coverage<'a> {
         Coverage {
             merge_functions,
-            branches: HashMap::new(),
+            fun_cov: HashMap::new(),
         }
     }
 
-    fn key(&self, fn_id: &'a FnId, span: &'a str) -> CoverageKey<'a> {
-        CoverageKey {
-            fn_id: if self.merge_functions {
-                strip_instance(fn_id)
-            } else {
-                fn_id as &str
-            },
-            branch_span: span,
-        }
+    pub fn fun<'b>(&'b mut self, fn_id: &'a FnId) -> &'b mut FnCoverage<'a> {
+        let k = if self.merge_functions { strip_instance(fn_id) } else { fn_id };
+        self.fun_cov.entry(k).or_insert_with(||FnCoverage::default())
     }
 
-    pub fn branch(&mut self, fn_id: &'a FnId, span: &'a str) -> &mut BranchCoverage {
-        let key = self.key(fn_id, span);
-        self.branches.entry(key).or_insert_with(BranchCoverage::default)
-    }
-
-    pub fn iter_sorted<'b>(&'b self) -> impl Iterator<Item = (&'a str, &'b BranchCoverage)> + 'b {
-        let mut keys = self.branches.keys().collect::<Vec<_>>();
+    pub fn iter_sorted<'b>(&'b self) -> impl Iterator<Item = (&'a str, &'b FnCoverage<'a>)> + 'b {
+        let mut keys = self.fun_cov.keys().collect::<Vec<_>>();
         keys.sort();
-        keys.into_iter().map(move |k| (k.branch_span, self.branches.get(k).unwrap()))
+        keys.into_iter().map(move |k| (*k, self.fun_cov.get(*k).unwrap()))
     }
 }
 
 /// If `s` ends with something that looks like a monomorphized instance disambiguator
-/// (`::inst0123456789abcdef[0]`), then remove that suffix.  If no such suffix is present, `s` is
+/// (`::_inst0123456789abcdef[0]`), then remove that suffix.  If no such suffix is present, `s` is
 /// returned unchanged.
 fn strip_instance<'a>(s: &'a str) -> &'a str {
     try_strip_instance(s).unwrap_or(s)
@@ -401,12 +409,19 @@ struct BranchCoverage {
     /// Whether this is a boolean branch.  If so, warnings will be formatted slightly differently
     /// (its values are rendered `false` and `true` instead of `0` and `1`).
     pub is_boolean: bool,
+    /// Did we visit this branch at all.
+    /// We don't report coverage warnings for branches
+    /// that never got reached, but we use them to compute
+    /// coverage stats.
+    pub visited: bool,
 }
 
 fn process<'a>(cov: &mut Coverage<'a>, fn_id: &'a FnId, report: &'a FnReport, trans: &'a FnTrans) {
     // Maps (branch ID, dest index) to the Core `BlockId` of the destination.
     let mut dest_map: HashMap<(u32, usize), BlockId> = HashMap::new();
     let mut visited_branches = HashSet::new();
+    let fn_cov = cov.fun(fn_id);
+    fn_cov.fn_called = !report.visited_blocks.is_empty();
 
     fn insert_dest(
         fn_id: &FnId,
@@ -460,11 +475,13 @@ fn process<'a>(cov: &mut Coverage<'a>, fn_id: &'a FnId, report: &'a FnReport, tr
         }
     }
 
-
-    let mut branch_ids = visited_branches.into_iter().collect::<Vec<_>>();
-    branch_ids.sort();
-    for branch_id in branch_ids {
-        let bt = &trans.branches[branch_id as usize];
+    // We go over all branches rather than just the visited ones, so
+    // we can report full stats about the coverage of a function.
+    // We record which ones were not visited so that we can report them as
+    // such and avoid spurious warnings. 
+    for (branch_id_32, bt) in trans.branches.iter().enumerate() {
+        let branch_id = branch_id_32 as u32;
+        let visited = visited_branches.contains(&branch_id);
 
         let dest_visited = |index| {
             let block_id = match dest_map.get(&(branch_id, index)) {
@@ -480,7 +497,8 @@ fn process<'a>(cov: &mut Coverage<'a>, fn_id: &'a FnId, report: &'a FnReport, tr
 
         match *bt {
             BranchTrans::Bool(_, ref span) => {
-                let bcov = cov.branch(fn_id, span);
+                let bcov = fn_cov.branch(span);
+                bcov.visited = visited;
                 bcov.is_boolean = true;
                 // The entries in `possible` and `seen` are discriminant values (0/false or
                 // 1/true), not indices.  In a boolean branch, the exit at index 0 is the true/1
@@ -500,7 +518,8 @@ fn process<'a>(cov: &mut Coverage<'a>, fn_id: &'a FnId, report: &'a FnReport, tr
             },
 
             BranchTrans::Int(ref vals, ref dests, ref span) => {
-                let bcov = cov.branch(fn_id, span);
+                let bcov = fn_cov.branch(span);
+                bcov.visited = visited;
                 for i in 0 .. dests.len() {
                     // If this destination is `Unreachable`, don't complain that the edge wasn't
                     // taken.  Branches to unreachable blocks are generated in exhaustive `match`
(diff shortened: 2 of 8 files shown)


crux-mir/README.md
```diff
@@ -135,14 +135,23 @@ To compile and test a single Rust program:
 * Clone this directory with `git clone https://github.com/GaloisInc/crucible.git` as we will need the `report-coverage` utility tool
 * In your crate run:
   ```
-  $ cargo crux-test --lib --  --branch-coverage --path-sat --output-directory test-coverage
+  $ EXPORT_ALL=true cargo crux-test --lib --  --branch-coverage --path-sat --output-directory test-coverage
```
RyanGlScott: I'm not super keen on making `EXPORT_ALL` a part of the user requirements for doing coverage reporting, as this environment variable is mostly intended to be an [internal implementation detail of `mir-json`](https://github.com/GaloisInc/mir-json/blob/dc0fb66d47d2eb0e320efb25dcedcbba510cdd2e/doc/rustc.md?plain=1#L84-L96). I wonder if we should instead add another flag to `crux-test` that turns on `EXPORT_ALL=true` and generates a `crux-mir` wrapper script that enables `--branch-coverage`.

crux-mir/report-coverage/src/main.rs
```diff
@@ -312,54 +312,62 @@ fn parse_branch(json: &Value) -> Result<BranchTrans, String> {
...
+#[derive(Default)]
+struct FnCoverage<'a> {
+    /// Did we call this function at all
+    fn_called: bool,
 
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
-struct CoverageKey<'a> {
-    fn_id: &'a str,
-    branch_span: &'a str,
+    /// Information about the branches in the functin
```
RyanGlScott:
```suggestion
    /// Information about the branches in the function
```

crux-mir/report-coverage/src/main.rs
```diff
@@ -747,35 +784,74 @@ impl Filter {
...
-        if is_boolean {
-            if !bcov.seen.contains(&0) {
-                reporter.warn(span, "branch condition never has value false");
+
+    let mut summary = HashMap::new();
+
+    for (fun, fn_cov) in cov.iter_sorted() {
+        
+        // hacky: skip core-library or crucible functions
+        if fun.starts_with("core/") || fun.starts_with ("crucible/") { continue }
```
RyanGlScott: Would the right way to do this be to only consider functions that come from the crate being tested itself?

crux-mir/README.md
yav: I modified `crux-test` to add `EXPORT_ALL=true` on ts own when it sees `--branch-coverage` so this shouldn't needed any more.

crux-mir/report-coverage/src/main.rs
```diff
@@ -401,12 +409,19 @@ struct BranchCoverage {
     /// Whether this is a boolean branch.  If so, warnings will be formatted slightly differently
     /// (its values are rendered `false` and `true` instead of `0` and `1`).
     pub is_boolean: bool,
+    /// Did we visit this branch at all.
+    /// We don't report coverage warnings for branches
+    /// that never got reached, but we use them to compute
+    /// coverage stats.
+    pub visited: bool,
```
spernsteiner: Is this field ever different from `self.seen.len() > 0 || self.default_seen`?

yav: I think so, unless the CFG has an invariant I am not aware of. In particular, I think it is possible that some of the alternatives in a branch that was never visited were visited at run time, through some other path in the CFG

spernsteiner: Is this thing only tracking which blocks were entered?  I thought it was actually recording which outgoing edges were taken at each branch.

yav: Previously the code used to iterate only over the visited branches, and count stats about them. This makes sense because we just wanted to warn when some alternative in a branch is not taken.  I modified it to iterate over all branches, so that we can report a stable coverage number in the summary.  The `visited` flag just tracks if a branch was reached at all so that we get the old warning behavior, but we count all branches in the function

spernsteiner: I was replying to this:

>I think it is possible that some of the alternatives in a branch that was never visited were visited at run time, through some other path in the CFG

I think this is not possible, because I think branch alternatives are only recorded in `seen`/`default_seen` if the branch is visited and exits through that alternative.  Reaching the same destination via a different route doesn't count.

Suppose you have this CFG:

```
    A
   / \
  B   C
 / \ /
D   E
```
And suppose you observe the paths `A -> B -> D` and `A -> C -> E`.  It will record that the branch alternative `B -> D` was seen and `B -> E` was not, even though E was visited through another path in the CFG.

Assuming I'm remembering the behavior here correctly, if B is visited at all, then at least one of its two outgoing edges must be marked as `seen`.  And if B is never visited, then neither edge will be `seen`.  In both cases, `visited` is the same as `seen.len() > 0 || default_seen`.

yav: Right, I see what you are saying.  With my changes above an alternative in a branch counts as visited if its target block is visited, which breaks that invariant.  So I added the flag, so when reporting warning we only consider alternatives whose branches were actually visited.  My thinking was that if the target of an alternative is visited we might as well count it as done, because you can't get to any more places with it, even though it might not have actually been visited.

I like your suggestion though, as it gives a coverage metric that's easier to explain, so I'll update the code so the invariant holds again.  Then the coverage metric will become "how many of the alternatives in the code were visited"

crux-mir/report-coverage/src/main.rs
yav: I don't think this is stored in the profiling data (`reportdata.js`) or the static info about the branches (`translation.json`) at the moment.  So I can think of a few options:
  (a) leave it as is,
  (b) modify the generated files to contain info about the crate containing the tests being run
  (c) add another flag, similar to `filter`, that will be something like "only show stats about functions in this crate"
  (d) same as (c) but modify the `filter` flag somehow

Any preferences?  My vote would be (a) or (c)

RyanGlScott: Without knowing much about the implementation details, my expectation would have been that the coverage tool would determine which functions to look at based on the the `tests` field of the MIR JSON. If that information isn't already present in `translation.json`, then it seems reasonable to include it. I think that corresponds to option (b).

It might also be worth having a separate flag to allow users to control which crates get profiling information generated for them, although this seems more like a quality-of-life improvement rather than something essential. I think that corresponds to option (c).

yav: Actually, it occurs to me that we can probably figure out the crate from the profiling data `reportdata.js`.  It contains all events from the profiler, so we can look at the function of the first visited block, and use that crate.

This should work as long as all the tests that we executed are in the same crate.  I am not sure if it is possible to somehow run tests from different crates with a single command, it seems unlikely.

I'll have a go at implementing this tomorrow.

crux-mir/report-coverage/src/main.rs
```diff
@@ -312,54 +323,67 @@ fn parse_branch(json: &Value) -> Result<BranchTrans, String> {
...
 }
 
+/// Info about the coverage of a funcion
+#[derive(Default)]
+struct FnCoverage<'a> {
+    /// Did we call this function at all
+    fn_called: bool,
+
+    /// Information about the branches in the function
+    branch_cov: HashMap<&'a str, BranchCoverage>,
```
RyanGlScott: It might be worth documenting that the `str`s in the `HashMap` are source spans, as this wasn't obvious to me at first glance.

crux-mir/report-coverage/src/main.rs
```diff
@@ -312,54 +323,67 @@ fn parse_branch(json: &Value) -> Result<BranchTrans, String> {
...
+    }
 
-#[derive(Clone, Copy, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
-struct CoverageKey<'a> {
-    fn_id: &'a str,
-    branch_span: &'a str,
+    pub fn iter_sorted<'b>(&'b self) -> impl Iterator<Item = (&'a str, &'b BranchCoverage)> + 'b {
+        let mut keys = self.branch_cov.keys().collect::<Vec<_>>();
+        keys.sort();
+        keys.into_iter().map(move |k| (*k, self.branch_cov.get(*k).unwrap()))
```
RyanGlScott: If I understand correctly, this is performing three conversion passes:

1. Convert `branch_cov` into a `Vec` of source spans.
2. Sort the `Vec`.
3. Convert the `Vec` into an `Iterator` of (source span, `BranchCoverage`) pairs by mapping over it.

I think we could save a pass by converting `branch_cov` to a `Vec` of (source span, `BranchCoverage`) pairs using `into_iter()`, calling `sort_by(|p, q| p.0.cmp(q.0))` to sort by the source span in each pair, and then converting it to an `Iterator` (without needing to call `map` an additional time).

crux-mir/report-coverage/src/main.rs
```diff
@@ -62,6 +62,10 @@ struct BlockId(pub String);
 #[derive(Clone, Debug, Default)]
 struct Report {
     fns: HashMap<FnId, FnReport>,
+    /// Only report coverage for the functions on this crates.
+    /// If this is empty, report coverage for al crates.
+    /// We use the first block in a profiling report to determine its crate.
+    report_crates: Vec<String>,
```
RyanGlScott: Under what circumstances will `report_crates` be empty? From the code below, it looks like this will always be populated when it encounters the first block. Is there ever a situation where there are no blocks?

crux-mir/report-coverage/src/main.rs
```diff
@@ -62,6 +62,10 @@ struct BlockId(pub String);
 #[derive(Clone, Debug, Default)]
 struct Report {
     fns: HashMap<FnId, FnReport>,
+    /// Only report coverage for the functions on this crates.
+    /// If this is empty, report coverage for al crates.
```
RyanGlScott:
```suggestion
    /// If this is empty, report coverage for all crates.
```

crux-mir/report-coverage/src/main.rs
```diff
@@ -91,11 +95,18 @@ fn parse_report_into(json: Value, r: &mut Report) -> Result<(), String> {
...
             .ok_or_else(|| format!("expected callgraph `events` field to be an array"))?;
+        let mut first_block = true;
         for evt in events {
             let evt = evt.as_object()
                 .ok_or_else(|| format!("expected event to be an object"))?;
             if evt.get("type").map_or(false, |j| j == "BLOCK") {
                 let fn_id = event_function(evt)?;
+                if first_block {
+                    if let Some(c) = get_crate(&fn_id) {
+                        r.report_crates.push(c);
```
RyanGlScott, line 106: If I'm reading this code correctly, `report_crates` is a `Vec` that can only ever have at most one element. Would it make more sense for `report_crates` to be an `Option<String>` instead to avoid the unrepresentable state of 2 or more crates?

crux-mir/report-coverage/src/main.rs
```diff
@@ -381,6 +405,24 @@ fn try_strip_instance<'a>(s: &'a str) -> Option<&'a str> {
...
+    let mut cs = s[end ..].chars();
+    if cs.next()? != ':' { return None }
+    Some(s.chars().take(start).chain(cs).collect())
+}
+
+fn get_crate(s: &str) -> Option<String> {
+    let end = s.find('/')?;
+    Some(s[..end].to_string())
+
+}
```
RyanGlScott: Style nit pick: remove the unnecessary newline at the end of this function:

```suggestion
fn get_crate(s: &str) -> Option<String> {
    let end = s.find('/')?;
    Some(s[..end].to_string())
}
```

crux-mir/report-coverage/src/main.rs
```diff
@@ -381,6 +405,24 @@ fn try_strip_instance<'a>(s: &'a str) -> Option<&'a str> {
...
+/// want to show in user facing outpus.
+fn strip_crate_hash(s: &str) -> Option<String>{
+    let start = s.find('/')? + 1;
+    let end = start + s[start ..].find(':')? + 1;
+    let mut cs = s[end ..].chars();
+    if cs.next()? != ':' { return None }
+    Some(s.chars().take(start).chain(cs).collect())
+}
+
+fn get_crate(s: &str) -> Option<String> {
```
RyanGlScott, line 425: Please add a docstring to this function stating what this does.

crux-mir/report-coverage/src/main.rs
```diff
@@ -612,6 +662,25 @@ impl Reporter {
...
 
+    fn coverage_stats(&self, fun: &str, called: bool, branch_seen: usize, branch_tot: usize) {
+        use std::io::Write;
+        use termcolor::{Color, ColorSpec, StandardStream, WriteColor};
+
+
+        let call_stat = if called { "✅" } else { "❌" }; 
+        let mut stdout = StandardStream::stdout(self.color_choice);
+        stdout.set_color(ColorSpec::new().set_fg(None)).unwrap();
+        let perc = if branch_tot == 0 { if called { 100 } else { 0 } } else { 100 * branch_seen / branch_tot };
```
RyanGlScott: I find this line of code somewhat hard to read since it packs multiple `if` expressions on the same line. How about splitting this up across multiple lines, e.g.,

```suggestion
        let perc =
            if branch_tot == 0 {
                if called { 100 } else { 0 }
            } else {
                100 * branch_seen / branch_tot
            }; 
```

crux-mir/report-coverage/src/main.rs
```diff
@@ -786,6 +904,15 @@ fn hash<H: Hash>(x: &H) -> u64 {
     hasher.finish()
 }
 
+/*
+Covergae information is computed from two files:
```
RyanGlScott:
```suggestion
Coverage information is computed from two files:
```

crux-mir/test/coverage/coverage_match.good
```diff
@@ -10,3 +10,7 @@ warning: branch condition never has value 20
 21 │     let y = match e {
    │             ^^^^^^^
 
+❌   0% coverage_match/E[0]::A[0]::{constant#0}[0]: 0/0
+❌   0% coverage_match/E[0]::B[0]::{constant#0}[0]: 0/0
+❌   0% coverage_match/E[0]::C[0]::{constant#0}[0]: 0/0
```
RyanGlScott, line 15: Hah, it took me a while to figure out what exactly wasn't being covered here. Looking at the original program:

https://github.com/GaloisInc/crucible/blob/31d05de0e3f742922d353af5ba6844e7b8a0b86c/crux-mir/test/coverage/coverage_match.rs#L1-L27

It appears that at the very least, the `crux_test` code paths involving `A` and `C` are being covered. That's not what the coverage report is complaining about, however. It turns out that `E[0]::A[0]::{constant#0}[0]` _et al._ are the numeric values associated with each enum variant (e.g., the `10` in `A = 10`). And sure enough, we don't make use of these numeric values anywhere! If you remove these numeric values, then `E[0]::A[0]::{constant#0}[0]` are no longer mentioned in the coverage report.

This is an interesting example, as it demonstrates a situation where the coverage reporting finds code that is genuinely not covered, but it isn't obvious from the report where the code actually lives. I wonder if we the coverage report should also report the source span for each function name to make this more obvious? I'm not sure if the source span for each function is readily available in the part of the code where the coverage is printed, however...

crux-mir/report-coverage/src/main.rs
yav: At the moment it shouldn't really be empty, but I was thinking that in the future we may want to add a flag to show stats for everything, which would be conveniently implemented as just setting this field to empty.

crux-mir/report-coverage/src/main.rs
yav: You can give multiple `reportdata.js` to the coverage tool, in which case there could be more than one.

crux-mir/report-coverage/src/main.rs
yav: This is how the code was before, I just split it up a bit.   I don't mind changing it, but I am not sure I follow why it eliminates a conversion: in both cases we make a vector, sort it, and then iterate over the result.

It seems to me the only difference is that in the one case we have a normal vector iterator, and in the other we have a `Map` iterator, but that doesn't seem any less efficient.  In fact, it seems to me that sorting the vector of pairs might be less efficient, because you'd have to move around the `BranchCoverage` information around as you are sorting the elements.

Or am I misunderstanding what's happening here?

crux-mir/report-coverage/src/main.rs
RyanGlScott: Ah, this is the part I was missing. In that case, a `Vec` seems like a reasonable choice.

crux-mir/report-coverage/src/main.rs
RyanGlScott: The part that seems unnecessarily inefficient is the `.map(move |k| (*k, self.branch_cov.get(*k).unwrap()))` call, as it requires performing an extra O(_n_ log(_n_)) pass to reassociate each source span with its `BranchCoverage`. But the whole reason that we have to reassociate them is because we removed their association by calling `branch_cov.keys()`! If we called `keys.into_iter()` instead (and performed sorting over pairs), then we would be able to skip this `.map(...)` call.

>  In fact, it seems to me that sorting the vector of pairs might be less efficient, because you'd have to move around the `BranchCoverage` information around as you are sorting the elements.

I'm not sure I follow what you mean here. [`Vec::sort_by`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_by) has the same time complexity has `Vec::sort`, and it only "moves around" `BranchCoverage` in the same way that it moves around the source spans (i.e., it moves them together as pairs).

yav: I see what you mean, I forgot that looking up in the HashMap is not constant time.  I was just worried that moving around the pairs, while the same complexity, will add a constant extra cost, but given that looking up is not constant, I'll change it up to just sort the pairs.

spernsteiner: > I forgot that looking up in the HashMap is not constant time

Looking up in a HashMap is constant time.  It's BTreeMap that has O(log n) lookups.

> `sort_by(|p, q| p.0.cmp(q.0))`

I believe you can use `sort_by_key(|x| x.0)` here.  The main reason to use `sort_by(... cmp ...)` over `sort_by_key` is that [the returned key can't borrow from the item in the slice](https://github.com/rust-lang/rust/issues/34162), but in this case the key is a pre-existing `&str`, which is `Copy`.

RyanGlScott: > Looking up in a HashMap is constant time.

TIL!

> I believe you can use sort_by_key(|x| x.0) here.

And TI(also)L about [`sort_by_key`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.sort_by_key). Good suggestions!

crux-mir/test/coverage/coverage_match.good
yav: I am having a lot of trouble understanding when these constants show up and when they don't.  For example:
```rust
enum T { A = 10, B = 20 }

fn f(x: T) -> u32 { x as u32 }

/// Check that ffs_fast and ffs_ref produce the same output on *every* input.
#[cfg_attr(crux, crux::test)]
fn test_num() -> u32 {
    f(T::A)
}
```
is reported like this:
```
❌   0% example_1/T[0]::A[0]::{constant#0}[0]: 0/0
❌   0% example_1/T[0]::B[0]::{constant#0}[0]: 0/0
✅ 100% example_1/f[0]: 0/0
✅ 100% example_1/test_num[0]: 0/0
```

Perhaps this is because the compiler inlined things?  Or is there some other way to "use" these numeric values?


I think maybe we should leave this as is, for the time being.

RyanGlScott: Hm, interesting. If you compile this program with `--print-mir`, you will see the following:

```
$ cabal -v0 run exe:crux-mir -- test.rs --print-mir
FNs
fn test/68eb543a::T[0]::A[0]::{constant#0}[0]() -> isize {
   let mut _0 : isize;
   bb0: {
      _0 = use(10);
      return;
   }
}
fn test/68eb543a::T[0]::B[0]::{constant#0}[0]() -> isize {
   let mut _0 : isize;
   bb0: {
      _0 = use(20);
      return;
   }
}
fn test/68eb543a::f[0](_1 : test/68eb543a::T[0]<>) -> u32 {
   let mut _0 : u32;
   let mut _2 : isize;
   let mut _3 : u8;
   let mut _4 : bool;
   let mut _5 : bool;
   let mut _6 : bool;
   bb0: {
      _2 = Discriminant(_1, isize);
      _3 = Cast(Misc, _2, u8);
      _4 = _3 >= 10;
      _5 = _3 <= 20;
      _6 = _4 & _5;
      Intrinsic[Assume](_6);
      _0 = Cast(Misc, _2, u32);
      return;
   }
}
fn test/68eb543a::test_num[0]() -> u32 {
   let mut _0 : u32;
   let mut _1 : test/68eb543a::T[0]<>;
   bb0: {
      _1 = AdtAg(test/68eb543a::T[0]::_adtc5a5405abab57fd5[0], 0, []);
      call(test/68eb543a::f[0](_1), (_0, bb1))
   }
   bb1: {
      return;
   }
}
ADTs
 Enum {_enumDiscrTy = TyInt USize} test/68eb543a::T[0]::_adtc5a5405abab57fd5[0][test/68eb543a::T[0] <>]( Variant( test/68eb543a::T[0]::A[0]
, Explicit(test/68eb543a::T[0]::A[0]::{constant#0}[0])
, []
, ConstKind
, 10
, True )
, Variant( test/68eb543a::T[0]::B[0]
, Explicit(test/68eb543a::T[0]::B[0]::{constant#0}[0])
, []
, ConstKind
, 20
, True ) )
TRAITs
VTABLEs
INTRINSICSs
test/68eb543a::T[0]::A[0]::{constant#0}[0] = IkItem test/68eb543a::T[0]::A[0]::{constant#0}[0]<<>>
test/68eb543a::T[0]::B[0]::{constant#0}[0] = IkItem test/68eb543a::T[0]::B[0]::{constant#0}[0]<<>>
test/68eb543a::f[0] = IkItem test/68eb543a::f[0]<<>>
test/68eb543a::test_num[0] = IkItem test/68eb543a::test_num[0]<<>>
STATICS
```

This shows that neither `T[0]::A[0]::{constant#0}[0]` nor `T[0]::B[0]::{constant#0}[0]` are mentioned directly in the syntax trees for the `test_num` or `f` functions. I would argue that they are "covered" in a different sense, however:

1. You'll note that `T[0]::A[0]::{constant#0}[0]` nor `T[0]::B[0]::{constant#0}[0]` are mentioned in the ADT definition for `T[0]::_adtc5a5405abab57fd5[0]` as `Explicit` numeric values that each variant takes on.
2. Moreover, in the following line in `test_num`'s definition:

   ```
   _1 = AdtAg(test/68eb543a::T[0]::_adtc5a5405abab57fd5[0], 0, []);
   ```

   The `crucible-mir` semantics for `AdtAg` will look up the numeric definition for `T[0]::_adtc5a5405abab57fd5[0]`'s 0th variant (i.e., `T[0]::A[0]::{constant#0}[0]`), pattern match on the `_0 = use(10);` assignment syntax in `T[0]::A[0]::{constant#0}[0]`'s definition (see [this code](https://github.com/GaloisInc/crucible/blob/5cc8c9f7ce3408e1b7650b14e62c0597192ec8df/crucible-mir/src/Mir/GenericOps.hs#L78-L81) in `crucible-mir`), extract out the `10` value, and use `10` as the variant's discriminant value. I claim that this should mark `T[0]::A[0]::{constant#0}[0]` as covered!

> I think maybe we should leave this as is, for the time being.

Certainly, I think fixing this is beyond the scope of this PR. That being said, I think we should open an issue to track this, since I expect that this will be a common point of confusion for users of this tool.

RyanGlScott (approved):

crux-mir/test/coverage/coverage_match.good
RyanGlScott: @yav opened https://github.com/GaloisInc/crucible/issues/1524 to track this.

PR #1515, opened 2025-09-02
Drop unnecessary `allow-newer` from `cabal.project`

langston-barrett:
The upstream issue was fixed.


cabal.GHC-9.10.1.config
```diff
@@ -38,8 +38,8 @@ constraints: any.BoundedChan ==1.0.3.0,
              any.bitvec ==1.1.5.0,
              bitvec +simd,
              any.bitwise ==1.0.0.1,
-             any.blaze-builder ==0.4.2.3,
-             any.boomerang ==1.4.9.3,
+             any.blaze-builder ==0.4.4.1,
+             any.boomerang ==1.4.9.4,
```
RyanGlScott, line 42: Any reason not to use the same `boomerang` version in the other `cabal.GHC-*.config` files as well?

PR #1522, opened 2025-09-04
`crux`: Print commit hash in `--version` output

### LINKED ISSUE #1416: `crux`: Print commit hash in `--version` output
Currently, `crux`'s `--version` command simply prints out the version number, e.g.,

```
$ ~/Software/crux-llvm-0.10/bin/crux-llvm --version
[Crux] version: 0.7.2, crux-llvm version: 0.10
```

In addition to this, it would also be useful to print out the commit hash, similarly to how Cryptol and SAW do it:

```
$ ~/Software/cryptol-3.3.0/bin/cryptol --version
Cryptol 3.3.0
Git commit 285c30d32e38ba50476260f7d5b4fac6afd16471
    branch release-3.3.0
FFI enabled

$ ~/Software/saw-1.3/bin/saw --version
1.3 (f887384 (HEAD detached at v1.3))
```

This would be particularly useful for nightly builds.

RyanGlScott:
This extends `crux`'s `--version` flag to also print out `git` commit hash information when available. By default, this consults the output of the `git` tool at compile time (by way of the `githash` Haskell library), but if `git` is not available locally, then we embed commit information from an external `crux.buildinfo.json` file. This allows embedding commit information in environments where `git` is not available (e.g., Docker image builds).

Fixes #1416.


.github/workflows/crux-mir-build.yml
```diff
@@ -243,6 +243,10 @@ jobs:
         with:
           submodules: true
 
+      - name: Generate crux.buildinfo.json
+        shell: bash
+        run: .github/ci.sh generate_buildinfo "${{ github.sha }}" main
```
langston-barrett: It looks as if `main` is treated as the name of a git branch elsewhere, so perhaps this should be [`${{ github.ref_name }}`](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts#github-context)? Or at least `master`, since this repo uses `master` as the main branch?

(Same comment applies to the `crux-llvm` workflow.)

langston-barrett (approved):

.github/workflows/crux-mir-build.yml
RyanGlScott: Good catch! This was a copy-paste error from bringing this code over from `grease`, which only builds its Docker images on a `main` branch. You are correct in that we should be doing something different here. I've opted for `${{ github.ref_name }}`, since we trigger Docker image builds on both the `master` branch and in release branches with different names.

Issue #1527, opened 2025-09-10
`crux-mir-test`: Add the ability to have per-test configuration tweaks, similar to `crux-llvm-test`


RyanGlScott:
One of the nice features of `crux-llvm`'s test suite is that each test case (be it a `.c` file, a `.bc` file, or otherwise) can optionally be accompanied by a `.config` file with the same basename. This `.config` file can override default Crux settings that are useful for testing certain properties, such as whether Crux works with a specific SMT solver.

We should add a similar capability to `crux-mir`'s test suite. For instance, this would be useful for testing a fix for https://github.com/GaloisInc/crucible/issues/1526, which requires a non-default solver to trigger.

PR #1528, opened 2025-09-10
`crux-mir`: Don't add failing assertions for unsatisfiable paths in `crucible::concretize`

### LINKED ISSUE #1526: `crux-mir` crashes when using `bitwuzla` solver (`user error (Could not pop from empty entry stack.)`)
Given this program:

```rs
// test.rs
extern crate crucible;
use crucible::*;

fn to_bits(x: u8) -> [bool; 8] {
    [
        x & (1 << 7) != 0,
        x & (1 << 6) != 0,
        x & (1 << 5) != 0,
        x & (1 << 4) != 0,
        x & (1 << 3) != 0,
        x & (1 << 2) != 0,
        x & (1 << 1) != 0,
        x & (1 << 0) != 0,
    ]
}

#[crux::test]
fn test() {
    let x = u8::symbolic("x");
    let expected = to_bits(x);
    let actual = to_bits(x);
    crucible_assert!(expected == actual, "{:?}", x);
}
```

`crux-mir` can verify it with the default choice of SMT solver:

```
$ crux-mir test.rs
test test/63647c0a::test[0]: [Crux] Attempting to prove verification conditions.
ok

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 1
[Crux]   Disproved: 0
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Valid.
```

If you pick `bitwuzla` as the SMT solver, however, then `crux-mir` crashes:

```
$ crux-mir test.rs -s bitwuzla
user error (Could not pop from empty entry stack.)test test/63647c0a::test[0]:
```

RyanGlScott:
Previously, if we encountered an unsatisfiable path condition when attempting to get the concrete model in `crucible::concretize`, it would add a failing Crucible assertion. This is not quite what we want, however, as this can wreak havoc if the call to `crucible::concretize` is in the middle of a solver frame, as seen in the oddities observed in #1526.

This patch tweaks the `crucible::concretize` implementation to simply return the original argument unchanged if the path condition is unsatisfiable. See the newly added comments for a justification of why this is sensible.

Fixes #1526.


RyanGlScott:
This PR does not add a test case for a somewhat silly reason: it's because the `crux-mir` test suite does not yet have the capability to add per-test tweaks that override Crux's default settings. See https://github.com/GaloisInc/crucible/issues/1527. If you'd like, I could develop that capability before landing this PR, although it would take some doing.

crux-mir/src/Mir/Overrides.hs
```diff
@@ -140,24 +140,40 @@ concretize (Just (SomeOnlineSolver bak)) = do
...
         cond <- getPathCondition bak
         result <- checkWithAssumptionsAndModel sp "concretize" [cond]
         case result of
-            Sat f -> return f
-            _ -> addFailedAssertion bak $
-                   GenericSimError "path is already unreachable"
-    let evalBase :: forall bt . BaseTypeRepr bt -> SymExpr sym bt -> IO (SymExpr sym bt)
-        evalBase btr v = evalGround v >>= groundToSym sym btr
+            Sat f -> return $ Just f
+            _ -> pure Nothing
```
samcowger: Nit: could you make these both be either `pure` or `return`?

crux-mir/src/Mir/Overrides.hs
```diff
@@ -140,24 +140,40 @@ concretize (Just (SomeOnlineSolver bak)) = do
...
+
+          regEval bak (\btpr exp -> liftIO $ evalBase btpr exp) tpr val
+
+        -- If the current path condition is not satisfiable, then return the
+        -- original argument unchanged. This is fine to do since the path will
+        -- be deemed unreachable anyway. Moreover, we don't want to simply
+        -- abort here, since doing so might cause the current assumption frame
+        -- to become unbalanced (see the TODO RGS test case).
+        Nothing ->
+          pure val
```
samcowger, line 177: Would it be appropriate/useful to mimic `addFailedAssertion`'s approach of adding a proof obligation for `False`, but then not abort afterwards? i.e., do everything except `abortExecBecause`:
https://github.com/GaloisInc/crucible/blob/71b789941031be94e3f2e7f1359e5e099cf679bf/crucible/src/Lang/Crucible/Backend.hs#L380-L390

samcowger:
> This PR does not add a test case for a somewhat silly reason: it's because the `crux-mir` test suite does not yet have the capability to add per-test tweaks that override Crux's default settings. See #1526. If you'd like, I could develop that capability before landing this PR, although it would take some doing.

I think you meant to reference #1527.

crux-mir/src/Mir/Overrides.hs
RyanGlScott: Yes, it's probably wise to do this to be on the safe side. I've updated the code accordingly.

PR #1533, opened 2025-09-10
Save the names of the test function when running with coverage

### LINKED ISSUE #1530: Incorrect heuristic for crate durin coverage reporting
When reporting coverage, we generally want to restrict the coverage stats to the functions in the crate containing the tests that being measured.   Unfortunately, at present the names of the tests are not stored anywhere in the data directly, so we used a heuristic that crate is determine from the first function that was executed by the test.

It would appear that this heuristic does not work in some cases as the first function executed can be some initialization function from `core`, and so we erroneously chose `core` as the crate to report stats for.

The solution here might be to just augment the data with the names of the test functions so we don't need to use any heuristic.
We should probably also add a flag to allow the user more control over what they want to see (e.g., if they want to see all function coverage, not just the functions in one crate)

yav:
This fixes #1530:  when running with coverage we save the names of the tests as an additional piece of data.  We then
use the names of the tests to determine which functions to include in the coverage statistics.

### DIFF
--- a/crux-mir/report-coverage/src/main.rs
+++ b/crux-mir/report-coverage/src/main.rs
@@ -62,10 +62,6 @@ struct BlockId(pub String);
 #[derive(Clone, Debug, Default)]
 struct Report {
     fns: HashMap<FnId, FnReport>,
-    /// Only report coverage for the functions on this crates.
-    /// If this is empty, report coverage for all crates.
-    /// We use the first block in a profiling report to determine its crate.
-    report_crates: Vec<String>,
 }
 
 #[derive(Clone, Debug, Default)]
@@ -95,18 +91,12 @@ fn parse_report_into(json: Value, r: &mut Report) -> Result<(), String> {
             .ok_or_else(|| format!("callgraph section has no `events` field"))?;
         let events = events.as_array()
             .ok_or_else(|| format!("expected callgraph `events` field to be an array"))?;
-        let mut first_block = true;
+
         for evt in events {
             let evt = evt.as_object()
                 .ok_or_else(|| format!("expected event to be an object"))?;
             if evt.get("type").map_or(false, |j| j == "BLOCK") {
                 let fn_id = event_function(evt)?;
-                if first_block {
-                    if let Some(c) = get_crate(&fn_id) {
-                        r.report_crates.push(c);
-                    }
-                }
-                first_block = false;
                 r.fns.entry(fn_id).or_insert_with(FnReport::default)
                     .visited_blocks.extend(event_blocks(evt)?);
             } else if evt.get("type").map_or(false, |j| j == "BRANCH") {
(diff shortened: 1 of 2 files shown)


crux-mir/report-coverage/src/main.rs
```diff
@@ -944,10 +934,12 @@ fn main() {
     let mut report = Report::default();
     let mut trans = None;
     let mut trans_hash = None;
+    let mut top_crates = HashSet::new();
 
     for report_path_str in m.values_of_os("input").unwrap() {
         let report_path = Path::new(report_path_str);
         let trans_path = report_path.with_file_name("translation.json");
+        let tests_path = report_path.with_file_name("tests.json");
```
RyanGlScott, line 942: Rather than emitting two separate JSON files, would it make more sense to combine the two files' contents into a single JSON dictionary?

yav: I did that at first, but this works better with how the coverage reporter code works.  In particular, the `translation.json` is exactly the same for every test, so when processing multiple outputs, the coverage reporter parses `translation.json` once and hashes it, and then if it encounters more `translation.json`s, it just checks that they have the same hash as the original one.  If we modified the `translation.json` to include the test names, then the files become different for silly reasons.

Btw, this behavior of the coverage reporter means that it won't work if you run it on tests from different crates, so this might be something we want to fix, if it is important, but even then, it is more convenient if the tests are separate.  The fix would be something like: parse the tests, determine the crate, and then process `translation.json` depending on the crate.   There's probably more stuff we'd need to do to combine the information from the different `translation.json`s too, so we should decide if this is a feature we want to support.

podhrmic:
EDIT: I might be having some version mistmatches, looking into it more

@yav I might be missing something here - I rebuilt `crux-mir` with this commit, and I am now getting the following errors:

```
---- ring/626ee24f::digest[0]::sha2[0]::rustcrypto_cryptol_test[0]::test_coverage_with_crux[0] counterexamples ----
[Crux]   internal: error: in ring/626ee24f::digest[0]::SHA1_FOR_LEGACY_USE_ONLY[0]
[Crux]   Translation error in ring/626ee24f::digest[0]::SHA1_FOR_LEGACY_USE_ONLY[0]: error building struct ring/626ee24f::digest[0]::Algorithm[0]: got Nothing for mandatory field 6 type:AnyRepr
Model:
[]
```

Is this some mismatch in the JSON format, or is there something else going on? The tests/proofs work just fine with the `master` version of `crux-mir`

yav:
Hm, weird, there should be nothing in the PR that causes this.  Let me try it out over here

RyanGlScott:
That error message looks like it might potentially be fallout from https://github.com/GaloisInc/crucible/pull/1503. Do you know what commit you were using before in which this problem didn't occur?

RyanGlScott:
On second thought, `git blame` suggests that that error message existed long before https://github.com/GaloisInc/crucible/pull/1503. In that case, I don't really know why the behavior would have changed—it would be good to come up with a minimal example of what is going on.

PR #1537, opened 2025-09-12
Add a tag to avoid relying on case

### LINKED ISSUE #1536: Support case insensitve file system when profiling the simulator
When running crucible with profiling enabled, we generate some files containing information about what the simulator did (`report_data.json`).  
For `crux-mir` and `crux-mir-comp` the locaiton of the files is derived from a hash of the crate and the name of the test being run.

Unfortunately, on systems with case insensitive file systems, such as MacOS and Windwos, files can overwrite each other, if the tests only differ in the case (e.g., test `f` and test `F`).

One possible workaround would be to hash the name of function and tag the location of the file with this additional hash.

yav:
Fixes #1536

### DIFF
--- a/crux-mir/README.md
+++ b/crux-mir/README.md
@@ -142,7 +142,7 @@ To compile and test a single Rust program:
   $ cargo run --manifest-path $PATH_TO_CRUCIBLE_REPO/report-coverage/Cargo.toml -- test-coverage/$YOUR_CRATE_NAME$/62f2dedb\:\:f\[0\]#aaa/report_data.js
   ```
 
-  Note that the `62f2dedb` part of this path will likely be different on your machine due to how `mir-json` works. And `$YOUR_CRATE_NAME` is the `name` specified in the crate's `Cargo.toml` file.  The `#aaa` suffix is a tag derived from the function name to avoid name collisions on case insensitive file systems.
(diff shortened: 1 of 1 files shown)


crux-mir/src/Mir/Language.hs
```diff
@@ -476,6 +476,25 @@ runTestsWithExtraOverrides initS bindExtra (cruxOpts, mirOpts) = do
       else
         return ExitSuccess
 
+-- | Add a prefix to the string, that will be different depending on the
+-- cases of the letters.  We use it to avoid name collision on case
+-- insensitive file systems.
```
RyanGlScott: Can you give some concrete examples of input/output pairs to make it more obvious how this function works?

crux-mir/src/Mir/Language.hs
```diff
@@ -476,6 +476,25 @@ runTestsWithExtraOverrides initS bindExtra (cruxOpts, mirOpts) = do
       else
         return ExitSuccess
 
+-- | Add a prefix to the string, that will be different depending on the
```
RyanGlScott: Based on the implementation, I think this should be:

```suggestion
-- | Add a suffix to the string, that will be different depending on the
```

RyanGlScott (approved):
Make sure to also update the relevant parts of the `crux-mir` README that are affected by this change in behavior:

https://github.com/GaloisInc/crucible/blob/9c8d9c78617b3666611af46bf98c5cbf566bdae1/crux-mir/README.md#L142

https://github.com/GaloisInc/crucible/blob/9c8d9c78617b3666611af46bf98c5cbf566bdae1/crux-mir/README.md#L175

Issue #1534, opened 2025-09-11
`crux-mir` can fail in the presence of unused, unsupported `static`s


samcowger:
Recall that Crucible does not generally support code directly involving Rust `union`s (https://github.com/GaloisInc/crucible/issues/1429). Consider this Rust:
```rs
// unions.rs
union Bad {
    x: u8,
    y: i8,
}

pub static FOO: Bad = Bad { x: 42 };

pub fn unused() -> u8 { unsafe { FOO.x } }

#[cfg_attr(crux, crux::test)]
pub fn used() -> u8 { 42 }
```
`crux-mir` on its own can run on this file and simulate/verify `used` without issue. SAW's `mir_verify`, however, can't - it fails with a translation error for `FOO`:
```
// unions.saw
enable_experimental;

m <- mir_load_module "unions.linked-mir.json";

print "Loaded MIR";

let spec = do {
    mir_execute_func [];
    mir_return (mir_term {{ 42 : [8] }});
};
mir_verify m "unions::used" [] false spec z3;
```
```sh
> saw-rustc unions.rs
> saw unions.saw
[20:12:16.761] Loading file "unions.saw"
[20:12:16.768] Loaded MIR
[20:12:16.768] Stack trace:
   (builtin) in mir_verify
   unions.saw:11:1-11:45 (at top level)
Symbolic execution failed.
Abort due to assertion failure:
  unions.rs:6:23: 6:36: error: in unions/d81e1794::FOO[0]
  Translation error in unions/d81e1794::FOO[0]: evalRval: Union types are unsupported, for unions/d81e1794::Bad[0]::_adtc5a5405abab57fd5[0]
```
If I remove `unused` entirely and regenerate the MIR, however, SAW is then able to verify `used`.

I hypothesize that this arises from the immutable `static` translation behavior described here: https://github.com/GaloisInc/saw-script/blob/c262eb84d6fb16af79b7833b43c3adf039c151a9/saw-central/src/SAWCentral/Crucible/MIR/Builtins.hs#L1818-L1866

RyanGlScott:
I suspect that the root issue is actually at the `crucible-mir` level, not at the SAW level. In your program above, you are lucky in that `mir-json`'s dead-code elimination is able to get rid of the static item entirely. You wouldn't be so lucky in this variation of your program, however:

```rs
// test.rs
union Bad {
    x: u8,
    y: i8,
}

pub static FOO: Bad = Bad { x: 42 };

[...]

PR #1538, opened 2025-09-16
`crucible-mir`: support constructing union values

samcowger:
These changes follow up on #1473 by allowing construction of union-type values, while still leaving meaningful manipulation of those values unsupported. The aim is to postpone more union-related errors past translation time, to allow Crucible to consume code that mentions, but does not use, unions. See the included tests for examples of code Crucible can now consume.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1364,8 +1364,23 @@ evalRval rv@(M.RAdtAg (M.AdtAg adt agv ops ty optField)) = do
             case adt^.adtkind of
                 M.Struct -> buildStruct adt es
                 M.Enum _ -> buildEnum adt (fromInteger agv) es
-                M.Union -> do
-                    mirFail $ "evalRval: Union types are unsupported, for " ++ show (adt ^. adtname)
+                -- This is a hack. crucible-mir doesn't _actually_ support union
+                -- types, and if you try simulating a code that performs an
+                -- operation on a union-typed value, then it will fail at
+                -- simulation time. Nevertheless, we intentionally choose to
+                -- give unions a dummy value here instead of outright failing so
+                -- that crucible-mir can translate some union-typed declarations
+                -- without crashing during MIR translation.
+                --
+                -- We apply `packAny` here to match the similar hack described
+                -- in the union case of `Mir.Trans.tyToRepr`, which gives
+                -- union-typed values a dummy representation of `AnyRepr`.
+                --
+                -- See #1429 for the larger issue of properly supporting union
+                -- types.
+                M.Union ->
+                  let dummyExpr = R.App $ usizeLit 0xdeadbeef
+                   in pure $ packAny $ MirExp knownRepr dummyExpr
       _ -> mirFail $ "evalRval: unsupported type for AdtAg: " ++ show ty
 evalRval (M.ThreadLocalRef did _) = staticPlace did >>= addrOfPlace
 
--- a/crucible-mir/src/Mir/TransTy.hs
+++ b/crucible-mir/src/Mir/TransTy.hs
@@ -303,7 +303,7 @@ tyToReprM ty = do
     Right repr -> return repr
     Left err -> mirFail ("tyToRepr: " ++ err)
 
--- Checks whether a type can be default-initialized.  Any time this returns
+-- | Checks whether a type can be default-initialized.  Any time this returns
 -- `True`, `initialValue` must also return `Just`.  Non-initializable ADT
 -- fields are wrapped in `Maybe` to support field-by-field initialization.
 canInitialize :: M.Collection -> M.Ty -> Bool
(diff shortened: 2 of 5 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -1364,8 +1364,23 @@ evalRval rv@(M.RAdtAg (M.AdtAg adt agv ops ty optField)) = do
...
+                --
+                -- We apply `packAny` here to match the similar hack described
+                -- in the union case of `Mir.Trans.tyToRepr`, which gives
+                -- union-typed values a dummy representation of `AnyRepr`.
+                --
+                -- See #1429 for the larger issue of properly supporting union
+                -- types.
+                M.Union ->
+                  let dummyExpr = R.App $ usizeLit 0xdeadbeef
+                   in pure $ packAny $ MirExp knownRepr dummyExpr
```
samcowger, line 1383: I mentioned in https://github.com/GaloisInc/crucible/issues/1534#issuecomment-3285856653 that we might want to make this dummy expression the same size as the union it replaces, since we have access to that size here, but I realized I wasn't sure when/why that would be necessary, so I left it as a simple `usizeLit`. I welcome opinions on whether this is appropriate.

spernsteiner, line 1383: My worry with this approach is that it allows some things that we can't support long-term, so when we implement unions properly with MirAggregate, we will get regressions.  For example:
```Rust
union U {
  x: i32,
  y: u8,
}
let u = if bool::symbolic("cond") {
  U { x: 1 }
} else {
  U { y: 2 }
};
// `u` is unused afterward
```
This would be accepted with this approach because both branches would produce `AnyRepr` wrapping `UsizeRepr`, and same-typed `AnyRepr`s can be muxed together (IIRC).  But the future `MirAggregate` version would produce `MirAggregateRepr`s wrapping `BVRepr 32` and `BVRepr 8`, which can't be muxed together.

I think the current `MirAggregate` implementation is already sufficient to implement unions properly.  But if you want to stick with this stopgap implementation, please change it to wrap the actual value instead of a dummy `UsizeRepr`, so that muxing fails in the same way that it would in the `MirAggregate` version.

RyanGlScott: I think this will matter if you put a union value in a data structure that is represented as a `MirAggregate` at simulation time. Currently, I believe only tuples use `MirAggregate`s, so it might be worth trying to see what happens if you pack/unpack union values into a tuple.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1364,8 +1364,23 @@ evalRval rv@(M.RAdtAg (M.AdtAg adt agv ops ty optField)) = do
...
+                -- give unions a dummy value here instead of outright failing so
+                -- that crucible-mir can translate some union-typed declarations
+                -- without crashing during MIR translation.
+                --
+                -- We apply `packAny` here to match the similar hack described
+                -- in the union case of `Mir.Trans.tyToRepr`, which gives
+                -- union-typed values a dummy representation of `AnyRepr`.
+                --
+                -- See #1429 for the larger issue of properly supporting union
+                -- types.
```
RyanGlScott, line 1380: This comment is very similar to the one [here](https://github.com/GaloisInc/crucible/blob/9c8d9c78617b3666611af46bf98c5cbf566bdae1/crucible-mir/src/Mir/TransTy.hs#L231-L239). I propose that we factor out this comments into a top-level Note (see an existing Note [here](https://github.com/GaloisInc/crucible/blob/9c8d9c78617b3666611af46bf98c5cbf566bdae1/crux/src/Crux/Version.hs#L69-L95) for inspiration) and reference the Note in both places.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -321,6 +323,9 @@ canInitialize col ty = case ty of
     M.TyClosure _ -> True
     M.TyAdt _ _ _
       | Just ty' <- tyAdtDef col ty >>= reprTransparentFieldTy col -> canInitialize col ty'
+      | Just adt <- tyAdtDef col ty
+      , M.Union <- adt ^. M.adtkind ->
+        False
```
RyanGlScott, line 328: Worth mentioning in the Note why the `Union` case returns `False` here.

crucible-mir/src/Mir/Trans.hs
spernsteiner: We don't support asking for the size of a Crucible type—anything that needs to know the size should look it up based on the `Mir.Ty` instead.  So there's no need to try to match the union's size here.

spernsteiner: > I think this will matter if you put a union value in a data structure that is represented as a `MirAggregate`

All of the `MirAggregate` operations take separate size, offset, and `TypeRepr` arguments.  They don't try to compute the size from the `TypeRepr`, but instead compute all three from `Mir.Ty` info.

samcowger:
This has been superseded by #1543.

Issue #1499, opened 2025-08-18
crux-mir: implement flattened aggregates


spernsteiner:
Implement the `MirAggregate` type, as described in https://github.com/GaloisInc/crucible/discussions/1498

Implementation plan:

* [x] Implement `MirAggregate` type
* [x] Use `MirAggregateType` for `TyTuple`
* [ ] Export layout data from mir-json (https://github.com/GaloisInc/mir-json/issues/184)
* [ ] Flatten nested `TyTuple`s into a single `MirAggregate`
* [x] Update saw-script
&nbsp;
* [ ] Use `MirAggregateType` for `TyStruct`
* [x] Use `MirAggregateType` for `TyArray`
* [ ] Use `MirAggregateType` for heap allocations
* [ ] Wrap all locals in `MirAggregateType`

Issues to revisit after this is done:

* https://github.com/GaloisInc/crucible/issues/1109#issuecomment-3205710205
* https://github.com/GaloisInc/crucible/issues/1429#issuecomment-3205732022
* https://github.com/GaloisInc/crucible/issues/1494#issuecomment-3205723941
* https://github.com/GaloisInc/crucible/issues/1497#issuecomment-3205719612

spernsteiner:
When adding struct support, we'll have to adjust the logic for accessing the last field of a `dyn Trait` custom DST.  Computing the offset of the last field requires loading the alignment from the `dyn Trait` vtable, as shown in the optimized LLVM IR for [this example](https://play.rust-lang.org/?version=stable&mode=release&edition=2024&gist=367526ae13caadb37a651eb272f3fae3):
```Rust
use std::fmt::Debug;
use std::mem;

pub struct S<T: ?Sized> {
    x: u8,
    y: u16,
    z: T,
}

#[unsafe(no_mangle)]
pub fn f(s: &S<dyn Debug>) -> &dyn Debug {
    &s.z
}
```

PR #1543, opened 2025-09-19
`crucible-mir`: `MirAggregate` representation for unions

### LINKED ISSUE #1429: crux-mir: Union types are unsupported
When running the following test some [mashalling code](https://gitlab-ext.galois.com/cryptol-saw-unclassified/cryptol-generated-specs/-/blob/main/rust/cryptol-crypto/src/marshal.rs?ref_type=heads) I get the following error:

```Rust
#[cfg(crux)]
mod crux_test {
    use super::*;
    extern crate crucible;
    use self::crucible::*;

    #[crux::test]
    fn test_dword_to_bytes() {
        let x = DWord::from_u64(64, u64::symbolic("a"));
        let y: [u8; 32] = dword_to_array(x.as_ref());
        let z = dword_from_bytes(&y);
        assert_eq!(x, z);
    }
}
```

```
[Crux] Overall status: Valid.
     Running unittests src/lib.rs (target/x86_64-unknown-linux-gnu/debug/deps/cryptol_crypto-570d949c326b9c61)
Union types are unsupported, for dword/db25dd55::core[0]::DWordData[0]::_adtc5a5405abab57fd5[0]
  error, called at src/Mir/TransTy.hs:229:9 in crucible-mir-0.4.0.0.99-292d9eb40ccccc366eff9fd1d981f756c5bcae5ea102e881c73b58ff25731799:Mir.TransTy
  tyToRepr, called at src/Mir/TransTy.hs:374:78 in crucible-mir-0.4.0.0.99-292d9eb40ccccc366eff9fd1d981f756c5bcae5ea102e881c73b58ff25731799:Mir.TransTyerror: test failed, to rerun pass `-p cryptol-crypto --lib`

```

Not sure if this is a crux-mir issue, or something that should be changed in `DWord` crate?

samcowger:
These changes give union-type values a `MirAggregate` representation in Crucible, leveraging work discussed in #1498 and #1499 and implemented in part in #1503.

This means Crucible can now comprehend code involving union introduction. It can also, to a limited extent, comprehend code involving union elimination - only so long as the union is interpreted according to the type of the value originally used to initialize it. That is, it does not generally support the use of unions as a way to reinterpret blocks of memory, à la `std::mem::transmute`. I've included some xfail tests attesting to this.

This supersedes #1538, as it implements a more durable approach to handling unions.

Whether or not this fixes #1429 is up for some debate. I would suggest we close that issue, since this meaningfully addresses the overall problem, and open a new one for the specific task of supporting memory reinterpretation.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1615,14 +1615,6 @@ evalPlaceProj ty pl@(MirPlace tpr ref meta) (M.PField idx fieldTy) = do
             -- Since `findReprTransparentField` returned `Just`, we know that
             -- fields aside from `tIdx` must be zero-sized, and thus contain no
             -- actual data.  So we can return a dummy reference here.
-            --
-            -- Also, for enum types, `#[repr(transparent)]` is only allowed on
-            -- single-variant enums, so we know `tIdx` refers to a field of
-            -- variant 0 (as with structs).
-            fieldTy <- case adt ^? M.adtvariants . ix 0 . M.vfields . ix idx . M.fty of
-                Just x -> return x
-                Nothing -> mirFail $ "impossible: accessed out of range field " ++
-                    show idx ++ " of " ++ show adt ++ "?"
             MirExp tpr' e <- initialValue fieldTy >>= \x -> case x of
                 Just x -> return x
                 Nothing -> mirFail $ "failed to produce dummy value of type " ++ show fieldTy
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1440,7 +1440,8 @@ union's fields. When interpreting this initialization:
   size of the union/`MirAggregate`.
 
 The type of a (subrange of a) `MirAggregate` is unspecified until it's written
-to, and fixed thereafter. This means that, once a union's `MirAggregate` is
+to, and fixed thereafter. This allows for unions to be default-initialized by an
+untyped `MirAggregate`. This also means that, once a union's `MirAggregate` is
 initialized with a field of a given type, we only support reading from the union
 via a field of the same type (which, in practice, generally means the same
 field). When reading from the union, we rely on the initialization behavior
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1438,6 +1438,17 @@ union's fields. When interpreting this initialization:
 - We declare that the given field spans the entire `MirAggregate` representing
   the union, even if the field type's size on its own would be smaller than the
   size of the union/`MirAggregate`.
+
+The type of a (subrange of a) `MirAggregate` is unspecified until it's written
+to, and fixed thereafter. This means that, once a union's `MirAggregate` is
+initialized with a field of a given type, we only support reading from the union
+via a field of the same type (which, in practice, generally means the same
+field). When reading from the union, we rely on the initialization behavior
+described above, by reading the entire `MirAggregate` starting from offset 0.
+
+To properly implement reinterpretation of union values at other types, we'd need
+to change the behavior of `MirAggregate` to support type-switching, and we'd
+need to mimic Rust's layout rules for unions.
 -}
 
 evalLvalue :: HasCallStack => M.Lvalue -> MirGenerator h s ret (MirExp s)
@@ -1589,7 +1600,7 @@ evalPlaceProj ty pl@(MirPlace tpr ref NoMeta) M.Deref = do
         let vtable = S.getStruct dynRefVtableIndex dynRef
         return $ MirPlace adtRepr adtPtr (DynMeta vtable)
 
-evalPlaceProj ty pl@(MirPlace tpr ref meta) (M.PField idx _mirTy) = do
+evalPlaceProj ty pl@(MirPlace tpr ref meta) (M.PField idx fieldTy) = do
   col <- use $ cs . collection
   case ty of
     CTyMaybeUninit _ -> do
@@ -1623,7 +1634,7 @@ evalPlaceProj ty pl@(MirPlace tpr ref meta) (M.PField idx _mirTy) = do
         case adt^.adtkind of
             Struct -> structFieldRef adt idx ref meta
             Enum _ -> mirFail $ "tried to access field of non-downcast " ++ show ty
-            Union -> mirFail $ "evalPlace (PField, Union) NYI"
+            Union -> unionFieldRef adt idx fieldTy ref
 
     M.TyDowncast (M.TyAdt nm _ _) i -> do
         adt <- findAdt nm
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1365,7 +1365,14 @@ evalRval rv@(M.RAdtAg (M.AdtAg adt agv ops ty optField)) = do
                 M.Struct -> buildStruct adt es
                 M.Enum _ -> buildEnum adt (fromInteger agv) es
                 M.Union -> do
-                    mirFail $ "evalRval: Union types are unsupported, for " ++ show (adt ^. adtname)
+                  fieldExpr <- case es of
+                    [e] -> pure e
+                    [] -> mirFail "evalRval: no union field initializer?"
+                    _ -> mirFail "evalRval: multiple union field initializers?"
+                  fieldIdx <- case optField of
+                    Just idx -> pure idx
+                    Nothing -> mirFail "evalRval: no union field initializer index?"
+                  buildUnion adt fieldIdx fieldExpr
       _ -> mirFail $ "evalRval: unsupported type for AdtAg: " ++ show ty
 evalRval (M.ThreadLocalRef did _) = staticPlace did >>= addrOfPlace
 
@@ -1374,6 +1381,64 @@ evalRval rv@(M.CopyForDeref lv) = evalLvalue lv
 
 evalRval rv@(M.ShallowInitBox op ty) = mirFail
     "evalRval: ShallowInitBox not supported"
+
+-- | Construct a `MirExp` representing a union, and initialize it with a
+-- particular field, specified by a field index and an expression. The
+-- expression's type must match the field's type in the provided `M.Adt`.
+buildUnion ::
+  M.Adt ->
+  -- | The index of the field being used to initialize the union
+  Int ->
+  MirExp s ->
+  MirGenerator h s ret (MirExp s)
+buildUnion unionAdt fieldIdx (MirExp fieldTpr fieldExpr) = do
+  unionFields <- case unionAdt ^. M.adtvariants of
+    [v] -> pure (v ^. M.vfields)
+    [] -> die "no variants?"
+    _ -> die "multiple variants?"
+
+  unionField <- case unionFields ^? ix fieldIdx of
+    Just field -> pure field
+    Nothing -> die $ "field index " <> show fieldIdx <> " out of range"
+
+  Some expectedFieldTpr <- tyToReprM (unionField ^. M.fty)
+  case testEquality expectedFieldTpr fieldTpr of
+    Just _ -> pure ()
+    Nothing -> die $
+      "expected field to have type " <> show expectedFieldTpr <>
+      ", but it was " <> show fieldTpr
+
+  emptyAg <- mirAggregate_uninit unionSize
+  fullAg <- mirAggregate_set fieldOffset fieldSize fieldTpr fieldExpr emptyAg
+  pure (MirExp MirAggregateRepr fullAg)
+  where
+    unionSize = fromIntegral (unionAdt ^. M.adtSize)
+
+    -- See Note [union representation]
+    fieldOffset = 0
+    fieldSize = unionSize
+
+    die :: String -> MirGenerator h s ret a
+    die s = mirFail ("buildUnion: "<>show (unionAdt ^. M.adtname)<>": "<>s)
+
+{- 
+Note [union representation]
+----------------------------------------
+
+Crucible represents Rust unions as `MirAggregate` values.
+
+A union's `MirAggregate` representation has the same size as the union itself,
+according to the `_adtSize` field of the `Mir.Mir.Adt` that describes it.
+
+A union is always initialized with a single expression representing one of the
+union's fields. When interpreting this initialization:
+- We declare that the given field appears at offset 0 in the `MirAggregate`,
+  even if the field would appear at a nonzero offset according to Rust's memory
+  model.
+- We declare that the given field spans the entire `MirAggregate` representing
+  the union, even if the field type's size on its own would be smaller than the
+  size of the union/`MirAggregate`.
+-}
 
 evalLvalue :: HasCallStack => M.Lvalue -> MirGenerator h s ret (MirExp s)
 evalLvalue lv = evalPlace lv >>= readPlace
--- a/crucible-mir/src/Mir/TransTy.hs
+++ b/crucible-mir/src/Mir/TransTy.hs
@@ -1240,6 +1240,47 @@ tupleFieldRef tys i tpr ref = do
     ref' <- mirRef_agElem_constOffset (fromIntegral i) 1 valTpr ref
     return $ MirPlace valTpr ref' NoMeta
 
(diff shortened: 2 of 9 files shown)


crucible-mir/src/Mir/TransTy.hs
```diff
@@ -228,16 +228,7 @@ tyToRepr col t0 = case t0 of
...
-        -- and if you try simulating a code that performs an operation on a
-        -- union-typed value, then it will fail at simulation time.
-        -- Nevertheless, we intentionally choose to use Any as a placeholder
-        -- type instead of failing so that crucible-mir can translate type
-        -- signatures mentioning union types without crashing during MIR
-        -- translation (e.g., in Mir.Trans.mkHandleMap).
-        --
-        -- See #1429 for the larger issue of properly supporting union types.
-        Right (Some C.AnyRepr)
+        Right (Some MirAggregateRepr)
```
RyanGlScott, line 232: Worth citing `Note [union representation]` here, I think.

RyanGlScott:
> I would suggest we close that issue, since this meaningfully addresses the overall problem, and open a new one for the specific task of supporting memory reinterpretation.

I'm aligned with this.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1438,6 +1438,17 @@ union's fields. When interpreting this initialization:
...
+The type of a (subrange of a) `MirAggregate` is unspecified until it's written
+to, and fixed thereafter. This means that, once a union's `MirAggregate` is
+initialized with a field of a given type, we only support reading from the union
+via a field of the same type (which, in practice, generally means the same
+field). When reading from the union, we rely on the initialization behavior
+described above, by reading the entire `MirAggregate` starting from offset 0.
+
+To properly implement reinterpretation of union values at other types, we'd need
+to change the behavior of `MirAggregate` to support type-switching, and we'd
+need to mimic Rust's layout rules for unions.
```
RyanGlScott: After opening an issue about supported type-reinterpreting unions, it would be worth citing that issue here.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1240,6 +1240,47 @@ tupleFieldRef tys i tpr ref = do
     ref' <- mirRef_agElem_constOffset (fromIntegral i) 1 valTpr ref
     return $ MirPlace valTpr ref' NoMeta
 
+-- | Acquire a reference to the union field specified by the given index and
+-- type. The field type must match the type specified in the provided `M.Adt`.
+--
+-- Additionally, the field type must match the expression type with which the
+-- union was initialized - see Note [union representation] in `Mir.Trans`.
```
RyanGlScott: Haddock nit pick: Haddock uses double quotes to hyperlink module names, not backticks (which are for identifiers):

```diff
--- union was initialized - see Note [union representation] in `Mir.Trans`.
+-- union was initialized - see Note [union representation] in "Mir.Trans".
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1375,6 +1382,76 @@ evalRval rv@(M.CopyForDeref lv) = evalLvalue lv
...
+  unionField <- case unionFields ^? ix fieldIdx of
+    Just field -> pure field
+    Nothing -> die $ "field index " <> show fieldIdx <> " out of range"
+
+  Some expectedFieldTpr <- tyToReprM (unionField ^. M.fty)
+  case testEquality expectedFieldTpr fieldTpr of
+    Just _ -> pure ()
+    Nothing -> die $
+      "expected field to have type " <> show expectedFieldTpr <>
+      ", but it was " <> show fieldTpr
```
RyanGlScott: There's quite a bit of code shared in common between this function (`buildUnion`) and `unionFieldRef` used for checking the union's variant, extracting its field, calling `testEquality`, etc. I propose that we factor this out into a `unionInfo :: Adt -> Int -> MirExp s -> UnionField` function, similar to the existing [`structInfo`](https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crucible-mir/src/Mir/TransTy.hs#L810-L860) and [`enumInfo`](https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crucible-mir/src/Mir/TransTy.hs#L945-L983) functions.

We could consider splitting out the `testEquality` check into an `expectUnionOrFail` function, similar to the existing [`expectStructOrFail` and `expectEnumOrFail` functions](https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crucible-mir/src/Mir/TransTy.hs#L1258-L1279), but that might be a bit overkill for now, given that there are only two call sites.

crux-mir/test/conc_eval/union/read_other.rs
```diff
@@ -0,0 +1,20 @@
+// FAIL: reading a non-active union field
+
+union U {
+    x: u16,
+    y: [u8; 2],
+}
```
spernsteiner, line 6: I think reading/writing a non-active field should work as long as the `TypeRepr`s match.  Can you add tests like this that try to convert between `u16` and `i16` or similar?

Another thing that would be interesting to test is two fields with the same `TypeRepr` but different sizes in bytes:
```Rust
// Using enums rather than structs because we don't plan to flatten enums into enclosing `MirAggregate`s
enum Small { X(u16) }
#[repr(align(8))]
struct Large { X(u16) }
```
This should pass right now, since everything is treated as having size 1, but once we add proper layout information, it might fail (depending on how strict the `MirAggregate` ops are about checking for matching field sizes).

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1375,6 +1382,76 @@ evalRval rv@(M.CopyForDeref lv) = evalLvalue lv
...
+according to the `_adtSize` field of the `Mir.Mir.Adt` that describes it.
+
+A union is always initialized with a single expression representing one of the
+union's fields. When interpreting this initialization:
+- We declare that the given field appears at offset 0 in the `MirAggregate`,
+  even if the field would appear at a nonzero offset according to Rust's memory
+  model.
+- We declare that the given field spans the entire `MirAggregate` representing
+  the union, even if the field type's size on its own would be smaller than the
+  size of the union/`MirAggregate`.
```
spernsteiner: The field size should be the actual size in bytes of the type of that field (or 1 for now).  If the program takes a pointer to the union field and does read/write operations on it, it will pass the field's size as the size argument to those operations, and some operations will fail if the requested size doesn't match the size of the `MirAggregateEntry`.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1375,6 +1382,76 @@ evalRval rv@(M.CopyForDeref lv) = evalLvalue lv
...
+    Just _ -> pure ()
+    Nothing -> die $
+      "expected field to have type " <> show expectedFieldTpr <>
+      ", but it was " <> show fieldTpr
+
+  emptyAg <- mirAggregate_uninit unionSize
+  fullAg <- mirAggregate_set fieldOffset fieldSize fieldTpr fieldExpr emptyAg
+  pure (MirExp MirAggregateRepr fullAg)
+  where
+    unionSize = fromIntegral (unionAdt ^. M.adtSize)
```
spernsteiner: I think this needs to be 1 for now.  Otherwise a union with all zero-sized fields will have total size 0 but will have a field covering bytes 0..1 (even ZSTs are treated as size 1 for now), which is illegal.  Might be worth adding a test for this (initialize an instance of `union U { x: (), y: () }`).

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1249,6 +1240,47 @@ tupleFieldRef tys i tpr ref = do
...
+    Just field -> pure field
+    Nothing -> die $ "field index " <> show fieldIdx <> " out of range"
+
+  Some expectedFieldTpr <- tyToReprM (unionField ^. M.fty)
+  Some fieldTpr <- tyToReprM fieldTy
+  case testEquality expectedFieldTpr fieldTpr of
+    Just _ -> pure ()
+    Nothing -> die $
+      "expected field to have type " <> show expectedFieldTpr <>
+      ", but it was " <> show fieldTpr
```
spernsteiner: Looks like you can just check whether `unionField ^. M.fty == fieldTy` here.  The type indexes of `expectedFieldTpr` and `fieldTpr` aren't connected to anything else, so there's no value in proving them equal (and indeed this code doesn't actually use the `Refl` that it gets from `testEquality`).

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1249,6 +1240,47 @@ tupleFieldRef tys i tpr ref = do
...
+
+  fieldRef <- mirRef_agElem_constOffset fieldOffset fieldSize fieldTpr unionRef
+  pure $ MirPlace fieldTpr fieldRef NoMeta
+  where
+    unionSize = fromIntegral (unionAdt ^. M.adtSize)
+
+    -- See Note [union representation] in `Mir.Trans`
+    fieldOffset = 0
+    fieldSize = unionSize
```
spernsteiner: Needs `unionSize` and `fieldSize` adjusted as in previous comments

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1449,7 +1481,12 @@ initialValue (M.TyAdt nm _ _) = do
                 Just (discr, var) -> do
                     fldExps <- mapM initField (var^.M.vfields)
                     Just <$> buildEnum' adt discr fldExps
-        M.Union -> return Nothing
+        M.Union ->
+            -- Unions are default-initialized to an untyped `MirAggregate` of an
+            -- appropriate size, like tuples. See Note [union representation] in
+            -- `Mir.Trans`.
+            let unionSize = fromIntegral (adt ^. M.adtSize)
```
spernsteiner: Needs `unionSize` adjusted as in previous comments

crucible-mir/src/Mir/Trans.hs
spernsteiner: Since this isn't using the `Refl`, I think you can just check `Some expectedFieldTpr == Some fieldTpr`

crucible-mir/src/Mir/TransTy.hs
samcowger: Done.

crucible-mir/src/Mir/Trans.hs
samcowger: Good call, done (and the issue is #1548).

crucible-mir/src/Mir/TransTy.hs
samcowger: Sigh - one day, I'll get Haddock right on the first try.

I moved the note to `TransTy`, so this reference to `Mir.Trans` was removed entirely.

crucible-mir/src/Mir/Trans.hs
samcowger: > I propose that we factor this out

Done in d3a8f67f929846e5fd2a2983fa784717ae1631d4 and e3c3fe1fb622ae33ba74216ce1fff4822dbec661.

> We could consider splitting out the testEquality check into an expectUnionOrFail function

I don't think that's necessary at this point, since I was able to centralize the necessary type checks in `unionInfo`.

crux-mir/test/conc_eval/union/read_other.rs
samcowger: > Can you add tests like this that try to convert between u16 and i16 or similar?

Done in d4ca95e625e7366003240c3b0001a23073f55cd5, which added tests and clarified `Note [union representation]` to explicitly call out this allowance.

> Another thing that would be interesting to test is two fields with the same `TypeRepr` but different sizes in bytes

I incorporated this test in 00d93679befb29f57dcc0233db2011d6a4bb8cad. It does indeed pass right now, but I left a comment on it describing how/why it might fail in the future.

crucible-mir/src/Mir/Trans.hs
samcowger: Changed to size 1 in 4612fe406b99cb2a11258199828875e39f447cd1. (For posterity, we discussed in person that the read/write issue was unlikely to crop up, since both union introduction and elimination make the same field size assumptions, but that this was better for consistency with other temporarily size/layout-agnostic `MirAggregate` behavior, e.g. every field in a tuple unconditionally having size 1.)

crucible-mir/src/Mir/TransTy.hs
samcowger: I think you're right, but I've since refactored this in d3a8f67f929846e5fd2a2983fa784717ae1631d4 in a way that made `testEquality` actually necessary.

crucible-mir/src/Mir/TransTy.hs
samcowger: Changed to size 1 in 4612fe406b99cb2a11258199828875e39f447cd1.

crucible-mir/src/Mir/Trans.hs
samcowger: I think you're right, but I've since refactored this in d3a8f67f929846e5fd2a2983fa784717ae1631d4 in a way that made `testEquality` actually necessary.

crucible-mir/src/Mir/TransTy.hs
samcowger: Changed to size 1 in 4612fe406b99cb2a11258199828875e39f447cd1.

crucible-mir/src/Mir/Trans.hs
samcowger: I changed this to size 1 in 4612fe406b99cb2a11258199828875e39f447cd1. However, I wasn't able to add a test case involving initializing a ZST union that resulted in the union actually being evinced in MIR.

samcowger:
Does this warrant an update to `crucible-mir`'s changelog, do you think?

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1011,6 +1003,50 @@ setEnumField adt i j (MirExp enumTpr e) (MirExp fldTpr e') = do
...
 
+-- | A description of an individual field of a union.
+data UnionInfo where
+  UnionInfo ::
+    -- | The overall size of the union
+    Word ->
+    -- | The size of this field
+    Word ->
+    -- | The offset of this field
+    Word ->
```
spernsteiner: All of the `MirAggregate` ops that take a field's layout info use the opposite ordering: `off` first, then `sz`.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1558,7 +1558,9 @@ evalPlaceProj ty pl@(MirPlace tpr ref meta) (M.PField idx _mirTy) = do
         case adt^.adtkind of
             Struct -> structFieldRef adt idx ref meta
             Enum _ -> mirFail $ "tried to access field of non-downcast " ++ show ty
-            Union -> mirFail $ "evalPlace (PField, Union) NYI"
+            Union -> do
+              info <- unionInfo adt idx
+              unionFieldRef info ref
```
spernsteiner: For consistency with `structFieldRef`/`enumFieldRef`, you could push the `unionInfo` call down into `unionFieldRef` (and `buildUnion`), making `unionInfo` purely an internal helper within `TransTy` like `structInfo` and `enumInfo` are

RyanGlScott:
> Does this warrant an update to `crucible-mir`'s changelog, do you think?

Yes, definitely! Having more support for union types is definitely worth calling out.

spernsteiner (approved):
Look good—thanks!  Just a couple final nitpicks regarding consistency with related code

crucible-mir/src/Mir/Trans.hs
samcowger: Good call, done in cc25d61c25c48c394d56af9ee9dc6460e70fb1eb.

crucible-mir/src/Mir/TransTy.hs
samcowger: Reversed in bdb1603c8f2c25d5597f1c24777525b79e87a2db.

samcowger:
> > Does this warrant an update to `crucible-mir`'s changelog, do you think?
> 
> Yes, definitely! Having more support for union types is definitely worth calling out.

Done in dc3e06bca6c7873a69aad7af08b53e9727a813c3.

PR #1546, opened 2025-09-22
`crux-mir`: Support concretizing static references

### LINKED ISSUE #1545: `crux-mir`: Support concretizing static references (i.e., references using `GlobalVar_RefRoot`)
`crux-mir` currently crashes if you try to `concretize` a static reference, as seen in the program below:

```rs
extern crate crucible;
use crucible::*;

#[crux::test]
pub fn f() -> u32 {
    let x: &'static u32 = &0;
    *crucible::concretize(x)
}
```
```
test.rs:7:6: 7:29: error: in test/d3320d23::f[0]
Unsupported feature: evaluation of GlobalVar_RefRoot is not yet implemented
Details:
  [("throwUnsupported",SrcLoc {srcLocPackage = "crux-mir-0.10.0.0.99-inplace", srcLocModule = "Mir.Overrides", srcLocFile = "src/Mir/Overrides.hs", srcLocStartLine = 353, srcLocStartCol = 9, srcLocEndLine = 353, srcLocEndCol = 25})]test test/d3320d23::f[0]:
```

Where the failing code arises from here:

https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crux-mir/src/Mir/Overrides.hs#L354

I propose that we make this work by adapting the existing code used to handle `RefCell_RefRoot`s [here](https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crux-mir/src/Mir/Overrides.hs#L317-L346).

RyanGlScott:
Fixes #1545.

### DIFF
--- a/crux-mir/test/symb_eval/concretize/static_ref.rs
+++ b/crux-mir/test/symb_eval/concretize/static_ref.rs
@@ -1,0 +1,13 @@
+// A regression test for #1545, which ensures that crux-mir is able to
+// concretize references to statics, i.e., references that are represented with
+// GlobalVar_RefRoot under the hood.
+
+extern crate crucible;
+use crucible::*;
+
+static X: u32 = 0;
+
+#[crux::test]
+pub fn f() -> u32 {
+    *crucible::concretize(&X)
+}
--- a/crux-mir/CHANGELOG.md
+++ b/crux-mir/CHANGELOG.md
@@ -25,6 +25,7 @@ This release supports [version
   `--skip` flag.
 * Fix a bug that could cause the `crucible::concretize` function to crash Crux
   when using the `bitwuzla`, `cvc4`, or `cvc5` solvers.
+* Allow calling `crucible::concretize` on `static` references.
 
 # 0.10 -- 2025-03-21
 
--- a/crux-mir/src/Mir/Overrides.hs
+++ b/crux-mir/src/Mir/Overrides.hs
@@ -58,7 +58,9 @@ import Lang.Crucible.Backend
     , singleEvent, addAssumptions, CrucibleEvent(..), backendGetSym
     , throwUnsupported, LabeledPred(..), addProofObligation )
 import Lang.Crucible.Backend.Online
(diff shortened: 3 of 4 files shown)


crux-mir/test/symb_eval/concretize/static_ref.rs
```diff
@@ -0,0 +1,12 @@
+// A regression test for #1545, which ensures that crux-mir is able to
+// concretize static references, i.e., references that are represented with
+// GlobalVar_RefRoot under the hood.
```
spernsteiner, line 3:
```suggestion
// A regression test for #1545, which ensures that crux-mir is able to
// concretize references to statics, i.e., references that are represented with
// GlobalVar_RefRoot under the hood.
```

"Static references" could mean `&'static T`, but not all `&'static T` refs are backed by `GlobalVar_RefRoot`:
```Rust
let x: &'static i32 = Box::leak(Box::new(123));
```
This will point to a heap allocation using `RefCell_RefRoot`.

crux-mir/test/symb_eval/concretize/static_ref.rs
```diff
@@ -0,0 +1,12 @@
...
+
+extern crate crucible;
+use crucible::*;
+
+#[crux::test]
+pub fn f() -> u32 {
+    let x: &'static u32 = &0;
```
spernsteiner: `&0` is guaranteed to get promoted to a static (by my read of the [Rust Reference](https://doc.rust-lang.org/reference/destructors.html#constant-promotion)), but it might be clearer to explicitly write a `static` item and take its address, since references to `static` items are the ones that use `GlobalVar_RefRoot`.

RyanGlScott: I've changed the test case to the following:

```rs
static X: u32 = 0;

#[crux::test]
pub fn f() -> u32 {
    *crucible::concretize(&X)
}
```

PR #1547, opened 2025-09-22
Allow casting safe pointers to unsafe ones

### LINKED ISSUE #1541: [crux-mir] Allow casting pointer to unsafe pointer
Since we don' track pointer safety in Crucible, this should be a no op.

yav:
Fixes #1541

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -956,6 +956,11 @@ evalCast' ck ty1 e ty2  = do
 
       (M.Transmute, _, _) -> transmuteExp e ty1 ty2
 
+      -- This casts from a safe pointer to an unsafe one.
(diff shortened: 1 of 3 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -956,6 +956,11 @@ evalCast' ck ty1 e ty2  = do
 
       (M.Transmute, _, _) -> transmuteExp e ty1 ty2
 
+      -- This casts from a safe pointer to an unsafe one.
+      -- Since we don't track safness this is just a no-op for now, but if
```
RyanGlScott:
```suggestion
      -- Since we don't track safeness this is just a no-op for now, but if
```

RyanGlScott (approved):

RyanGlScott:
Any reason not to merge this, @yav?

Issue #1549, opened 2025-09-23
Failure reading structs with 0-length arrays


glguy:
Something's wrong with how we treat zero-length arrays.

```
struct S {
    int x; // we need at least two fields beside z
    int y;
    char z[]; // we have to have z generate the trailing [0 x i8]
};

struct S s;

void doit() {
    s.y = 0; // We have to set at least one field
}
```

```
m <- llvm_load_module "example.bc";

llvm_verify m "doit" [] false
  do {
    llvm_alloc_global "s";

    llvm_points_to (llvm_global "s")
        (llvm_packed_struct_value
            [ llvm_term {{ 0 : [32] }}
            , llvm_term {{ 0 : [32] }}
            , llvm_term {{ zero : [0][8] }}
            ]);

    llvm_execute_func [];

    Y <- llvm_fresh_var "Y" (llvm_alias "struct.S");
    llvm_points_to (llvm_global "s") (llvm_term Y);
  }
  rme;
```

```
[16:20:44.600] Loading file "doit.saw"
[16:20:44.687] Verifying doit...
[16:20:44.688] Simulating doit...
[16:20:44.688] Stack trace:
   (builtin) in llvm_verify
   doit.saw:3:1-19:6 (at top level)
at doit.saw:17:5
error when loading through pointer that appeared in the override's points-to precondition(s):
Precondition:
  Pointer: @s
  Pointee: let { x@1 = Prelude.Vec 32 Prelude.Bool
      }
   in fresh:Y#1520 : ([32], [32], [0][8])

  Assertion made at: doit.saw:17:5
Failure reason:
  When reading through pointer: (2, 0x0:[64])
  in the  postcondition of an override
  Tried to read something of size: 8
  And type: { i32, i32, [0 x i8] }
  Found 1 possibly matching allocation(s):
  - GlobalAlloc 2 0x8:[64] Mutable 4-byte-aligned s
```

RyanGlScott:
I claim that this is ultimately a bug in `crucible-llvm`'s memory model. With some effort, I managed to reproduce the same bug in Crux-LLVM, using a hand-written LLVM bitcode file that precisely sets up the memory model operations needed to mimic what goes on in SAW:

```llvm
; test.ll

%struct.S = type { i32, i32, [0 x i8] }

define i32 @main() {
  ; Allocate a ptr large enough to hold a %struct.S value.
  %sp = alloca %struct.S

[...]

PR #1551, opened 2025-09-24
Unconditionally succeed when loading 0-sized values

glguy:
These values have no representation in memory and thus do not need to depend on what's in memory at all.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Common.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Common.hs
@@ -511,6 +511,8 @@ valueLoad ::
   ValueView {- ^ view of stored value -} ->
   ValueCtor (ValueLoad Addr)
 valueLoad lo ltp so v
+  -- The non-zero test ensures that 0-byte loads are always overlapping
+  -- with the most recent store so that they're not spuriously rejected.
   | le <= so && nonZeroLoad = ValueCtorVar (OldMemory lo ltp) -- Load ends before store
   | se <= lo && nonZeroLoad = ValueCtorVar (OldMemory lo ltp) -- Store ends before load
     -- Load is before store.
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Common.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Common.hs
@@ -511,8 +511,8 @@ valueLoad ::
   ValueView {- ^ view of stored value -} ->
   ValueCtor (ValueLoad Addr)
 valueLoad lo ltp so v
-  | le <= so = ValueCtorVar (OldMemory lo ltp) -- Load ends before store
-  | se <= lo = ValueCtorVar (OldMemory lo ltp) -- Store ends before load
+  | le <= so && nonZeroLoad = ValueCtorVar (OldMemory lo ltp) -- Load ends before store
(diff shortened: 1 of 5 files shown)


glguy:
See #1549

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -707,6 +708,10 @@ readMem :: forall sym w.
   Alignment ->
   Mem sym ->
   IO (PartLLVMVal sym)
+readMem sym _ _ _ tp _ _
+  -- no actual memory read happens when reading a value with
+  -- no memory representation
```
RyanGlScott: It would be worth clarifying that this isn't just an optimization—this can avoid actual memory load failures that happen when a 0-byte load occurs on a write boundary (citing the relevant test case as evidence that this can happen).

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -741,6 +746,20 @@ readMem sym w gsym l tp alignment m = do
...
+-- | Construct a value of a type that has no in-memory representation at all
+buildEmptyVal :: StorageType -> Maybe (LLVMVal sym)
+buildEmptyVal t =
+  case storageTypeF t of
+    Struct fields -> LLVMValStruct <$> traverse emptyStorageField fields
+    Array n eltTy
+      | n == 0 -> Just (LLVMValArray eltTy V.empty)
+      | otherwise -> LLVMValArray eltTy . V.replicate (fromIntegral n) <$> buildEmptyVal eltTy
+    -- Note: Bitvectors can't be empty; they have a non-zero size invariant
+    _ -> Nothing
```
RyanGlScott: Can you fill out the remaining constructors of `StorageTypeF` here rather than using a wildcard pattern? That way, GHC's pattern-match coverage checker will warn us in case we add new constructors to `StorageTypeF` in the future.

RyanGlScott (commented):
Please add the test case from https://github.com/GaloisInc/crucible/issues/1549#issuecomment-3329051933 to the `crux-llvm` test suite. Please add an entry in the `crucible-llvm` changelog as well.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Common.hs
```diff
@@ -511,8 +511,8 @@ valueLoad ::
   ValueView {- ^ view of stored value -} ->
   ValueCtor (ValueLoad Addr)
 valueLoad lo ltp so v
-  | le <= so = ValueCtorVar (OldMemory lo ltp) -- Load ends before store
-  | se <= lo = ValueCtorVar (OldMemory lo ltp) -- Store ends before load
+  | le <= so && nonZeroLoad = ValueCtorVar (OldMemory lo ltp) -- Load ends before store
+  | se <= lo && nonZeroLoad = ValueCtorVar (OldMemory lo ltp) -- Store ends before load
```
RyanGlScott, line 517: Please leave a comment about why we need to check for `nonZeroLoad`s (namely, that we want to treat 0-byte loads as always overlapping with the most recent store so that they're not spuriously rejected).

RyanGlScott (approved):
Thanks! I've pushed a commit which (I believe) should fix the CI failure.

One small request for additional documentation, but otherwise LGTM.

PR #1552, opened 2025-09-29
`crucible-mir`: Properly support const generics

RyanGlScott:
This contains the `crucible-mir`-side changes needed to properly support const
generics, as detailed in https://github.com/GaloisInc/mir-json/issues/64.
Specifically, the `TyConst` data constructor now supports a `ConstVal` field
that describes the constant used to instantiate a const generic parameter. This
has no impact on the semantics of `crucible-mir` itself, but this `ConstVal`
can be used by downstream tools that wish to distinguish different
instantiations of const generic functions. (The primary use case is SAW's
`mir_find_adt` function.)

This bumps the `mir-json` submodule to bring in the changes from
https://github.com/GaloisInc/mir-json/pull/173. Because this requires a bump
to the MIR JSON schema version, corresponding changes to the version number
must also be made here.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -1,7 +1,7 @@
 # next -- TBA
 
 This release supports [version
-3](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#3) of
+5](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#5) of
 `mir-json`'s schema.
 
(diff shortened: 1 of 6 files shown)


spernsteiner (approved):

RyanGlScott:
The CI revealed a bug in https://github.com/GaloisInc/mir-json/pull/173: it was filling out `"needs_drop": null` for type-level constants, but the JSON parser in `crucible-mir` requires this to be a boolean. I think I will change https://github.com/GaloisInc/mir-json/pull/173 to fill out `"needs_drop": false` for type-level constants instead.

PR #1555, opened 2025-10-03
Ground evaluation, concretization, and unique concretization

### LINKED ISSUE #1550: Ground evaluation, concretization, and unique concretization
What4 has three ways to turn symbolic expressions (`SymExpr`s) into ground (i.e., non-symbolic) values (`GroundValue`s). In ascending order of how much work they do:

1. [`What4.Expr.GroundEval`](https://github.com/GaloisInc/what4/blob/e8e5cb58da2c933fdc77ade8d737689b0e9aba74/what4/src/What4/Expr/GroundEval.his) defines models (`GroundEvalFn`s) that take `SymExpr`is and return `GroundValue`s
2. [`What4.Concretize.concretize`](https://github.com/GaloisInc/what4/blob/e8e5cb58da2c933fdc77ade8d737689b0e9aba74/what4/src/What4/Concretize.his#L43-L49) queries the solver for a model and returns the `GroundValue` for a particular `SymExpr`
3. [`What4.Concretize.uniquelyConcretize`](https://github.com/GaloisInc/what4/blob/e8e5cb58da2c933fdc77ade8d737689b0e9aba74/what4/src/What4/Concretize.his#L75-L83) does all of (2), but also asks the solver if this is the *only* possible value for the expression, in *any* model.

The latter two are primarily intended for use *during* symbolic execution. The former is widely used *after* symbolic execution to provide counterexamples.

Crucible has a few relevant pieces of functionality.

1. [`Lang.Crucible.Concretize`](https://github.com/GaloisInc/crucible/blob/2e0e02b95d7d057eae13945a9e0875994c55757b/crucible/src/Lang/Crucible/Concretize.his), which is akin to `What4.Expr.GroundEval`.
1. Crux-MIR's [`crucible::concretize`](https://github.com/GaloisInc/crucible/blob/2e0e02b95d7d057eae13945a9e0875994c55757b/crux-mir/src/Mir/Overrides.his#L130), which is intended to be used at symex runtime as an override. It doesn't return a separate "ground" value type, but traverses symbolic expressions, transforming them in-place.
1. Crux's [`ModelView`](https://github.com/GaloisInc/crucible/blob/2e0e02b95d7d057eae13945a9e0875994c55757b/crux/src/Crux/Goal.his#L304). Frankly, it's not clear to me how this works.
1. Various parts of Crucible-LLVM that aren't immediately relevant.

In summary:

- Crucible is missing functionality analogous to What's (2) and (3) above.
- `Lang.Crucible.Concretize` and the functions therein are named disanalogously with the corresponding functionality (1) in What4.

We should fix the disanalogy in naming and add the analogous functionality to Crucible, likely inside `Lang.Crucible.Concretize`. Crucible's version of (2) and (3) should gather the current path condition from the backend and assume it in a new frame before obtaining a model.

langston-barrett:
This commit brings Crucible's terminology around grounding and concretization in line with What4's, and adds some new functionality in that area. See module comment on `Lang.Crucible.Concretize`.

Fixes #1550.


langston-barrett:
Needs:

- [x] Remaining cases of `eqRegValue` (deferred to issue + TODO)
- [x] Ideally, more cases in `asConcRegValue` (will add issue number and TODO if we go forward with this)
- [x] Changelog entry
- [x] Overrides in `crucible-syntax` for (uniquely) concretizing `Bool`s
- [x] Tests in `crucible-cli`
  - [x] Concretizing a `Bool` after `assume!`ing that it's `#t` and branching on it only takes the true path
  - [ ] ~~Same with uniquely concretizing~~
  - [ ] ~~Unique concretization fails on a fresh `Bool`~~

crucible/CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Add `FunctorF`, `FoldableF`, and `TraversableF` instances for `CrucibleEvent`,
   `CrucibleAssumption`, and `CrucibleAssumptions`.
 * Add `gcAddTopLevelAssume`, for making top-level assumptions.
+* Rename functions in `Lang.Crucible.Concretize` to match What4's conventions.
```
RyanGlScott, line 14: Which particular functions were renamed?

crucible/src/Lang/Crucible/Simulator/RegValue.hs
```diff
@@ -361,3 +363,41 @@ muxVariant sym recf ctx = \p x y ->
...
+  RegValue sym tp ->
+  RegValue sym tp ->
+  IO (Pred sym)
+eqRegValue sym tp x y =
+  case tp of
+    -- Base types
+    BoolRepr -> eqPred sym x y
+    BVRepr _width -> bvEq sym x y
+    ComplexRealRepr -> cplxEq sym x y
+    FloatRepr _ -> isEq sym x y
```
RyanGlScott: Using `isEq` does work, but it might be worth using `floatEq` instead for symmetry with the other cases that use type-specific equality functions.

crucible-cli/src/Lang/Crucible/CLI.hs
```diff
@@ -1,4 +1,4 @@
-{-# LANGUAGE DataKinds #-}
+ {-# LANGUAGE DataKinds #-}
```
RyanGlScott: This whitespace change was introduced here, only to be reverted in a later commit in the PR. Consider fixing up this commit to just not make this whitespace change in the first place?

crucible/src/Lang/Crucible/Simulator/RegValue.hs
langston-barrett: That yields `Could not deduce ‘SymInterpretedFloatType sym flt ~ BaseFloatType fpp0’`

RyanGlScott: Ah, I forgot that `crucible`'s `FloatRepr` doesn't correspond to `what4`'s `BaseFloatRepr`. In that case, I think the actual thing I was looking for is `What4.InterpretedFloatingPoint`'s [`iFloatEq`](https://hackage-content.haskell.org/package/what4-1.7/docs/What4-InterpretedFloatingPoint.html#v:iFloatEq).

langston-barrett: Nice find, thanks!

PR #1557, opened 2025-10-08
llvm: Additional documentation for merging LLVM memories

langston-barrett:
Plus a new assertion to help check the subtle invariant about calls to `brachMem` and `mergeMem` being well-bracketed.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
@@ -76,6 +76,7 @@ module Lang.Crucible.LLVM.MemModel.Generic
 
 import           Prelude hiding (pred)
 
+import qualified Control.Exception as X
 import           Control.Lens
 import           Control.Monad
 import           Control.Monad.State.Strict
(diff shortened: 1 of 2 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Generic.hs
```diff
@@ -1606,21 +1607,41 @@ freeMem sym w (LLVMPointer blk off) m loc =
...
 mergeMem :: IsExpr (SymExpr sym) => Pred sym -> Mem sym -> Mem sym -> Mem sym
 mergeMem c x y =
   case (x^.memState, y^.memState) of
-    (BranchFrame _ _ a s, BranchFrame _ _ b _) ->
-      let s' = s & memStateAddChanges (muxChanges c a b)
-      in x & memState .~ s'
+    (BranchFrame as ws a s, BranchFrame as' ws' b _) ->
+      -- The memories to be merged must have originated from the same memory,
+      -- and in particular, should have matching alloc/write counts.
+      X.assert (as == as' && ws == ws') $
```
langston-barrett, line 1642: I'd be happy to make this into a `panic` (likely by adding a helper to the `Panic` module), just let me know if that's desired.

RyanGlScott: I think using `assert` is fine, since this isn't something that we'd want to check (and incur a slight performance penality) unless we are building without optimizations.

Issue #1556, opened 2025-10-08
`crux-mir`: Allow passing custom options to `mir-json` when compiling a Rust file


RyanGlScott:
When you invoke `crux-mir` on an `.rs` file, it will first compile the `.rs` file by invoking `mir-json` with a default set of command-line options. These default options are reasonable most of the time, but there are situations where one might want to provide different options (e.g., changing the optimization level by passing `-Copt-level=...`). Currently, there isn't a very direct way of accomplishing this—you would have to first compile a MIR JSON file separately, and then pass the resulting JSON file to `crux-mir`.

We should take inspiration from `crux-llvm`, which has a dedicated `--clang-opts` option (along with a `CLANG_OPTS` environment variable) for controlling what additional options to pass `clang`. Having something akin to `--clang-opts`/`CLANG_OPTS` for `crux-mir` would be very convenient.

sand7000:
Here is my plan of attack before I dig in too far:

Add support for passing arguments to `mir-json` via a repeatable flag `--mir-json-arg`. This makes argument boundaries explicit and avoids the need for ad-hoc string splitting.

```
crux-mir foo.rs \
  --mir-json-arg -Copt-level=3 \
  --mir-json-arg --edition=2021 \
  --mir-json-arg --extern \
  --mir-json-arg foo=path\ with\ spaces/libfoo.rlib
```

From there I would add two things.:

1. Running `crux-mir` with `-v` would cause crux-mir to print the exact `mir-json` command it invokes. This is purely for transparency/debugging, flags are forwarded verbatim.
2. If the user supplies an `--edition` flag via `--mir-json-arg` the default edition flag currently added by crux-mir would be omitted.

Look reasonable?

RyanGlScott:
This looks pretty reasonable to me. Some additional thoughts:

* If you type `crux-mir --mir-json-arg -Copt-level=3`, I'm not sure whether Crux's command-line parser will interpret that as `crux-mir "--mir-json-arg=-Copt-level=3"` (which is the intended behavior) or `crux-mir "--mir-json-arg" "-Copt-level=3"` (i.e., treating `-Copt-level=3` as a separate `crux-mir` argument, which is not the intended behavior). Depending on how Crux's CLI parser works, we may have to instruct users to explicitly write `--mir-json-arg="..."` to avoid the latter behavior.
[...]

PR #1558, opened 2025-10-08
`crux-mir`: Always print error messages upon `mir-json` compilation failures

### LINKED ISSUE #1084: Make `crux-mir` print compiler errors from `rustc`
If you run `crux-mir` on a malformed `.rs` file, e.g.,

```rs
syntax_error()
```

Then by default, this is the error that `crux-mir` will give you:

```
$ cabal run exe:crux-mir -- err.rs
user error (Error 101 while running mir-json on err.rs)
```

It's only if you run `crux-mir` with higher verbosity settings that the actual problem is made clear:

```
$ cabal run exe:crux-mir -- err.rs --sim-verbose=3
Generating err.mir
error: expected one of `!` or `::`, found `(`
 --> err.rs:1:13
  |
1 | syntax_error()
  |             ^ expected one of `!` or `::`

error: aborting due to previous error

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: ErrorGuaranteed(())', src/bin/mir-json.rs:65:89
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
user error (Error 101 while running mir-json on err.rs)
```

In this small example, it's somewhat obvious what the issue is, but in larger examples, it can be befuddling to see nothing but `Error 101` until you realize what is actually going on. It would be preferable for `crux-mir` to print these sorts of `rustc` errors by default, without needing to resort to higher verbosity levels.

RyanGlScott:
Previously, compiler errors would only be shown if the user explicitly opted into them using `--sim-verbose=3` or higher. This is a fairly counterintuitive default setting, as it means that users who run `crux-mir` on a malformed Rust file (a fairly common occurrence) will have to scratch their heads until they realize why `crux-mir` isn't showing them what went wrong.

This patch changes `crux-mir` to unconditionally show compiler error messages when compilation fails, regardless of the `--sim-verbose` level. I have adopted a convention for formatting the standard output/standard error of `mir-json` that is very similar to what `crux-llvm` uses when it unsucessfully compiles something using a `clang` subprocess.

Fixes #1084.

### DIFF
--- a/crux-mir/src/Mir/Generate.hs
+++ b/crux-mir/src/Mir/Generate.hs
@@ -35,6 +35,9 @@ import Mir.Mir
 import Mir.ParseTranslate (parseMIR)
 import Mir.PP()
 
+import qualified Crux.Config.Common as Crux (CruxOptions(..), OutputOptions(..))
+import qualified Crux.Config.Load as Crux (ColorOptions(..))
+
 import Debug.Trace
 
 
@@ -60,11 +63,18 @@ mirJsonOutFile rustFile = rustFile -<.> "mir"
 getRlibsDir :: (?defaultRlibsDir :: FilePath) => IO FilePath
 getRlibsDir = maybe ?defaultRlibsDir id <$> lookupEnv "CRUX_RUST_LIBRARY_PATH"
 
-compileMirJson :: (?defaultRlibsDir :: FilePath) => Bool -> FilePath -> IO ()
-compileMirJson keepRlib rustFile = do
+compileMirJson :: (?defaultRlibsDir :: FilePath) => Crux.CruxOptions -> Bool -> FilePath -> IO ()
+compileMirJson cruxOpts keepRlib rustFile = do
     let outFile = rustFile -<.> "bin"
 
     rlibsDir <- getRlibsDir
+    -- rustc produces colorful error messages, so preserve the colors whenever
+    -- possible when printing the error messages back out to the user.
+    let colorOpts
+          | Crux.noColorsErr $ Crux.colorOptions $ Crux.outputOptions cruxOpts
+          = []
+          | otherwise
+          = ["--color=always"]
     -- TODO: don't hardcode -L library path
     let cp =
           Proc.proc "mir-json" $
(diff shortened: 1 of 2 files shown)


crux-mir/src/Mir/Generate.hs
```diff
@@ -77,23 +77,26 @@ compileMirJson keepRlib quiet rustFile = do
...
     case ec of
-        ExitFailure cd -> fail $
-            "Error " ++ show cd ++ " while running mir-json on " ++ rustFile
+        ExitFailure cd -> fail $ unlines $
+            [ "Error " ++ show cd ++ " while running mir-json on " ++ rustFile
+            , "*** Standard out:"
+            ] ++
+            [ "   " ++ l | l <- lines sout ] ++
+            [ "*** Standard error:" ] ++
+            [ "   " ++ l | l <- lines serr ]
```
spernsteiner: FWIW I don't think rustc or mir-json prints anything to stdout during normal builds.  Rustc errors/warnings and env_logger messages all go to stderr.

crux-mir/src/Mir/Generate.hs
```diff
@@ -77,23 +77,26 @@ compileMirJson keepRlib quiet rustFile = do
...
             , "-Z", "ub-checks=false"
             , "-o", outFile]
-    let cp' = if not quiet then cp else
-            (cp { Proc.std_out = Proc.NoStream, Proc.std_err = Proc.NoStream })
-    ec <- Proc.withCreateProcess cp' $ \_ _ _ ph -> Proc.waitForProcess ph
+    (ec, sout, serr) <- Proc.readCreateProcessWithExitCode cp ""
     case ec of
-        ExitFailure cd -> fail $
-            "Error " ++ show cd ++ " while running mir-json on " ++ rustFile
+        ExitFailure cd -> fail $ unlines $
```
samcowger: Embedding this output in `fail` seems to subject it to `crux-mir`'s colorization, which feels like a bummer, given how richly colored `rustc` errors are. Local experimentation suggests we could avoid this if we
- add `--color=always` to the `mir-json` invocation,
- `hPutStrLn stderr` this message, and
- fail out afterwards with `fail "compilation error"` or some such simple message.

Does that sound plausible and worthwhile to you?

samcowger (approved):
Thank you for this!

crux-mir/src/Mir/Generate.hs
RyanGlScott: Indeed, I think you're right. I mainly did it this way to be consistent with `crux-llvm`, but I think it would be reasonable to deviate from `crux-llvm`'s conventions a little bit here for the sake of making the output a bit more concise.

crux-mir/src/Mir/Generate.hs
RyanGlScott: I like this train of thought, although I'm a bit cautious about being overly opinionated on which flags to pass `rustc`. In particular, always passing `--color=always` to `rustc` would mean that the compiler output would _always_ be colorized, even if a user set the [`NO_COLOR` environment variable](https://no-color.org/). Even if https://github.com/GaloisInc/crucible/issues/1556 were implemented, a user would have no way to override this setting, since passing `--color` to `rustc` twice causes an error, e.g.,

```
$ rustc --color=always --color=never test.rs
error: Option 'color' given more than once
```

Perhaps the thing to do here is to control whether `--color=always` is passed or not based on Crux's `--no-color` flag. That should be reasonably consistent with how other bits of colorized output work in Crux.

samcowger: Ah, fair - that sounds like a reasonable approach to me. Is that simple enough to fold into this PR, or would you prefer that we open an issue about it and address it separately?

RyanGlScott: I'll try implementing this as part of this PR.

RyanGlScott: I've pushed a new version of the PR which preserves colors in `rustc` error messages. Let me know what you think.

samcowger: LGTM, much appreciated!

PR #1559, opened 2025-10-08
llvm: Assert that assertions are enabled in the test suite

langston-barrett:
Otherwise, they will rarely, if ever, be checked! See `Note [Asserts]`.


langston-barrett:
[EDIT]: Hah, this exposes that the assertion that I wrote in my previous PR was mistaken!

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Common.hs
```diff
@@ -467,7 +467,9 @@ loadBitvector lo lw so v = do
         let d = lo - so
         -- Store is before load.
         valueLoad lo ltp lo (SelectSuffixBV d (sw - d) v)
-      | otherwise -> assert (lo == so && lw < sw) $
+      | otherwise ->
+        -- TODO: This assertion can fail, uncovered by #1559.
```
langston-barrett: If we decide to move forward with this, I will file a bug and refer to it here.

RyanGlScott: Please do!

langston-barrett:
I agree, though I only found two uses outside of `crucible-llvm` so it's probably not super urgent.
```
crucible/src/Lang/Crucible/Simulator/Evaluation.hs
crucible/src/Lang/Crucible/CFG/SSAConversion.hs
```
(via `rg 'assert [^b]' cru*/**/*.hs`)

PR #1561, opened 2025-10-08
Execution features to record and replay traces

langston-barrett:
This PR introduces two execution features:

- The first records a "trace" (i.e., `SymSequence`) of `ProgramLoc`s as they are executed
- The second *replays* traces by aborting paths that stray from the given trace

These can be used to guide the simulator down paths of interest. Some example use-cases:

- Concolic testing, whereby you begin the simulator with concrete data and see what path it takes, but then re-execute a prefix of the path on symbolic data and ask for a model that takes a different branch.
- Ensuring the accuracy of concretization by replaying a path to an assertion failure with concretized data.
- Interactively debugging an assertion failure by only replaying the path to the failure (with the same symbolic data).

I'm planning to do a bit more testing in a downstream application before merging.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
+++ b/crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
@@ -167,7 +167,8 @@ replayTraceLength simState = do
 -- | An 'C.ExecutionFeature' to record traces.
 --
 -- During execution this logs program locations to a Crucible global variable.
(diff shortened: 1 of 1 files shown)


langston-barrett:
Okay, testing in a downstream application indicates that these are working as expected.

langston-barrett:
The intent is that you use `getRecordedTrace`, which is the only way for an API consumer to construct a `RecordedTrace`. It would be good to say as much in the docs, so I'll do so!

PR #1563, opened 2025-10-09
Support the exact_div intrinsic

### LINKED ISSUE #1273: `crucible-mir`: Support the `exact_div` intrinsic
`crucible-mir` does not currently support the [`exact_div`](https://doc.rust-lang.org/std/intrinsics/fn.exact_div.html) family of intrinsics, but it reasonably could. `exact_div(x, y)` behaves like integer division, except that it produces undefined behavior when `x % y != 0` or `y == 0` or `x == T::MIN && y == -1`. Among other uses, this is used in the implementation of [`array_chunks`](https://doc.rust-lang.org/std/primitive.slice.html#method.array_chunks).

sand7000:
## Description

Implements translation and semantics for the [`core::intrinsics::exact_div`](https://doc.rust-lang.org/std/intrinsics/fn.exact_div.html) intrinsic.  
This intrinsic performs integer division that triggers undefined behavior if the division is not exact (i.e., leaves a nonzero remainder) or if the operation overflows.  
Both signed and unsigned integer types are supported.

Includes corresponding tests in `crux-mir` to verify both exact and UB cases, and updates the changelog accordingly.

## Verification

Without the code change, the new tests failed with:

```
[Crux]   Translation error in exact_div/280fa3d4::f[0]: callExp: Don't know how to call core/9999e705::intrinsics[0]::exact_div[0]::_inste4581d0a71164688[0]
```

With the new implementation, both tests pass as expected.

Fixes #1273

### DIFF
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -946,42 +946,42 @@ unchecked_shr =
 -- This operation performs integer division that triggers undefined behavior
 -- if the division has a nonzero remainder or overflows.
 -- Supports both signed and unsigned integer types.
-makeExactDiv :: String -> CustomRHS
-makeExactDiv name =
+makeExactDiv :: CustomRHS
+makeExactDiv =
   \_substs -> Just $ CustomOp $ \opTys ops -> case (opTys, ops) of
     ([TyInt _ , TyInt _ ], [e1, e2]) ->
-      checkExact name (Just Signed) e1 e2
+      evalExactDiv Signed e1 e2
     ([TyUint _, TyUint _], [e1, e2]) ->
-      checkExact name (Just Unsigned) e1 e2
+      evalExactDiv Unsigned e1 e2
     _ ->
-      mirFail $ "BUG: invalid arguments to " ++ name ++ ": " ++ show (opTys, ops)
+      mirFail $ "BUG: invalid arguments to exact_div: " ++ show (opTys, ops)
 
-checkExact ::
-  String ->
-  Maybe ArithType ->
+-- Evaluate the `exact_div` intrinsic for signed or unsigned integers.
+-- Raises UB if the division overflows or is not exact.
+evalExactDiv ::
+  ArithType ->
   MirExp s -> MirExp s ->
   MirGenerator h s ret (MirExp s)
-checkExact name arithType e1 e2 = do
-  (q, overflow) <- evalBinOp Div arithType e1 e2
-  (r, _)        <- evalBinOp Rem arithType e1 e2
+evalExactDiv arithType e1 e2 = do
+  (q, overflow) <- evalBinOp Div (Just arithType) e1 e2
+  (r, _)        <- evalBinOp Rem (Just arithType) e1 e2
   case (q, r) of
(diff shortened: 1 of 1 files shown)


crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -941,6 +942,47 @@ unchecked_shr =
...
+    ([TyInt _ , TyInt _ ], [e1, e2]) ->
+      checkExact name (Just Signed) e1 e2
+    ([TyUint _, TyUint _], [e1, e2]) ->
+      checkExact name (Just Unsigned) e1 e2
+    _ ->
+      mirFail $ "BUG: invalid arguments to " ++ name ++ ": " ++ show (opTys, ops)
+
+checkExact ::
+  String ->
+  Maybe ArithType ->
```
RyanGlScott: This function is only ever called with `Just`, so for the sake of avoiding unrepresentable states, I'd favor changing the type of this argument from a `Maybe ArithType` to an `ArithType`.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -941,6 +942,47 @@ unchecked_shr =
...
+makeExactDiv name =
+  \_substs -> Just $ CustomOp $ \opTys ops -> case (opTys, ops) of
+    ([TyInt _ , TyInt _ ], [e1, e2]) ->
+      checkExact name (Just Signed) e1 e2
+    ([TyUint _, TyUint _], [e1, e2]) ->
+      checkExact name (Just Unsigned) e1 e2
+    _ ->
+      mirFail $ "BUG: invalid arguments to " ++ name ++ ": " ++ show (opTys, ops)
+
+checkExact ::
```
RyanGlScott: The name "`checkExact`" implies that this does nothing but perform a validity check of some kind, but it also does more by computing the quotient. Perhaps we should rename this function to something like "`evalExactDiv`" for consistency with the `evalBinOp` function.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -941,6 +942,47 @@ unchecked_shr =
     , makeUncheckedArith "unchecked_shr" Shr
     )
 
+-- Implement the [`core::intrinsics::exact_div`] intrinsic.
+-- This operation performs integer division that triggers undefined behavior
+-- if the division has a nonzero remainder or overflows.
+-- Supports both signed and unsigned integer types.
+makeExactDiv :: String -> CustomRHS
```
RyanGlScott: Is there any need to parameterize this function by a `String` argument? There's only one call site to this function, so in practice, `name` is always `exact_div`.

crucible-mir/src/Mir/TransCustom.hs
sand7000: Good point, fixed.

crucible-mir/src/Mir/TransCustom.hs
sand7000: Good point, fixed.

crucible-mir/src/Mir/TransCustom.hs
sand7000: You are right, fixed.

Issue #1564, opened 2025-10-10
`crucible-mir`: Incorrect handling of `AgElem_RefPath` in pointer equality/overlap checking


RyanGlScott:
Consider the following program:

```rs
#[cfg(not(with_main))]
#[cfg_attr(crux, crux::test)]
pub fn main() {
    let x: (i32, i32) = (0, 1);
    let p0 = &x as *const (i32, i32);
    let q0: *const (u8, u8) = p0.cast();
    let p1: &i32 = unsafe { &(*p0).0 };
    let q1: &u8  = unsafe { &(*q0).0 };
    let p2 = p1 as *const i32;
    let q2 = q1 as *const u8;
    assert!(p2 == q2.cast());
}
```

The assertion ought to pass, as both `p2` and `q2` should have the same memory address, but at different types (`*const i32` and `*const u8`, respectively). We can see that this assertion passes when you compile it with `rustc` normally ([playground link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=f8ad8b109ac757c98d07072d34e78768)), but it fails when you simulate it using `crux-mir`:

```
$ cabal -v0 run exe:crux-mir -- test.rs
test test/3485a068::main[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/3485a068::main[0] counterexamples ----
[Crux] Found counterexample for verification goal
[Crux]   test.rs:11:5: 11:29 !test.rs:11:5: 11:29: error: in test/3485a068::main[0]
[Crux]   panicking::panic, called from test/3485a068::main[0]

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Invalid.
```

The culprit ultimately lies in how pointer equality is defined in `crucible-mir`:

https://github.com/GaloisInc/crucible/blob/0217d6fa66c0c21de6ef2e1e5b1501864b92437b/crucible-mir/src/Mir/Intrinsics.hs#L1855-L1879

`refPathEq`, which checks if two references/pointers have the same path, is missing a case for `AgElem_RefPath`, which is used in the program above to access the first element of a tuple (i.e., a `MirAggregate`). As a result, this function returns `False`, thereby deeming the two pointer addresses to be inequal when they are in fact equal.

[...]

Issue #1544, opened 2025-09-19
`crucible-mir`: enable, un-disable, and address more GHC warnings


samcowger:
Several modules in `crucible-mir` manually manage GHC warnings with a module-level `OPTIONS_GHC` pragma. For instance, in `Mir.Trans`: https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crucible-mir/src/Mir/Trans.hs#L26-L29

In that module in particular, `-fno-warn-name-shadowing` and `-fno-warn-unused-matches` each hide around 60 warnings that I think are worth addressing. They, and most if not all other ad-hoc `-fno` flags in the package, probably ought to be removed.

`crucible-mir-debug` appears to be the most modern (as in, recently created) package in this repo, and it enables a well-considered and extensive range of warnings and errors, so I think we could follow its lead: https://github.com/GaloisInc/crucible/blob/45648cc1ebc834fea38f126e6cac869101189fd0/crucible-mir-debug/crucible-mir-debug.cabal#L17-L110

PR #1565, opened 2025-10-10
`Mir.Intrinsics`: Add missing cases for `AgElem_RefPath`

RyanGlScott:
These cases allow `crucible-mir` to properly simulate pointer equality and
overlap checking when accessing aggregate elements is involved. See the
`test/conc_eval/tuple/ref_path_equality.rs` test case for an example of where
proper `AgElem_RefPath` support in pointer equality checking is crucial.

Fixes https://github.com/GaloisInc/crucible/issues/1564.

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -1879,13 +1879,18 @@ refPathEq sym (VectorAsMirVector_RefPath tpr1 p1) (VectorAsMirVector_RefPath tpr
   | Just Refl <- testEquality tpr1 tpr2 = refPathEq sym p1 p2
 refPathEq sym (ArrayAsMirVector_RefPath tpr1 p1) (ArrayAsMirVector_RefPath tpr2 p2)
   | Just Refl <- testEquality tpr1 tpr2 = refPathEq sym p1 p2
-refPathEq sym (AgElem_RefPath off1 sz1 _tpr1 p1) (AgElem_RefPath off2 sz2 _tpr2 p2)
-  | sz1 == sz2 = do
+refPathEq sym (AgElem_RefPath off1 _sz1 _tpr1 p1) (AgElem_RefPath off2 _sz2 _tpr2 p2) = do
     offEq <- liftIO $ bvEq sym off1 off2
-    -- NB: Don't check the TypeReprs for equality, as pointers with the same
-    -- memory addresses can have different types if pointer casting is involved
-    -- (see the crux-mir/test/conc_eval/tuple/ref_path_equality.rs test case
-    -- for an example).
+    -- NB: Don't check the following for equality:
+    --
+   --
+    -- * The TypeReprs (_tpr{1,2}), as pointers with the same memory addresses
+    --   can have different types if pointer casting is involved (see the
+    --   crux-mir/test/conc_eval/tuple/ref_path_equality.rs test case for an
+    --   example).
+    --
+    -- * The sizes (_sz{1,2}), as pointers of different types may have
+    --   different layout sizes.
     pEq <- refPathEq sym p1 p2
     liftIO $ andPred sym offEq pEq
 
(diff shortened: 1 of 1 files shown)


crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -29,7 +29,7 @@
 {-# LANGUAGE UndecidableInstances #-}
 
 {-# OPTIONS_GHC -Wincomplete-patterns -Wall #-}
-{-# OPTIONS_GHC -fno-warn-incomplete-patterns -fno-warn-unused-imports #-}
+{-# OPTIONS_GHC -fno-warn-unused-imports #-}
```
samcowger, line 32: Obligatory plug for #1544 🙃

crux-mir/test/conc_eval/tuple/ref_path_equality.rs
```diff
@@ -0,0 +1,20 @@
...
+// deem two pointers to be equal even when their types differ, which is
+// something that can occur when pointer type casting is involved.
+
+#[cfg_attr(crux, crux::test)]
+pub fn crux_test() {
+    let x: (i32, i32) = (0, 1);
+    let p0 = &x as *const (i32, i32);
+    let q0: *const (u8, u8) = p0.cast();
+    let p1: &i32 = unsafe { &(*p0).0 };
+    let q1: &u8  = unsafe { &(*q0).0 };
```
samcowger, line 22: This may be overly pedantic (and/or wrong), but I'm not sure Rust's layout rules guarantee that `(*p0).0` and `(*q0).0` have the same address here. Can't Rust choose to store the field we know as `.0` either before or after `.1` in memory, and make one choice for `(u8, u8)` and another for `(i32, i32)`?

crucible-mir/src/Mir/Intrinsics.hs
RyanGlScott: I, too, yearn for `-Wall`.

crux-mir/test/conc_eval/tuple/ref_path_equality.rs
RyanGlScott: Yes, I think this is a bit coupled to the specific choices that today's `rustc` happens to make when laying out tuples, and I don't believe that there are any strong guarantees that tomorrow's `rustc` will make the same choices. At the same time, I don't know of another way to observe the behavior under test.

We could make a note of this fact in the comments of the test case, saying that we may need to adjust the specifics of the implementation to match any changes to Rust's tuple layout in the future. Alternatively, we could remove this test case entirely, although I'm less excited about this option.

samcowger: I would only go so far as to add a comment, and only then if you feel motivated to do so - I certainly don't want to remove the test case entirely. Out of curiosity, though, why do we need to do this with tuples in the first place? As compared to, say:
```rs
    let x: i32 = 0;
    let p: *const i32 = &x as *const i32;
    let q: *const u8 = p.cast();
    assert!(p == q.cast());
```

samcowger: Oh, right, because we're specifically testing the behavior of aggregates. Please disregard!

RyanGlScott: `crucible-mir` correctly handles that example, even without the changes in this PR. The reason that we need tuples is because the bug is only observable when the pointers being checked for equality have `AgElem_RefPath` somewhere in their paths, and currently, that is only possible to construct when you have a pointer to a tuple.

This presents another possibility: wait until we change the representations of other types to use `MirAggregate` as well, and then rewrite this test case to use another type with a more stable layout (e.g., a `#[repr(C)]` struct). That would require holding off on adding a regression test for #1564 for a while, however.

RyanGlScott: For now, I've settled for adding a disclaimer in the comments on this test case.

samcowger: Better to go with this, I think.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1876,7 +1879,24 @@ refPathEq sym (VectorAsMirVector_RefPath tpr1 p1) (VectorAsMirVector_RefPath tpr
   | Just Refl <- testEquality tpr1 tpr2 = refPathEq sym p1 p2
 refPathEq sym (ArrayAsMirVector_RefPath tpr1 p1) (ArrayAsMirVector_RefPath tpr2 p2)
   | Just Refl <- testEquality tpr1 tpr2 = refPathEq sym p1 p2
-refPathEq sym _ _ = return $ falsePred sym
+refPathEq sym (AgElem_RefPath off1 sz1 _tpr1 p1) (AgElem_RefPath off2 sz2 _tpr2 p2)
+  | sz1 == sz2 = do
```
spernsteiner: I don't think we want to require `sz1 == sz2` here.  Currently all sizes are 1, but once we add proper layout support, the `*const i32` vs `*const u8` example will use size 4 vs size 1, and thus will be considered unequal.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2118,7 +2143,24 @@ refPathOverlaps sym path1 path2 = do
...
+    go (AgElem_RefPath off1 sz1 _tpr1 _ `RrpCons` rrp1)
+        (AgElem_RefPath off2 sz2 _tpr2 _ `RrpCons` rrp2)
+      | sz1 == sz2 = do
+        offEq <- liftIO $ bvEq sym off1 off2
+        -- NB: Don't check the TypeReprs for equality, as pointers with the
+        -- same memory addresses can have different types if pointer casting is
+        -- involved (see the crux-mir/test/conc_eval/tuple/ref_path_equality.rs
+        -- test case for an example).
+        pEq <- go rrp1 rrp2
+        liftIO $ andPred sym offEq pEq
```
spernsteiner: Once layout is added, I think this will need to check whether `off1 .. off1 + sz1` overlaps `off2 .. off2 + sz2`.  Previous cases didn't need logic like this because in `MirVector`, `StructType`, etc, all sub-locations of the value are disjoint.

crucible-mir/src/Mir/Intrinsics.hs
RyanGlScott: I've removed this check and added a comment explaining why.

crucible-mir/src/Mir/Intrinsics.hs
RyanGlScott: I've reworked this to check for overlap instead of just checking if the offsets are equal.

PR #1568, opened 2025-10-14
Move initial trace into personality

2over12:
this setup allows clients to change the initial trace

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
+++ b/crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
@@ -24,7 +24,7 @@ module Lang.Crucible.Simulator.RecordAndReplay (
   replayFeature,
   initialTrace,
   traceGlobal,
-  emptyReplayTrace
+  emptyRecordedTrace
 ) where
 
 import Control.Exception qualified as X
@@ -135,10 +135,8 @@ locAsStr sym = do
   let txtLoc = Text.pack (show loc)
   W4.stringLit sym (W4.UnicodeLiteral txtLoc)
 
-
 emptyRecordedTrace :: sym -> IO (RecordedTrace sym)
-emptyRecordedTrace sym =  RecordedTrace <$> CSSS.nilSymSequence sym
-   
+emptyRecordedTrace sym = RecordedTrace <$> CSSS.nilSymSequence sym
 
 getRecordTrace ::
   HasRecordState p p sym ext rtp =>


crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
```diff
@@ -273,6 +274,19 @@ getRecordedTrace globals (RecordState g) sym evalBool = do
         Just (W4.UnicodeLiteral s') -> pure s'
         Nothing -> panic "getRecordedTrace" ["Non-literal trace element?"]
 
+{- | Inserts a recorded trace into the state's replay trace variable
+The replay feature will follow this trace if it is enabled
+-}
+insertReplayTrace ::
+  (HasReplayState p p sym ext rtp) =>
+  C.SimState p sym ext rtp f args ->
+  C.RegValue sym TraceType ->
```
langston-barrett, line 295: If we expose this function as-is, we'd at least want to also export `TraceType`. However, the API of this module was carefully crafted to *not* expose the underlying representation for the sake of implementation flexibility. I wonder if we could instead expose a more limited API that takes a `[W4.ProgramLoc]`, i.e., a concrete trace, rather than a `RegValue sym TraceType` (a symbolic trace). What do you think?

crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
```diff
@@ -61,17 +64,23 @@ newtype RecordState p sym ext rtp
   = RecordState (C.GlobalVar TraceType)
     -- ^ constructor intentionally not exported
 
+{- | A trace from 'recordFeature', processed and ready for consumption by
+'replayFeature'.
+-}
+newtype RecordedTrace sym
+  = RecordedTrace (C.RegValue sym TraceType)
+
```
langston-barrett, line 72: A bit odd that this moved around the file and changed comment styles - this kind of noise makes it (very slightly) harder to interpret a diff, so it's nice to avoid if possible. No changes requested, just a note for the future.

crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
```diff
@@ -61,17 +64,23 @@ newtype RecordState p sym ext rtp
...
 --
 -- * @p@: see 'C.cruciblePersonality'
 -- * @sym@: instance of 'Lang.Crucible.Backend.IsSymInterface'
 -- * @ext@: language extension, see "Lang.Crucible.CFG.Extension"
 -- * @rtp@: type of the simulator return value
 type ReplayState :: Type -> Type -> Type -> Type -> Type
-newtype ReplayState p sym ext rtp
-  = ReplayState (C.GlobalVar TraceType)
+data ReplayState p sym ext rtp
+  = ReplayState {_traceGlobal :: (C.GlobalVar TraceType), _initialTrace :: (RecordedTrace sym)}
```
langston-barrett: I would find this easier to read if each field were on its own line:
```suggestion
  = ReplayState
    { _traceGlobal :: (C.GlobalVar TraceType)
    , _initialTrace :: (RecordedTrace sym)
    }
```

crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
```diff
@@ -61,17 +64,23 @@ newtype RecordState p sym ext rtp
...
     -- ^ constructor intentionally not exported
-
+Lens.makeLenses ''ReplayState
 -- | Constructor for 'RecordState'
```
langston-barrett, line 88:
```suggestion
Lens.makeLenses ''ReplayState

-- | Constructor for 'RecordState'
```

crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
```diff
@@ -122,6 +131,11 @@ locAsStr sym = do
   let txtLoc = Text.pack (show loc)
   W4.stringLit sym (W4.UnicodeLiteral txtLoc)
 
+
+emptyReplayTrace :: sym -> IO (RecordedTrace sym)
+emptyReplayTrace sym =  RecordedTrace <$> CSSS.nilSymSequence sym
```
langston-barrett: For consistency with the type name:
```suggestion
emptyRecordedTrace :: sym -> IO (RecordedTrace sym)
emptyRecordedTrace sym =  RecordedTrace <$> CSSS.nilSymSequence sym
```

crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
```diff
@@ -273,6 +284,19 @@ getRecordedTrace globals (RecordState g) sym evalBool = do
...
+-}
+insertReplayTrace ::
+  (HasReplayState p p sym ext rtp) =>
+  C.SimState p sym ext rtp f args ->
+  C.RegValue sym TraceType ->
+  C.SimState p sym ext rtp f args
+insertReplayTrace st v = do
+  let simCtx = st ^. C.stateContext
+  let g = simCtx ^. C.cruciblePersonality . replayState . traceGlobal
+  st & C.stateGlobals %~ C.insertGlobal g v
```
langston-barrett, line 300: Now that this is no longer exported it also is basically noise in the diff (see above comment).

langston-barrett (approved):
I think this is definitely a nice change to make this code easier to work with downstream, just some minor style suggestions.

crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs
2over12: i could be wrong but i think this is so it's above where the lense is created

langston-barrett: I see, then this comment only applies to changing the comment style!

PR #1569, opened 2025-10-16
Equality of `SymSequence`s

langston-barrett:
Needed for #1550. This code is complicated enough that it needs some Hedgehog tests before merging.


langston-barrett:
Closing until I have some time to revisit.

PR #1572, opened 2025-10-17
concurrency: Classy lenses for personality

### LINKED ISSUE #1570: concurrency: Classy lenses for personality
`crucible-concurrency` currently hardcodes [`ThreadExec`](https://github.com/GaloisInc/crucible/blob/0c3e694a13f264121df6ff1e38d2c623d89a0386/crucible-concurrency/src/Crucibles/ExploreTypes.hs#L27) as the Crucible [personality](https://github.com/GaloisInc/crucible/blob/0c3e694a13f264121df6ff1e38d2c623d89a0386/crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs#L1328) in a lot of places. This is inflexible, because it doesn't allow `crucible-concurrency` to be embedded into a larger application with other components in the personality. We should replace this hard-coding using a "classy lenses" approach, as is done in the `RecordAndReplay` feature:

https://github.com/GaloisInc/crucible/blob/0c3e694a13f264121df6ff1e38d2c623d89a0386/crucible/src/Lang/Crucible/Simulator/RecordAndReplay.hs#L87-L96

The reason I actually need this is that I want to perform the same transformation for `crucible-debug` (taking the state of the debugger out of an `IORef` and putting it in the personality), but this adds a constraint to Crux's interface that the personality contain the debugger state. Since `ThreadExec` does not, in fact, contain a debugger state, I need to use a different personality type even within `crucible-concurrency`.

langston-barrett:
Fixes #1570.

### DIFF
--- a/crucible-concurrency/src/Crucibles/ExploreTypes.hs
+++ b/crucible-concurrency/src/Crucibles/ExploreTypes.hs
@@ -54,7 +54,7 @@ data Exploration p alg ext ret sym = Exploration
   }
 makeLenses ''Exploration
 
--- | A class for Crucible personality types @p@ which contain a
+-- | A class for Crucible personality types @p@ which contain an
 -- 'Exploration'.
 class HasExploration p alg ext ret sym | p -> alg ext ret sym where
   exploration :: Lens' p (Exploration p alg ext ret sym)


crucible-concurrency/src/Crucibles/ExploreTypes.hs
```diff
@@ -51,17 +54,41 @@ data Exploration alg ext ret sym = Exploration
   }
 makeLenses ''Exploration
 
-stateExpl :: Simple Lens (SimState (ThreadExec alg sym ext ret) sym ext r f a) (ThreadExec alg sym ext ret)
-stateExpl = stateContext.cruciblePersonality
+-- | A class for Crucible personality types @p@ which contain a
+-- 'Exploration'.
```
RyanGlScott: Grammar police:

```suggestion
-- | A class for Crucible personality types @p@ which contain an
-- 'Exploration'.
```

RyanGlScott (approved):

PR #1574, opened 2025-10-17
`crux-mir`: Support printing `MirVector`s in `dump_rv` override

RyanGlScott:
The override for `crucible::dump_rv` was missing a single `crucible-mir`-specific intrinsic, `MirVector`. This adds support for it. With this, the following program:

```rs
// test.rs
extern crate crucible;
use crucible::*;
use crucible::alloc::reallocate;
use crucible::array::Array;

pub fn main() {
    let mut x = <[u8; 2]>::symbolic("x");
    crucible::dump_rv("x0", x);
    let mut xp = x.as_mut_ptr();
    crucible::alloc::reallocate(xp, 5);
    unsafe {
        *xp.offset(2) = 42;
        if x[0] == 0 {
            *xp.offset(3) = 27;
        }
    }
    crucible::dump_rv("x1", x);

    let mut y = <Array<u8>>::symbolic("y");
    y = y.update(2, 42);
    y = y.update(3, 27);
    crucible::dump_rv("y", y);
}
```

Now prints:

```
test test/10cba7c8::main[0]: x0 = [cx@7:bv, cx@8:bv]
x1 = [cx@7:bv, cx@8:bv, 0x2a:[8], 0x1b:[8] (if eq 0x0:[8] cx@7:bv), <uninit>]
y = arrayMap [BV 2::[64]] 0x2a:[8] [BV 3::[64]] 0x1b:[8] cy@14:a
ok

[Crux] Overall status: Valid.
```

### DIFF
--- a/crux-mir/src/Mir/Overrides.hs
+++ b/crux-mir/src/Mir/Overrides.hs
@@ -625,20 +625,45 @@ mirReferencePrettyFn = IntrinsicPrettyFn $ \Ctx.Empty ref -> PP.viaShow ref
 mirAggregatePrettyFn :: forall sym. IsSymInterface sym =>
   IntrinsicPrettyFn sym MirAggregateSymbol
 mirAggregatePrettyFn = IntrinsicPrettyFn $ \Ctx.Empty (MirAggregate totalSize m) ->
-  let renderVal :: forall tpr ann. TypeRepr tpr -> RegValue sym (MaybeType tpr) -> PP.Doc ann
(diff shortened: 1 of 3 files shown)


samcowger:
It looks like this program would cause `crux-mir` to fail, absent your changes. Given that, do you think it's worth including as a test case?

RyanGlScott:
Probably, although I'm not entirely convinced that the output of the particular test case I tried is stable. Let me cook up a variant of this that doesn't use fresh symbolic names to avoid this problem...

Issue #1482, opened 2025-07-30
`crux-mir`: Create `crucible-mir-debug` and use it in `--debug` mode


RyanGlScott:
https://github.com/GaloisInc/crucible/pull/1283 added a `crucible-llvm-debug` library that adds `crucible-llvm`–specific extensions to `crucible-debug`. https://github.com/GaloisInc/crucible/pull/1297 continued by adding a `--debug` flag to `crux-llvm` that drops into a Crucible debugger session when simulation starts. It would be convenient to have the same capabilities for `crux-mir`. Specifically, we should:

1. Create a `crucible-mir-debug` library that implements `crucible-mir`–specific extensions to `crucible-debug`. For now, this can be trivial by making `MIRCommand` and empty data type. In the future, we may want to expand this with MIR-specific commands.
2. Copy over the same `--debug` logic from `crux-llvm` to `crux-mir` (see the code added in https://github.com/GaloisInc/crucible/pull/1297).

RyanGlScott:
Actually, it turns out that _all_ Crux frontends currently support `--debug`, including `crux-mir`! The catch is that there are currently no MIR-specific debugging commands available in `crux-mir`. I'll leave this issue open as a reminder to create `crucible-mir-debug` so that we can have this MIR-specific functionality from the debugger.

RyanGlScott:
In order to integrate `crucible-debug`'s S-expression syntax-related functionality into `crux-mir`, we would first need something akin to `crucible-mir-syntax` (see https://github.com/GaloisInc/crucible/issues/1483).

PR #1575, opened 2025-10-17
debug: Overrides for interacting with the debugger

### LINKED ISSUE #1571: debug: Overrides for interacting with the debugger
I would like to write some overrides for interacting with the debugger. The primary motivation would be an override `debug : Unit -> Unit` that, when called, starts a session of the `crucible-debug` debugger. This would be a convenient way to drop into the debugger at a certain point in the program under test, like [x86_64's `int3`](https://en.wikipedia.org/wiki/INT_(x86_instruction)#INT3) instruction. This will require a few steps:

- Take the state of the debugger out of the `IORef` it lives in now and instead store it in the Crucible personality using the "classy lenses" approach
- Adapt Crux (and `crucible-concurrency`, and Crux-LLVM, and Crux-MIR) to use a personality that stores a debugger state
- Create helper functions for modifying the debugger state held in the personality, both explicitly via the `SimContext` (e.g., for execution features) and in the `OverrideSim` monad (for overrides)
- Define the override(s)
- (Optionally) define corresponding LLVM override(s) and use it in Crux-LLVM

Some other overrides that might be interesting:

1. Parse and execute an arbitrary debugger command (as a string literal)
2. Dedicated overrides for specific debugger commands (e.g., `backtrace`)
3. ???

langston-barrett:
Fixes #1571. Changes:

- Put the debugger state in the Crucible personality instead of an `IORef`
- Modify Crux to require debugger state in the personality, adapt Crux{,-LLVM,-MIR,Go,Wasm,JVM} to those changes
- Export helpers that allow other code to modify the debugger state (e.g., starting or stopping it)
- Create overrides that use those helpers, use those overrides in `crucible-cli`
- Change `crucible-cli` to only accept programs that return `Unit` (necessary for the above)

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Simulate.hs
+++ b/crux-llvm/src/Crux/LLVM/Simulate.hs
@@ -225,7 +225,7 @@ setupFileSim halloc llvm_file llvmOpts bak _maybeOnline =
   do let sym = backendGetSym bak
      memVar <- mkMemVar "crux:llvm_memory" halloc
 
-     simctx_ <- (setupSimCtxt halloc bak (memOpts llvmOpts) memVar)
+     simctx_ <- setupSimCtxt halloc bak (memOpts llvmOpts) memVar
      let simctx = simctx_ { printHandle = view outputHandle ?outputConfig }
 
      prepped <- prepLLVMModule llvmOpts halloc bak llvm_file memVar


langston-barrett:
Running the following program in `crucible-cli` will drop you into the debugger between the prints!

    (defun @main () Unit
      (start start:
        (print "before!\n")
        (funcall @dbg)
        (print "after!\n")
        (return ())))

crucible-debug/src/Lang/Crucible/Debug/Personality.hs
```diff
@@ -0,0 +1,72 @@
...
+import Lang.Crucible.Debug.Context qualified as Ctxt
+import Lang.Crucible.Debug.Inputs qualified as Inps
+import Lang.Crucible.Debug.Statement (Statement)
+import Lang.Crucible.Simulator.ExecutionTree qualified as C
+
+-- | A class for Crucible personality types @p@ which contain a
+-- 'RecordState'. This execution feature is polymorphic over
+-- 'RecordState' so that downstream users can supply their own
+-- personality types that extend 'RecordAndReplay' further.
+class HasContext p cExt sym ext t | p -> cExt sym ext t where
```
RyanGlScott: Instead of `RecordState`, should these Haddocks mention `Context`?

A similar question about the mention of `RecordAndReplay` here, since I think that is a copy-paste leftover.

crucible-go/tool/Main.hs
```diff
@@ -62,9 +74,20 @@ simulateGo copts _opts =
...
+              let cExts = Debug.voidExts
+              inps <- Debug.defaultDebuggerInputs cExts
+              dbgCtx <-
+                Debug.initCtx
+                  cExts
+                  (Debug.IntrinsicPrinters MapF.empty)
+                  inps
+                  Debug.defaultDebuggerOutputs
+                  CT.UnitRepr
+              let p = Personality dbgCtx
```
RyanGlScott, line 86: This appears to set up the necessary machinery for `crucible-go` to use a debugger-related override, but I'm not actually clear if it's possible to _use_ such an override from within `crucible-go`. Is this possible? If not, do we need to set up this additional machinery?

Similar questions apply to `crucible-jvm` and `crucible-wasm` as well.

crux-llvm/src/Crux/LLVM/Simulate.hs
```diff
@@ -199,8 +225,8 @@ setupFileSim halloc llvm_file llvmOpts bak _maybeOnline =
   do let sym = backendGetSym bak
      memVar <- mkMemVar "crux:llvm_memory" halloc
 
-     let simctx = (setupSimCtxt halloc bak (memOpts llvmOpts) memVar)
-                  { printHandle = view outputHandle ?outputConfig }
+     simctx_ <- (setupSimCtxt halloc bak (memOpts llvmOpts) memVar)
```
RyanGlScott: Redundant parentheses:

```diff
-     simctx_ <- (setupSimCtxt halloc bak (memOpts llvmOpts) memVar)
+     simctx_ <- setupSimCtxt halloc bak (memOpts llvmOpts) memVar
```

crucible-cli/src/Lang/Crucible/CLI.hs
```diff
@@ -145,13 +202,16 @@ simulateProgramWithExtension mkExt fn theInput outh profh opts hooks =
                       , parsedProgForwardDecs = fds
                       } = parsedProg
                 case find isMain cs of
-                  Just (AnyCFG mn@(C.Reg.cfgArgTypes -> Ctx.Empty)) ->
+                  Just (AnyCFG mn)
+                    | Ctx.Empty <- C.Reg.cfgArgTypes mn
+                    , CT.UnitRepr <- C.Reg.cfgReturnType mn ->
```
RyanGlScott, line 207: What is the motivation behind requiring the `main` function to return type `Unit`? (I'm not necessarily opposed to the idea, but it's unclear to me what part of the PR relates to this.)

RyanGlScott:
Note that I included commit https://github.com/GaloisInc/crucible/pull/1575/commits/eeea4a5098fbb405c26a1ab38523a9d1ef054cf3 from this PR into a different PR that I recently landed (https://github.com/GaloisInc/crucible/pull/1574) in order to make the CI work again.

crucible-debug/src/Lang/Crucible/Debug/Personality.hs
langston-barrett: Whoops, thanks for the catch!

crucible-go/tool/Main.hs
langston-barrett: Good question - this is not just necessary for the overrides, but also just for using the debugger generally. All of these frontends support the debugger through Crux's `--debug` flag (just as you [discovered](https://github.com/GaloisInc/crucible/issues/1482#issuecomment-3144486954) that Crux-MIR does).

Basically, this setup used to live in the `debugger` function, which stashed it in an `IORef`. For it to be part of the personality, clients need to set it up. Furthermore, each client is likely to want to move towards setting it up differently. For example, Crux-LLVM will want LLVM-related debugger commands (#1576), and `crucible-cli` optionally uses a passed-in `Handle` for output rather than the default debugger outputs.

crucible-cli/src/Lang/Crucible/CLI.hs
langston-barrett: It's an API chicken-and-egg problem. The debugger (and hence, the personality) needs to know that the overall return type of the simulator is `RegValue sym t` for a particular `t` (I don't fully remember why this was necessary but it definitely was important for things to work out right). `setupOverridesHook` needs to know the personality, and runs before parsing. We only have access to the return type of the main CFG *after* parsing. Thus, it's easiest to just fix it to be `Unit`.

langston-barrett:
> Note that I included commit [eeea4a5](https://github.com/GaloisInc/crucible/commit/eeea4a5098fbb405c26a1ab38523a9d1ef054cf3) from this PR into a different PR that I recently landed (#1574) in order to make the CI work again.

Noted, I went ahead and rebased onto `master`.

RyanGlScott (approved):

PR #1577, opened 2025-10-17
Consistently use same warning/error conventions for all `crucible-mir`-related packages

RyanGlScott:
This copies over the warning/error style from `crucible-mir-debug` over to `crucible-mir` and `crux-mir` for the sake of consistently handling warnings/errors the same across all MIR-related packages. This also fixes all of the warnings that were uncovered in the process of doing so.

Fixes https://github.com/GaloisInc/crucible/issues/1544.

### DIFF
--- a/crucible-mir/crucible-mir.cabal
+++ b/crucible-mir/crucible-mir.cabal
@@ -86,7 +86,6 @@ common shared
     -Werror=unrecognised-warning-flags
     -Werror=unsupported-calling-conventions
     -Werror=unsupported-llvm-version
-    -Werror=unticked-promoted-constructors
     -Werror=unused-do-bind
     -Werror=unused-imports
     -Werror=unused-record-wildcards
--- a/crucible-cli/crucible-cli.cabal
+++ b/crucible-cli/crucible-cli.cabal
@@ -84,7 +84,6 @@ common shared
     -Werror=unrecognised-warning-flags
     -Werror=unsupported-calling-conventions
     -Werror=unsupported-llvm-version
-    -Werror=unticked-promoted-constructors
     -Werror=unused-do-bind
     -Werror=unused-imports
     -Werror=unused-record-wildcards
--- a/crucible-concurrency/crucible-concurrency.cabal
+++ b/crucible-concurrency/crucible-concurrency.cabal
@@ -86,7 +86,6 @@ common warns
     -Werror=unrecognised-warning-flags
     -Werror=unsupported-calling-conventions
     -Werror=unsupported-llvm-version
-    -Werror=unticked-promoted-constructors
     -Werror=unused-do-bind
     -Werror=unused-imports
     -Werror=unused-record-wildcards
--- a/crucible-debug/crucible-debug.cabal
+++ b/crucible-debug/crucible-debug.cabal
@@ -85,7 +85,6 @@ common shared
     -Werror=unrecognised-warning-flags
     -Werror=unsupported-calling-conventions
     -Werror=unsupported-llvm-version
-    -Werror=unticked-promoted-constructors
     -Werror=unused-do-bind
     -Werror=unused-imports
     -Werror=unused-record-wildcards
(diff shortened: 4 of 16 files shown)


crucible-mir/crucible-mir.cabal
```diff
@@ -14,11 +15,106 @@ maintainer:          spernsteiner@galois.com
...
+  -- A list of warnings and the GHC version in which they were introduced is
+  -- available here:
+  -- https://ghc.gitlab.haskell.org/ghc/doc/users_guide/using-warnings.html
+
+  -- Since GHC 9.6 or earlier:
+  ghc-options:
+    -Wall
+    -Werror=ambiguous-fields
+    -Werror=compat-unqualified-imports
+    -Werror=deferred-type-errors
```
spernsteiner, line 40: Is there a way to make these error only in CI?  `-Werror` can be quite annoying during development, since it forces you to fix trivial issues like unused variables while you're still trying to figure out whether your proposed solution actually works.

langston-barrett: Agreed that warnings can introduce friction to local development. I think there's a trade-off here: it can also be annoying to not realize that things that were only warnings locally will be turned into errors in CI. To locally disable `-Werror`, you can add this to `cabal.project.local`:
```
program-options
    ghc-options: -Wwarn
```

RyanGlScott:
Well this is annoying. The CI is failing because recent versions of GHC incorrectly trigger `-Werror=operator-whitespace` errors on code that looks like [this](https://github.com/GaloisInc/crucible/actions/runs/18602712297/job/53044779624):

```
src/Mir/Intrinsics.hs:2881:73: error: [GHC-40798] [-Woperator-whitespace, Werror=operator-whitespace]
Error:     The suffix use of a ‘:<>:’ might be repurposed as special syntax
      by a future language extension.
    Suggested fix: Add whitespace around ‘:<>:’.
     |
2881 |     ('Text "MethodSpecBuilderType expects no arguments, but was given" ':<>: 'ShowType ctx)
     |                                                                         ^^^^
```

This issue been reported upstream in [GHC#23297](https://gitlab.haskell.org/ghc/ghc/-/issues/23297). I am aware of two workarounds, but neither are very satisfying:

1. Remove the tick in `':<>:`. That is, write `('Text "MethodSpecBuilderType expects no arguments, but was given" :<>: 'ShowType ctx)` instead. That no longer triggers a `-Werror=operator-whitespace` error, but it instead triggers a `-Werror=unticked-promoted-constructor` due to the lack of a tick. Ugh.
2. Instead of writing `':<>:`, write `' :<>:`, with a space in between the tick (`'`) and the promoted constructor (`:<>:`). That avoids both `-Werror=operator-whitespace` and `-Werror=unticked-promoted-constructor`, but it looks quite ugly.

I wonder if we should just remove the use of `-Werror=operator-whitespace` or `-Werror=unticked-promoted-constructor` instead, at least until [GHC#23297](https://gitlab.haskell.org/ghc/ghc/-/issues/23297) is fixed.

RyanGlScott:
After some further investigation, I've discovered that `-Wunticked-promoted-constructor` was removed from `-Wall` in GHC 9.4, and that the _opposite_ style (omitting ticks on promoted data constructors unless it is absolutely necessary) is now encouraged. See [GHC#20531](https://gitlab.haskell.org/ghc/ghc/-/issues/20531). As such, I am inclined to remove the use of `-Werror=unticked-promoted-constructor` and to fix the `-Werror=operator-whitespace` warnings by simply removing the ticks from uses of `':<>:`.

crucible-mir/crucible-mir.cabal
RyanGlScott: I also prefer to err on the side of making the set of warnings in both the CI and local development the same, although I don't feel strongly about this. My primary motivation in this PR was to enable `-Wall`, so if we find the use of `-Werror` in `{crucible,crux}-mir` to be contentious, then I can always revert it (that part is in a separate commit).

spernsteiner: I'm fine with leaving `-Werror` enabled by default—@langston-barrett's `cabal.project.local` suggestion sounds like it will work for me.

samcowger (approved):
Thanks for doing this!

PR #1579, opened 2025-10-20
Property-based tests for `SymSequence`

langston-barrett:
This PR adds differential testing of the `SymSequence` API as compared to a `[]`-backed model. This will be helpful for #1569.

### DIFF
--- a/crucible/test/helpers/SymSequence.hs
+++ b/crucible/test/helpers/SymSequence.hs
@@ -1,0 +1,300 @@
+{-# LANGUAGE EmptyDataDeriving #-}
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+
+module SymSequence (tests) where
+
+import Control.Monad.IO.Class (liftIO)
+import Data.Foldable qualified as F
+import Data.List qualified as List
+import Data.Maybe qualified as Maybe
+import Data.Parameterized.Nonce qualified as Nonce
+import Data.Parameterized.Some (Some(Some))
+import Hedgehog (Gen)
+import Hedgehog qualified as H
+import Hedgehog.Gen qualified as Gen
+import Hedgehog.Range qualified as Range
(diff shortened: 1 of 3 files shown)


crucible/test/helpers/SymSequence.hs
```diff
@@ -0,0 +1,297 @@
...
+  AsSeq sym a = a
+
+-- | An interaction with the 'SymSequence' API
+data Op a t where
+  -- Generic functions
+  OTrue :: Op a Bool
+  OFalse :: Op a Bool
+  OFst :: Op a (l, r) -> Op a l
+  OSnd :: Op a (l, r) -> Op a r
+  OElem:: a -> Op a (Elem a)
```
RyanGlScott: Style nit pick:

```suggestion
  OElem :: a -> Op a (Elem a)
```

crucible/test/helpers/SymSequence.hs
```diff
@@ -0,0 +1,297 @@
...
+    Nothing -> error "non-constant predicate?"
+
+---------------------------------------------------------------------
+-- Op
+
+data Elem a deriving Show
+
+data List a deriving Show
+
+-- TODO: Replace with `Seq` for performance
```
RyanGlScott, line 89: Do you plan to resolve this TODO before landing this?

crucible/test/helpers/SymSequence.hs
```diff
@@ -0,0 +1,297 @@
...
+
+---------------------------------------------------------------------
+-- Helpers
+
+mkBackend :: IO (Some SomeBackend)
+mkBackend = do
+  sym <- newExprBuilder FloatIEEERepr EmptyExprBuilderState Nonce.globalNonceGenerator
+  Some . SomeBackend <$> newSimpleBackend sym
+
+asConstPred ::
```
RyanGlScott, line 71: Worth documenting that this has a precondition that the supplied `Pred` is concrete.

crucible/test/helpers/SymSequence.hs
```diff
@@ -0,0 +1,297 @@
...
+-- Tests
+
+tests :: TT.TestTree
+tests =
+  TTH.testProperty
+    "propSame"
+    -- This is a big API, so we want adequate coverage (default is 100)
+    (H.withTests 4096 propSame)
+
+propSame :: H.Property
```
RyanGlScott, line 48: Perhaps worth a brief Haddock comment here describing the property that this is testing.

crucible/test/helpers/SymSequence.hs
langston-barrett: Eh, it's not a huge deal... The tests are slower than they need to be but not so slow as to be a problem. I could file an issue, but that feels like overkill. Mind just leaving this one as a TODO?

RyanGlScott: Sure, fine by me.

PR #1580, opened 2025-10-20
crux-mir: use MirAggregateRepr for arrays

spernsteiner:
This branch changes the representation of `TyArray` from `MirVectorRepr tpr` to `MirAggregateRepr`.  An array type `[T; N]` is `N * size_of::<T>()` bytes in size, laid out as `N` consecutive elements of size `size_of::<T>()`.  This change also affects slices indirectly, but all of the underlying `MirReference` operations have been updated to support `MirAggregateRepr`, so most `TySlice`-related code remains unchanged.

We still aren't using real layout data from rustc, so for now we assume `size_of::<T>()` is always 1.

This branch does not remove `MirVectorRepr`, as it's still used in the Rust bindings for Crucible's `VectorType` and `SymbolicArrayType` (`crucible::vector::Vector<T>` and `crucible::array::Array<T>`).  Some `MirReference` operations that previously supported only `MirVectorRepr` now support both `MirVectorRepr` and `MirAggregateRepr`, so slices backed by `VectorType`/`SymbolicArrayType` continue to work.  In the future it might be worth giving these primitives separate `MirAggregateRepr`, (Crucible) `VectorRepr`, and `SymbolicArrayRepr` cases and removing `MirVectorRepr` entirely.

The corresponding saw-script PR is GaloisInc/saw-script#2722

### DIFF
--- a/crux-mir/test/conc_eval/alloc/multi_type.rs
+++ b/crux-mir/test/conc_eval/alloc/multi_type.rs
@@ -1,0 +1,29 @@
+// FAIL: reads/writes still expect a u64 target, even after cast
+// Test storing values of different types in a single allocation.
+
+#[cfg_attr(crux, crux::test)]
+pub fn crux_test() -> u32 {
+    unsafe {
+        let mut mem = Vec::<u64>::with_capacity(1);
+        let ptr = mem.as_mut_ptr().cast::<u8>();
+        let p1: *mut u8 = ptr;
+        let p2: *mut u8 = ptr.add(1);
+        let p3: *mut u32 = ptr.add(4).cast::<u32>();
+
+        *p1 = 1;
+        *p2 = 2;
+        *p3 = 3;
+
+        *p1 += 10;
+        *p2 += 10;
+        *p3 += 10;
+
+        *p1 as u32 + *p2 as u32 + *p3
+    }
+}
+
+#[cfg(with_main)]
+pub fn main() {
+    println!("{:?}", crux_test());
+}
+


crux-mir/test/symb_eval/alloc/out_of_bounds.good
```diff
@@ -5,7 +5,7 @@ failures:
 ---- out_of_bounds/<DISAMB>::crux_test[0] counterexamples ----
 [Crux] Found counterexample for verification goal
 [Crux]   test/symb_eval/alloc/out_of_bounds.rs:12:9: 12:24: error: in out_of_bounds/<DISAMB>::crux_test[0]
-[Crux]   vector index out of range: the length is 10 but the index is BV 12
+[Crux]   no value at offset BV 12
```
RyanGlScott: This regression in error message quality is a bit unfortunate. Do you think it's possible to tweak these errors to make it a little more obvious that this happens because of an out-of-bounds index?

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1360,6 +1406,27 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
   MirAggregate_Uninit ::
     !Word ->
     MirStmt f MirAggregateType
+  -- | Create an empty `MirAggregate`, where the size is given as an expression
+  -- of `UsizeType`.  The size must still be a concrete expression at symbolic
+  -- execution time.
+  MirAggregate_UninitSym ::
+    !(f UsizeType) ->
+    MirStmt f MirAggregateType
```
RyanGlScott, line 1393: Is there an advantage to having separate `MirAggregate_Uninit` and `MirAggregate_UninitSym` constructors? What if we instead consolidated them both into `MirAggregate_UninitSym`?

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1997,6 +2094,59 @@ mirVector_resizeIO bak _tpr mirVec newLenSym = do
...
+mirAggregate_uninitSymIO ::
+    IsSymBackend sym bak =>
+    bak ->
+    RegValue sym UsizeType ->
+    IO (RegValue sym MirAggregateType)
+mirAggregate_uninitSymIO bak szSym = do
+  sz <- case asBV szSym of
+    Just x -> return (BV.asUnsigned x)
+    Nothing -> addFailedAssertion bak $ Unsupported callStack $
+      "Attempted to allocate aggregate of symbolic size"
```
RyanGlScott: This size-is-concrete check is repeated three different times throughout `Mir.Intrinsics`, so it would be worth factoring this part out into a helper function.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -172,27 +172,38 @@ transConstVal _ty (Some (IsizeRepr)) (ConstInt i) =
 -- staticSlicePlace does the first four of these actions; addrOfPlace
 -- does the last two.
 --
-transConstVal (M.TyRef _ _) (Some MirSliceRepr) (M.ConstSliceRef defid len) = do
-    place <- staticSlicePlace len defid
+transConstVal (M.TyRef (M.TySlice ty) _) (Some MirSliceRepr) (M.ConstSliceRef defid len) = do
+    Some tpr <- tyToReprM ty
+    place <- staticSlicePlace len (Some tpr) defid
+    addr <- addrOfPlace place
+    return addr
```
RyanGlScott: Slightly simpler:

```suggestion
    addrOfPlace place
```

(And similarly for the `TyStr` case immediately below.)

crucible-mir/src/Mir/Trans.hs
```diff
@@ -359,14 +370,20 @@ staticPlace did = do
...
+    M.DefId ->
+    MirGenerator h s ret (MirPlace s)
+staticSlicePlace len (Some tpr) did = do
     sm <- use $ cs.staticMap
     case Map.lookup did sm of
         Just (StaticVar gv) -> do
             let tpr_found = G.globalType gv
-            Some tpr <- case tpr_found of
-                MirVectorRepr tpr -> return $ Some tpr
+            -- TODO: why is `() <- ...` needed here?
```
RyanGlScott, line 378: GHC's type inference involving GADT pattern matching has some notable limitations, so this doesn't surprise me. I would have wrote this code the same way.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -572,6 +567,8 @@ buildTupleMaybeM tys xs = do
     col <- use $ cs . collection
     ag <- mirAggregate_uninit (fromIntegral $ length tys)
     ag' <- foldM
+        -- TODO: `ty` is unused
```
RyanGlScott: Indeed, this is something that I discovered (and fixed) in #1577.

crux-mir/test/symb_eval/alloc/out_of_bounds.good
spernsteiner: We could maybe expand it to something like "no value at bytes 48..52, in aggregate of size 40".  (For now it would actually say 12..13 and size 10, but once we use proper layout/size information, the offsets here will no longer conveniently match the array indices.)

Currently we don't record whether a given `AgElem_RefPath` came from a tuple field access, array indexing, or some other operation, so we can't easily distinguish those cases in the error message here.  But we could maybe carry around some metadata describing the origin.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: I rebased over #1577, which fixed this.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1974,6 +2071,59 @@ mirVector_resizeIO bak _tpr mirVec newLenSym = do
...
+  let sym = backendGetSym bak
+  len <- case asBV lenSym of
+    Just x -> return (BV.asUnsigned x)
+    Nothing -> addFailedAssertion bak $ Unsupported callStack $
+      "Attempted to allocate aggregate of symbolic size"
+  let totalSize = fromIntegral len * elemSz
+  let entries =
+        [(fromIntegral i * fromIntegral elemSz,
+          MirAggregateEntry elemSz elemTpr (justPartExpr sym elemVal))
+        | i <- [0 .. len]]
```
samcowger:
```suggestion
        | i <- init [0 .. len]]
```
Otherwise, I think this results in `len + 1` elements, no?

crux-mir/test/symb_eval/alloc/out_of_bounds.good
spernsteiner: In 24e152057d6d12c163068a0612f684e8043a5209, I changed the message to "no value at offset 12, in aggregate of size 10".

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: I combined the two in d68ce07185175c5f0fbd019092321ecc6da22d51.  The combined operation is named `MirAggregate_Uninit`, for consistency with `MirVector_Uninit`.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Added a helper in c436814469c317688856c3a980c8c1d4d43a6775, and used it for the similar `MirVector` cases as well.

crucible-mir/src/Mir/Trans.hs
spernsteiner: Fixed in 32da691b3c8cee2a81cfecf162e274077690ef38

crucible-mir/src/Mir/Trans.hs
```diff
@@ -3127,6 +3151,25 @@ vectorCopy tpr ptr0 len = do
...
+  Word ->
+  MirGenerator h s ret (G.Expr MIR s MirAggregateType)
+aggregateCopy_constLen tpr ptr0 len size = do
+  ag <- mirAggregate_uninit_constSize (fromIntegral len * size)
+  ag' <- foldM
+    (\ag' i -> do
+       ptr <- mirRef_offset ptr0 (S.app $ usizeLit $ fromIntegral i)
+       elt <- readMirRef tpr ptr
+       mirAggregate_set (fromIntegral i * size) size tpr elt ag')
+    ag [0 .. len - 1]
```
```suggestion
    ag (init [0 .. len])
```
To attend to the case where `len == 0`

crucible-mir/src/Mir/Trans.hs
```diff
@@ -685,10 +700,13 @@ transUnaryOp uop op = do
...
 buildRepeat op size = do
-    (MirExp tp e) <- evalOperand op
+    MirExp tpr e <- evalOperand op
     let n = fromInteger size
-    expr <- mirVector_fromVector tp $ S.app $ E.VectorReplicate tp (S.app $ E.NatLit n) e
-    return $ MirExp (MirVectorRepr tp) expr
+    ag <- mirAggregate_uninit_constSize n
+    ag' <- foldM
+        (\ag' i -> mirAggregate_set i 1 tpr e ag')
+        ag [0 .. n - 1]
```
```suggestion
        ag (init [0 .. n])
```
To attend to the case where `n == 0`

crucible-mir/src/Mir/Trans.hs
```diff
@@ -166,27 +166,36 @@ transConstVal _ty (Some (IsizeRepr)) (ConstInt i) =
...
+
+transConstVal _ty (Some MirAggregateRepr) (M.ConstStrBody bs) = do
     let bytes = map (\b -> R.App (eBVLit (knownNat @8) (toInteger b))) (BS.unpack bs)
-    let vec = R.App $ E.VectorLit u8Repr (V.fromList bytes)
-    mirVec <- mirVector_fromVector u8Repr vec
-    return $ MirExp (MirVectorRepr u8Repr) mirVec
+    ag <- mirAggregate_uninit_constSize (fromIntegral $ length bytes)
+    ag' <- foldM
+        (\ag' (i, b) -> mirAggregate_set i 1 knownRepr b ag')
+        ag (zip [0..] bytes)
```
samcowger, line 184: Should this mention `TODO: hardcoded size=1` somewhere? (Same question applies to the similar use of `mirAggregate_set` a few lines down.)

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -927,6 +935,48 @@ writeMirAggregateWithSymOffset bak iteFn off sz tpr val ag
...
+      -- no overlap (and no value at `off`).
+      | off' + sz' <= off -> Right Unassigned
+      | otherwise -> die $ "partial overlap: " ++ show off
+          ++ " vs " ++ show off' ++ ".." ++ show (off' + sz')
+  where
+    die msg = Left $ "bad MirAggregate lookup: " ++ msg
+
+-- | Resize a `MirAggregate`.  If the new size is larger, the extra space will
+-- be left uninitialized.  If the new size is smaller, any entries that extend
+-- past the new end point will be discarded.
```
samcowger, line 970: I see that this supports the case in which an aggregate's new endpoint appears in the middle of an existing value (thereby discarding that value). Is this a reasonable thing to be able to do to an aggregate?

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -927,6 +935,48 @@ writeMirAggregateWithSymOffset bak iteFn off sz tpr val ag
...
 
+mirAggregate_lookup ::
+  Word ->
+  TypeRepr tp ->
+  MirAggregate sym ->
+  Either String (RegValue sym (MaybeType tp))
+mirAggregate_lookup off tpr (MirAggregate totalSize m) = do
+  -- We return @Right Unassigned@ if the offset is valid but there's no value
+  -- there, and @Left errorMessage@ if offset is invalid (in the middle of some
+  -- entry) or the type @tpr@ is incorrect.
```
samcowger: This might be worth promoting to a docstring for this function.

crucible-mir/src/Mir/Trans.hs
samcowger: Also, should this mention `TODO: hardcoded size=1`?

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1270,10 +1274,10 @@ array_from_ref = (["core", "array", "from_ref", "crucible_array_from_ref_hook"],
                 -- output are aliases.
                 Some elemRepr <- tyToReprM elemTy
                 elemVal <- readMirRef elemRepr elemRef
-                let vecVal = R.App (E.VectorLit elemRepr (V.fromList [elemVal]))
-                mirVecVal <- mirVector_fromVector elemRepr vecVal
-                mirVecRef <- constMirRef (MirVectorRepr elemRepr) mirVecVal
-                pure (MirExp MirReferenceRepr mirVecRef)
+                ag <- mirAggregate_uninit_constSize 1
+                ag' <- mirAggregate_set 0 1 elemRepr elemVal ag
```
samcowger, line 1279: Should this mention `TODO: hardcoded size=1`?

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1226,16 +1229,18 @@ array_from_slice = (["core","slice", "{impl}", "as_array", "crucible_array_from_
     \_substs -> Just $ CustomOpNamed $ \fnName ops -> do
         fn <- findFn fnName
         case (fn ^. fsig . fsreturn_ty, ops) of
-            ( TyAdt optionMonoName _ (Substs [TyRef (TyArray ty _) Immut]),
-              [MirExp MirSliceRepr e, MirExp UsizeRepr eLen] ) -> do
+            ( TyAdt optionMonoName _ (Substs [TyRef (TyArray ty tyLen) Immut]),
+              [MirExp MirSliceRepr e, _] ) -> do
```
samcowger, line 1233: Unless I'm mistaken, this change makes the second parameter to this function entirely useless, since it's unmentioned in [the Rust implementation](https://github.com/GaloisInc/mir-json/blob/88cf801f8204983d78a84311bb11d84517ffb867/libs/core/src/slice/mod.rs#L865-L878), so we should remember to remove it. I'm happy to make an issue about/PR effecting that, if you agree.

samcowger (commented):
Have these changes introduced/changed any behavior (other than changes to output-printing) that can be confirmed with a regression test? It feels like the answer might be no, but I wanted to double check.

qsctr:
> We still aren't using real layout data from rustc, so for now we assume `size_of::<T>()` is always 1.

What further work is needed to support this? As of #1487 we have the layout data from `mir-json` in the `Collection`.

spernsteiner:
> As of #1487 we have the layout data from `mir-json` in the `Collection`.

That PR adds only the size and alignment, but for structs and tuples we also need the offset of each field.  For consistency, I've left arrays with size=1 as well, even though just the element size is sufficient in that case.  This also lets us put off some of the trickier bits of `MirAggregateType`/`MirVectorType` pointer arithmetic interop to a later PR.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: I don't expect it to be common, but it is legal in concrete Rust to e.g. allocate 4 bytes, write a `u32`, reallocate down to 2 bytes, and read a `u16`.  We don't currently support type conversion on pointer reads, so even without the reallocate step, this still wouldn't work in crux-mir.  I figured it's okay to convert that into a different kind of error ("no value at offset") in this unusual case.  This lets us keep the current invariant (each entry covers a sub-range of `0 .. totalSize`) and ensures we still error on cases that are illegal in concrete Rust (e.g. write `u32`, shrink to 2 bytes, read `u32`) without requiring extra logic on pointer reads.

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Sure, go ahead.  It looks like the argument has actually been redundant for quite a while, since the same value is also available as the `TyArray` length.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Made it a doc comment in 4c8159426c3cddd45f20030214adba25df9de84b

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Yikes, good catch.  This was building a `MirAggregate` with an extra entry extending beyond its `totalSize`.  Fixed in ab73fe40860202f13ce18ee5531526e43a3c06e3.

crucible-mir/src/Mir/Trans.hs
spernsteiner: Fixed in ab73fe40860202f13ce18ee5531526e43a3c06e3

crucible-mir/src/Mir/Trans.hs
spernsteiner: Fixed in ab73fe40860202f13ce18ee5531526e43a3c06e3.  Added the comment in a6603a474d26bce61ecacf7f70e065574c1f1a07.

crucible-mir/src/Mir/Trans.hs
spernsteiner: Added the comment in a6603a474d26bce61ecacf7f70e065574c1f1a07.  Note there are still a bunch of unmarked places that assume size=1, since I didn't think to start marking them this way while I was implementing the previous tuple changes.

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Added the comment in a6603a474d26bce61ecacf7f70e065574c1f1a07

spernsteiner:
> Have these changes introduced/changed any behavior (other than changes to output-printing) that can be confirmed with a regression test?

I thought this change might be enough to allow multiple types in a single allocation, but that still fails for other reasons.  I added the code I tried as an expected-fail test case for now.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Alright, I'll do that once this lands.

PR #1581, opened 2025-10-20
Add pointer arithmetic intrinsics (offset, arith_offset, ptr_offset_from, ptr_offset_from_unsigned) and tests (Closes #1180)

### LINKED ISSUE #1180: `crucible-mir`: Re-implement overrides for `get_unchecked` slice indexing
Given this Rust code:

```rs
fn bar(bytes: &[u8]) -> u32 {
    return 0u32;
}

pub fn foo(arr: &[u8; 4]) -> u32 {
    bar(&arr[0..4])
}
```

The following spec will crash SAW 1.1:

```
enable_experimental;

let ref_to_fresh n ty = do {
    x <- mir_fresh_var n ty;
    r <- mir_alloc ty;
    mir_points_to r (mir_term x);
    return (x, r);
};

let foo_spec = do {
    (x, r_x) <- ref_to_fresh "x" (mir_array 4 mir_u8);
    mir_execute_func [r_x];
};

m <- mir_load_module "test.linked-mir.json";
mir_verify m "test::foo" [] false foo_spec z3;
```
```
$ ~/Software/saw-1.1/bin/saw test.saw
[11:57:58.326] Loading file "/home/ryanscott/Documents/Hacking/SAW/test.saw"
[11:57:58.332] Verifying test/d9b5c637::foo[0] ...
[11:57:58.340] Simulating test/d9b5c637::foo[0] ...
[11:57:58.341] Stack trace:
"mir_verify" (/home/ryanscott/Documents/Hacking/SAW/test.saw:16:1-16:11)
Symbolic execution failed.
  /home/ryanscott/.rustup/toolchains/nightly-2023-01-23-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/ptr/const_ptr.rs:473:43: 473:48 !/home/ryanscott/.rustup/toolchains/nightly-2023-01-23-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/core/src/slice/index.rs:386:54: 386:69: error: in core/73237d41::slice[0]::index[0]::{impl#4}[0]::get_unchecked[0]::_instaddce72e1232152c[0]
  Translation error in core/73237d41::slice[0]::index[0]::{impl#4}[0]::get_unchecked[0]::_instaddce72e1232152c[0]: callExp: Don't know how to call core/73237d41::intrinsics[0]::{extern#0}[0]::offset[0]::_instaddce72e1232152c[0]
```

I used SAW to reproduce this crash (it is surprisingly difficult to reproduce using `crux-mir`), but the issue really lies in `crucible-mir`. The problem is that the [`get_unchecked`](https://doc.rust-lang.org/std/slice/trait.SliceIndex.html#tymethod.get_unchecked) slice-indexing function (as well as its cousin, [`get_unchecked_mut`](https://doc.rust-lang.org/std/slice/trait.SliceIndex.html#tymethod.get_unchecked_mut)) are too low-level for `crucible-mir` to handle at the moment.

In the previous Rust nightly that `crucible-mir` supported (`2020-03-22`), we put custom overrides in place to handle these functions—see https://github.com/GaloisInc/crucible/commit/349feee3c6229b8e4fa477ddfb555c8220a8e42d for how these were handled. Unfortunately, these overrides weren't ported over when we upgraded to the `2023-01-23` nightly. I believe that porting these overrides to the more recent nightly would fix this issue. Note that nowadays, the `get_unchecked{,_mut}` functions now live in `core::slice::index::{impl}` rather than `core::slice::{impl}`.

sand7000:
# Intrinsics added

* [core::intrinsics::offset](https://doc.rust-lang.org/std/intrinsics/fn.offset.html)
* [core::intrinsics::arith_offset](https://doc.rust-lang.org/std/intrinsics/fn.arith_offset.html)
* [core::intrinsics::ptr_offset_from](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from.html)
* [core::intrinsics::ptr_offset_from_unsigned](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from_unsigned.html)

# Test coverage

* Each intrinsic has tests for normal (positive), negative, zero, and out-of-bounds pointer arithmetic.
* Mutable pointer variants are tested where relevant.

Closes #1180

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -12,6 +12,23 @@ This release supports [version
 * Add an intrinsic for [`size_of_val`](https://doc.rust-lang.org/std/intrinsics/fn.size_of_val.html),
   which computes the size of a value in bytes. This works for all sized types
   as well as a limited number of unsized types (currently, only slices).
+* Add an intrinsic for [`offset`](https://doc.rust-lang.org/std/intrinsics/fn.offset.html),
+  which performs pointer offset arithmetic with bounds checking, mirroring the
+  semantics of Rust's `core::intrinsics::offset`.
+* Add an intrinsic for [`arith_offset`](https://doc.rust-lang.org/std/intrinsics/fn.arith_offset.html),
+  which performs wrapping pointer offset arithmetic (allowing arithmetic beyond
+  the end of objects without UB), matching the behavior of Rust's
+  `core::intrinsics::arith_offset`.
+* Add an intrinsic for [`ptr_offset_from`](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from.html),
+  which computes the offset (in elements) between two pointers, mirroring
+  Rust's `core::intrinsics::ptr_offset_from`.
+* Add an intrinsic for [`ptr_offset_from_unsigned`](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from_unsigned.html),
+  which computes the offset (in elements, as an unsigned integer) between two
+  pointers.  This intrinsic matches the semantics of Rust's
+  core::intrinsics::ptr_offset_from_unsigned, and requires that the first
+  pointer is greater than or equal to the second.  If the offset would be
+  negative (i.e., the first pointer is less than the second), this is undefined
+  behavior.
 
 # 0.5 -- 2025-11-09
 
@@ -56,20 +73,6 @@ This release supports [version
   division has a nonzero remainder. This is implemented for both signed and
   unsigned integers and mirrors the semantics of Rust's
   `core::intrinsics::exact_div`.
-* Add an intrinsic for [`offset`](https://doc.rust-lang.org/std/intrinsics/fn.offset.html),
-  which performs pointer offset arithmetic with bounds checking, mirroring the
-  semantics of Rust's `core::intrinsics::offset`.
-* Add an intrinsic for [`arith_offset`](https://doc.rust-lang.org/std/intrinsics/fn.arith_offset.html),
-  which performs wrapping pointer offset arithmetic (allowing arithmetic beyond
-  the end of objects without UB), matching the behavior of Rust's
-  `core::intrinsics::arith_offset`.
-* Add an intrinsic for [`ptr_offset_from`](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from.html),
-  which computes the offset (in elements) between two pointers, mirroring
-  Rust's `core::intrinsics::ptr_offset_from`.
-* Add an intrinsic for [`ptr_offset_from_unsigned`](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from_unsigned.html),
-  which computes the offset (in elements, as an unsigned integer) between two
-  pointers, returning zero on negative offsets, matching the semantics of
-  Rust's `core::intrinsics::ptr_offset_from_unsigned`.
 
 # 0.4 -- 2025-03-21
 


crucible-mir/CHANGELOG.md
```diff
@@ -45,6 +45,20 @@ This release supports [version
...
+* Add an intrinsic for [`arith_offset`](https://doc.rust-lang.org/std/intrinsics/fn.arith_offset.html),
+  which performs wrapping pointer offset arithmetic (allowing arithmetic beyond
+  the end of objects without UB), matching the behavior of Rust's
+  `core::intrinsics::arith_offset`.
+* Add an intrinsic for [`ptr_offset_from`](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from.html),
+  which computes the offset (in elements) between two pointers, mirroring
+  Rust's `core::intrinsics::ptr_offset_from`.
+* Add an intrinsic for [`ptr_offset_from_unsigned`](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from_unsigned.html),
+  which computes the offset (in elements, as an unsigned integer) between two
+  pointers, returning zero on negative offsets, matching the semantics of
```
RyanGlScott: Where does it say that this returns zero on negative offsets? I can't find anything about this in the documentation for [`ptr_offset_from_unsigned`](https://doc.rust-lang.org/std/intrinsics/fn.ptr_offset_from_unsigned.html) or the [`offset_from_unsigned`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset_from_unsigned-1) built on top of it. Moreover, the actual implementation in `Mir.TransCustom` suggests that it _errors_ on negative offsets, which I think is closer to what it is intended to do.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -518,6 +522,30 @@ ptr_offset_from = (["core", "ptr", "const_ptr", "{impl}", "offset_from"], ptr_of
...
+      let unsignedOffset = signedOffset
+      return $ MirExp UsizeRepr unsignedOffset
+    _ -> mirFail $ "bad arguments for ptr_offset_from_unsigned: " ++ show ops
+  _ -> Nothing
+
+ptr_offset_from_unsigned :: (ExplodedDefId, CustomRHS)
+ptr_offset_from_unsigned =
+  ( ["core", "intrinsics", "ptr_offset_from_unsigned"]
+  , ptr_offset_from_unsigned_impl
+  )
```
RyanGlScott, line 550: Since it is very nearly the same function, I think it would be worth adding an override for [`core::ptr::const_ptr::offset_from_unsigned`](https://doc.rust-lang.org/std/primitive.pointer.html#method.offset_from_unsigned-1) as well.

sand7000: I checked the Rust version used in the mir-json Docker image: 

https://github.com/GaloisInc/mir-json/blob/master/Dockerfile#L16 

which is currently nightly-2025-02-16. In this version, `core::ptr::const_ptr::offset_from_unsigned` does not appear to be available in the source tree under `src/ptr/const_ptr.rs`. 

Since our tests and rlibs are utilize this toolchain, I can add but not test an override for `core::ptr::const_ptr::offset_from_unsigned`. Do we support a newer version of Rust in mir-json than is being used in the Dockerfile or should I leave out `core::ptr::const_ptr::offset_from_unsigned` for now?

crucible-mir/CHANGELOG.md
sand7000: > Where does it say that this returns zero on negative offsets?

It doesn't. I got confused on my first reading, subsequently fixed my implementation it but failed to fix the description in the CHANGELOG. I will do so,

crucible-mir/src/Mir/TransCustom.hs
RyanGlScott: Oops, good catch! (I really should have checked this before suggesting it.)

Since it's infeasible to test, I'm fine with leaving this out.

PR #1584, opened 2025-10-21
llvm-debug: Export `LLVMCommand` so it can be mentioned in types

langston-barrett:


RyanGlScott (approved):
Oh yes, good catch!

This is slightly orthogonal to this PR, but would you also be willing to export `MIRCommand` from [`Lang.Crucible.MIR.Debug`](https://github.com/GaloisInc/crucible/blob/10056d63b2de6cda48db9269afa4669723a85150/crucible-mir-debug/src/Lang/Crucible/MIR/Debug.hs) as well (for the same reasons)?

langston-barrett:
Done!

PR #1588, opened 2025-10-22
llvm-cli: Add test for `isnan`

langston-barrett:
Not sure why we're getting `0` for the NaNs...?

### DIFF
--- a/crucible-llvm-cli/test-data/isnan.cbl
+++ b/crucible-llvm-cli/test-data/isnan.cbl
@@ -1,0 +1,25 @@
+; Floats produced with:
+;
+;     import GHC.Float
+;     import Numeric
+;     showString "0x" $ showHex (castDoubleToWord64 42.0) ""
+;     showString "0x" $ showHex (castDoubleToWord64 (0.0 / 0.0)) ""
+
+(defun @main () Unit
+  (start start:
+    (let g (resolve-global "isnan"))
+    (let h (load-handle (Ptr 32) ((FP Double)) g))
+
+    (let i42 (bv 64 0x4045000000000000))
+    (let f42 (binary-to-fp Double i42))
+    (let n (funcall h f42))
+    (let n-off (ptr-offset 32 n))
+    (assert! (equal? n-off (bv 32 0)) "42.0 is not NaN")
+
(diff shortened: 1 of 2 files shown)


crucible-llvm-cli/test-data/isnan.cbl
```diff
@@ -0,0 +1,30 @@
...
+;         return hex(int("{:032b}".format(bits), base=2))
+;     print(float_to_bin(42.0))
+;     print(float_to_bin(nan))
+
+(defun @main () Unit
+  (start start:
+    (let g (resolve-global "isnan"))
+    (let h (load-handle (Ptr 32) ((FP Double)) g))
+
+    (let i42 (bv 64 0x42280000))
```
RyanGlScott: I don't think this is right. `0x42280000` is the bit-level representation of `42.0` when represented as a single-precision float (i.e., 32 bits), not a double-precision float (i.e., 64 bits). You can see this in GHCi:

```
λ> :m + GHC.Float Numeric
λ> showString "0x" $ showHex (castFloatToWord32 42.0) ""
"0x42280000"
λ> showString "0x" $ showHex (castDoubleToWord64 42.0) ""
"0x4045000000000000"
```

Similar comments apply to the NaN values below.

langston-barrett: Ah, thanks! Fixed.

crucible-llvm-cli/test-data/isnan.cbl
```diff
@@ -0,0 +1,25 @@
...
+    (let f42 (binary-to-fp Double i42))
+    (let n (funcall h f42))
+    (let n-off (ptr-offset 32 n))
+    (assert! (equal? n-off (bv 32 0)) "42.0 is not NaN")
```
RyanGlScott, line 17: This error message claims that the `0x42280000` value above is the bit-level representation of `42.0` as a double-precision float, but as noted in https://github.com/GaloisInc/crucible/pull/1588/files#r2452645167, that's not the case. Consider changing the value above to `0x4045000000000000`?

langston-barrett: Ah shoot, sorry! Fixed.

RyanGlScott (approved):

PR #1590, opened 2025-10-23
`crucible-llvm`: Support vectorized `llvm.{umin,umax,smin,max}` intrinsics

### LINKED ISSUE #1589: `crucible-llvm`: Support vectorized `llvm.u{min,max}` and `llvm.s{min,max}` intrinsics
`crux-llvm`'s [`T1177.c`](https://github.com/GaloisInc/crucible/blob/b9bd2bac457568875423c017dd6113a6162e6f06/crux-llvm/test-data/golden/T1177.c) test case passes with Clang 14, but not with Clang 16 or later. Some investigation reveals that this is because Clang 16+ is more likely to optimize code to uses of vectorized [`llvm.umin`](https://llvm.org/docs/LangRef.html#llvm-umin-intrinsic) intrinsics (as well as its cousins, vectorized [`llvm.umax`](https://llvm.org/docs/LangRef.html#llvm-vector-reduce-umax-intrinsic), [`llvm.smin`](https://llvm.org/docs/LangRef.html#llvm-vector-reduce-smin-intrinsic), and [`llvm.smax`](https://llvm.org/docs/LangRef.html#llvm-vector-reduce-smax-intrinsic)). Here is a minimized version of `T1177.c` that demonstrates the issue with vectorized `llvm.umin` in particular:

```c
// test.c
#include <crucible.h>
#include <stddef.h>
#include <stdint.h>

uint32_t __attribute__((noinline)) reduce_umin(const uint8_t* in, size_t len) {
  uint32_t x = UINT32_MAX;

  for (size_t i = 0; i < len; i++) {
    uint32_t y = in[i];
    x = (x < y) ? x : y;
  }

  return x;
}

int main(void) {
  const uint8_t unsigned_in[8] = {0, 1, 2, 3, 4, 5, 6, 7};
  check(reduce_umin(unsigned_in, 8) == 0);
  return 0;
}
```

`crux-llvm` can simulate this with Clang 14:

```
$ CLANG=clang-14 cabal -v0 run exe:crux-llvm -- -O2 test.c
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 1
[Crux]   Disproved: 0
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Valid.
```

With Clang 16 or later, however, it fails:

```
$ CLANG=clang-18 cabal -v0 run exe:crux-llvm -- -O2 test.c
[Crux] *** debug executable: results/test/debug-11
[Crux] *** break on line: 11
[Crux]   test.c:11:9: error: in reduce_umin
[Crux]   Details:
[Crux]     Attempting to load callable function with type: <4 x i32>(<4 x i32>, <4 x i32>)
[Crux]       Via pointer: Global symbol "llvm.umin.v4i32" (7, 0x0:[64])
[Crux]     In memory state:
[Crux]       Stack frame reduce_umin
[Crux]       Stack frame main
[Crux]         Allocations:
[Crux]           StackAlloc 12 0x8:[64] Mutable 8-byte-aligned internal
[Crux]         Writes:
[Crux]             12 |->   *(12, 0x0:[64]) := 0x706050403020100:[64]
[Crux]       Base memory
[Crux]         Allocations:
[Crux]           GlobalAlloc 11 0x7:[64] Immutable 1-byte-aligned [global variable  ] .str
[Crux]           GlobalAlloc 10 0x0:[64] Immutable 1-byte-aligned [defined function ] main
[Crux]           GlobalAlloc 9 0x0:[64] Immutable 1-byte-aligned [defined function ] reduce_umin
[Crux]           GlobalAlloc 8 0x0:[64] Immutable 1-byte-aligned [external function] llvm.vector.reduce.umin.v4i32
[Crux]           GlobalAlloc 7 0x0:[64] Immutable 1-byte-aligned [external function] llvm.umin.v4i32
[Crux]           GlobalAlloc 6 0x0:[64] Immutable 1-byte-aligned [external function] llvm.umin.i32
[Crux]           GlobalAlloc 5 0x0:[64] Immutable 1-byte-aligned [external function] llvm.dbg.value
[Crux]           GlobalAlloc 4 0x0:[64] Immutable 1-byte-aligned [external function] llvm.dbg.assign
[Crux]           GlobalAlloc 3 0x0:[64] Immutable 1-byte-aligned [external function] crucible_assert
[Crux]           GlobalAlloc 2 0x0:[64] Immutable 1-byte-aligned [external function] llvm.lifetime.end.p0
[Crux]           GlobalAlloc 1 0x0:[64] Immutable 1-byte-aligned [external function] llvm.lifetime.start.p0
[Crux]         Writes:
[Crux]             11 |->   *(11, 0x0:[64]) := "test.c\NUL"
[Crux]     in context:
[Crux]       reduce_umin
[Crux]       main
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

The problematic intrinsic has this type signature:

```llvm
declare <4 x i32> @llvm.umin.v4i32(<4 x i32>, <4 x i32>)
```

We will need to add a `crucible-llvm` override for this. ([There are already overrides for the unvectorized versions.](https://github.com/GaloisInc/crucible/blob/b9bd2bac457568875423c017dd6113a6162e6f06/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs#L241-L252))

There are of course many more vectorized intrinsics that `crucible-llvm` could conceivably support, but for now, I propose that we limit ourselves to what we need to make the `crux-llvm` test suite pass.

RyanGlScott:
Doing so allows `crux-llvm`'s `T1177.c` test case to pass with LLVM 16 or later, as Clang 16+ is more likely to generate uses of these intrinsics in optimized code.

Fixes #1589.

### DIFF
--- a/crucible-llvm/CHANGELOG.md
+++ b/crucible-llvm/CHANGELOG.md
@@ -13,6 +13,8 @@
 * Fix a bug that would cause the simuator to compute incorrect results for the
   `llvm.is.fpclass` intrinsic. (Among other things, this is used to power the
   `isnan` function in Clang 17 or later.)
+* Support vectorized versions of the `llvm.u{min,max}` and `llvm.s{min,max}`
+  intrinsics.
 
 # 0.7.1 -- 2025-03-21
 
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs
@@ -307,6 +307,23 @@ poly1_vec_llvm_overrides =
     , Poly1VecLLVMOverride $ \vecSz intSz ->
         SomeLLVMOverride (llvmVectorReduceUmin vecSz intSz)
     )
+
+  , ("llvm.smax"
+    , Poly1VecLLVMOverride $ \vecSz intSz ->
+        SomeLLVMOverride (llvmSmaxVector vecSz intSz)
+    )
+  , ("llvm.smin"
+    , Poly1VecLLVMOverride $ \vecSz intSz ->
+        SomeLLVMOverride (llvmSminVector vecSz intSz)
+    )
+  , ("llvm.umax"
+    , Poly1VecLLVMOverride $ \vecSz intSz ->
+        SomeLLVMOverride (llvmUmaxVector vecSz intSz)
+    )
+  , ("llvm.umin"
+    , Poly1VecLLVMOverride $ \vecSz intSz ->
+        SomeLLVMOverride (llvmUminVector vecSz intSz)
+    )
   ]
 
 ------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


RyanGlScott:
I like the idea of this in principle, but I've encountered an obstacle in trying to add such a test case. Even something small like this currently fails to simulate with `crucible-llvm-cli`:

```racket
(defun @main () Unit
  (start start:
    (let umin-g (resolve-global "llvm.umin.v4i32"))
    (let umin-h (load-handle (Vector (Ptr 32)) ((Vector (Ptr 32)) (Vector (Ptr 32))) umin-g))
    (return ())))
```
```
$ cabal -v0 run exe:crucible-llvm -- simulate test-data/vectorized-min-max.cbl
==== Begin Simulation ====

==== Finish Simulation ====
==== Proof obligations ====

Prove:
  test-data/vectorized-min-max.cbl:3:17: error: in main
  Global symbol not allocated
  Details:
    Global symbol "llvm.umin.v4i32" has no associated allocation
  false
COUNTEREXAMPLE
```

This is because `crucible-llvm-syntax` only registers overrides for "basic" LLVM intrinsics, i.e., intrinsics for which we can enumerate the names of the intrinsics itself. It does not register overrides for infinite families of LLVM intrinsics like `llvm.umin.v*i*`, however, because it would need to know ahead of time which particular instantiation to use in order to register it. Should we register `llvm.umin.v4i32`? `llvm.umin.v4i64`? `llvm.umin.v8i32`? `llvm.umin.v8i64`? It's not clear, and picking one at the expense of others would feel somewhat arbitrary.

langston-barrett:
Hmm, that makes sense. I know that for actual LLVM modules, `crucible-llvm` registers overrides based on the `declare`s therein. Perhaps we could do something similar for `crucible-llvm-cli`? But we certainly shouldn't hold up this PR on that.

RyanGlScott:
> I know that for actual LLVM modules, `crucible-llvm` registers overrides based on the `declare`s therein.

Right, it relies on inspecting the bitcode to figure out which particular instantiations of polymorphic overrides to use.

> Perhaps we could do something similar for `crucible-llvm-cli`?

`crucible-llvm-cli` doesn't have access to bitcode, so we'd need a different approach. Perhaps it would suffice to add a command-line argument that supplied the names of additional LLVM intrinsics to register?

> But we certainly shouldn't hold up this PR on that.

Agreed, this is best left for a separate issue.

langston-barrett:
I filed an issue along those lines: https://github.com/GaloisInc/crucible/issues/1591

PR #1592, opened 2025-10-23
llvm-syntax: Support "polymorphic" LLVM overrides

### LINKED ISSUE #1591: llvm-syntax: Register polymorphic overrides
`crucible-llvm-syntax` currently registers the "basic" libc and LLVM overrides:

https://github.com/GaloisInc/crucible/blob/b9bd2bac457568875423c017dd6113a6162e6f06/crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax/Overrides.hs#L38

These are contrasted with the ["polymorphic"](https://github.com/GaloisInc/crucible/blob/b9bd2bac457568875423c017dd6113a6162e6f06/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs#L191) overrides, which include vectorized intrinsics such as `llvm.umin.v4i32`. These are really infinite *families* of overrides, and a decision must be made ahead of time on which ones to register.

Tools like Crux-LLVM register these polymorphic overrides based on the LLVM `declare`s found in the LLVM module. We could do something similar in `crucible-llvm-syntax`, as `llvm-syntax` also supports forward declarations (which are also spelled `declare`).

langston-barrett:
Fixes #1591. The implementation is slightly hacky, in that we reverse-engineer an LLVM type from a Crucible type and we leave out many cases. However, it's an upgrade on the status quo, and allows us to test "polymorphic" LLVM intrinsics without an actual bitcode file. Plus, `crucible-llvm-{cli,syntax}` is a developer-facing tool, so I'm not terribly worried about polish.

### DIFF
--- a/crucible-llvm-cli/src/Lang/Crucible/LLVM/CLI.hs
+++ b/crucible-llvm-cli/src/Lang/Crucible/LLVM/CLI.hs
@@ -106,6 +106,7 @@ withLlvmHooks k = do
               let ?memOpts = Mem.defaultMemOptions
               let ?intrinsicsOpts = defaultIntrinsicsOptions
 
+              -- TODO(#1594): Deduplicate these override names using a `Map`
               Monad.forM_ (Map.toList fwdDecs) $ \(nm, C.SomeHandle hdl) -> do
                 case nm of
                   "read-bytes" -> tryBindTypedOverride hdl (StrOv.readBytesOverride mvar)


crucible-llvm-syntax/src/Lang/Crucible/LLVM/Syntax/Overrides.hs
```diff
@@ -48,5 +73,104 @@ registerLLVMOverrides bak llvmCtx = do
...
+          TFC.toListFC Some.Some fieldAssn
+      Just $ L.Struct fieldTys
+    C.SymbolicArrayRepr{} -> Nothing
+    C.SymbolicStructRepr{} -> Nothing
+    C.UnitRepr -> Just (L.PrimType L.Void)
+    C.VariantRepr{} -> Nothing
+    -- HACK: It's not clear how to support other vector sizes, because Crucible
+    -- vectors don't have the length as part of the type. We arbitrarily choose
+    -- 4 here so that we can write tests for vectorized intrinsics of width 4
+    -- in crucible-llvm-cli.
```
RyanGlScott: Cite https://github.com/GaloisInc/crucible/issues/1593 here.

crucible-llvm-cli/test-data/ub/025.cbl
```diff
@@ -1,6 +1,8 @@
 ; A pointer is used to call a function whose type is not compatible with the
 ; referenced type (6.3.2.3).
 
+(declare @malloc ((sz (Bitvector 64))) (Ptr 64))
```
RyanGlScott, line 4: This goes through the effort of `declare`'ing `malloc`, but then never actually uses it! That's a bit of a shame. Any reason not to use the forward declaration below in `main`?

crucible-llvm-cli/test-data/ptr.cbl
```diff
@@ -1,3 +1,5 @@
+(declare @malloc ((sz (Bitvector 64))) (Ptr 64))
```
RyanGlScott, line 1: Similar question here.

crucible-llvm-cli/src/Lang/Crucible/LLVM/CLI.hs
```diff
@@ -104,14 +105,17 @@ withLlvmHooks k = do
...
-              Monad.forM_ (Map.toList fds) $ \(nm, C.SomeHandle hdl) -> do
+
+              Monad.forM_ (Map.toList fwdDecs) $ \(nm, C.SomeHandle hdl) -> do
                 case nm of
                   "read-bytes" -> tryBindTypedOverride hdl (StrOv.readBytesOverride mvar)
                   "read-c-string" -> tryBindTypedOverride hdl (StrOv.readCStringOverride mvar)
                   "write-bytes" -> tryBindTypedOverride hdl (StrOv.writeBytesOverride mvar)
                   "write-c-string" -> tryBindTypedOverride hdl (StrOv.writeCStringOverride mvar)
                   _ -> pure ()
+              let otherOvs = ["read-bytes", "read-c-string", "write-bytes", "write-c-string"]
```
RyanGlScott: The duplication of strings between this `case` expression and `otherOvs` below it is somewhat unfortunate. Perhaps we can define a `Map` of strings to overrides so that we only need to write out the strings once and for all?

crucible-llvm-cli/test-data/ub/025.cbl
langston-barrett: The program *does* use the `malloc` override, just indirectly. It loads the function pointer from memroy, and then calls it at an incorrect type. The test is itself a demonstration of this kind of UB. With these changes, it is now necessary to explicitly `declare` functions you want in the LLVM memory.

crucible-llvm-cli/test-data/ptr.cbl
langston-barrett: Same answer, this test demonstrates that `malloc` has a function allocation in the LLVM memory, and needs to `declare` it for that to be the case.

crucible-llvm-cli/src/Lang/Crucible/LLVM/CLI.hs
langston-barrett: Fixed, though I'm seeing a GHC panic locally...
```
<no location info>: error:
    panic! (the 'impossible' happened)
  GHC version 9.10.2:
	lookupIdSubst
  $dIP_acbr
  InScope {k_a9i6 ha_a9ia mvar_a9je bak_a9RO _ha_a9RP fwdDecs_a9RQ
           mem_a9XI tyCtx_a9ZW p_aaz8 sym_aaz9 ext_aaza args_aazb ret_aazc
           ext_aaNr a_ab8t $dIP_ab8N $dKnownNat_ab8P p_abA6 sym_abA7 bak_abA8
           rtp_abA9 a_abAa r_abAb t_abAc st_abAd fs_abAe $dIsSymBackend_abAf
           $d~_abAg $dIP_ac3d co_aciH $dCTuple2_aciJ
           $dIsInterpretedFloatSymExprBuilder_aciZ ds_ddwK a_ddwO c_ddwP
           n_ddwQ tryBindTypedOverride}
  Call stack:
        callStackDoc, called at compiler/GHC/Utils/Panic.hs:190:37 in ghc-9.10.2-b8ed:GHC.Utils.Panic
        pprPanic, called at compiler/GHC/Core/Subst.hs:196:17 in ghc-9.10.2-b8ed:GHC.Core.Subst
    panic, called at compiler/GHC/Utils/Error.hs:507:29 in ghc-9.10.2-b8ed:GHC.Utils.Error


Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

langston-barrett: And indeed, the panic is happening in CI too. Mind if we revert this commit and file an issue saying that we should re-add it when this panic is fixed?

langston-barrett: See #1594.

RyanGlScott: Ugh. Fair enough, there's not much we can do about it if GHC barfs.

PR #1595, opened 2025-10-24
`crucible-llvm`: Support LLVM 17

RyanGlScott:
This adds notional support for simulating LLVM 17 bitcode in `crucible-llvm`. To accomplish this, we bump the `llvm-pretty-bc-parser` submodule to bring in the changes from https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/261, which are necessary to handle the new `DIAssignID` metadata value that was introduced in LLVM 17.

This submodule bump also brings along a partial fix for https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/258, which slightly changes the shape of metadata values that can appear as arguments to LLVM debug intrinsics. As it turns out, there is at least one `crux-llvm` test (`shrink.ll`) that is sensitive to this change. To allow this test to continue to pass, this patch also updates the debug intrinsic argument-translation logic to account for the possibility that metadata references (instead of raw metadata values) may appear as arguments now.

### DIFF
--- a/crucible-llvm-cli/src/Lang/Crucible/LLVM/CLI.hs
+++ b/crucible-llvm-cli/src/Lang/Crucible/LLVM/CLI.hs
@@ -101,6 +101,7 @@ withLlvmHooks k = do
                     , _llvmTypeCtx = tyCtx
                     , llvmGlobalAliases = Map.empty
                     , llvmFunctionAliases = Map.empty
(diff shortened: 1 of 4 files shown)


langston-barrett (approved):
LGTM once the upstream PR is merged. Thanks!

PR #1600, opened 2025-10-28
`crucible-llvm`: Adapt to revamped `LayoutSpec`s in `llvm-pretty`

RyanGlScott:
This bumps the following submodules:

* `llvm-pretty`: https://github.com/GaloisInc/llvm-pretty/pull/167
* `llvm-pretty-bc-parser`: https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/313

The `llvm-pretty` PR makes some backwards-incompatible changes to the `LayoutSpec` data type, which require code changes in `crucible-llvm` to accommodate. In particular, the following changes must be made:

* `PointerSize` now stores its fields in an auxiliary `PointerSize` data type, which requires some minor code changes to adjust to.

* `{Integer,Vector,Float,StackObj}Size` now store their fields in an auxiliary `Storage` data type, which requies some minor code changes to adjust to.

* `AggregateSize` no longer stores a size, as this was a mistake: data layout strings only store the `abi` and `pref` for aggregate alignments. As such, it doesn't make sense to use `AlignInfo` to store aggregate alignments.

  The `aggInfo` function has been removed in favor of `aggregateAlignment`, a `Lens` that retrieves an `Alignment` instead of an `AlignInfo`.

* The code in `Lang.Crucible.LLVM.DataLayout` has been updated to account for the new `ProgramAddrSpace`, `GlobalAddrSpace`, `AllocaAddrSpace`, and `NonIntegralPointerSpaces` data constructors of `LayoutSpec`. For now, these do not have any material effect on `crucible-llvm`, as `crucible-llvm` always assumes the use of the default address space of 0. (See `doc/limitations.md` for more on this point.)

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
@@ -42,7 +42,6 @@ import Control.Lens
 import Control.Monad.State.Strict
 import Data.Map (Map)
 import qualified Data.Map as Map
-import Data.Maybe (fromMaybe)
 import Data.Word (Word32)
 import qualified Text.LLVM as L
 import Numeric.Natural
@@ -131,11 +130,6 @@ floatAlignment :: DataLayout -> Natural -> Maybe Alignment
 floatAlignment dl w = Map.lookup w t
   where AT t = dl^.floatInfo
 
--- | Get the basic alignment for aggregate types.
-aggregateAlignment :: DataLayout -> Alignment
-aggregateAlignment dl =
-  fromMaybe noAlignment (findExact 0 (dl^.aggInfo))
-
 -- | Return maximum alignment constraint stored in tree.
 maxAlignmentInTree :: AlignInfo -> Alignment
 maxAlignmentInTree (AT t) = foldrOf folded max noAlignment t
@@ -165,12 +159,12 @@ data DataLayout
    = DL { _intLayout :: EndianForm
         , _stackAlignment :: !Alignment
         , _functionPtrAlignment :: !Alignment
+        , _aggregateAlignment :: !Alignment
         , _ptrSize     :: !Bytes
         , _ptrAlign    :: !Alignment
         , _integerInfo :: !AlignInfo
         , _vectorInfo  :: !AlignInfo
         , _floatInfo   :: !AlignInfo
-        , _aggInfo     :: !AlignInfo
         , _stackInfo   :: !AlignInfo
         , _layoutWarnings :: [L.LayoutSpec]
         }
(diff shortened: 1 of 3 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
```diff
@@ -289,14 +281,14 @@ fromSize i | i < 0 = error $ "Negative size given in data layout."
...
 -- | Insert alignment into spec.
-setAtBits :: Lens' DataLayout AlignInfo -> L.LayoutSpec -> Int -> Int -> State DataLayout ()
-setAtBits f spec sz a =
-  case fromBits a of
+setAtBits :: Lens' DataLayout AlignInfo -> L.LayoutSpec -> L.Storage -> State DataLayout ()
+setAtBits f spec st =
+  case fromBits (L.alignABI (L.storageAlignment st)) of
     Left{} -> layoutWarnings %= (spec:)
-    Right w -> f . at (fromSize sz) .= Just w
+    Right w -> f . at (fromSize ((L.storageSize st))) .= Just w
```
langston-barrett: Looks like one too many layers of parens perhaps?

crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
```diff
@@ -337,6 +330,17 @@ addLayoutSpec ls =
       -- should revisit this.
       L.FunctionPointerAlign _ a -> setBits functionPtrAlignment ls a
       L.Mangling _      -> return ()
+      -- Currently, we assume that only the default address space (0) is used,
+      -- and we ignore all other address space-related layout specs.
+      -- See doc/limitations.md for more discussion.
+      L.ProgramAddrSpace {} -> return ()
+      L.GlobalAddrSpace {} -> return ()
+      L.AllocaAddrSpace {} -> return ()
+      L.NonIntegralPointerSpaces {} -> return ()
```
langston-barrett, line 339: We don't necessarily have to make a change here and now, but should we consider raising some kind of error up to the user if there is something specified in the data layout that we don't support?

crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
RyanGlScott: Yeah, I must have been listening to Sunn O))) and got carried away with the parentheses. I've changed this to simply `fromSize (L.storageSize st))`.

crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
RyanGlScott: I'm unsure if we'd want to go as far as raising an error, since Clang does emit non-zero address spaces in practice. For instance, in x86 data layouts, Clang [reserves address spaces 270 through 272](https://reviews.llvm.org/D64931) for implementing mixed pointer sizes, as this manifests as non-zero address spaces in the data layout. We could consider just warning instead erroring, but even that would be annoying, since it's a warning that the user would be completely unable to take action on (save for manually modifying the bitcode).

I'm inclined to leave this as is, but if you'd prefer, I could open a `crucible` issue where we could discuss this further.

langston-barrett: All good, thanks for explaining! I assumed that Clang didn't emit non-zero address spaces by default but sounds like I was mistaken.

kquick (approved):

PR #1603, opened 2025-10-29
`crucible-llvm`: Support flags in cast-related instructions (`zext`, `uitofp`, and `trunc`)

RyanGlScott:
LLVM 18, 19, and 20 add flags to various cast-related instructions:

* `nneg` flags in `zext` instructions (added in LLVM 18) and `uitofp` instructions (added in LLVM 19)
* `nuw`/`nsw` flags in `trunc` instructions (added in LLVM 20)

See also https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/306.

This makes the necessary changes to `crucible-llvm` to implement the semantics of these flags. This bumps the following submodules:

* https://github.com/GaloisInc/llvm-pretty/pull/169
* https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/317

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
@@ -52,6 +52,7 @@ module Lang.Crucible.LLVM.Translation.Constant
   , testBreakpointFunction
   ) where
 
+import qualified Control.Exception as X
 import           Control.Lens( to, (^.) )
 import           Control.Monad
 import           Control.Monad.Except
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
```diff
@@ -805,7 +805,9 @@ evalConv expr op mt x = case op of
       , DoubleConst d <- x
       -> return $ IntConst w (BV.mkBV w (truncate d))
 
-    L.UiToFp
+    -- NB: The `nneg` is ignored, as this flag is never enabled in constant
+    -- expressions, only in instructions.
```
langston-barrett: Should we perhaps `assert` that the flag is not enabled? (Here and elsewhere)

RyanGlScott: Yes, good call. I've converted these to `assert`ions.

Issue #1605, opened 2025-10-30
`crucible-llvm`: Fully align `getelementptr` attribute semantics with LLVM Language Reference


RyanGlScott:
While `crucible-llvm` generally does a good job of implementing the semantics of LLVM's `getelementptr` instruction, there are some corner cases where `crucible-llvm`'s behavior differs from the specification in the [LLVM Language Reference](https://releases.llvm.org/20.1.0/docs/LangRef.html#getelementptr-instruction). In particular, `getelementptr` has a number of optional attributes that enable additional validity checks, both during the calculation of the address that `getelementptr` returns, as well as in subsequent memory loads from or stores to that address. At present, `crucible-llvm` implements its own approach to `getelementptr` that implements some subset of what of the attributes check for, but if we want to be fully compliant with the LLVM Language Reference, then we will need to tighten up `crucible-llvm`'s implementation in various ways.

Here is the list of `getelementptr` attributes, along with my understanding of how `crucible-llvm` currently implements (or doesn't implement) each one:

## `nusw`

The LLVM Language Reference says:

> For `nusw` (no unsigned signed wrap):
>
> * If the type of an index is larger than the pointer index type, the truncation to the pointer index type preserves the signed value (`trunc nsw`).
> * The multiplication of an index by the type size does not wrap the pointer index type in a signed sense (`mul nsw`).
> * The successive addition of each offset (without adding the base address) does not wrap the pointer index type in a signed sense (`add nsw`).
> * The successive addition of the current address, truncated to the pointer index type and interpreted as an unsigned number, and each offset, interpreted as a signed number, does not wrap the pointer index type.

`crucible-llvm`s `getelementptr` semantics do not check whether the `nusw` flag is enabled or not. In practice, the semantics already perform some validity checks that involved signed integer overflow, and these likely overlap with the checks described above. Figuring out if all of the checks are implemented will require some auditing, as we do not literally implement the `getelementptr` semantics in terms of the semantics of the `{add,mul,trunc} nsw` instructions, as suggested in the LLVM Language Reference. The following parts of the code will need to be audited:

* https://github.com/GaloisInc/crucible/blob/ff139fe2e7fab0a36b76be48a9950fb3399d58c9/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs#L506-L511
[...]

PR #1606, opened 2025-10-30
`crucible-llvm`: Adapt to LLVM debug records and `getelementptr` attributes 

RyanGlScott:
This brings in the changes from the following submodules:

* https://github.com/GaloisInc/llvm-pretty/pull/168
* https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/316

These contain two notable changes:

* Instead of using debug intrinsic functions, LLVM now records debug info using [debug records](https://llvm.org/docs/RemoveDIsDebugInfo.html). As such, some of the functionality which inspects intrinsic functions in order to track debug information in `crucible-llvm` had to be updated to also look at debug records.
* `getelementptr` has has several new attributes which describe additional validity checks that it should perform when calculating the address that it returns. As noted in https://github.com/GaloisInc/crucible/issues/1605, however, figuring out which checks exactly correspond to which parts of `crucible-llvm`'s `getelementptr` semantics is not simple, and moreover, many of the checks are already performed by default in the semantics.

  As such, I am opting not to change the `getelementptr` semantics, instead making a note of https://github.com/GaloisInc/crucible/issues/1605 is relevant-looking parts of the code. We can always tighten this up later if it becomes important, but for now, the existing implementation is correct enough.

These changes go a long ways towards being able to support LLVM versions up to 19. I will plan to properly support LLVM 19 in the `crux-llvm` test suite in a separate patch.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Errors/Poison.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Errors/Poison.hs
@@ -226,11 +226,11 @@ explain =
       ]
 
     -- The following explanation is a bit unsatisfactory, because it is specific
-    -- to how we treat this instruction in Crucible.
+    -- to how we treat this instruction in Crucible. (See also #1605.)
     GEPOutOfBounds _ _ -> cat $
       [ "Calling `getelementptr` resulted in an index that was out of bounds for the"
       , "given allocation (likely due to arithmetic overflow), but Crucible currently"
-      , "treats all GEP instructions as if they had the `inbounds` flag set."
+      , "treats all GEP instructions as if they had the `inbounds` attribute set."
       ]
 
 details :: forall sym ann.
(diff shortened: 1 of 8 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Errors/Poison.hs
```diff
@@ -226,11 +226,11 @@ explain =
       ]
 
     -- The following explanation is a bit unsatisfactory, because it is specific
-    -- to how we treat this instruction in Crucible.
+    -- to how we treat this instruction in Crucible. (See also #1605.)
     GEPOutOfBounds _ _ -> cat $
       [ "Calling `getelementptr` resulted in an index that was out of bounds for the"
       , "given allocation (likely due to arithmetic overflow), but Crucible currently"
-      , "treats all GEP instructions as if they had the `inbounds` flag set."
+      , "treats all GEP instructions as if they had the `inbounds` atribute set."
```
kquick: s/atribute/attribute/

kquick (approved):

PR #1608, opened 2025-10-31
Assign unique type ids to each type

glguy:


crucible-mir/src/Mir/Generator.hs
```diff
@@ -187,6 +187,8 @@ data CollectionState
       -- but with different hashes. Most of the time, however, this list will
       -- contain exactly one disambiguator per crate name.
       _crateHashesMap :: !(Map Text (NonEmpty Text)),
+      -- | Preallocated type IDs per instantiated type for dynamic casts
+      _tyIdMap        :: !(Map Ty Int),
```
RyanGlScott, line 193: Also worth saying that this map is used in exactly one spot (in the custom override for `type_id`)

crucible-mir/src/Mir/Generator.hs
```diff
@@ -877,8 +879,12 @@ mirAggregate_set ::
...
 mirAggregate_set off sz tpr val ag = G.extensionStmt $ MirAggregate_Set off sz tpr val ag
 
-
-
+getTypeId :: Ty -> MirGenerator h s ret Int
+getTypeId ty = do
+    m <- use (cs . tyIdMap)
+    case Map.lookup ty m of
+      Just tyId -> pure tyId
+      Nothing -> mirFail "Ty not allocated a type id!"
```
RyanGlScott: Should this `Nothing` case panic instead of `mirFail`? My understanding is that this should never happen, since every conceivable type in the program should be in the domain of the `tyIdMap`.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2976,6 +2976,10 @@ mkCrateHashesMap
     f m = map (\did -> (did ^. M.didCrate, did ^. M.didCrateDisambig :| []))
               (Map.keys m)
 
+-- | Generate a map assigning a unique integer to every type in the program
+mkTyIdMap :: M.Collection -> Map Ty Int
+mkTyIdMap col = Map.fromList (zip (Map.keys (col ^. layouts)) [1..])
```
spernsteiner, line 2981: It looks like `TypeId::of::<T>()` is a `const fn`, so in the future we may need to change this to get the actual `TypeId` values from mir-json so that the values produced at run time will match the ones rustc produces during const evaluation.

PR #1609, opened 2025-11-03
`crucible-debug`: Only create `.crucible-debug` file when debugger requests input

### LINKED ISSUE #1604: Crux always creates a `.crucible-debug` file, even when the debugger isn't used
Recent versions of Crux eagerly create a `.crucible-debug` file, even when the debugger is never actually used. As an example, if you run `crux-llvm` on a simple file like this:

```c
// test.c
int main(void) {
  return 0;
}
```

Then you can find an empty `.crucible-debug` file created in your current working directory after invoking `crux-llvm`:

```
$ cabal -v0 run exe:crux-llvm -- test.c
$ cat .crucible-debug
```

I have also observed something similar with `crux-mir-comp`.

I (and other folks who are new to Crux) find this behavior surprising. I think we should hold off on creating a `.crucible-debug` file unless the debugger is explicitly invoked.

RyanGlScott:
Fixes #1604.

### DIFF
--- a/crucible-debug/src/Lang/Crucible/Debug/Inputs.hs
+++ b/crucible-debug/src/Lang/Crucible/Debug/Inputs.hs
@@ -20,6 +20,7 @@ module Lang.Crucible.Debug.Inputs
   , defaultDebuggerInputs
   ) where
 
+import Control.Concurrent.Extra (once)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad qualified as Monad
 import Control.Monad.Reader qualified as Reader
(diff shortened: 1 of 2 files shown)


crucible-debug/src/Lang/Crucible/Debug/Inputs.hs
```diff
@@ -156,9 +156,12 @@ defaultDebuggerInputs ::
...
-  initIsocline
   pure $
     parseInputsWithRetry
       cExts
-      (Text.pack <$> lift readIsocline)
+      (do -- NB: We delay initializing Isocline until the first input is
+          -- requested. This ensures that a .crucible-debug file won't be
+          -- created unless a Crux user specifically requests the debugger.
+          liftIO initIsocline
+          Text.pack <$> lift readIsocline)
```
langston-barrett: I think this calls `initIsocline` *every* time input is requested. That might be fine, but it might be a bit safer to wrap this in something like
```haskell
import Data.IORef

once :: IO () -> IO (IO ())
once act = do
  doneRef <- newIORef False
  pure $ do
    done <- readIORef doneRef
    if done
    then pure ()
    else do
      writeIORef doneRef True
      act
```

langston-barrett: (Wrap the call to `initIsocline`, that is)

RyanGlScott: Done in 2a95dcb3a86bec9ecc8fb8927a25f49b5be10f3a.

PR #1613, opened 2025-11-04
`crucible-mir`: Add basic `size_of_val` override

### LINKED ISSUE #1612: `crucible-mir`: Support the `size_of_val` intrinsic
It would be convenient to be able to simulate the [`size_of_val`](https://doc.rust-lang.org/std/intrinsics/fn.size_of_val.html) intrinsic:

```rs
pub const unsafe fn size_of_val<T>(ptr: *const T) -> usize;
```

For instance, `size_of_val` is directly invoked [here](https://github.com/Lokathor/bytemuck/blob/da748163ea203f80098b6bdc754c54ebc535364c/src/internal.rs#L384) in the [`bytemuck`](https://docs.rs/bytemuck/latest/bytemuck/) crate, which in turn is a dependency of [`binrw`](https://docs.rs/binrw/latest/binrw/).

As the comments for the higher-level [`std::mem::size_of_val`](https://doc.rust-lang.org/std/mem/fn.size_of_val.html) and [`std::mem::size_of_val_raw`](https://doc.rust-lang.org/std/mem/fn.size_of_val_raw.html) functions indicate, the semantics of `size_of_val::<T>(x)` are similar to that of `size_of::<T>()`, except that `size_of_val` also works for unsized types such as slices and trait objects, e.g.,

```rs
let x: [u8; 13] = [0; 13];
let y: &[u8] = &x;
assert_eq!(13, size_of_val(y));
```

Supporting slice types `[T]` would be straightforward, as all that would be required is to take the slice length and multiply it by `size_of::<T>()`. Supporting trait objects (e.g., `dyn Trait`) or custom DSTs would be trickier, however. We would need to get the size (and maybe alignment) of the last field, and then we would need to add the sizes of the remaining fields. (Note that for the `bytemuck` use case above, it would suffice to only support slices, and not trait objects or custom DSTs.)

RyanGlScott:
For now, this override does not support trait objects or custom DSTs. This is still enough to be useful in practice despite these limitations.

Fixes #1612.

### DIFF
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -34,6 +34,7 @@ import           Data.String (fromString)
 import Data.Text (Text)
 import qualified Data.Text as Text
 import qualified Data.Vector as V
+import qualified Prettyprinter as PP
 
 import Control.Monad
 import Control.Lens ((^.), at, use)
@@ -111,6 +112,7 @@ customOpDefs = Map.fromList $ [
                          , rotate_left
                          , rotate_right
                          , size_of
+                         , size_of_val
                          , min_align_of
                          , intrinsics_assume
                          , assert_inhabited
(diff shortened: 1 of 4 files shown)


crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1132,6 +1134,68 @@ size_of = (["core", "intrinsics", "size_of"], \substs -> case substs of
...
+                TyDynamic {} -> unsupportedTraitObject ty
+                TyAdt {} -> unsupportedCustomDst ty
+                _ -> panic "size_of_val"
+                       ["Unexpected DynRefRepr type", show (PP.pretty ty)]
+
+            -- All other cases should correspond to sized types. For these
+            -- cases, computing the value's size is equivalent to computing the
+            -- type's size.
+            MirReferenceRepr ->
+                getLayoutFieldAsMirExp "size_of_val" laySize ty
```
spernsteiner, line 1177: Extern types (`TyForeign`), which are unsized but have no metadata, would presumably also fall into this case.  Extern types are unstable, and it sounds like the Rust devs are likely to implement the "`Sized` hierarchy" feature first, which will (among other things) statically prevent calling `size_of_val` on extern types.  So we can probably get away with ignoring them here.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1132,6 +1134,68 @@ size_of = (["core", "intrinsics", "size_of"], \substs -> case substs of
...
+    -- by the size of the element type. Note that slice element types are
+    -- always sized, so we do not need to call size_of_val recursively here.
+    sizeOfSlice ::
+      Ty -> R.Expr MIR s MirSlice -> MirGenerator h s ret (MirExp s)
+    sizeOfSlice ty e = do
+        let usizeWidth = knownNat @SizeBits
+        let len = getSliceLen e
+        sz <- getLayoutFieldAsExpr "size_of_val" laySize ty
+        pure $ MirExp (C.BVRepr usizeWidth)
+             $ R.App $ E.BVMul usizeWidth len sz
```
samcowger: Leveraging `UsizeRepr` and `usizeMul` would let us be a bit more concise here:
```suggestion
        let len = getSliceLen e
        sz <- getLayoutFieldAsExpr "size_of_val" laySize ty
        pure $ MirExp UsizeRepr $ R.App $ usizeMul len sz
```

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1132,6 +1134,68 @@ size_of = (["core", "intrinsics", "size_of"], \substs -> case substs of
     _ -> Nothing
     )
 
+size_of_val :: (ExplodedDefId, CustomRHS)
+size_of_val = (["core", "intrinsics", "size_of_val"], \substs -> case substs of
+    Substs [_] -> Just $ CustomOp $ \opTys ops -> case (opTys, ops) of
+        -- We first check whether the underlying type is sized or not.
+        ([TyRawPtr ty Immut], [MirExp tpr e]) -> case tpr of
```
samcowger: Do we need to account for the `([TyRef ty Immut], [MirExp tpr e])` case?

crucible-mir/src/Mir/TransCustom.hs
RyanGlScott: TIL about [extern types](https://rust-lang.github.io/rfcs/1861-extern-types.html).

crucible-mir/src/Mir/TransCustom.hs
RyanGlScott: I assume that you mean `TyRef ty Mut` instead of `TyRef ty Immut`? (The latter is already what is being used in the code.) If so, then no—the type signature for `size_of_val` takes a `*const` argument, so I don't think it's ever possible for the mutability to be `Mut`.

Thinking about this some more, however, there's really not much of a reason this code needs to match on the mutability at all. I'll go ahead and change `Mut` to a wildcard pattern.

RyanGlScott:
I've opened #1614 to track supporting trait objects/custom DSTs specifically. I suspect that we'll need largely the same machinery to support both, so I've opted to combine both of them into a single issue. I've also cited #1614 in the relevant parts of this PR that are affected by the lack of trait object/custom DST support.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Ah - I meant what I said, but I now realize why it didn't make much sense. I was looking at the type signature for [`std::mem::size_of_val`](https://doc.rust-lang.org/std/mem/fn.size_of_val.html) (and wondering how matching on `TyRawPtr` worked in the first place), instead of [`core::intrinsics::size_of_val`](https://doc.rust-lang.org/beta/core/intrinsics/fn.size_of_val.html).

I have no opinion on matching mutability vs not, though I don't know how we'd ever see `Mut` here.

PR #1615, opened 2025-11-04
`crucible-mir`: support `needs_drop` intrinsic

### LINKED ISSUE #1611: `crucible-mir`: Support the `needs_drop` intrinsic
It would be convenient to be able to simulate the [`needs_drop`](https://doc.rust-lang.org/std/intrinsics/fn.needs_drop.html) intrinsic:

```rs
pub const fn needs_drop<T>() -> bool;
```

For instance, this intrinsic is transitively invoked [here](https://github.com/Manishearth/array-init/blob/60635047ab2800832319829840c68df03c39945b/src/lib.rs#L236) in the [`array_init`](https://docs.rs/array-init/latest/array_init/) crate, which in turn is a dependency of [`binrw`](https://docs.rs/binrw/latest/binrw/).

As the documentation for [`std::mem::needs_drop`](https://doc.rust-lang.org/std/mem/fn.needs_drop.html) (the stable version of this intrinsic) indicates, it would be perfectly safe for a `needs_drop` override to always return `true`, but with the drawback that this would create a bit of a mismatch between what gets verified and what actually gets run. Alternatively, we could instead implement a more precise override that returns exactly what `rustc` does by leveraging the `rustc` API's [`needs_drop`](https://doc.rust-lang.org/1.86.0/nightly-rustc/rustc_middle/ty/struct.Ty.html#method.needs_drop) function, which is what powers the `needs_drop` intrinsic under the hood. After calling the `needs_drop` API function, we could store the result somewhere in the MIR JSON for a `Ty`, similarly to how we already store a `Ty`'s layout information ([here](https://github.com/GaloisInc/mir-json/blob/666f9f6797107238f206df5ca2233cbb0c756ced/src/analyz/ty_json.rs#L601-L625)).

samcowger:
Fixes #1611.

Marked as draft while https://github.com/GaloisInc/mir-json/pull/183 is pending.

### DIFF
--- a/crux-mir/test/conc_eval/drop/needs_drop.rs
+++ b/crux-mir/test/conc_eval/drop/needs_drop.rs
@@ -1,0 +1,20 @@
+use std::mem::needs_drop;
+
+struct Foo(u32);
+
+impl Drop for Foo {
+    fn drop(&mut self) {}
+}
+
+struct Bar(u32);
+
+#[cfg_attr(crux, crux::test)]
+fn crux_test() {
+    assert!(needs_drop::<Foo>());
+    assert!(!needs_drop::<Bar>());
+    assert!(!needs_drop::<u32>());
+}
+
+fn main() {
+    println!("{:?}", crux_test());
+}


crucible-mir/src/Mir/Mir.hs
```diff
@@ -295,6 +304,15 @@ data Collection = Collection {
...
 } deriving (Show, Eq, Ord, Generic)
 
+-- | Information about a type (minus its name), as seen in the @tys@ table in
+-- @mir-json@-generated JSON.
+data TyInfo = TyInfo
+    { _tiLayout :: Maybe Layout
+    , _tiNeedsDrop :: Bool
+    , _tiTy :: Ty
+    }
+    deriving (Show, Eq, Ord, Generic)
```
RyanGlScott, line 314: It would be worth introducing lenses for these for consistency with the other data types in this module...

crucible-mir/src/Mir/ParseTranslate.hs
```diff
@@ -83,7 +85,7 @@ uninternMir col =
     tyMap = fmap (uninternTys unintern) (col ^. namedTys)
     unintern name = case M.lookup name tyMap of
         Nothing -> error $ "missing " ++ show name ++ " in type map"
-        Just (ty, _) -> ty
+        Just tyInfo -> _tiTy tyInfo
```
RyanGlScott: ...so that you can use `tyInfo ^. tiTy` here. (And similarly wherever `_ti*` fields are used as record selectors in this PR.)

crux-mir/test/conc_eval/drop/needs_drop.rs
```diff
@@ -0,0 +1,22 @@
+#![feature(core_intrinsics)]
+
+use core::intrinsics::needs_drop;
```
RyanGlScott: To fix the test suite error that the CI caught, it might be more direct just to do:


```suggestion
use std::intrinsics::needs_drop;
```

RyanGlScott:
I'll plan to land https://github.com/GaloisInc/crucible/pull/1613 after this one, since there is a decent chance that the two PRs will conflict with each other.

crux-mir/test/conc_eval/drop/needs_drop.rs
RyanGlScott: My apologies, I misread what the error message was saying. Rather, instead of using `{core,std}::intrinsics::needs_drop` (both of which will cause `rustc` to bark about it being unstable), we should instead use the stabilized [`std::mem::needs_drop`](https://doc.rust-lang.org/std/mem/fn.needs_drop.html) function defined in terms of the former.

samcowger: You beat me to the punch. I agree that `std::mem` is probably best.

PR #1619, opened 2025-11-07
`crucible-mir`: Improve error messages involving MIR JSON schema version mismatches

### LINKED ISSUE #1618: `crucible-mir`: Perform MIR JSON schema version check before parsing the rest of the JSON
`crucible-mir` includes a MIR JSON schema version check [here](https://github.com/GaloisInc/crucible/blob/59043ece0ee32945166039b8f6159127c43af832/crucible-mir/src/Mir/ParseTranslate.hs#L58-L66), which is intended to give more obvious feedback if you try to use a MIR JSON file with an incompatible schema. Unfortunately, this check doesn't work very well in practice, as `crucible-mir` is more likely to throw a random JSON parsing error before it ever performs the version check.

To see what I mean, I decided to use a version of `crux-mir` built with commit https://github.com/GaloisInc/crucible/commit/59043ece0ee32945166039b8f6159127c43af832 (which uses schema version `5`) and run it on a small MIR JSON file built with `mir-json` commit https://github.com/GaloisInc/mir-json/commit/666f9f6797107238f206df5ca2233cbb0c756ced (which uses schema version `4`). Given this Rust file:

```rs
// test.rs
#[crux::test]
pub fn crux_test() -> u8 {
    0
}
```

Running `crux-mir` yields:

```
user error (JSON Decoding of test.all.mir failed: Error in $.tys[0]: key "needs_drop" not found)
```

Urk, that error message is terrible! To see why the version check didn't file here, let's look at the code surrounding the check:

https://github.com/GaloisInc/crucible/blob/59043ece0ee32945166039b8f6159127c43af832/crucible-mir/src/Mir/ParseTranslate.hs#L53-L66

This code parses the entirety of the MIR JSON file _before_ it checks the version, which is more eager than we want. Ideally, we would partially decode the JSON first to get the version number, check that it matches our expectations, and only then decode the rest once we've confirmed that it's the right version.

RyanGlScott:
Previously, `parseMIR` would parse the entirety of a MIR JSON file _before_ checking to see if the JSON schema version number is supported. This is almost never the right order in which to do things, because if you are parsing a file using an unsupported schema version, then parsing the JSON file will almost certainly fail with an inscrutable error before we ever get to the schema version check.

This commit rearranges the order in which these checks are performed. We now carefully set things up in `parseMIR` such that is parses the schema version number and validates it before it parses any of the other parts of the JSON file. `parseMIR` now only parses the remainder of the JSON file after it has confirmed that the schema version is supported.

Fixes #1618.

### DIFF
--- a/crucible-mir/src/Mir/ParseTranslate.hs
+++ b/crucible-mir/src/Mir/ParseTranslate.hs
@@ -113,7 +113,7 @@ parseMIR path f = do
         fail $ unlines
           [ path ++ " uses an unsupported mir-json schema version: "
                  ++ show actualSchemaVersion
-          , "This crux-mir release only supports schema version "
+          , "This release only supports schema version "
             ++ show supportedSchemaVersion ++ "."
           , "(See https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md"
           , "for more details on what the schema version means.)"


crucible-mir/crucible-mir.cabal
```diff
@@ -115,7 +115,7 @@ library
   import: shared
 
   build-depends: base >= 4.11 && < 5,
-                 aeson < 2.3,
+                 aeson >= 2.0 && < 2.3,
```
RyanGlScott, line 118: Reviewers' note: I raised the lower version bounds on `aeson` to `>= 2.0` because `Mir.ParseTranslate` now explicitly requires the `Data.Aeson.KeyMap` module, which was first introduced in `aeson-2.0`.

RyanGlScott:
With these changes, parsing the example from https://github.com/GaloisInc/crucible/issues/1618#issue-3600040835 with a mismatched schema version now yields the following error message:

```
user error (test.all.mir uses an unsupported mir-json schema version: 4
This crux-mir release only supports schema version 5.
(See https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md
for more details on what the schema version means.)
)
```

crucible-mir/crucible-mir.cabal
samcowger: I don't mind this bump, but if you feel it's undesirable, I think you could do the same lookup via [`(.:)`](https://hackage.haskell.org/package/aeson-2.2.3.0/docs/Data-Aeson.html#v:.:) instead.

RyanGlScott: I don't think it's particularly undesirable, as `aeson-2.0` is pretty well established at this point (having first been released in 2021), and other parts of the github.com/GaloisInc ecosystem already use `>=2.0` as a lower bound (e.g., `cryptol` does it [here](https://github.com/GaloisInc/cryptol/blob/b727e80daaca0738395fe1d714b6bffc35f68e76/cryptol.cabal#L51)).

RyanGlScott:
You raise a good point about the performance impact of parsing to an intermediate `J.Value`. Normally, I'd be inclined to say that the performance impact of this would be negligible, but given that we are parsing very large JSON files, it's likely that the performance difference _would_ be measurable here. As such, I'll implement your suggested refactoring to ensure that the common case doesn't degrade in performance.

RyanGlScott:
I've pushed a new version of the PR that implements @spernsteiner's suggested refactoring.

crucible-mir/src/Mir/ParseTranslate.hs
```diff
@@ -46,30 +50,79 @@ import Debug.Trace
...
+          J.Error msg ->
+            fail $ path ++ " has a \"version\" field, but it is not a number: " ++ msg
+          J.Success actualSchemaVersion -> pure actualSchemaVersion
+      -- As a sanity-check, ensure that the number which "version" maps to is a
+      -- supported JSON schema version. If not, error out early.
+      unless (actualSchemaVersion == supportedSchemaVersion) $
+        fail $ unlines
+          [ path ++ " uses an unsupported mir-json schema version: "
+                 ++ show actualSchemaVersion
+          , "This crux-mir release only supports schema version "
```
sauclovian-g: Isn't this also the code path used by SAW? Maybe just say "This release"...?

RyanGlScott: Yes, excellent point. I've dropped the mention of `crux-mir` here.

PR #1629, opened 2025-11-12
Add an overview of how Crucible works

yav:


langston-barrett:
This seems like a useful high-level overview, any reason not to integrate it into the Haddocks?

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+    * `CrucibleType` is the list of types
+    * `TypeRepr` is the datatype of types, used when we want to store or pass
+      around Crucible types.
+
+
+## Values
+
+The inhabitants of each type are specified via the type function `RegValue`
+(`Simulator.RegValue`).  We also have `RegValue'` which is just
+a newtype wrapper `RegValue`, so that we can partially apply it.
```
langston-barrett, line 19: I'd imagine that this sentence would be a bit opaque to those not already familiar with `-XTypeFamilies`. I'm not sure exactly what the best solution is, though, because I can't find the prohibition on partial application mentioned anywhere in [the official docs](https://ghc.gitlab.haskell.org/ghc/doc/users_guide/exts/type_families.html).

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+
+An important subset of the Crucible types are the base types (see `BaseToType`),
+which is a backend specific type for the symbolic expression we can construct
+(see `SymExpr` in `what4`).  Only these types may contain variables.
+In practice, we only have one backend, namely `ExprBuidler` defined in `what4`,
```
langston-barrett, line 24: Crucible defines a typeclass called [`IsSymBackend`](https://github.com/GaloisInc/crucible/blob/5e610f0ff44fc080ef2d875c4a6c48ab401541df/crucible/src/Lang/Crucible/Backend.hs#L225), but `ExprBuilder` is *not* an instance of that class, so I'm worried that the use of the word "backend" will be misleading.

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+(`What4.Expr.Builder`) which uses `what4`'s `Expr` as the type of symbolic
+expressions.
+
+Also, in some cases we use `RegEntry` (`Simulator.RegMap`) which
+is just a pair of `RegValue` and it's associated `TypeRep`.
+
+There's also `BaseTerm` (`CFG.Expr`), which is similar to `RegEntry` but
+for base types---it contains a `BaseTypeRepr` and a value of the corresponding
+base type.  XXX: It is actually parameterized, so that it could be used with other
+things than values, but I am not sure if this generality is used.
```
langston-barrett, line 34: Huh, I actually didn't know about `BaseTerm`. That makes me a little suspicious that it might not be an important enough part of using Crucible's API to call out in this high-level overview.

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+The expression language for `CFG.Core` is the type `App` in `CFG.Expr`.
+
+
+
+## The Symbolic Simulator
+
+The state of a running simulator is described in `Simulator.ExecutionTree`:
+  * `ExecState` is the current state of execution.
+     We start with `InitialState`, and keep performing steps until we get
+     to a `ResultState`
```
langston-barrett:
```suggestion
     to a `ResultState`.
```

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+  * As the simulator executes, it keeps track of its state in `SimState`.
+  * `SimContext` is the part of the state that persists across branches
+    (e.g, after we explore the `then` part of an `if` statement, we'll
+    roll back some of the state changes before simulating the `else` part,
+    but the stuff in `SimContexts` persists).  An important part of the
+    `SymContext` is the simulator's backend (`Backend`), which is how the
```
langston-barrett:
```suggestion
    `SimContext` is the simulator's backend (`Backend`), which is how the
```

langston-barrett: Crucible does not actually define a symbol (i.e., type or class) named `Backend`.

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+is described by `evalApp` in `Simulator.Evaluation`.
+
+A lot of useful functionality relevant to the simulator can be accessed
+from module `Simulator`.
+
+The simulator supports global mutable references.  Our tools use one such
+reference to store a language specific memory model.  For example, our LLVM
+tools are in `crucible-llvm`, the memory model is in `LLVM.MemModel` (with
+more modules in `LLVM.MemModel.*`), and the function that makes the global
+variable that contains the memory model is `LLVM.MemModel.mkMemVar`.
```
langston-barrett: Crucible has a both (1) global variables and (2) a `ReferenceType`. I'm concerned that this paragraph conflates the two, perhaps just:
```suggestion
The simulator supports mutable global variables.  Our tools use one such
global to store a language specific memory model.  For example, our LLVM
frontend is in `crucible-llvm`, the memory model is in `LLVM.MemModel` (with
more modules in `LLVM.MemModel.*`), and the function that makes the global
variable that contains the memory model is `LLVM.MemModel.mkMemVar`.
```

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+References to Crucible modules should all start with `Lang.Crucible`, which
+we omit for brevity.
+
+## Types
+
+The types of the Crucible language are defined in `Types`.
+  Of interest:
+    * `CrucibleType` is the list of types
+    * `TypeRepr` is the datatype of types, used when we want to store or pass
+      around Crucible types.
```
spernsteiner, line 12: I would specify: `CrucibleType` is the type-level representation of types, and `TypeRepr` is the value-level singleton (indexed by a `CrucibleType`) used to match on types at run time.  Otherwise it's a bit unclear from this description how the two are different.

doc/overview.md
```diff
@@ -0,0 +1,79 @@
...
+
+An important subset of the Crucible types are the base types (see `BaseToType`),
+which is a backend specific type for the symbolic expression we can construct
+(see `SymExpr` in `what4`).  Only these types may contain variables.
+In practice, we only have one backend, namely `ExprBuidler` defined in `what4`,
+(`What4.Expr.Builder`) which uses `what4`'s `Expr` as the type of symbolic
+expressions.
+
+Also, in some cases we use `RegEntry` (`Simulator.RegMap`) which
+is just a pair of `RegValue` and it's associated `TypeRep`.
```
spernsteiner:
```suggestion
is just a pair of `RegValue` and it's associated `TypeRepr`.
```

doc/overview.md
langston-barrett: In that case, it might also be worth linking to the `parameterized-utils` README, which explains the concept of a singleton type.

doc/overview.md
yav: Probably not super important, but it is used in Crucible's expression language in the operations for symbolic arrays, so it might be worth mentioning.

langston-barrett:
Looks like this was moved to #1633.

PR #1630, opened 2025-11-12
ci: capture more `crux-mir` output for `symb_eval` golden tests

samcowger:
Including, most notably, how many goals (if any) were generated and proved/disproved in the course of verification. This could have helped avoid e.g. the performance regression addressed in #1625.

### DIFF
--- a/crux-mir/test/Test.hs
+++ b/crux-mir/test/Test.hs
@@ -74,7 +74,7 @@ runCrux rustFile outHandle mode =
     -- The timeout is temporarily increased even further due to a performance
     -- regression (#627).  This keeps CI from breaking while we investigate.
     -- TODO: revert the timeout to 180 once performance is fixed
-    let quiet = True
+    let quiet = False
     let outOpts = (Crux.outputOptions defaultCruxOptions)
                     { Crux.simVerbose = 0
                     , Crux.quietMode = quiet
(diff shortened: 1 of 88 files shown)


samcowger:
> Truth be told, I'm not sure I like having this enabled by default. The problem is that the number of proof goals in a test case can change unpredictably based on lower-level changes to implementation details, and my fear is that we may need to frequently update expected test output based on knock-on changes of this sort.

I'm sympathetic to the concern of test output churn, and I agree that it's a plausible effect of these changes. I think it's easy enough to update test output, but I wouldn't want that to become a constant chore. However - and this may just be naive optimism on my part - it seems just as plausible to me that this _wouldn't_ produce major churn. If we end up seeing only a handful of tests change from time to time, or many tests change in identical ways, then it seems feasible (and appropriate) to review those changes, and see whether we find them expected/acceptable.

> What if we had a way to opt into emitting proof goals for certain test cases where it matters, but otherwise leaving it disabled by default?

I am a bit leery of an opt-in choice, since I suspect that test authors wouldn't readily think to opt in or not, and that we'd instead opt in only after finding and fixing a problem like #1625's.

Two possible (not necessarily mutually exclusive) compromises come to mind:
- We try this out for a few days/weeks/PRs, and we feel very free to revert it if it becomes annoying.
- We default to emitting proof goals, but implement a way to opt _out_ of that on a per-test basis, so that the decision to do so is necessarily subject to review.

Do either of those strike you as reasonable?

PR #1631, opened 2025-11-12
CI: Bump from `macos-14` to `macos-15`

RyanGlScott:
This also requires bumping the `what4-solvers` snapshot to a more recent version that supports `macos-15`.

### DIFF
--- a/crux-llvm/test-data/golden/abd-test-file-32.cvc5.good
+++ b/crux-llvm/test-data/golden/abd-test-file-32.cvc5.good
@@ -1,3 +1,14 @@
+SKIP_TEST
+
+This test is disabled because it takes an extraordinarily long time to complete
+with `cvc5-1.2.0` or later (as reported upstream in
+https://github.com/cvc5/cvc5/issues/12234).
+
+TODO(#1635): Re-enable this test once the upstream issue has been fixed and is
+available in a what4-solvers release.
+
+-----
+
 [Crux] Found counterexample for verification goal
 [Crux]   test-data/golden/abd-test-file-32.c:11:0: error: in main
 [Crux]   crucible_assert
--- a/crux-llvm/test-data/golden/abd-test-maxint-32.cvc5.good
+++ b/crux-llvm/test-data/golden/abd-test-maxint-32.cvc5.good
@@ -1,3 +1,13 @@
+SKIP_TEST
+
+This test is disabled because it loops forever with `cvc5-1.2.0` or later (as
+reported upstream in https://github.com/cvc5/cvc5/issues/12234).
+
+TODO(#1635): Re-enable this test once the upstream issue has been fixed and is
+available in a what4-solvers release.
+
+-----
+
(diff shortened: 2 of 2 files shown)


.github/Dockerfile-crux-llvm
```diff
@@ -61,7 +61,7 @@ RUN case ${TARGETPLATFORM} in \
         printf "Unsupported architecture: %s\n" "${TARGETPLATFORM}" >&2 \
         exit 1 ;; \
     esac && \
-    curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20250606/ubuntu-22.04-${WHAT4_SOLVERS_ARCH}-bin.zip"
+    curl -o solvers.zip -sL "https://github.com/GaloisInc/what4-solvers/releases/download/snapshot-20251112/ubuntu-22.04-${WHAT4_SOLVERS_ARCH}-bin.zip"
```
langston-barrett: While we're in the neighborhood, I usually like to be explicit in places like `Dockerfile`s and CI systems and use long flags. Also, we're missing a few of the flags that improve the security of cURL, like `--proto`.
```
curl \
  --fail \
  --location \
  --proto '=https' \
  --show-error \
  --silent \
  --tlsv1.2 \
```
Feel free to ignore this though, we can make this change at any time.

langston-barrett (approved):

RyanGlScott:
Hm, there is a regression in the `abd-test-maxint-32.c` test case, which uses `cvc5`:

```
    abd-test-maxint-32.c
      solver=cvc5
        loop-merging=loop
          clang-range=pre-clang15
            cvc5 1.3.1
              abd-test-maxint-32
terminated
terminated
crux-llvm-test: SignalException 15
                abd-test-maxint-32 crux run:  
```

This was most likely caused by the change from `cvc5-1.1.1` to `1.3.1` when upgrading `what4-solvers`. I'll need to take a closer look to see what went wrong.

RyanGlScott:
The problem appears to be a regression in recent versions of `cvc5`, starting with `1.2.0`. I've opened https://github.com/cvc5/cvc5/issues/12234 upstream.

This raises a question of what to do in the meantime. Since `--get-abducts` is a somewhat experimental Crux feature, I am inclined to simply remove the problematic test for now. Luckily, there are still some abduction-related test cases that do complete in a reasonable amount of time with recent CVC5 versions, so we can ensure that the basics of `--get-abducts` continue to work.

.github/Dockerfile-crux-llvm
RyanGlScott: Done in 384c9b8dfc4860feb26eb142d10066173d08b4a7.

langston-barrett:
I think removing the test and filing a GitHub issue to restore it later is a reasonable solution for now.

.github/Dockerfile-crux-llvm
langston-barrett: Thanks!

PR #1633, opened 2025-11-13
Documentaion: add an overview on how Crucible works

yav:


crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
+{- | This module is only for documentation purposes, and provides a high
+level overview of Crucible aimed at developers. -}
+{-# OPTIONS_GHC -Wno-unused-imports #-}
+module Lang.Crucible.README where
```
langston-barrett, line 4: Just to be explicit that this module doesn't (and shouldn't) define anything 
```suggestion
module Lang.Crucible.README () where
```

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
+{- | This module is only for documentation purposes, and provides a high
+level overview of Crucible aimed at developers. -}
```
langston-barrett, line 2: nbd, but the "aimed at developers" part feels a bit redundant since this is embedded in the API docs.

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+* 'CrucibleType' is the Haskell type-level description of all Crucible types
+* 'TypeRepr' are the associated value-level singletons, which
+  are used when we pass around types, or store them in data structures.
+-}
+
+-- * Crucible Values
+
+{- $
+The inhabitants of each type are specified via the type function 'RegValue'.
+We also have 'RegValue'' which is just a newtype wrapper around `RegValue`,
```
langston-barrett: Does the quoting style here result in a correct hyperlink? If not, could use backticks. Might be a bit more readable anyway:
```suggestion
We also have `RegValue'` which is just a newtype wrapper around `RegValue`,
```

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+because in Haskell type families may not be partially applied but `newtypes` can.
+
+An important subset of the Crucible types are the base types (see 'BaseToType'),
+which is a backend specific type for the symbolic expression we can construct
```
langston-barrett: The use of "backend" here might be a bit confusing considering that the same word is used to mean something else later in this document (`IsSymBackend`).

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+(see 'SymExpr' in [what4](https://github.com/Galoisinc/what4)).
+Only these types may contain variables. In practice, we always `what4`'s
```
langston-barrett: If you just want monospacing and don't expect linking, I think `@` is the correct syntax.
```suggestion
Only these types may contain variables. In practice, we always use @what4@'s
```

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+* 'TypeRepr' are the associated value-level singletons, which
+  are used when we pass around types, or store them in data structures.
+-}
+
+-- * Crucible Values
+
+{- $
+The inhabitants of each type are specified via the type function 'RegValue'.
+We also have 'RegValue'' which is just a newtype wrapper around `RegValue`,
+because in Haskell type families may not be partially applied but `newtypes` can.
```
langston-barrett:
```suggestion
because in Haskell type families may not be partially applied but @newtypes@ can.
```

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+which is a backend specific type for the symbolic expression we can construct
+(see 'SymExpr' in [what4](https://github.com/Galoisinc/what4)).
+Only these types may contain variables. In practice, we always `what4`'s
+'Expr' type to represent symbolic expressions.
+
+Also, in some cases we use 'RegEntry' which
+is just a pair of a 'RegValue' and it's associated 'TypeRepr'.
+
+There's also 'BaseTerm', which is similar to 'RegEntry' but
+for base types---it contains a `what4`'s 'BaseTypeRepr' and a value of the corresponding
```
langston-barrett:
```suggestion
for base types---it contains a @what4@'s 'BaseTypeRepr' and a value of the corresponding
```

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+base type (usually;  the type is parameterized on exactly what we package
+with the type).
```
langston-barrett, line 55: "the type" is used twice in this sentence to refer to different things. Perhaps "`BaseTerm` is parameterized on..."?

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+  
+-- * Symbolic Simulator
+  
+{- $
+The state of a running simulator is described in "Lang.Crucible.Simulator.ExecutionTree":
+
+  * 'ExecState' is the current state of execution.
+     We start with 'InitialState', and keep performing steps until we get
+     to a 'ResultState'.
+  * As the simulator executes, it keeps track of its state in 'SimState'.
```
langston-barrett: It would be helpful to contrast `SimState` with `ExecState` here --- I believe that the former is (often) contained in the latter?

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+  * 'SimContext' is the part of the state that persists across branches
+    (e.g, after we explore the @then@ part of an @if2 statement, we'll
+    roll back some of the state changes before simulating the @else@ part,
+    but the stuff in `SimContexts` persists).  An important part of the
```
langston-barrett:
```suggestion
    but the data in `SimContext` persists).  An important part of the
```

langston-barrett (approved):
It might be nice to mention Crucible syntax extensions, but not a blocker. Minor comments inline but generally looks good to me. Thanks!

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+  
+{- $
+The state of a running simulator is described in "Lang.Crucible.Simulator.ExecutionTree":
+
+  * 'ExecState' is the current state of execution.
+     We start with 'InitialState', and keep performing steps until we get
+     to a 'ResultState'.
+  * As the simulator executes, it keeps track of its state in 'SimState'.
+  * 'SimContext' is the part of the state that persists across branches
+    (e.g, after we explore the @then@ part of an @if2 statement, we'll
```
samcowger:
```suggestion
    (e.g, after we explore the @then@ part of an @if@ statement, we'll
```

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+because in Haskell type families may not be partially applied but `newtypes` can.
+
+An important subset of the Crucible types are the base types (see 'BaseToType'),
+which is a backend specific type for the symbolic expression we can construct
+(see 'SymExpr' in [what4](https://github.com/Galoisinc/what4)).
+Only these types may contain variables. In practice, we always `what4`'s
+'Expr' type to represent symbolic expressions.
+
+Also, in some cases we use 'RegEntry' which
+is just a pair of a 'RegValue' and it's associated 'TypeRepr'.
```
samcowger:
```suggestion
is just a pair of a 'RegValue' and its associated 'TypeRepr'.
```

crucible/src/Lang/Crucible/README.hs
```diff
@@ -0,0 +1,112 @@
...
+To evaluate a 'CFG' we evaluate the statements as described in
+"Lang.Crucible.Simulator.EvalStmt" (details in 'stepStmt', 'stepTerm').
+Details about expressions evaluation are in 'evalApp' in "Lang.Crucible.Simulator.Evaluation".
+
+A lot of useful functionality relevant to the simulator can be accessed
+from module "Lang.Crucible.Simulator".
+
+The simulator supports mutable global variables.  Our tools use one such
+global to store a language specific memory model, which records information
+about various memory operations.
```
samcowger, line 102: Can you provide a code reference to an instance of this global, or of a memory model it contains?

langston-barrett: The global is created at runtime (it's a Crucible global variable, not a Haskell one). Might be good to provide a link to https://github.com/GaloisInc/crucible/blob/428e20cfffa28601caad4495b3b116cab22b3a7c/crucible-llvm/doc/memory-model.md

crucible/src/Lang/Crucible/README.hs
yav: The way haddock works, if you do this there's no documentation generated.  If we want it to be like that, we'd have to redo the file to move parts of it in the export list.  I don't think it's worth doing it, tbh.

crucible/src/Lang/Crucible/README.hs
yav: As far as I understand the type we use for the symbolic expressions is determined by that same backend.   Am I misunderstanding how things work?

crucible/src/Lang/Crucible/README.hs
yav: In the previous version I had the LLVM code as an example, but I am not sure how to link to if from here.  The problem is that it is in a different pacakge (`crucible-llvm`) and the dependency is backward (i.e., `curcible-llvm` depends on `crucible` but this note is in `crucible`).   I could put a link to the source file in the repo, but this has the drawback that the link will go stale if the code in `crucible-llvm` changes, and it'd be hard to spot. 

For reference, this is the variable in question https://github.com/GaloisInc/crucible/blob/428e20cfffa28601caad4495b3b116cab22b3a7c/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs#L537

yav: I think I'll add the link @langston-barrett suggested, it is useful info anyway, and it is likely to not change that much

crucible/src/Lang/Crucible/README.hs
langston-barrett: Gotcha, TIL!

crucible/src/Lang/Crucible/README.hs
langston-barrett: Ah, I guess the devil is in the details here.

The type of symbolic expressions is `SymExpr sym`. As you note, in practice this is always `SymExpr (ExprBuilder ...)`, i.e., `Expr`. So if you call `sym` a "backend", then in this sense the type of symbolic expressions is determined by the backend. I probably would *not* call `sym` a "backend" though. I would call it an expression builder. I would make that distinction because we have the `IsSymBackend` class, where the `bak` parameter is the "backend", not `sym`. `bak` is instantiated to things like `OnlineBackend`. Now, `IsSymBackend sym bak` *does* have a fundep `bak -> sym`. So that's another way you could say that the type of expressions is determined by the backend: `bak` determines `sym`, which determines `SymExpr`. But that's pretty indirect. And it's worth noting that both the simple and online backends do in fact fix the same `sym`, and so the same `SymExpr`, so in this sense the type of symbolic expressions really doesn't *meaningfully* depend on `bak`.

Now, the actual sorts of expressions you put in your `SymExpr`s probably *do* need to depend somewhat on which SMT solver you're using: you wouldn't want to make symbolic expressions that are not supported by the available theories. So in that sense, the expressions depend on the backend.

Does that help? Sorry for the dense comment!

langston-barrett: In any case, this isn't crucial and I would be fine with the current wording.

PR #1634, opened 2025-11-13
`crux-llvm-test`: Make tests pass with LLVMs up to version 20

RyanGlScott:
Now that we have landed several fixes needed to allow `crucible-llvm` to reason about LLVM versions up to 20, this updates the `crux-llvm` test suite to ensure that all of its test cases are able to pass with LLVM up to 20. Due to the all of the work we have already put in, only two test cases needed to be updated to be compatible with LLVM 19 and 20. See the individual commit messages for more details.

### DIFF
--- a/crux-llvm/README.md
+++ b/crux-llvm/README.md
@@ -50,7 +50,7 @@ We have tested `crux-llvm` most heavily with GHC 9.4, GHC 9.6, GHC 9.8,
 and `cabal` version 3.10. We recommend Yices 2.6.x, and Z3
 4.8.x. Technically, only one of Yices or Z3 is required, and CVC4 is
 also supported. However, in practice, having both tends to be
-convenient. Finally, LLVM versions from 3.6 through 19 are likely to
+convenient. Finally, LLVM versions from 3.6 through 20 are likely to
 work well, and any failures with versions in that range should be
 [reported as bugs](https://github.com/GaloisInc/crucible/issues).
 
--- a/crux-llvm/test-data/golden/T1549.ll
+++ b/crux-llvm/test-data/golden/T1549.ll
@@ -5,7 +5,20 @@ define i32 @main() {
   %sp = alloca %struct.S
 
   ; Write a %struct.S value to the pointer whose fields are all zeroes.
-  store %struct.S { i32 0, i32 0, [0 x i8] [] }, ptr %sp
+  ;
+  ; NB: Make sure _not_ to use the empty array value `[0 x i8] []` or
+  ; `[0 x i8] zeroinitializer` for the third field, because for whatever reason,
+  ; Clang 20+ normalizes these to `poison` values. This, in turn, will cause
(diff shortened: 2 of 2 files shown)


langston-barrett (approved):
I'm a bit surprised not to see a change to the CI configuration - do we just use whichever version of Clang+LLVM happens to be installed on the CI runner?

Test changes look fine. The minor discrepancies between LLVM versions are another indication that it might be helpful to look towards a solution like [Oughta](https://github.com/GaloisInc/oughta) at some point.

RyanGlScott:
> I'm a bit surprised not to see a change to the CI configuration - do we just use whichever version of Clang+LLVM happens to be installed on the CI runner?

Specifically, the CI tests LLVM 14, as seen [here](https://github.com/GaloisInc/crucible/blob/7d30b26b89b43d732534a519609fe86644fd3c57/.github/ci.sh#L54-L83). I agree that this is not thorough, although we haven't yet figured out what would be a good way to test more LLVM versions without blowing up the `crux-llvm` CI job times.

> The minor discrepancies between LLVM versions are another indication that it might be helpful to look towards a solution like [Oughta](https://github.com/GaloisInc/oughta) at some point.

Agreed.

langston-barrett:
> Specifically, the CI tests LLVM 14, as seen [here](https://github.com/GaloisInc/crucible/blob/7d30b26b89b43d732534a519609fe86644fd3c57/.github/ci.sh#L54-L83). I agree that this is not thorough, although we haven't yet figured out what would be a good way to test more LLVM versions without blowing up the crux-llvm CI job times.

Yeah, I see how testing *multiple* LLVM versions might be tough for sure. Is there a reason that we specifically test LLVM 14? If we choose to test only a single LLVM version, my instinct would be to prefer the latest supported version. At the very least, that would ensure that each LLVM version is tested in CI *at some point in time*. Moreover, it seems reasonable to expect that users will default to more recent versions. (On the other hand, we might expect users to default to versions that are available in major package repositories, like Ubuntu. I think it would be reasonable to tie our CI version to that, but we might want to make a note of that if so.)

I don't think that the versions of LLVM in Ubuntu's default repositories should limit us, LLVM distributes a script `llvm.sh` capable of installing a wide range of versions on a wide range of OSes.

RyanGlScott:
> Is there a reason that we specifically test LLVM 14?

No particular reason other than it is convenient to install on all of the OSes that CI currently supports. This matters to some degree because we currently have an Ubuntu 22.04 CI job, and Ubuntu 22.04's package manager doesn't have access to later versions of LLVM. I agree that this is not an entirely satisfying explanation, however, and it would be nicer to pick a more recent LLVM version now (even if we have to jump through some hoops to actually download a later version on certain OSes).

RyanGlScott:
I'm very late in acting on this, but I've opened https://github.com/GaloisInc/crucible/issues/1663 to track the task of moving the `crux-llvm` CI to a more recent LLVM version.

PR #1647, opened 2025-11-24
Vec arrays

yav:
This PR wraps the vectors used to implement Crucible vector values with a newtype, which should allow us to support more representations in the future (e.g., vectors backed by SMT arrays, which would be useful for uninterpreted functions returning vectors)

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/VecValue.hs
+++ b/crucible/src/Lang/Crucible/Simulator/VecValue.hs
@@ -19,6 +19,8 @@ module Lang.Crucible.Simulator.VecValue
   , vecValTake
   , vecValDrop
   , vecValAppend
+  , vecValMap
+  , vecValZipWith
 
     -- * Observers
   , vecValToVec
@@ -30,6 +32,7 @@ module Lang.Crucible.Simulator.VecValue
   , vecValGetEntryConcrete
   , vecValHead
   , vecValLast
+  , vecValFold
 
   
 
@@ -88,9 +91,10 @@ instance VecMonad sym IO where
   vecPre = pure . truePred
 
 -- | Type suitable for indexing into vectors, or getting the size of a vector.
--- We expect to have 2 instances for this class:
---  * one for natural numbers,
+-- Instances:
+--  * one for symbolic natural numbers,
 --  * one for bit-vector types (i.e., usize),
+-- * one for concrete natural numbers
 class IsExprBuilder sym => VecSize sym t where
   vecSizeIsConcrete :: sym -> t -> Maybe Int
   vecSizeLit        :: sym -> Natural -> IO t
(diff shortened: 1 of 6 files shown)


crucible/src/Lang/Crucible/Simulator/VecValue.hs
```diff
@@ -0,0 +1,492 @@
...
+import What4.Partial
+import What4.Expr(BVExpr)
+import qualified Data.BitVector.Sized as BV
+import Lang.Crucible.Types
+import Lang.Crucible.Backend
+import Lang.Crucible.Simulator.SimError
+
+-- | We expect this to have one instance: for `RegValue'`.
+-- We don't use that type directly to avoid recursive module dependencies.
+class IsExprBuilder sym => IsRegValue sym f | f -> sym where
```
RyanGlScott, line 72: Do you always expect this class to have exactly one instance? If so, my preference would be to move the definition of `RegValue'` to another module to avoid recursive module dependencies, rather than complicating all of the type signatures in this module with yet another type class.

crucible/src/Lang/Crucible/Simulator/VecValue.hs
```diff
@@ -0,0 +1,492 @@
...
+-- One for 'IO', used by the symbolic evaluator,
+-- and one for 'MuxLeafT' used by `crucible-mir`
+class MonadIO m => VecMonad sym m where
+
+  -- | We use this for partial operations (e.g., indexing a vector out of bounds).
+  vecAbort :: IsSymBackend sym bak => bak -> SimErrorReason -> m a
+
+  -- | A pre-condition on the current operation.  Assertions we generate
+  -- are conditional on this.
+  vecPre   :: IsExprBuilder sym => sym -> m (Pred sym)
```
RyanGlScott, line 87: For symmetry with the `VecSize` class, it might be worth factoring out the `IsExprBuilder sym` constraint into a superclass, i.e.,


```suggestion
class (IsExprBuilder sym, MonadIO m) => VecMonad sym m where

  -- | We use this for partial operations (e.g., indexing a vector out of bounds).
  vecAbort :: IsSymBackend sym bak => bak -> SimErrorReason -> m a

  -- | A pre-condition on the current operation.  Assertions we generate
  -- are conditional on this.
  vecPre   :: sym -> m (Pred sym)
```

crucible/src/Lang/Crucible/Simulator/VecValue.hs
```diff
@@ -0,0 +1,492 @@
...
+  vecSizeLeq sym x y = pure (backendPred sym (x <= y))
+
+
+
+--------------------------------------------------------------------------------
+
+
+
+data VecVal (f :: CrucibleType -> Type) (tp :: CrucibleType) where
+  VecVal :: V.Vector (f tp) -> VecVal f tp
```
RyanGlScott, line 146: The name of this module is "`VecValue`", but this data type is named "`VecVal`". For consistency, it would be nice to use the same name throughout. My personal preference would be "`VecValue`", as it more closely aligns with the names of other types in Crucible (e.g., `RegValue` and `AtomValue`).

langston-barrett:
@yav Are there still plans to work on this or should we close?

yav:
I think we can close it, I ended up doing things in a different way.

PR #1650, opened 2025-11-26
Pretty-print counterexample models in crux-mir (Fixes #1440)

### LINKED ISSUE #1440: `crux-mir`: Print counterexamples in a user friendly fashion
`crux-mir` currently prints counterexamples in JSON format:

```
$ cargo crux-test --lib --  -m midpoint_overflow
...
Model:
[{"name": "a","loc": null,"val": "0xa730e7fe","bits": "32"},{"name": "b","loc": null,"val": "0x58cf1803","bits": "32"}]
```

While we *could* pretty-print the JSON output, and get a result similar to this:

```
$ cargo crux-test --lib --  -m midpoint_overflow
...
Model:
[
    {
        "name": "a",
        "loc": null,
        "val": "0xa730e7fe",
        "bits": "32"
    },
    {
        "name": "b",
        "loc": null,
        "val": "0x58cf1803",
        "bits": "32"
    }
]
```

it is still a result that is difficult to grok. This issue should capture the desired counter-example format, and various pros and cons.
Here is a straw-man suggestion:

> As a Rust developer, I want to know the variable name and variable value. I don't necessarily care about the location, as that is captured in the error message above. Bitsize is also irrelevant, as the variable type carries that information.
> Here is a hypothetical output:

```
$[example-5] cargo crux-test --lib -- -m
...
---- example_4/b839819f::crux_test[0]::midpoint_overflow[0] counterexamples ----
[Crux]   src/lib.rs:5:12: 5:24: error: in example_4/b839819f::find_midpoint[0]
[Crux]   attempt to compute `copy _1 + copy _2`, which would overflow
Model:
a = -0x80000000 (signed), 0x80000000 (unsigned), -2147483648 (decimal)
b = -0x80000000 (signed), 0x80000000 (unsigned), -2147483648 (decimal)
```

sand7000:
Added a prettyprinter-based rendering pipeline in Crux.Model.

### DIFF
--- a/crux-mir/src/Mir/Language.hs
+++ b/crux-mir/src/Mir/Language.hs
@@ -436,7 +436,7 @@ runTestsWithExtraOverrides initS bindExtra (cruxOpts, mirOpts) = do
                logGoal gs
                when (showModel mirOpts) $ do
                   outputLn "Model:"
-                  outputLn (show (Crux.renderModel m))
+                  outputLn (show (Crux.prettyModel m))
 
     when anyFailed $ do
         outputLn ""
--- a/crux/src/Crux/Model.hs
+++ b/crux/src/Crux/Model.hs
@@ -1,13 +1,13 @@
 -- | This file is almost exactly the same as crucible-c/src/Model.hs
 
 {-# Language DataKinds #-}
+{-# Language OverloadedStrings #-}
 {-# Language PolyKinds #-}
 {-# Language Rank2Types #-}
 {-# Language TypeFamilies #-}
 {-# Language TypeApplications #-}
 {-# Language TypeOperators #-}
 {-# Language ScopedTypeVariables #-}
-{-# Language OverloadedStrings #-}
 
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
@@ -172,8 +172,8 @@ prettyBVEnt n e =
       ]
 
 -- Human-readable model as a Prettyprinter 'Doc'.
-renderModel :: ModelView -> Doc ann
-renderModel m =
+prettyModel :: ModelView -> Doc ann
+prettyModel m =
   PP.vsep
     (MapF.foldrWithKey
        (\ty vals docs -> prettyVals ty vals ++ docs)


sand7000:
With the changes made so far the output for example 5 is now:

```
---- example_5/87c02ed6::crux_test[0]::midpoint_overflow[0] counterexamples ----
[Crux]   src/lib.rs:5:12: 5:24: error: in example_5/87c02ed6::find_midpoint[0]
[Crux]   attempt to compute `copy _1 + copy _2`, which would overflow
Model:
a = -0x80000000 (signed), 0x80000000 (unsigned), -2147483648 (decimal)
b = -0x80000000 (signed), 0x80000000 (unsigned), -2147483648 (decimal)

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
error: test failed, to rerun pass `--lib`
```

RyanGlScott:
This is looking great, @sand7000! Let me know when this is ready for review.

sand7000:
I also tested how the new pretty-printer behaves with extremely large bitvector values to make sure it still looked nice. To do that, I added a small `u128` example that forces the solver to choose `u128::MAX`. This produces both the very large unsigned decimal and the two’s-complement signed interpretation.

Here is the output:

```
---- u128_max/d1f2bbdc::crux_test[0]::wide_u128_counterexample[0] counterexamples ----
[Crux]   MIR assertion at src/lib.rs:11:9:
[Crux]          x = 340282366920938463463374607431768211455
Model:
x = -0x1 (signed), 0xffffffffffffffffffffffffffffffff (unsigned), -1 (decimal)
```

which still looks good to me. If anyone has additional edge cases they’d like me to test I’m happy to.

crux/src/Crux/Model.hs
```diff
@@ -112,3 +116,73 @@ valsJS ty (Vals xs) =
...
+prettyBVEnt :: (1 <= w) => NatRepr w -> Entry (BV w) -> Doc ann
+prettyBVEnt n e =
+  let v   = entryValue e
+      sg  = showBVLiteralSigned   n v
+      un  = showBVLiteralUnsigned n v
+      dec = showBVLiteralDecimal  n v
+  in
+    PP.hsep
+      [ PP.pretty (entryName e)
+      , PP.pretty ("=" :: String)
```
RyanGlScott: If you enable `{-# LANGUAGE OverloadedStrings #-}` at the top of this file, then you can shorten this line to just:


```suggestion
      , "="
```

And similarly for other occurrences of `PP.pretty ("..." :: String)` in this file.

crux/src/Crux/Model.hs
```diff
@@ -112,3 +116,73 @@ valsJS ty (Vals xs) =
...
+       (\ty vals docs -> valsDoc ty vals ++ docs)
+       []
+       (modelVals m)
+    )
+
+-- Convenience: render a model to 'String' for CLI output.
+renderModel :: ModelView -> String
+renderModel m =
+  let layout = PP.layoutPretty PP.defaultLayoutOptions
+  in T.unpack (renderStrict (layout (modelDoc m)))
```
RyanGlScott: Note that you will get the same behavior if you simply define this function as `renderModel m = show (modelDoc m)`.

That being said, I propose that we instead give `renderModel` the type `ModelView -> Doc ann` (and then delete `modelDoc`, which would become redundant). The reason is that `Doc` is a richer type than `String`, and as a matter of API design, we would like to give users the flexibility to choose how to display the `Doc` in call sites. (For example, some users might want to embed the output of `renderModel` in another `Doc`, which is much easier to do if you have access to the original `Doc`.)

Since it's pretty straightforward to convert a `Doc` to a `String` (just call `show`), I don't think we need to expose an API function of type `ModelView -> String`.

crux/src/Crux/Model.hs
```diff
@@ -112,3 +116,73 @@ valsJS ty (Vals xs) =
...
+          | Just Refl <- testEquality eb (knownNat @11)
+          , Just Refl <- testEquality sb (knownNat @53)
+          -> prettyEnt' showDoubleLiteral
+
+        BaseRealRepr ->
+          -- same semantics as valsJS: print reals via toDouble
+          prettyEnt' (show . toDouble)
+
+        _ ->
+          error ("Type not implemented: " ++ show ty)
```
RyanGlScott, line 142: The cases here are very closely in sync with the cases in `valsJS`. I think that's fine, but if we update one of these functions in the future, then we should remember to update the other one at the same time. To make it more likely that we remember to do this in the future, I think it would be helpful to add comments above the cases in each function saying something akin to "Make sure to keep these cases in sync with those in <name of other function>".

crux/src/Crux/Model.hs
```diff
@@ -112,3 +116,73 @@ valsJS ty (Vals xs) =
 modelJS :: ModelView -> IO JS
 modelJS m =
   jsList . concat <$> sequence (MapF.foldrWithKey (\k v xs -> valsJS k v : xs) [] (modelVals m))
+
+-- Pretty-print all entries in a model for a given base type.
+valsDoc :: BaseTypeRepr ty -> Vals ty -> [Doc ann]
+valsDoc ty (Vals xs) =
```
RyanGlScott: A common convention for naming functions that perform pretty-printing is `pretty*` or `pp*`, so I would propose naming this function something like `prettyVals` instead. That would also be more in sync with the `prettyEnt'` and `prettyBVEnt` functions you've defined below.

crux-mir/src/Mir/Language.hs
```diff
@@ -437,8 +436,7 @@ runTestsWithExtraOverrides initS bindExtra (cruxOpts, mirOpts) = do
                logGoal gs
                when (showModel mirOpts) $ do
                   outputLn "Model:"
-                  mjs <- Crux.modelJS m
-                  outputLn (Crux.renderJS mjs)
+                  outputLn (Crux.renderModel m)
```
RyanGlScott: Now that you've changed how `crux-mir`'s output looks, make sure to also update this part of the `crux-mir` README that references how the output should look:

https://github.com/GaloisInc/crucible/blob/36881ffe9698522d90a86b49d084c076dadd58e3/crux-mir/README.md#L297-L298

crux/src/Crux/Model.hs
sand7000: Much cleaner, fixed.

crux/src/Crux/Model.hs
sand7000: Done.

crux/src/Crux/Model.hs
sand7000: Comments added in both.

crux/src/Crux/Model.hs
sand7000: Renamed.

crux-mir/src/Mir/Language.hs
sand7000: Done.

crux/src/Crux/Model.hs
sauclovian-g: In SAW we're slowly undertaking a bulk cleanup such that all the functions producing `Doc` are called `pretty` and all the functions that produce strings are called `pp`. I would kind of suggest adopting the same convention in Crucible, since this way around matches the prettyprinter library's naming.

FWIW

crux/src/Crux/Model.hs
```diff
@@ -7,6 +7,7 @@
 {-# Language TypeApplications #-}
 {-# Language TypeOperators #-}
 {-# Language ScopedTypeVariables #-}
+{-# Language OverloadedStrings #-}
```
RyanGlScott: Keeping in line with the existing convention in this module, could you put this language pragma in alphabetical order?

RyanGlScott (approved):
LGTM, aside from David's suggestion and one minor quibble.

crux/src/Crux/Model.hs
sand7000: Done.

crux/src/Crux/Model.hs
sand7000: Sounds good, thanks. I renamed `renderModel` to `prettyModel` so the new code follows your convention.

PR #1655, opened 2025-12-02
`crucible-mir`: Simplify `VarInfo` and reduce it to a single data constructor

### LINKED ISSUE #1652: `crucible-mir`: Simplify `VarInfo`
`crucible-mir` tracks the origin of each MIR variable using the `VarInfo` data type, which is defined as:

https://github.com/GaloisInc/crucible/blob/d516dcfdb55118a1f7804a67fe03fff56d3cebb7/crucible-mir/src/Mir/Generator.hs#L251-L254

In practice, `VarReference` is by far the most commonly used of the three data constructors. In fact, I claim that the other two data constructors (`VarRegister` and `VarInfo`) are vestigial:

* `VarRegister` doesn't appear to be constructed _anywhere_, save for [this unreachable code](https://github.com/GaloisInc/crucible/blob/d516dcfdb55118a1f7804a67fe03fff56d3cebb7/crucible-mir/src/Mir/Trans.hs#L2165).
* A comment [here](https://github.com/GaloisInc/crucible/blob/d516dcfdb55118a1f7804a67fe03fff56d3cebb7/crucible-mir/src/Mir/Generator.hs#L248-L249) suggests that `VarAtom`s are used for function arguments, that doesn't appear to be the case in practice. The only use site for `VarAtom`s in practice is [here](https://github.com/GaloisInc/crucible/blob/d516dcfdb55118a1f7804a67fe03fff56d3cebb7/crucible-mir/src/Mir/Generator.hs#L582), which in turn is only used when constructing tuple clone shims.

I propose that we remove `VarRegister` and `VarInfo` entirely and use `VarReference` for all variables. Doing so would allow us to greatly simplify the implementation of `VarInfo`.

RyanGlScott:
This greatly simplifies the definition of use sites of the `VarInfo` data type. In particular, this eliminates the mostly unused `VarRegister` and `VarAtom` data constructors, reducing `VarInfo` to just a single data constructor. This also performs some light cleanup of what remains of `VarInfo` (for instance, redefining it as a record data type for simplicity's sake).

Fixes #1652.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -2146,7 +2146,7 @@ initLocals localVars = forM_ localVars $ \v -> do
 cleanupLocals :: MirGenerator h s ret ()
 cleanupLocals = do
     vm <- use varMap
-    forM_ (Map.elems vm) $ \(Some vi) ->
+    forM_ vm $ \(Some vi) ->
         G.readReg (varInfoReg vi) >>= dropMirRef
 
 buildLabelMap :: forall h s ret. M.MirBody -> MirGenerator h s ret (LabelMap s)


crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -2175,8 +2175,7 @@ cloneShimTuple tys parts = CustomMirOp $ \ops -> do
     -- the individual parts require `&A`, `&B`, `&C`, computed as `&arg.0`.
     let fieldRefRvs = zipWith (\ty i ->
             Ref Shared (LProj (LProj lv Deref) (PField i ty)) "_") tys [0..]
-    fieldRefExps <- mapM evalRval fieldRefRvs
-    fieldRefOps <- zipWithM (\ty expr -> makeTempOperand (TyRef ty Immut) expr) tys fieldRefExps
+    let fieldRefOps = map Temp fieldRefRvs
     clonedExps <- zipWithM (\part op -> callExp part [op]) parts fieldRefOps
```
RyanGlScott: Looking at this again, it might be simpler just to fuse this call to `map` with the subsequent call to `zipWithM`:


```suggestion
    clonedExps <- zipWithM (\part rv -> callExp part [Temp rv]) parts fieldRefRvs
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2148,31 +2134,20 @@ initLocals localVars = forM_ localVars $ \v -> do
...
     varMap %= Map.insert name varinfo
 
 -- | Deallocate RefCells for all locals in `varMap`.
 cleanupLocals :: MirGenerator h s ret ()
 cleanupLocals = do
     vm <- use varMap
-    forM_ (Map.elems vm) $ \(Some vi) -> case vi of
-        VarReference _ reg -> G.readReg reg >>= dropMirRef
-        _ -> return ()
+    forM_ (Map.elems vm) $ \(Some vi) ->
```
samcowger: This might avoid evincing an intermediate list:
```suggestion
    forM_ vm $ \(Some vi) ->
```

crucible-mir/src/Mir/TransCustom.hs
samcowger: Agreed re fusion.

Do you know why we went to the trouble of applying `evalRval` to `fieldRefRvs` before, and/or why we no longer need to now?

RyanGlScott: Honestly, I have no idea why the old code went with the approach of evaluating the rvalues to `MirExp`s and then binding the results to temporary variables (via Crucible `Atom`s). It seems far more direct just to pass the rvalues directly to `clone` methods (without introducing temporary variables) as is done here.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -2175,8 +2175,7 @@ cloneShimTuple tys parts = CustomMirOp $ \ops -> do
     -- the individual parts require `&A`, `&B`, `&C`, computed as `&arg.0`.
     let fieldRefRvs = zipWith (\ty i ->
             Ref Shared (LProj (LProj lv Deref) (PField i ty)) "_") tys [0..]
-    fieldRefExps <- mapM evalRval fieldRefRvs
-    fieldRefOps <- zipWithM (\ty expr -> makeTempOperand (TyRef ty Immut) expr) tys fieldRefExps
+    let fieldRefOps = map Temp fieldRefRvs
```
spernsteiner: I think this use of `Temp` will fail if the callee is another tuple clone shim, since `cloneShimTuple` requires `Move` or `Copy` above.  Can you add a test that does something like `((1, 2), 3).clone()`?  In that case I'd expect the clone shim for `((i32, i32), i32)` to call the one for `(i32, i32)`.

RyanGlScott: Ouch, good catch. Here is a failing test case:

```rs
// test.rs
pub fn dup<C: Clone>(x: C) -> (C, C) {
    (x.clone(), x)
}

#[crux::test]
pub fn f() -> ((u32, u32), u32) {
    dup(((1, 2), 3)).0
}
```
```
test test/5769e644::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/5769e644::f[0] counterexamples ----
[Crux]   test.rs:3:6: 3:15: error: in test/5769e644::dup[0]::_instd79c0ac80ac695e5[0]
[Crux]   Translation error in test/5769e644::dup[0]::_instd79c0ac80ac695e5[0]: cloneShimTuple: expected lvalue operand, but got Temp (Ref {_rbk = Shared, _rvar = LProj (LProj (LBase (Var {_varname = "_3", _varmut = Mut, _varty = TyRef (TyTuple [TyTuple [TyUint B32,TyUint B32],TyUint B32]) Immut, _varIsZST = False})) Deref) (PField 0 (TyTuple [TyUint B32,TyUint B32])), _rregion = "_"})

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

RyanGlScott: What's even scarier is that there actually _is_ a test case in `crux-mir`'s test suite that claims to test this property:

https://github.com/GaloisInc/crucible/blob/191fc0a0fd3c902cfc9734846e43499767aba8fc/crux-mir/test/conc_eval/tuple/clone_rec.rs#L2-L7

Why didn't CI turn this up? It's because `rustc` is [optimizing away the calls to the clone shims](https://play.rust-lang.org/?version=nightly&mode=debug&edition=2024&gist=5425411b0a787dc575dea59a4545a86c)! Thankfully, using the `dup` trick above suffices to preserve the call to `clone`, so I'll update all of the test cases that rely on the presence of `clone` to use `dup` instead.

RyanGlScott: I've pushed a new version of this PR that fixes the issue by adding a special case for `Temp (Ref _ lv _)` alongside the existing cases for `Move` and `Copy`. It's a bit ad hoc, but then again, the code already expects the input operand to be of a certain shape anyway, so this is arguably no less ad hoc than it was before.

PR #1662, opened 2025-12-03
`{crucible,crux}-mir`: Fully remove `MirVector` in favor of `MirAggregate`

### LINKED ISSUE #1648: `{crucible,crux}-mir`: Fully remove `MirVector` in favor of `MirAggregate`
https://github.com/GaloisInc/crucible/pull/1580 started the process of switching out the internal representation of MIR arrays from the `MirVector` intrinsic type to the `MirAggregate` intrinsic type. This process is very nearly complete, and I was tempted to remove `MirVector` entirely. To wit, that would entail removing the following:

* [`MirVectorSymbol`, `MirVectorType`, `MirVectorRepr`, `MirVectorFam`, and `MirVector`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/Intrinsics.hs#L1143-L1170)
* [`Index_RefPath`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/Intrinsics.hs#L371-L375), [`VectorAsMirVector_RefPath`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/Intrinsics.hs#L380-L384), and [`ArrayAsMirVector_RefPath`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/Intrinsics.hs#L385-L389) (whose types all depend on `MirVector`)
* [`MirRef_VectorAsMirVector` and `MirRef_ArrayAsMirVector`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/Intrinsics.hs#L1431-L1438) (for similar reasons)
* [`MirVector_Uninit`, `MirVector_FromVector`, `MirVector_FromArray`, and `MirVector_Resize`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/Intrinsics.hs#L1517-L1533) (for similar reasons)

Unfortunately, removing these is not simply a matter of removing dead code, as there are a handful of code paths that still rely on these vestigial types. Namely:

* `crucible-mir`'s custom overrides for the following functions rely on `MirVector`-based operations:
  * [`vector_as_slice` and `vector_as_mut_slice`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/TransCustom.hs#L328-L351)
  * [`array_as_slice` and `array_as_mut_slice`](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crucible-mir/src/Mir/TransCustom.hs#L426-L447)
* `crux-mir`'s `regEval` function concretizes slices by [creating a `MirVector_Vector` value](https://github.com/GaloisInc/crucible/blob/c76f025200342aef5e850883b372cac8c160d073/crux-mir/src/Mir/Overrides.hs#L229-L230).

In order to complete the process of removing `MirVector` in favor of `MirAggregate`, we will need to migrate the code paths above to use `MirAggregate`-based equivalents. See https://github.com/GaloisInc/crucible/pull/1580 for inspiration on how this might be accomplished.

RyanGlScott:
This completes the migration from `MirVector` to `MirAggregate` that was started in #1580. For the most part, this simply entails removing dead code. There were a handful of places in the code that were still using `MirVector` operations, however, and these needed to be migrated:

* `crux-mir`'s `regEval` function needed to replace a use of `MirVector_Vector` with an equivalent `MirAggregate` expression. This proved fairly straightforward.

* `crucible-mir`'s custom overrides for `vector_as_slice{,_mut}` and `array_as_slice{,_mut}` needed to be reimplemented from scratch using custom operations for indexing into `Vector`s and `Array`s.

Fixes #1648.

### DIFF
--- a/crucible-mir/src/Mir/TransTy.hs
+++ b/crucible-mir/src/Mir/TransTy.hs
@@ -1522,14 +1522,23 @@ initialValue (M.TyAdt nm _ _) = do
             -- for details, including some regarding this choice of size.
             let unionSize = 1
             in Just . MirExp MirAggregateRepr <$> mirAggregate_uninit_constSize unionSize
-
-
-
-initialValue (M.TyFnPtr _) = return $ Nothing
 initialValue (M.TyFnDef _) = return $ Just $ MirExp C.UnitRepr $ R.App E.EmptyApp
 initialValue M.TyNever     = return $ Just $ MirExp C.UnitRepr $ R.App E.EmptyApp
-initialValue (M.TyDynamic _) = return $ Nothing
-initialValue _ = return Nothing
+
+-- Remaining `Nothing` cases
+initialValue (M.TyRef {}) = return Nothing
+initialValue (M.TyRawPtr {}) = return Nothing
+initialValue (M.TyFnPtr {}) = return Nothing
+initialValue (M.TyDynamic {}) = return Nothing
+initialValue (M.TySlice {}) = return Nothing
+initialValue (M.TyStr {}) = return Nothing
+initialValue (M.TyFloat {}) = return Nothing
+initialValue (M.TyDowncast {}) = return Nothing
+initialValue (M.TyForeign {}) = return Nothing
+initialValue (M.TyConst {}) = return Nothing
+initialValue (M.TyLifetime {}) = return Nothing
+initialValue (M.TyErased {}) = return Nothing
+initialValue (M.TyInterned {}) = return Nothing
 
 initField :: M.Field -> MirGenerator h s ret (Maybe (MirExp s))
 initField (M.Field _name ty) = initialValue ty


crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2447,14 +2129,20 @@ refPathOverlaps sym path1 path2 = do
         -- test case for an example).
         pEq <- go rrp1 rrp2
         liftIO $ andPred sym overlaps pEq
+    go (VectorAsMirAggregate_RefPath tpr1 _ `RrpCons` rrp1)
+        (VectorAsMirAggregate_RefPath tpr2 _ `RrpCons` rrp2)
+      | Just Refl <- testEquality tpr1 tpr2 = go rrp1 rrp2
+    -- TODO RGS: Is ignoring the length here OK?
+    go (ArrayAsMirAggregate_RefPath _len1 tpr1 _ `RrpCons` rrp1)
+        (ArrayAsMirAggregate_RefPath _len2 tpr2 _ `RrpCons` rrp2)
+      | Just Refl <- testEquality tpr1 tpr2 = go rrp1 rrp2
```
RyanGlScott: Not sure about this one

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -3076,21 +2717,94 @@ readRefPath bak iTypes v = \case
...
+  let sym = backendGetSym bak
+  len <- concreteAllocSize bak lenSym
+  -- TODO: hardcoded size=1
+  let sz = 1
+  liftIO $ foldM
+    (\arr' idx -> do
+      idxSym <- bvLit sym knownNat $ BV.mkBV knownNat idx
+      v <- mirAggregate_getIO bak (fromInteger @Word idx) sz (baseToType btp) ag
+      arrayUpdate sym arr' (Empty :> idxSym) v)
+    arr [0..len-1]
```
RyanGlScott: These four functions (`vectorToMirAggregate`, `mirAggregateToVector`, `arrayToMirAggregate`, and `mirAggregateUpdateArray`) are by far the most questionable part of this patch. In fact, I'm pretty sure I've gotten the details of at least one of `arrayToMirAggregate` or `mirAggregateUpdateArray` wrong, since the `test/symb_eval/array/slice.rs` test case currently fails with:

```
$ cabal -v0 run exe:crux-mir -- test/symb_eval/array/slice.rs
test slice/a2cb15a6::crux_test[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- slice/a2cb15a6::crux_test[0] counterexamples ----
[Crux]   ./libs/core/src/clone.rs:581:25: 581:30 !./libs/core/src/clone.rs:588:5: 593:6: error: in core/8f9ca59b::clone[0]::impls[0]::{impl#14}[0]::clone[0]
[Crux]   no value at offset 10, in aggregate of size 10
[Crux]   ./libs/core/src/clone.rs:581:25: 581:30 !./libs/core/src/clone.rs:588:5: 593:6: error: in core/8f9ca59b::clone[0]::impls[0]::{impl#14}[0]::clone[0]

[Crux] Goal status:
[Crux]   Total: 2
[Crux]   Proved: 0
[Crux]   Disproved: 2
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -395,6 +380,18 @@ data MirReferencePath sym :: CrucibleType -> CrucibleType -> Type where
...
     MirReferencePath sym tp_base tp
+  -- | Present `&mut Vector` as `&mut MirAggregate`.
+  VectorAsMirAggregate_RefPath ::
+    !(TypeRepr tp) ->
+    !(MirReferencePath sym tp_base (VectorType tp)) ->
+    MirReferencePath sym tp_base MirAggregateType
+  -- | Present `&mut Array` as `&mut MirAggregate`.
+  ArrayAsMirAggregate_RefPath ::
+    -- | Size of array in elements (must be concrete)
+    !(RegValue sym UsizeType) ->
```
RyanGlScott: See `vectorToMirAggregate` and `mirAggregateToVector` below for the place where this must-be-concrete restriction is needed (when calling `concreteAllocSize`). Strictly speaking, I don't believe that this restriction was required before, although there do not appear to be any test cases that use a symbolic `Array` size.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Looking at the test, nothing should be accessing offset 10.  I wonder if `arr.as_slice(0, 10)` is actually returning a slice with length 11?

RyanGlScott: That seems plausible, but I am struggling to imagine how a slice of length 11 is being created. I can only think of a handful of places where we are specifying the lengths of things:

* In the `as_slice` custom override:

  https://github.com/GaloisInc/crucible/blob/d222dd65b77c7db9a51946f420fc26cbb6a2cbe2/crucible-mir/src/Mir/TransCustom.hs#L441

  Here, `len` is just `10`, so I doubt that this is the culprit.
* In the implementations of `arrayToMirAggregate`:

  https://github.com/GaloisInc/crucible/blob/d222dd65b77c7db9a51946f420fc26cbb6a2cbe2/crucible-mir/src/Mir/Intrinsics.hs#L2774-L2785

  And `mirAggregateUpdateArray`:

  https://github.com/GaloisInc/crucible/blob/d222dd65b77c7db9a51946f420fc26cbb6a2cbe2/crucible-mir/src/Mir/Intrinsics.hs#L2797-L2807

  The former creates a `MirAggregate` of length `len` (again, that should just be 10), and both functions retrieve elements of a `MirAggregate` from indexes `0` to `len-1` (i.e., `0` to `9`). That sounds fine at a first glance...

I am likely misunderstanding how something works internally. More investigation is required.

spernsteiner: I agree, all of that sounds fine.  Another way to get an access at offset 10 is for `len - 1` to underflow, but `len` shouldn't be zero in this test case.

Do you know which line of the test is causing the bad access?

RyanGlScott: It's specifically this line:

https://github.com/GaloisInc/crucible/blob/4d88a65b10c3b6bcee9bdc3251611360346bdd35/crux-mir/test/symb_eval/array/slice.rs#L13

The other two lines before it run without issues:

https://github.com/GaloisInc/crucible/blob/4d88a65b10c3b6bcee9bdc3251611360346bdd35/crux-mir/test/symb_eval/array/slice.rs#L11-L12

spernsteiner: Oh, `as_slice` takes a `start` and `len`, not `start` and `end`, so `arr.as_slice(5, 10)` covers indices 5..15

RyanGlScott: Does it? That would suggest that `arr.as_slice(5, 10).iter().cloned().sum::<i32>()` should be equal to `5 + 6 + 7 + 8 + 9 + 10 + 11 + 12 + 13 + 14` (`95`), rather than `5 + 6 + 7 + 8 + 9` (`35`, or `45 - 10`) as the `crucible_assert!` seems to indicate.

spernsteiner: That's how it works in the `array_as_slice_impl` override:

https://github.com/GaloisInc/crucible/blob/d222dd65b77c7db9a51946f420fc26cbb6a2cbe2/crucible-mir/src/Mir/TransCustom.hs#L433

And the library code in mir-json calls the arguments `start` and `len`.

In the test, the loop only writes indices 0..10, so 10..15 are unset and must be defaulting to zero for some reason.  (I would have expected them to be unconstrained, causing the assert to fail.)

RyanGlScott: Oh, I think I misunderstood how `iter()` interacts with `start`. I had assumed that `iter()` would know that the "starting index" is equal to `start` and would therefore only count up to index `len - 1`. But that isn't actually how it works—`iter()` doesn't record a starting index at all, so it simply iterates `len` times, without ever checking that the pointer offset eventually exceeds `len - 1`. A bit surprising, but that is likely what is going on.

> 10..15 are unset and must be defaulting to zero for some reason. (I would have expected them to be unconstrained, causing the assert to fail.)

This is because the `Array` was created with [`Array::zeroed`](https://github.com/GaloisInc/crucible/blob/4d88a65b10c3b6bcee9bdc3251611360346bdd35/crux-mir/test/symb_eval/array/slice.rs#L7). It would only be unconstrained if we did something like:

```diff
diff --git a/crux-mir/test/symb_eval/array/slice.rs b/crux-mir/test/symb_eval/array/slice.rs
index 728665c2a..8637b0d9c 100644
--- a/crux-mir/test/symb_eval/array/slice.rs
+++ b/crux-mir/test/symb_eval/array/slice.rs
@@ -4,7 +4,7 @@ use crucible::*;

 #[crux::test]
 fn crux_test() -> i32 {
-    let mut arr = Array::<i32>::zeroed();
+    let mut arr = Array::<i32>::symbolic("arr");
     for i in 0..10 {
         arr = arr.update(i, i as i32);
     }
```

RyanGlScott: Knowing this, I suppose we need to change the implementation of `arrayToMirAggregate` or `mirAggregateUpdateArray` to track the value of `start` and initialize the `MirAggregate` using the first `len + start` elements of the SMT array, not the first `len` elements. (I think?)

RyanGlScott: In fact, I suppose there is a broader question of just how large the `MirAggregate` backing a call to `array.as_slice(start, len)` should even be. In theory, it's possible that someone could use pointer arithmetic to update the pointer underlying the returned slice to a different offset. If you then attempt to read `len` elements from that new offset, should it be expected to succeed?

I suspect that with the old way `array.as_slice(...)` was implemented, the resulting memory loads would always succeed, since loading from the slice would always defer directly to an SMT array (and loading from an SMT array always succeeds). With the new way that `array.as_slice(...)` is implemented, you instead have to load from an intermediate `MirAggregate`, and because `MirAggregate`s cannot have infinite size, there will always be some offset at which loading from the `MirAggregate` will fail.

I think that we'll likely need to modify `array.as_slice(start, len)` a bit to solve this problem. Specifically, I propose that:

* `array.as_slice(start, len)` should be backed by a `MirAggregate` with `start + len` elements.
* The first `start` elements of the `MirAggregate` will be uninitialized, and the subsequent `len` elements will be initialized with the corresponding elements from the SMT array.
* Attempting to read any elements outside of the range `[start, start + len)` will result in a memory error. As such, anyone who attempts to do pointer arithmetic on the slice's underlying pointer must take care. (We may want to update the doc comments for `as_slice` to state this explicitly as a precaution.) 

Does this sound reasonable?

spernsteiner: Based on this discussion, I'm wondering whether `ArrayAsMirAggregate_RefPath` is actually the right approach.

Previously, we would start with a reference to `UsizeArrayRepr tpr` and apply `ArrayAsMirVector_RefPath` and `Index_RefPath` to go from `UsizeArrayRepr tpr` -> `MirVectorRepr tpr` -> `tpr`.  Pointer arithmetic would adjust the index inside the final `Index_RefPath`.  Instead of changing the intermediate type from `MirVectorRepr tpr` to `MirAggregateRepr`, we could remove that step completely, and instead have separate `ArrayIndex_RefPath` and `VectorIndex_RefPath` that go directly from `UsizeArrayRepr tpr` / `VectorRepr tpr` to `tpr`.  The implementation of pointer arithmetic in `mirRef_offsetWrapLeaf` already has separate cases for `Index_RefPath` and `AgElem_RefPath`, so we could add new cases for these new path elements.

One advantage of this approach (which I think also applied to the previous `MirVectorRepr` version) is it doesn't require us to convert the array to a finite representation and back.  This means unsafe pointer operations that go outside the bounds of the slice would work.

RyanGlScott: Great idea! I've pushed a new version of this PR which implements this idea. This did turn up a couple of additional questions whose answers I'm not sure of, but I'll post these as separate review threads.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1519,6 +1519,8 @@ initialValue (M.TyArray _ size) = do
 --
 -- NB: When re-enabling this, also re-enable the TyRef case of `canInitialize`
 {-
+-- TODO RGS: The code below mentions MirVectorRepr repeatedly, which has now
+-- been removed. What should we do about this?
```
RyanGlScott: I'd be interested to hear your thoughts about this chunk of commented-out `initialValue` code, now that it has thoroughly bitrotted after the changes in this PR.

crucible-mir/src/Mir/TransCustom.hs
RyanGlScott, line 1356: I'm not entirely sure about this one, as I'm not sure what the relationship between `MirAggregate` flattening and using `Vector`s/`Array`s in the input slice's data pointer would be. (Perhaps I'm misunderstanding?)

spernsteiner: The current plan is to leave vectors/arrays unflattened, so there would be a single aggregate entry containing the entire vector/array.  This is similar to the way we intend to handle enums.  So the issue described in the comment would still apply, despite the `MirVector` -> `Vector`/`Array` split.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: These cases have been commented out for over 5 years, across multiple toolchain upgrades.  We should probably just delete them.  In theory we could uncomment them now that `MirReferenceRepr` supports the required mux operation, but the only benefit of that is avoiding a few `MaybeRepr` wrappers, which is not that big a win.

RyanGlScott:
SAW passes with these changes (see https://github.com/GaloisInc/saw-script/pull/2867), so I think this is ready for review.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2616,28 +2350,31 @@ mirRef_peelIndexIO bak iTypes (MirReferenceMux ref) =
...
-        MirVector_PartialVector pv -> return $ V.length pv
-        MirVector_Array _ -> leafAbort $ Unsupported callStack $
-            "can't compute allocation length for MirVector_Array, which is unbounded"
-    len <- liftIO $ bvLit sym knownNat $ BV.mkBV knownNat $ fromIntegral lenInt
+    let lenInteger = toInteger $ V.length parentVec
+    len <- liftIO $ bvLit sym knownNat $ BV.mkBV knownNat lenInteger
     return (idx, len)
+mirRef_indexAndLenLeaf _bak _gs _iTypes (MirReference _tpr _root (ArrayIndex_RefPath {})) =
+    leafAbort $ Unsupported callStack
+        "can't compute allocation length for MirVector_Array, which is unbounded"
```
samcowger:
```suggestion
        "can't compute allocation length for Array, which is unbounded"
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2616,28 +2350,31 @@ mirRef_peelIndexIO bak iTypes (MirReferenceMux ref) =
...
--- Unlike `peelIndex`, this operation is valid on non-`Index_RefPath`
--- references (on which it returns `(0, 1)`) and also on `MirReference_Integer`
--- (returning `(0, 0)`).
+-- Note that unlike `peelIndex`:
+--
+-- * This operation is *not* valid on `ArrayIndex_RefPath`s, as SMT arrays do
+--   not have a finite length.
+--
+-- * This operation is valid other `MirReference` references (on which it
+--   returns `(0, 1)`) and also on `MirReference_Integer` (returning `(0, 0)`).
```
samcowger: While you're here:
```suggestion
--   returns @(0, 1)@) and also on `MirReference_Integer` (returning @(0, 0)@).
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2616,28 +2350,31 @@ mirRef_peelIndexIO bak iTypes (MirReferenceMux ref) =
 -- the length of that allocation.  This is useful for determining the amount of
 -- memory accessible through all valid offsets of `ref`.
 --
--- Unlike `peelIndex`, this operation is valid on non-`Index_RefPath`
--- references (on which it returns `(0, 1)`) and also on `MirReference_Integer`
--- (returning `(0, 0)`).
+-- Note that unlike `peelIndex`:
+--
+-- * This operation is *not* valid on `ArrayIndex_RefPath`s, as SMT arrays do
```
samcowger: If you hoped to italicize `not` here, you'd need to say `/not/`.

Issue #1369, opened 2025-04-10
Support `async fn` in crux-mir


podhrmic:
Currently `async fn f()` construct is not supported in crux-mir. It would be great to add this capability, as neither verus nor kani currently supports it (kani has an *experimental* support, but as of 0.61 it doesn't really work). Not an urgent need, but having this within a few months would be useful.

From @spernsteiner 

> it would probably be feasible to add some support, enough that you could test an async function with block_on.  This would be subject to the usual caveats that you can't actually do I/O inside the test, and anything with tricky data-dependent control flow is likely to confuse the solver.

podhrmic:
An example I was starting from was https://tokio.rs/tokio/tutorial/hello-tokio

RyanGlScott:
See https://github.com/GaloisInc/mir-json/issues/199#issuecomment-3608386682 for a discussion how how MIR represents the internals of coroutines, as well as a sketch of how what their Crucible type representation might be.

RyanGlScott:
As of https://github.com/GaloisInc/crucible/pull/1686, `crux-mir` now has rudimentary support for simulating `async` functions. As such, I think this issue has outlived its usefulness, so I will opt to close it.

There is more work to be done in supporting the types of code patterns commonly seen in `async`-heavy libraries such as `tokio`, however. For instance, `tokio` frequently uses unsafe code that is difficult for `crucible-mir` to model at the moment. This will likely be improved when `MirAggregate` flattening is implemented, so we will track that task in https://github.com/GaloisInc/crucible/issues/1499.

Issue #1668, opened 2025-12-09
`crux-mir-test`: Invoke `rustc` with a newer language edition by default


RyanGlScott:
Currently, `crux-mir` invokes `mir-json` with `--edition=2021` by default, as seen [here](https://github.com/GaloisInc/crucible/blob/0925e99635451f06f24c1063ea2c21f965111b50/crux-mir/src/Mir/Generate.hs#L84). The `crux-mir` test suite is not in sync with this choice, however. When the `crux-mir` test suite invokes `rustc` on a test file to check if the normal output of a Rust binary matches the output of `crux-mir`, it omits an explicit language edition, as seen here:

https://github.com/GaloisInc/crucible/blob/0925e99635451f06f24c1063ea2c21f965111b50/crux-mir/test/Test.hs#L251-L254

As a result, this `rustc` invocation will behave as though you had selected `--edition=2015`. For most test cases, this is fine, but this poses a problem if you need to add a test case that relies on Rust features that were introduced in a newer language edition. For instance, `async` support is only available in the 2018 edition of the language or later (see https://github.com/GaloisInc/crucible/issues/1369), so as things stand, it is effectively impossible to add a test case that relies on post-2015 language features.

At a minimum, we should ensure that the default language editions used in both the `mir-json` and `rustc` invocations above are the same. Namely, we should make the `rustc` invocation in the test suite use `--edition=2021` as well. We might also consider making both use `--edition=2024` by default. This would not be a suitable choice for all use cases, but there are other issues open about making it possible to override these defaults:

* Overriding the options passed to `mir-json`: https://github.com/GaloisInc/crucible/issues/1556
* Overriding the options passed to `rustc` in the test suite: https://github.com/GaloisInc/crucible/issues/1527

PR #1669, opened 2025-12-09
`crucible-mir`: Limited support for translating coroutines

### LINKED ISSUE #1626: `crucible-mir`: Emit a proper translation error when simulating coroutines
Simulating the following program with `crux-mir`:

```rs
#![feature(coroutines, coroutine_trait, stmt_expr_attributes)]

use std::ops::Coroutine;
use std::pin::Pin;

#[crux::test]
pub fn f() {
    let mut coroutine = #[coroutine] || {
        yield 1;
        return "foo"
    };

    let p = Pin::new(&mut coroutine);
    p.resume(());
}
```

Yields the following, terrible error message:

```
user error (JSON decoding of test.all.mir failed: Error in $.fns[0].body.blocks[0].block.data[0].rhs.akind: unimp: Coroutine)
```

At a minimum, we should emit a translation error here instead of a confusing JSON parsing-related error message. This would also allow `crux-mir` to simulate more programs where coroutines are used, but not directly from the function being used as a simulation entrypoint.

RyanGlScott:
Although we stop short of full support for `async`-related coroutine code (see issue #1369 for that task), `crucible-mir` now has the ability to translate coroutine closure expression and a limited ability to translate coroutine expressions. This means that we can translate code involving coroutines without `crucible-mir` crashing outright, instead giving a nicer `Coroutines not currently supported` error message in the event that a user tries to simulate a coroutine expression.

This bumps the `mir-json` submodule to bring in the corresponding changes from https://github.com/GaloisInc/mir-json/pull/205. As a consequence, this requires bumping the supported MIR JSON schema version.

Fixes #1626.

### DIFF
--- a/crucible-mir/src/Mir/TransTy.hs
+++ b/crucible-mir/src/Mir/TransTy.hs
@@ -1551,6 +1551,7 @@ initialValue (M.TyInterned {}) = return Nothing
 initialTupleValue ::
   HasCallStack => [M.Ty] -> MirGenerator h s ret (Maybe (MirExp s))
 initialTupleValue tys =
+  -- TODO: hardcoded size=1
   Just . MirExp MirAggregateRepr <$> mirAggregate_uninit_constSize (fromIntegral $ length tys)
 
 initField :: M.Field -> MirGenerator h s ret (Maybe (MirExp s))


RyanGlScott:
Note that this does not add any test cases. This is for an unfortunate reason: any test case that would exercise the new functionality would require Rust language features that are not available in `--edition=2015`, which is what the test suite uses when compiling `conc_eval` test cases with `rustc`. See https://github.com/GaloisInc/crucible/issues/1668. We would need to address that issue first if we wanted to test this new code.

That being said, even if we did, the results wouldn't be very interesting. Essentially all programs one could write involving `coroutine`s will still crash at simulation time anyway, as we do not currently have a good story for translating coroutine aggregates.

spernsteiner (approved):

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1395,6 +1400,12 @@ evalRval (M.CopyForDeref lv) = evalLvalue lv
 evalRval (M.ShallowInitBox {}) = mirFail
     "evalRval: ShallowInitBox not supported"
 
+evalTupleRval :: HasCallStack => [Operand] -> MirGenerator h s ret (MirExp s)
+evalTupleRval ops = do
+  let tys = map typeOf ops
+  exps <- mapM evalOperand ops
+  buildTupleMaybeM tys (map Just exps)
```
samcowger: Save a few characters:
```suggestion
  buildTupleM tys exps
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1537,9 +1544,15 @@ initialValue (M.TyDowncast {}) = return Nothing
...
 initialValue (M.TyConst {}) = return Nothing
 initialValue (M.TyLifetime {}) = return Nothing
+initialValue (M.TyCoroutine {}) = return Nothing
 initialValue (M.TyErased {}) = return Nothing
 initialValue (M.TyInterned {}) = return Nothing
 
+initialTupleValue ::
+  HasCallStack => [M.Ty] -> MirGenerator h s ret (Maybe (MirExp s))
+initialTupleValue tys =
+  Just . MirExp MirAggregateRepr <$> mirAggregate_uninit_constSize (fromIntegral $ length tys)
```
samcowger, line 1555: Using `length tys` as the aggregate size implicitly hardcodes type size to be 1 byte; would you mind leaving a note to that effect?
```suggestion
initialTupleValue ::
  HasCallStack => [M.Ty] -> MirGenerator h s ret (Maybe (MirExp s))
initialTupleValue tys =
  -- TODO: hardcoded size=1
  Just . MirExp MirAggregateRepr <$> mirAggregate_uninit_constSize (fromIntegral $ length tys)
```

samcowger (approved):

Issue #1663, opened 2025-12-04
`crux-llvm` CI: Test with a more recent LLVM version


RyanGlScott:
Currently, the `crux-llvm` CI tests with LLVM 14, as seen here:

https://github.com/GaloisInc/crucible/blob/4d88a65b10c3b6bcee9bdc3251611360346bdd35/.github/ci.sh#L54-L83

This is pretty old by this point, especially since LLVM is now currently at version 21. We should update the CI to use a more recent LLVM version, perhaps targeting the latest LLVM version that `crux-llvm` currently supports.

One hiccup in this plan is that on Ubuntu CI jobs, we currently install Clang/LLVM using `apt`, and Ubuntu's packages usually will only include a very specific range of LLVM versions, which may be exclude the latest versions. To fix this problem, we should instead download Clang/LLVM using https://apt.llvm.org/. That page contains an automatic installation script (`llvm.sh`) that installs the appropriate PPA for a given Ubuntu version and can then allow you to install a range of Clang/LLVM versions, including the latest versions. This would make the LLVM installation story for Ubuntu much more future-proof.

PR #1675, opened 2025-12-12
CI: Use apt.llvm.org to download LLVM, test with LLVM 20

RyanGlScott:
Fixes https://github.com/GaloisInc/crucible/issues/1663.

### DIFF
--- a/doc/dev.md
+++ b/doc/dev.md
@@ -38,3 +38,34 @@ When adding support for the new GHC version to Crucible itself, complete the fol
 - [ ] Remove the old version's `cabal.GHC-X.Y.Z.config` file
 - [ ] Remove outdated CPP `ifdef`s that refer to the dropped version
 - [ ] Remove outdated `if` stanzas in the Cabal files
+
+## LLVM versions
+
+The `crux-llvm` CI currently runs the against one fixed, recent version of
+LLVM. We should aim to update this LLVM version from time to time as new
+versions of LLVM are released.
+
+When updating the version of LLVM that the `crux-llvm` CI uses, complete the
+following steps:
+
+- [ ] Bump the `llvm-pretty` and `llvm-pretty-bc-parser` submodule commits to
+      versions that support the desired version of LLVM.
+- [ ] Ensure that the `crucible-llvm-*` and `crux-llvm` packages build against
+      the new versions of `llvm-pretty` and `llvm-pretty-bc-parser`. If not,
+      update the code as needed.
+- [ ] Ensure that the `crux-llvm` test suite passes with the desired version of
+      LLVM. If not, investigate why specific test cases fail. There is no
+      one-size-fits-all-formula for fixing these sorts of test failures, but
+      often times this will require changes to `llvm-pretty`,
+      `llvm-pretty-bc-parser`, `crucible-llvm`, or a combination thereof.
+- [ ] Update `crux-llvm/README.md` to mention that a newer LLVM version is now
+      supported (near the sentence "... LLVMs versions from X to Y are likely
+      to work well ...").
+- [ ] Go to the vendored-in `.github/llvm.sh` script and check to see if there
+      have been any changes to the script upstream (follow the URL in the
+      comments near the top of the script). If there have been changes, then
+      copy over the changes to the vendored-in script and commit them.
+- [ ] In `.github/ci.sh`, update `LINUX_LLVM_VER` and `MACOS_LLVM_VER` to the
+      desired LLVM version.
+- [ ] In `.github/Dockerfile-crux-llvm`, update `LLVM_VER` to the desired LLVM
+      version.


RyanGlScott:
The CI _almost_ passes, except that the macOS CI job fails on the [`T1177.c`](https://github.com/GaloisInc/crucible/blob/9fa8f120326f3d5accca726b9c35d225ba2bb0ca/crux-llvm/test-data/golden/T1177.c) test case, as seen in [this job log](https://github.com/GaloisInc/crucible/actions/runs/20153296807/job/57850466519#step:18:49):

```
    T1177.c
      solver=z3
        loop-merging=loop
          clang-range=recent-clang
            z3 4.8.14
              T1177
                T1177 crux run:                          OK (0.70s)
                T1177 crux output:                       FAIL
                  test/Test.hs:340:
                  MISMATCH for expected (test-data/golden/T1177.z3.good)
                             and actual (test-data/golden/T1177.z3.z3.out) output:
                      F-expect>|[Crux] Overall status: Valid.
                      F-actual>|[Crux] Found counterexample for verification goal
                      F-actual>|[Crux]   test-data/golden/T1177.c:48:1: error: in reduce_add
                      F-actual>|[Crux]   unsupported LLVM value: ValPoison of type i32
                      F-actual>|[Crux] Overall status: Invalid.
                  
                  Use -p '/loop-merging=loop.clang-range=recent-clang.z3 4.8.14.T1177.T1177 crux output/' to rerun this test only.
```

It appears that LLVM's InstCombine pass generates bitcode that looks like this:

```llvm
%50 = insertelement <8 x i32> <i32 poison, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0>, i32 %48, i64 0
```

This is unfortunate from a `crucible-llvm` perspective, as `crucible-llvm` essentially has [no ability to reason about code involving poison values](https://github.com/GaloisInc/crucible/issues/366). As such, I'm really not sure what to do about this...

langston-barrett:
Is it generating that because there is UB in the code? If so, perhaps we should fix it?

RyanGlScott:
> Is it generating that because there is UB in the code?

If there is, it's not at all obvious to me where. Here is a minimized version of `T1177.c`, which produces the LLVM bitcode seen above:

```c
#include <stddef.h>
#include <stdint.h>

uint32_t reduce_add(const uint8_t* in, size_t len) {
  uint32_t x = 0;

  for (size_t i = 0; i < len; i++) {
    x = x + in[i];
  }

  return x;
}
```

(Compile this with `/opt/homebrew/opt/llvm\@20/bin/clang test.c -O2 -emit-llvm -S` on macOS.)

The LLVM Language reference says [the following](https://releases.llvm.org/20.1.0/docs/LangRef.html#poison-values) about poison values:

> This means that immediate undefined behavior occurs if a poison value is used as an instruction operand that has any values that trigger undefined behavior. Notably this includes (but is not limited to):
>
> * The pointer operand of a [`load`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-load), [`store`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-store) or any other pointer dereferencing instruction (independent of address space).
> * The divisor operand of a `udiv`, `sdiv`, `urem` or `srem` instruction.
> * The condition operand of a [`br`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-br) instruction.
> * The callee operand of a [`call`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-call) or [`invoke`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-invoke) instruction.
> * The parameter operand of a [`call`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-call) or [`invoke`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-invoke) instruction, when the function or invoking call site has a noundef attribute in the corresponding position.
> * The operand of a [`ret`](https://releases.llvm.org/20.1.0/docs/LangRef.html#i-ret) instruction if the function or invoking call site has a noundef attribute in the return value position.

`insertelement` isn't explicitly called out here, so is it undefined behavior to overwrite a vector element that contains a `poison` value? The discussion on [this LLVM Discourse thread](https://discourse.llvm.org/t/documentation-of-detailed-poison-semantics/81243) would suggest that it's not.

RyanGlScott:
Alive2 also seems to agree with this interpretation of `insertelement`'s interaction with `poison` (see https://github.com/AliveToolkit/alive2/pull/343).

RyanGlScott:
One other thing to note: `insertelement <8 x i32> <i32 poison, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0>, i32 %48, i64 0` is a bit interesting in that the `poison` value is never actually stored into a register, which means that with the way `crucible-llvm`'s translation works, we could conceivably translate this without ever needing to reason about `poison` in the Crucible simulator. This is great, because simulation is one the tricky parts that I'm not quite sure how to do correctly. (Compare how we [simulate `undef` values](https://github.com/GaloisInc/crucible/blob/ec039bbf84abdd71ea6a78eca30b68f569b477d9/crucible/src/Lang/Crucible/Simulator/Evaluation.hs#L716-L718), which is pretty dodgy-looking.)

Unfortunately, that isn't the only way that `poison` appears in the `T1177.c` bitcode. Here is another minimized snippet that results in a slightly different bitcode pattern:

```c
#include <stddef.h>
#include <stdint.h>

uint32_t reduce_umax(const uint8_t* in, size_t len) {
  uint32_t x = 0;

  for (size_t i = 0; i < len; i++) {
    uint32_t y = in[i];
	x = (x > y) ? x : y;
  }

  return x;
}
```

Compile this with `/opt/homebrew/opt/llvm\@20/bin/clang test.c -O2 -emit-llvm -S` on macOS, and you will see the following bitcode in the resulting `test.ll` file:

```llvm
  %50 = insertelement <8 x i32> poison, i32 %48, i64 0
  %51 = shufflevector <8 x i32> %50, <8 x i32> poison, <8 x i32> zeroinitializer
```

Just like `insertelement <8 x i32> <i32 poison, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0>, i32 %48, i64 0`, this will result in a vector where all of the elements are zero. Unfortunately, due to the way that this code is written, the `%50` register ends up with a "temporary" `poison` value stored inside it, which `crucible-llvm` translates all the way to the simulator level. As such, we can't as easily work around the problem in this example—we'd actually need to implement faculties for reasoning about the `poison` value as it is passed around during simulation.

RyanGlScott:
After some thought, here is my current plan:

* Add support for translating things like `insertelement <8 x i32> <i32 poison, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0>, i32 %48, i64 0` to Crucible. As noted in https://github.com/GaloisInc/crucible/pull/1675#issuecomment-3662280745, this approach has plenty of limitations, but this would nevertheless allow us to support strictly more programs than before.
* Replace the `T1177.c` test case with a series of `crucible-llvm-cli` test cases. The original point of `T1177.c` was to test the semantics of vectorized LLVM intrinsics (e.g., `llvm.vector.reduce.or.v4i32`), but it's now being tangled up in the whims of Clang's optimization passes. Let's nip this in the bud by simply testing the intrinsics directly in `crucible-llvm-syntax`, which will never sneakily introduce `poison` values without our notice.

RyanGlScott:
I've pushed a separate commit which replaces `crux-llvm`'s `T1177.c` test case with equivalent `crucible-llvm-cli` test cases. That is enough to make the CI pass, so I'll mark this as ready for review.

I've also opened https://github.com/GaloisInc/crucible/pull/1684, which implements my idea for rudimentary `poison` constant support. This is perhaps a more contentious change, so I've opted to leave that outside of the PR in case we decide that we don't want to adopt that change.

.github/ci.sh
```diff
@@ -53,25 +53,15 @@ test() {
...
-      LINUX_LLVM_VER=14
-    else
-      echo "Don't know what LLVM version to use for $LINUX_LLVM_VER."
-      exit 1
-    fi
-    sudo apt-get update -q && sudo apt-get install -y "clang-$LINUX_LLVM_VER" "llvm-$LINUX_LLVM_VER-tools"
+    LINUX_LLVM_VER=20
+    wget https://apt.llvm.org/llvm.sh
+    chmod +x llvm.sh
+    sudo ./llvm.sh "$LINUX_LLVM_VER"
```
sauclovian-g: Can we do some kind of check on the downloaded script before executing it as root (even on a GH runner)? E.g. check that it has an expected hash.

wget does check certificates, but that doesn't protect against broken/partial downloads, let alone blunders or hacking upstream.

RyanGlScott: As far as I am aware, apt.llvm.org doesn't provide expected hashes for `llvm.sh`. Moreover, I think it's a moving target, so it likely wouldn't suffice to just compute the hash ourselves and check it in to the repo (unless we wanted to constantly chase a moving target).

Some alternatives are:

* Instead of downloading this script through apt.llvm.org, download a fixed version of the script from https://github.com/opencollab/llvm-jenkins.debian.net/blob/master/llvm.sh, where it is maintained upstream. This would allow us to compute a fixed hash, but at the expense of freezing the script going forward until we update the URL.
* Don't use the `llvm.sh` script at all, and instead manually do the work to set up the `apt` repositories, which come with signature files. Note that this would require doing slightly different things on each supported Ubuntu version, since the URLs for each Ubuntu version are different.

langston-barrett: > Instead of downloading this script through apt.llvm.org, download a fixed version of the script from https://github.com/opencollab/llvm-jenkins.debian.net/blob/master/llvm.sh, where it is maintained upstream. This would allow us to compute a fixed hash, but at the expense of freezing the script going forward until we update the URL.

I like this idea!

kquick: The third alternative is: store a copy of `llvm.sh` in the `.github` directory of this repo manually rather than having it auto-fetched for each CI and Docker run.  The only time this script is likely to change is when a new LLVM version is released, so we can note this as an "update llvm.sh from apt.llvm.org" task in the developer docs (https://github.com/GaloisInc/crucible/blob/master/doc/dev.md) [which should be enhanced to discuss new LLVM versions anyhow].  The script license appears to be Apache-2.0, which would permit us to host a copy here.

 Additional benefits are that it allows us to control the update frequence and visually inspect the script for naughtiness.

RyanGlScott: I've vendored in a specific snapshot of `llvm.sh` instead of downloading it afresh on each CI run. I've also added a new section to `doc/dev.md` describing the steps needed to update the `crux-llvm` CI's tested version of LLVM, including a step for updating the vendored-in copy of `llvm.sh`. Let me know what you think!

.github/llvm.sh
```diff
@@ -0,0 +1,247 @@
+#!/bin/bash
+
+# This script is adapted from the upstream code at
+# https://github.com/opencollab/llvm-jenkins.debian.net/blob/f8b8349a485dec4f879724c6aa76e554c358fb59/llvm.sh.
+# We choose to vendor in a specific copy of the script so that we do not have
+# to download it on each CI run. This avoids issues related to broken/partial
+# downloads or unexpected changes to the code upstream.
+#
+# If you update this script (e.g., when updating the crux-llvm CI to use a more
+# recent LLVM version, make sure to update the commit hash above). The rest of
```
sauclovian-g:
```suggestion
# If you update this script (e.g., when updating the crux-llvm CI to use a more
# recent LLVM version), make sure to update the commit hash above. The rest of
```

doc/dev.md
```diff
@@ -38,3 +38,34 @@ When adding support for the new GHC version to Crucible itself, complete the fol
 - [ ] Remove the old version's `cabal.GHC-X.Y.Z.config` file
 - [ ] Remove outdated CPP `ifdef`s that refer to the dropped version
 - [ ] Remove outdated `if` stanzas in the Cabal files
+
+## LLVM versions
```
langston-barrett, line 42: Very nice, thank you!

langston-barrett (approved):

Issue #1674, opened 2025-12-11
Nightly `crux-mir` regression: creating 0-length array with `array::from_fn()` fails


RyanGlScott:
A recent nightly version of `crux-mir` (built from commit 594305a3726478752247e8489ce1dcbe69948346) fails when simulating the following program:

```rs
// test.rs
#[crux::test]
pub fn f() -> [u8; 0] {
    std::array::from_fn(|_i| 0)
}
```
```
$ cabal -v0 run exe:crux-mir -- test.rs
test test/58c1c704::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/58c1c704::f[0] counterexamples ----
[Crux] Found counterexample for verification goal
[Crux]   ./libs/core/src/ptr/const_ptr.rs:1155:18: 1155:28: error: in core/729aed48::ptr[0]::const_ptr[0]::{impl#0}[0]::read[0]::_inst30a4a1a6d21e7481[0]
[Crux]   read requires a reference to IntrinsicRepr MirAggregate [], but got a reference to UnitRepr
[Crux] Found counterexample for verification goal
[Crux]   ./libs/core/src/ptr/const_ptr.rs:1155:18: 1155:28: error: in core/729aed48::ptr[0]::const_ptr[0]::{impl#0}[0]::read[0]::_inst30a4a1a6d21e7481[0]
[Crux]   attempted to read empty mux tree

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] Goal status:
[Crux]   Total: 2
[Crux]   Proved: 0
[Crux]   Disproved: 2
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Invalid.
```

This used to work. Curiously, this error only occurs if the array length is 0; all other array lengths work without issue.

RyanGlScott:
I wonder if this is fallout from https://github.com/rust-lang/rust/commit/8362508989c8941ccbba26aaa8c787cc492c6fef (introduced in Rust 1.90.0), which applied the following change to the definition of `MaybeUninit::assume_init()`:

```diff
diff --git a/library/core/src/mem/maybe_uninit.rs b/library/core/src/mem/maybe_uninit.rs
index 63a479ed8dd4e..fc35e54bb0dcc 100644
--- a/library/core/src/mem/maybe_uninit.rs
+++ b/library/core/src/mem/maybe_uninit.rs
[...]

PR #1677, opened 2025-12-12
`{crucible,crux}-mir`: Represent `()`, `!`, and `FnDef`s with `MirAggregateType`

RyanGlScott:
Previously, these were represented with `UnitType`. This has the unfortunate consequence that different zero-sized types were represented with different Crucible type representations, as empty arrays are represented with `MirAggregateType` instead. This poses an issue for some code in the internals of `array::from_fn::<0>()`, which uses a `union` type that tries to reinterpret a field of type `()` (represented with `UnitType`) as a length-0 array (represented with `MirAggregateType`), and these Crucible type representations are not compatible.

This changes the Crucible type representation of `()` (as well as `!` and `FnDef`s, two other forms of zero-sized types) to instead use `MirAggregateType`. This means that most zero-sized types now use `MirAggregateType`, with the notable exception of empty structs (see https://github.com/GaloisInc/crucible/issues/1499). Still, things are much more uniform than they were before, which is a nice improvement.

One slight annoyance brought about by this change is that several custom overrides which previously returned `()` implicitly now must explicitly return a `MirAggregate`. To make this slightly easier, I have introduced helper functions `mirAggregate_zst{,IO,Sim}`, which can be used to construct zero-sized `MirAggregate` values in such situations.

While I was in town, I noticed that although `TyFnDef` is a zero-sized type, the `isZeroSized` function was incorrectly returning `False` for `TyFnDef`s. This has now been corrected.

Fixes https://github.com/GaloisInc/crucible/issues/1674.

### DIFF
--- a/crucible-mir/src/Mir/TransTy.hs
+++ b/crucible-mir/src/Mir/TransTy.hs
@@ -45,6 +45,7 @@ import Data.Parameterized.Some
 
 
 -- crucible
+import qualified Lang.Crucible.Panic as P
 import qualified Lang.Crucible.Types as C
 
 import qualified Lang.Crucible.CFG.Expr as E
@@ -334,9 +335,23 @@ canInitialize col ty = case ty of
       | otherwise -> True
     -- Others
     M.TyArray {} -> True
-    -- TODO: workaround for a ref init bug - see initialValue for details
-    --M.TyRef ty' _ -> canInitialize col ty'
-    _ -> False
+
+    M.TyFnDef {} -> False
+    M.TyNever {} -> False
+    M.TyRef {} -> False
+    M.TyRawPtr {} -> False
+    M.TyFnPtr {} -> False
+    M.TyDynamic {} -> False
+    M.TySlice {} -> False
+    M.TyStr {} -> False
+    M.TyFloat {} -> False
+    M.TyDowncast {} -> False
+    M.TyForeign {} -> False
+    M.TyConst {} -> False
+    M.TyLifetime {} -> False
+    M.TyCoroutine {} -> False
+    M.TyErased {} -> False
+    M.TyInterned {} -> False
 
 isUnsized :: M.Ty -> Bool
 isUnsized ty = case ty of
@@ -354,10 +369,30 @@ isZeroSized col = go
       M.TyClosure tys -> all go tys
       M.TyCoroutineClosure tys -> all go tys
       M.TyArray elemTy n -> n == 0 || go elemTy
-      M.TyAdt name _ _ | Just adt <- col ^? M.adts . ix name -> adt ^. M.adtSize == 0
+      M.TyAdt name _ _
+        | Just adt <- col ^? M.adts . ix name -> adt ^. M.adtSize == 0
+        | otherwise -> P.panic "isZeroSized" ["unknown ADT", show name]
       M.TyFnDef {} -> True
       M.TyNever -> True
-      _ -> False
+
+      M.TyBool {} -> False
+      M.TyChar {} -> False
+      M.TyInt {} -> False
+      M.TyUint {} -> False
+      M.TyRef {} -> False
+      M.TyRawPtr {} -> False
+      M.TyFnPtr {} -> False
+      M.TyDynamic {} -> False
+      M.TySlice {} -> False
+      M.TyStr {} -> False
+      M.TyFloat {} -> False
+      M.TyDowncast {} -> False
+      M.TyForeign {} -> False
+      M.TyConst {} -> False
+      M.TyLifetime {} -> False
+      M.TyCoroutine {} -> False
+      M.TyErased {} -> False
+      M.TyInterned {} -> False
 
 
 -- | Get the "ABI-level" function arguments for @sig@, which determines the


crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1075,6 +1075,7 @@ buildStructAssign' ctx0 es0 = go ctx0 $ reverse es0
         Ctx.AssignExtend ctx' fldr -> case (fldr, optExp) of
             (FieldRepr (FkInit tpr), Nothing) ->
                 case tpr of
+                    -- TODO RGS: Is this right?
                     C.UnitRepr -> continue ctx' rest tpr (R.App $ E.NothingValue tpr)
                     _ -> Left $ "got Nothing for mandatory field " ++ show (length rest) ++ " type:" ++ show tpr
```
RyanGlScott: I'm not sure about this `UnitRepr` case. Interestingly, I opted to not change this, thinking that the test suite would surely turn up an example of where something goes wrong, except that the test case actually _passes_ without updating this case! In light of this, I'm wondering if the `UnitRepr` case is even reachable at all, given that we should no longer be using `UnitRepr` to represent any types in `crucible-mir`.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -356,6 +355,7 @@ isZeroSized col = go
       M.TyCoroutineClosure tys -> all go tys
       M.TyArray elemTy n -> n == 0 || go elemTy
       M.TyAdt name _ _ | Just adt <- col ^? M.adts . ix name -> adt ^. M.adtSize == 0
+      M.TyFnDef {} -> True
       M.TyNever -> True
       _ -> False
```
samcowger: It's a bit tangential to this PR, but could you perhaps go in and list the remaining constructors here?

crucible-mir/src/Mir/TransTy.hs
samcowger: Hmm, it's not clear to me why/whether this case is necessary. I see it comes from b4f6235aaefdc3dcf98745484497027736a55a7b, but I can't easily see why that change was needed at that time. Did you try removing it altogether?

RyanGlScott:
> I hypothesize that the former arises from this use of `UnitType` to represent the return types of static initializers

Yes, that's right. This is solely used for defining a Crucible-internal function, which means that we can give it whatever type we want (we aren't forced to use a Rust type for it). We could give it the return type `MirAggregate`, but there is no pressing need to do so, given that none of the translated Rust code ever attempts to use the value that this internal function returns.

A similar rationale applies for `Mir{Write,Drop}Ref`—these don't have corresponding Rust types, so there's not a pressing need for this to use `MirAggregate`.

I could go either way on this. It might be nice to use `MirAggregate` here for consistency's sake, although it does require making the corresponding definitions a bit more verbose (you have to explicitly return a `MirAggregate` value instead of implicitly returning `()`).

crucible-mir/src/Mir/TransTy.hs
RyanGlScott: I think I will remove this case outright. The CI results from https://github.com/GaloisInc/saw-script/pull/2915 (which demonstrate SAW making use of these changes) reveal that nothing ever triggers this code path, which is pretty convincing evidence that this is dead code.

RyanGlScott: In order to convince myself that this case can never be reached, I discovered (or rather, re-discovered) the following:

1. MIR can sometimes initialize enums without providing explicit assignments for all of its fields, e.g.,

   ```rs
   let mut _0 : Result<(), i32>;
   discr(_0) = 0
   // The field of Ok is implicitly initialized to ()
   ```
2. When translating an enum value in `crucible-mir`, this means that sometimes there will be fewer fields than field types, which poses a challenge for translation. The `inferElidedVariantFields` function looks at each of the field types, and if there are any missing fields corresponding to zero-sized types, then it fills in a default field value. The `isZeroSized` function determines if a type is zero-sized, and the `initialValue` computes the default value, returning `Just` if a default value can be obtained and `Nothing` otherwise.
3. Meanwhile, the `tyToFieldRepr` function is responsible for determining the `FieldRepr` (i.e., the Crucible type representation) for each field's type. There are two cases: one for types that can be reasonably initialized (`FkInit`), and another for types where that isn't reasonable (`FkMaybe`). The function which determines whether a type can be reasonably initialized or not is `canInitialize`.
4. `buildStructAssign'` takes the output of both `tyToFieldRepr` and `inferElidedVariantFields` and zips them together. In order for the `UnitRepr` case in `buildStructAssign'` (which we suspect is unreachable) to actually be reachable, we would need to find a MIR type for which (a) `canInitialize` returns `True`, (b) `isZeroSized` returns `True`, and (c) `initialValue` returns `Nothing`.

   As far as I can tell, there is exactly one such type: a `TyAdt` which is an uninhabited enums. On the other hand, `buildStructAssign'` is only used for constructing values, and you can't construct an uninhabited enum in the first place! As such, I think this case is unreachable regardless.

I'll add a comment which summarizes the reasoning above.

RyanGlScott: Actually, the comments [here](https://github.com/GaloisInc/crucible/blob/ec039bbf84abdd71ea6a78eca30b68f569b477d9/crucible-mir/src/Mir/TransTy.hs#L314-L315) claim that if `canInitialize` returns `True`, then `initialValue` should also return `Just`, suggesting that one of the two functions' handling of uninhabited enums is wrong. (Most likely `canInitialize`, which [returns `True` for enums](https://github.com/GaloisInc/crucible/blob/ec039bbf84abdd71ea6a78eca30b68f569b477d9/crucible-mir/src/Mir/TransTy.hs#L335), even uninhabited ones.)

RyanGlScott: I've removed the unreachable `UnitRepr` case in https://github.com/GaloisInc/crucible/pull/1677/commits/b9669411c104b18c94b5ba16fb1e2f3f81de47d8.

crucible-mir/src/Mir/TransTy.hs
RyanGlScott: Yes, good call. I've done this in https://github.com/GaloisInc/crucible/pull/1677/commits/a59e561799d808bab1948aae8f78ebe9d15e0f5c. I also applied the same treatment to the closely related `canInitialize` function.

PR #1678, opened 2025-12-13
crux-mir-test: align rustc edition with `mir-json` defaults

### LINKED ISSUE #1668: `crux-mir-test`: Invoke `rustc` with a newer language edition by default

sand7000:
The crux-mir test suite now invokes `rustc` with the same default Rust edition (2021) as `mir-json`, avoiding mismatches between oracle execution and MIR generation and allowing tests that rely on post-2015 language features.

Fixes #1668

### DIFF
--- a/crux-mir/src/Mir/Defaults.hs
+++ b/crux-mir/src/Mir/Defaults.hs
@@ -1,7 +1,15 @@
-module Mir.Defaults (defaultRustEditionFlag) where
+module Mir.Defaults
+  ( defaultRustEdition
+  , defaultRustEditionFlag
+  ) where
 
+-- | The default Rust language edition used by crux-mir when invoking Rust
+-- tooling (e.g., mir-json or rustc) on a standalone @.rs@ file.
+--
+-- Cargo-managed builds continue to use the edition specified in @Cargo.toml@.
 defaultRustEdition :: String
 defaultRustEdition = "2021"
 
+-- | The @--edition@ flag corresponding to 'defaultRustEdition'.
 defaultRustEditionFlag :: String
 defaultRustEditionFlag = "--edition=" ++ defaultRustEdition


sand7000:
I added `Mir.Defaults` to define the default Rust edition flag in a single place and reuse it consistently in both the `mir-json` invocation and the test suite’s `rustc` invocation, preventing future mismatches.

I added `test/conc_eval/async/async.rs` as a minimal test that relies on a post-2015 Rust language feature.

Before this PR, running:

```
cabal test crux-mir-test --test-options='--pattern async'
```

failed during the compilation step:

```
rustc compilation failed for async
error[E0670]: `async fn` is not permitted in Rust 2015
...
test/Test.hs:106: failed to compile and run
```

After these changes, the test fails for the *correct* reason (on the Crux side), and it will become a passing regression test once async support is implemented (#1369).

crux-mir/src/Mir/Defaults.hs
```diff
@@ -0,0 +1,7 @@
+module Mir.Defaults (defaultRustEditionFlag) where
+
+defaultRustEdition :: String
+defaultRustEdition = "2021"
```
RyanGlScott, line 11: It would be worth adding some Haddocks to say in what sense this is the "default" Rust language edition. In particular, this is the default edition that `crux-mir` itself uses when invoking the compiler on a standalone `.rs` file. (Not to be confused with `rustc`'s default language edition, which is `2015`.)

crux-mir/src/Mir/Defaults.hs
```diff
@@ -0,0 +1,7 @@
+module Mir.Defaults (defaultRustEditionFlag) where
```
RyanGlScott: I would favor exporting `defaultRustEdition` as well. This would allow for maximum flexibility in call sites—for instance, one could imagine wanting to pass the `rust-{2018,2021,2024}-compatibility` lint warning, and having `defaultRustEdition` available would make it easier to do so.

crux-mir/test/Test.hs
```diff
@@ -27,6 +27,7 @@ import           Test.Tasty.ExpectedFailure (expectFailBecause)
 import           Text.Regex.Base ( makeRegex, matchM )
 import           Text.Regex.Posix.ByteString.Lazy ( Regex )
 
+import Mir.Defaults (defaultRustEditionFlag)
 import qualified Mir.Language as Mir
```
RyanGlScott, line 31: For consistency with the style used in the rest of the module, please add some whitespace here:


```suggestion
import           Text.Regex.Posix.ByteString.Lazy ( Regex )

import           Mir.Defaults (defaultRustEditionFlag)
import qualified Mir.Language as Mir
```

crux-mir/src/Mir/Defaults.hs
sand7000: Added comments.

crux-mir/src/Mir/Defaults.hs
sand7000: Done.

PR #1683, opened 2025-12-16
Allow passing custom options to `mir-json` when compiling a Rust file. Closes #1556.

### LINKED ISSUE #1556: `crux-mir`: Allow passing custom options to `mir-json` when compiling a Rust file

sand7000:
`crux-mir` always calls `mir-json` with a fixed set of flags (including a default `--edition`), and there is no direct way to override or extend that. This PR addresses that by adding a repeatable flag to forward extra arguments to `mir-json`.

### DIFF
--- a/crux-mir/src/Mir/Generate.hs
+++ b/crux-mir/src/Mir/Generate.hs
@@ -16,8 +16,10 @@
 module Mir.Generate(generateMIR) where
 
 import Control.Monad (when)
+
 import Data.List (stripPrefix, isPrefixOf)
 
+import qualified Data.Foldable as Foldable
 import qualified Data.ByteString.Lazy as B
 import qualified Data.ByteString as BS
 
@@ -34,7 +36,7 @@ import GHC.Stack
 import Lang.Crucible.Panic (panic)
 
 import Mir.Mir
-import Mir.Options (MIROptions(..), mirJsonArgsList)
+import Mir.Options (MIROptions(..))
 import Mir.Defaults (defaultRustEditionFlag)
 import Mir.ParseTranslate (parseMIR)
 import Mir.PP()
@@ -72,6 +74,9 @@ buildMirJsonArgs :: [String]   -- ^ base/default args (no edition)
                  -> [String]
 buildMirJsonArgs base extras =
   let hasEdition  = any isEditionArg extras
+      -- rustc rejects multiple --edition flags, so if the user supplies
+      -- an explicit --edition via --mir-json-arg, we must *not* also add
+      -- the default --edition flag here.
       editionPart = if hasEdition then [] else [defaultRustEditionFlag]
   in base ++ editionPart ++ extras
 
@@ -110,7 +115,7 @@ compileMirJson cruxOpts mirOpts keepRlib rustFile = do
              , "-o", outFile
              ]
 
-    let extraMirArgs = mirJsonArgsList mirOpts
+    let extraMirArgs = Foldable.toList (mirJsonArgs mirOpts)
 
     -- Add default edition if needed, then append extraMirArgs
     let finalArgs = buildMirJsonArgs baseArgs extraMirArgs
--- a/crux-mir/src/Mir/Options.hs
+++ b/crux-mir/src/Mir/Options.hs
@@ -2,10 +2,8 @@ module Mir.Options
   ( MIROptions(..)
   , defaultMirOptions
   , mirConfig
-  , mirJsonArgsList
   ) where
 
-import qualified Data.Foldable as Foldable
 import           Data.Sequence (Seq)
 import qualified Data.Sequence as Seq
 import qualified Data.Text as Text
@@ -99,6 +97,3 @@ mirConfig = Config
                 Right opts { mirJsonArgs = mirJsonArgs opts Seq.|> v })
         ]
     }
-
-mirJsonArgsList :: MIROptions -> [String]
-mirJsonArgsList = Foldable.toList . mirJsonArgs


sand7000:
I tested the new `--mir-json-arg` behavior on `crux-mir/test/conc_eval/async/async.rs`.

First, with no explicit edition override, `crux-mir` still passes the default
edition (`--edition=2021`) and prints the full `mir-json` invocation at
higher `--sim-verbose` levels:

```bash
$ crux-mir -d 2 async.rs
[crux-mir] mir-json async.rs -L /home/esandberg/projects/mir-json/rlibs --crate-type=rlib --extern rustc_literal_escaper=/home/esandberg/projects/mir-json/rlibs/librustc_literal_escaper.rlib --extern getopts=/home/esandberg/projects/mir-json/rlibs/libgetopts.rlib --extern panic_unwind=/home/esandberg/projects/mir-json/rlibs/libpanic_unwind.rlib --extern compiler_builtins=/home/esandberg/projects/mir-json/rlibs/libcompiler_builtins.rlib --extern int512=/home/esandberg/projects/mir-json/rlibs/libint512.rlib --extern rustc_std_workspace_alloc=/home/esandberg/projects/mir-json/rlibs/librustc_std_workspace_alloc.rlib --extern adler2=/home/esandberg/projects/mir-json/rlibs/libadler2.rlib --extern alloc=/home/esandberg/projects/mir-json/rlibs/liballoc.rlib --extern object=/home/esandberg/projects/mir-json/rlibs/libobject.rlib --extern rustc_demangle=/home/esandberg/projects/mir-json/rlibs/librustc_demangle.rlib --extern unwind=/home/esandberg/projects/mir-json/rlibs/libunwind.rlib --extern crucible=/home/esandberg/projects/mir-json/rlibs/libcrucible.rlib --extern test=/home/esandberg/projects/mir-json/rlibs/libtest.rlib --extern cfg_if=/home/esandberg/projects/mir-json/rlibs/libcfg_if.rlib --extern libc=/home/esandberg/projects/mir-json/rlibs/liblibc.rlib --extern std=/home/esandberg/projects/mir-json/rlibs/libstd.rlib --extern std_detect=/home/esandberg/projects/mir-json/rlibs/libstd_detect.rlib --extern addr2line=/home/esandberg/projects/mir-json/rlibs/libaddr2line.rlib --extern miniz_oxide=/home/esandberg/projects/mir-json/rlibs/libminiz_oxide.rlib --extern rustc_std_workspace_std=/home/esandberg/projects/mir-json/rlibs/librustc_std_workspace_std.rlib --extern gimli=/home/esandberg/projects/mir-json/rlibs/libgimli.rlib --extern core=/home/esandberg/projects/mir-json/rlibs/libcore.rlib --extern byteorder=/home/esandberg/projects/mir-json/rlibs/libbyteorder.rlib --extern rustc_std_workspace_core=/home/esandberg/projects/mir-json/rlibs/librustc_std_workspace_core.rlib --extern hashbrown=/home/esandberg/projects/mir-json/rlibs/libhashbrown.rlib --extern memchr=/home/esandberg/projects/mir-json/rlibs/libmemchr.rlib --extern panic_abort=/home/esandberg/projects/mir-json/rlibs/libpanic_abort.rlib --extern bytes=/home/esandberg/projects/mir-json/rlibs/libbytes.rlib --extern proc_macro=/home/esandberg/projects/mir-json/rlibs/libproc_macro.rlib --color=always --cfg crux --cfg crux_top_level -Z ub-checks=false -o async.bin --edition=2021
... (existing translation failures about Coroutines not yet supported) ...
```

This matches the previous behavior: the test still fails because `crux-mir`
doesn’t support async/coroutines yet.

With an explicit edition override via `--mir-json-arg`, `crux-mir` drops the
default edition flag and forwards the user-supplied one instead causing (in this example) the compilation step to fail due to incompatible rust edition:

```bash
$ crux-mir -d 2 --mir-json-arg="--edition=2015" async.rs
[crux-mir] mir-json async.rs ... --edition=2015
error[E0670]: `async fn` is not permitted in Rust 2015
...
user error (Error 101 while running mir-json on async.rs)
```

So:

* default behavior remains unchanged (`--edition=2021` when no override is given), and
* `--mir-json-arg` both affects the `mir-json` invocation and is reflected in
  the printed command line.

crux-mir/src/Mir/MirJsonArgs.hs
```diff
@@ -0,0 +1,25 @@
...
+-- | Given base mir-json args (without any edition flag) and extra args
+--   from the user, build the final argv.
+--
+-- Policy:
+--   * If the user supplies any --edition (or --edition=...) in extras,
+--     we DO NOT add the default edition.
+--   * Otherwise, we add the default --edition flag.
+buildMirJsonArgs :: [String]   -- ^ base/default args (no edition)
+                 -> [String]   -- ^ extra args (from MIROptions.mirJsonArgs)
+                 -> [String]
```
RyanGlScott: Putting this function in its own module seems a bit overkill, as it's only called in a single place. I propose that we just make this a helper function in `Mir.Generate`.

crux-mir/src/Mir/Language.hs
```diff
@@ -571,6 +573,11 @@ mirConfig = Crux.Config
         , GetOpt.Option []  ["cargo-test-file"]
             "treat trailing args as --test-filter instead of FILES, like `cargo test`; load program from this file instead (used by `cargo crux test`)"
             (GetOpt.ReqArg "file" (\v opts -> Right opts { cargoTestFile = Just v }))
+
+        , GetOpt.Option [] ["mir-json-arg"]
+            "pass an argument to mir-json (can be repeated)"
+            (GetOpt.ReqArg "ARG" $ \v opts ->
+                Right opts { mirJsonArgs = mirJsonArgs opts ++ [v] })
```
RyanGlScott: This appends new entries to the end of the list each time, and since `(++)` is an O(n) operation, it will take O(n^2) time to process all of the `mir-json-arg`s. I think it would be worth using a different data structure with a cheaper append operation (e.g., [`Data.Sequence`](https://hackage-content.haskell.org/package/containers-0.8/docs/Data-Sequence.html#t:Seq)) instead of a list to avoid this problem.

crux-mir/src/Mir/Language.hs
```diff
@@ -247,7 +247,7 @@ runTestsWithExtraOverrides initS bindExtra (cruxOpts, mirOpts) = do
                 ifs -> error $ "expected exactly 1 input file, but got " ++ show (length ifs) ++ " files"
 
     -- Load the MIR collection
-    col <- generateMIR cruxOpts filename False
+    col <- generateMIR cruxOpts (mirJsonArgs mirOpts) filename False
```
RyanGlScott: A bit of a personal stylistic preference, but I would rather `generateMIR` take a `MIROptions` value instead of a `[String]`, as it is more obvious what the purpose of the former type is than the latter.

crux-mir/src/Mir/Generate.hs
```diff
@@ -66,8 +66,8 @@ mirJsonOutFile rustFile = rustFile -<.> "mir"
 getRlibsDir :: (?defaultRlibsDir :: FilePath) => IO FilePath
 getRlibsDir = maybe ?defaultRlibsDir id <$> lookupEnv "CRUX_RUST_LIBRARY_PATH"
 
-compileMirJson :: (?defaultRlibsDir :: FilePath) => Crux.CruxOptions -> Bool -> FilePath -> IO ()
-compileMirJson cruxOpts keepRlib rustFile = do
+compileMirJson :: (?defaultRlibsDir :: FilePath) => Crux.CruxOptions -> [String] -> Bool -> FilePath -> IO ()
```
RyanGlScott: Similarly, I think it the type of this function would be more self-evident if it took a `MIROptions` value instead of `[String]`.

crux-mir/README.md
```diff
@@ -128,6 +128,20 @@ To compile and test a single Rust program:
...
+`mir-json` with `--mir-json-arg` (repeatable). For example:
+
+```bash
+$ crux-mir test/conc_eval/prim/add1.rs \
+    --mir-json-arg="--edition=2015" \
+    --mir-json-arg="-Copt-level=3"
+```
+
+If any `--mir-json-arg` supplies an `--edition` flag, `crux-mir` will not add
+its default `--edition` flag.
```
RyanGlScott: I don't think the README mentions the idea that `crux-mir` passes a default `--edition` flag to `mir-json` flag, which might make this sentence a bit confusing if you lack context. Perhaps it would be worth changing this sentence to something like "By default, `crux-mir` will pass `--edition=2021` to `mir-json`, but you can override this if you pass`--mir-json-arg="--edition=...".`"

crux-mir/src/Mir/MirJsonArgs.hs
sand7000: I agree, I had imagined using it in Test.hs as well but your description of per test case config files seems different enough that this won't be re-used there. I will move it.

crux-mir/README.md
sand7000: Updated.

crux-mir/src/Mir/Language.hs
sand7000: That introduces a circular dependence between Language.hs and Generate.hs so I moved MIROptions into a new Module called Options.hs. I considered combining Options.hs and Defaults.hs and can still do so if you like that idea.

crux-mir/src/Mir/Generate.hs
sand7000: Fixed.

crux-mir/src/Mir/Language.hs
sand7000: Done.

crux-mir/src/Mir/MirJsonArgs.hs
sand7000: Done.

crux-mir/src/Mir/Options.hs
```diff
@@ -0,0 +1,104 @@
...
+
+        , GetOpt.Option [] ["mir-json-arg"]
+            "pass an argument to mir-json (can be repeated)"
+            (GetOpt.ReqArg "ARG" $ \v opts ->
+                Right opts { mirJsonArgs = mirJsonArgs opts Seq.|> v })
+        ]
+    }
+
+mirJsonArgsList :: MIROptions -> [String]
+mirJsonArgsList = Foldable.toList . mirJsonArgs
```
RyanGlScott: Personally, this doesn't meet my [Fairbairn threshold](https://wiki.haskell.org/Fairbairn_threshold) of usefulness to warrant its own API function. I'd rather just call `buildMirJsonArgs baseArgs (Foldable.toList (mirJsonArgs ...))` at the call site where this function is used.

crux-mir/src/Mir/Generate.hs
```diff
@@ -66,8 +67,20 @@ mirJsonOutFile rustFile = rustFile -<.> "mir"
...
 getRlibsDir = maybe ?defaultRlibsDir id <$> lookupEnv "CRUX_RUST_LIBRARY_PATH"
 
-compileMirJson :: (?defaultRlibsDir :: FilePath) => Crux.CruxOptions -> Bool -> FilePath -> IO ()
-compileMirJson cruxOpts keepRlib rustFile = do
+buildMirJsonArgs :: [String]   -- ^ base/default args (no edition)
+                 -> [String]   -- ^ extra args (from MIROptions.mirJsonArgs)
+                 -> [String]
+buildMirJsonArgs base extras =
+  let hasEdition  = any isEditionArg extras
+      editionPart = if hasEdition then [] else [defaultRustEditionFlag]
```
RyanGlScott, line 80: This line deserves a comment, I think. It's somewhat non-obvious that `rustc` will throw an error if you supply the `--edition` flag twice, so it is worth commenting that this line exists to ensure that user-specified language editions work as expected.

RyanGlScott (approved):
Great! Two more minor suggestions, but otherwise I think this is ready to land.

crux-mir/src/Mir/Language.hs
RyanGlScott: Good call, I think having a dedicated `Mir.Options` module is wise. I am fine with having `Mir.Defaults` be separate, given that `defaultRustEdition` doesn't correspond to a `crux-mir` CLI option (which is the main purpose of `Mir.Options`).

PR #1684, opened 2025-12-17
`crucible-llvm`: Extremely limited support for `poison` constants

RyanGlScott:
LLVM features a `poison` constant, which can have any type and theoretically appear anywhere in an LLVM program. While Clang has historically avoided generating `poison` constants in well-defined programs, more recent versions of Clang have become more aggressive in inserting `poison` constants as a result of optimization passes. For instance, the macOS Homebrew version of Clang 20 optimizes the `T1177.c` in such a way that bitcode of the following form is generated:

```llvm
%50 = insertelement <8 x i32> <i32 poison, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0, i32 0>, i32 %48, i64 0
```

Note the `poison` value in the first vector lane, which is then immediately overwritten by the `insertelement` instruction. Prior to this patch, `crucible-llvm` had absolutely no support for `poison` constants, which meant that the bitcode above would crash `crucible-llvm`. This is an unfortunate state of affairs, but the reality of the situation is that getting all the details of tracking `poison` correctly is quite tricky. (See #366.) Nevertheless, it feels like we could surely do at least slightly better here.

This patch is a small step in the direction of improving support for `poison` constants. Like with `undef` values, we now add a dedicated construct for `poison` values in the `LLVMVal`, `LLVMExpr`, and `LLVMConst` data types, which allow us to reason about `poison` constants at a very rudimentary level. We also add `LLVM_{BV,Float}Poison` extension expressions that allow us to put `poison` values in the Crucible simulator. For now, attempting to actually simulate one of these `LLVM_{BV,Float}Poison` values will result in an error, as we only support `poison` constants that are immediately overwritten (like in the LLVM code snippet above). Accurately tracking `poison` values throughout simulation is left as future work.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
@@ -304,44 +304,48 @@ undefExpand :: ( ?err :: String -> a
             -> MemType
             -> (forall tp. Proxy# arch -> TypeRepr tp -> Expr LLVM s tp -> a)
             -> a
-undefExpand _archProxy (IntType w) k =
-  case mkNatRepr w of
-    Some w' | Just LeqProof <- isPosNat w' ->
-      k proxy# (LLVMPointerRepr w') $
-         BitvectorAsPointerExpr w' $
-         App $ BVUndef w'
+undefExpand archProxy t k =
+  case t of
+    IntType w ->
+      case mkNatRepr w of
+        Some w' | Just LeqProof <- isPosNat w' ->
+          k proxy# (LLVMPointerRepr w') $
+             BitvectorAsPointerExpr w' $
+             App $ BVUndef w'
 
-    _ -> ?err $ unwords ["illegal integer size", show w]
-undefExpand _archProxy (PtrType _tp) k =
-   k proxy# PtrRepr $ BitvectorAsPointerExpr PtrWidth $ App $ BVUndef PtrWidth
-undefExpand _archProxy PtrOpaqueType k =
-   k proxy# PtrRepr $ BitvectorAsPointerExpr PtrWidth $ App $ BVUndef PtrWidth
-undefExpand _archProxy (StructType si) k =
-   unpackArgs (map UndefExpr tps) $ \archProxy ctx asgn ->
-     k archProxy (StructRepr ctx) (mkStruct ctx asgn)
- where tps = map fiType $ toList $ siFields si
-undefExpand archProxy (ArrayType n tp) k =
-  llvmTypeAsRepr tp $ \tpr ->
-    unpackVec
-      archProxy
-      tpr
-      (replicate (fromIntegral n) (UndefExpr tp))
-      (k proxy# (VectorRepr tpr))
-undefExpand archProxy (VecType n tp) k =
-  llvmTypeAsRepr tp $ \tpr ->
-    unpackVec
-      archProxy
-      tpr
-      (replicate (fromIntegral n) (UndefExpr tp))
-      (k proxy# (VectorRepr tpr))
-undefExpand _archProxy FloatType k =
-  k proxy# (FloatRepr SingleFloatRepr) (App (FloatUndef SingleFloatRepr))
-undefExpand _archProxy DoubleType k =
-  k proxy# (FloatRepr DoubleFloatRepr) (App (FloatUndef DoubleFloatRepr))
-undefExpand _archProxy X86_FP80Type k =
-  k proxy# (FloatRepr X86_80FloatRepr) (App (FloatUndef X86_80FloatRepr))
-undefExpand _archPrxy MetadataType _ =
-  ?err $ unwords ["cannot undef expand metadata type"]
+        _ -> ?err $ unwords ["illegal integer size", show w]
+    PtrType _tp ->
+      k proxy# PtrRepr $
+        BitvectorAsPointerExpr PtrWidth $ App $ BVUndef PtrWidth
+    PtrOpaqueType ->
+      k proxy# PtrRepr $
+        BitvectorAsPointerExpr PtrWidth $ App $ BVUndef PtrWidth
+    StructType si ->
+      unpackArgs (map UndefExpr tps) $ \structArchProxy ctx asgn ->
+        k structArchProxy (StructRepr ctx) (mkStruct ctx asgn)
+      where tps = map fiType $ toList $ siFields si
+    ArrayType n tp ->
+      llvmTypeAsRepr tp $ \tpr ->
+        unpackVec
+          archProxy
+          tpr
+          (replicate (fromIntegral n) (UndefExpr tp))
+          (k proxy# (VectorRepr tpr))
+    VecType n tp ->
+      llvmTypeAsRepr tp $ \tpr ->
+        unpackVec
+          archProxy
+          tpr
+          (replicate (fromIntegral n) (UndefExpr tp))
+          (k proxy# (VectorRepr tpr))
+    FloatType ->
+      k proxy# (FloatRepr SingleFloatRepr) (App (FloatUndef SingleFloatRepr))
+    DoubleType ->
+      k proxy# (FloatRepr DoubleFloatRepr) (App (FloatUndef DoubleFloatRepr))
+    X86_FP80Type ->
+      k proxy# (FloatRepr X86_80FloatRepr) (App (FloatUndef X86_80FloatRepr))
+    MetadataType ->
+      ?err $ unwords ["cannot undef expand metadata type"]
 
 poisonExpand :: ( ?err :: String -> a
                 , HasPtrWidth (ArchWidth arch)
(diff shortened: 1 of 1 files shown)


crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
```diff
@@ -339,6 +339,9 @@ data LLVMConst where
   -- | The @undef@ value is quite strange. See: The LLVM Language Reference,
   -- § Undefined Values.
   UndefConst    :: !MemType -> LLVMConst
+  -- | The @undef@ value is quite poison. See: The LLVM Language Reference,
+  -- § Undefined Values.
```
sauclovian-g: I think this may have gotten scrambled :-)

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
```diff
@@ -355,13 +358,16 @@ instance Show LLVMConst where
...
+      (PoisonConst mem)   -> ["PoisonConst", show mem]
       (StringConst bs)    -> ["StringConst", show bs]
 
 -- | The interesting cases here are:
 --  * @IntConst@: GHC can't derive this because @IntConst@ existentially
 --    quantifies the integer's width. We say that two integers are equal when
 --    they have the same width *and* the same value.
 --  * @UndefConst@: Two @undef@ values aren't necessarily the same...
+--  * @PoisonConst@: A comparison involving a @poison@ value returns @poison@,
+--    so we conservatively return 'False' here.
```
sauclovian-g: That's not really conservative (imagine something that, for example, checks `v1 /= v2` to see if array indexes overlap)... it is probably not remotely safe to do so without a lot more work, but my inclination would be to panic here.

Maybe for now just change the wording...

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/MemLog.hs
```diff
@@ -677,6 +677,8 @@ concLLVMVal _ _ v@LLVMValString{} = pure v
 concLLVMVal _ _ v@LLVMValZero{} = pure v
 concLLVMVal _ _ (LLVMValUndef st) =
   pure (LLVMValZero st) -- ??? does it make sense to turn Undef into Zero?
```
sauclovian-g: Not immediately related, but: this seems like the sort of thing that should have an issue to remind someone to investigate sometime.

sauclovian-g (approved):
Looks generally plausible

sauclovian-g:
Also, I would think that executing poison is relatively straightforward: for any of the ops where it's defined to cause UB, fail, and otherwise return another poison...

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
RyanGlScott: I agree that making `poison == poison` return `False` is pretty dodgy. Then again, making `undef == undef` always return `False` is also dodgy...

Truth be told, I'm not sure why this `Eq Const` instance even exists. As far as I can tell, it is completely unused. My inclination would be to remove it entirely, given that it seems impossible to always compute a definite `True`-or-`False` answer in the presence of `undef` and `poison`.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
RyanGlScott: Clearly, my brain wasn't braining when I copy-pasted this—thanks for spotting it. Fixed.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/MemLog.hs
RyanGlScott: I've opened https://github.com/GaloisInc/crucible/issues/1685 for this. I've also mentioned this in the comments.

RyanGlScott:
> Also, I would think that executing poison is relatively straightforward: for any of the ops where it's defined to cause UB, fail, and otherwise return another poison...

It's worth noting here that the constructs introduced here (`LLVMValPoison`, `PoisonExpr`, and `PoisonConst`) are really only suited for tracking literal `poison` values that appear directly in LLVM code. Most interesting uses of poison involve `poison` involve storing poison into registers/memory and then loading them back out in subsequent instructions, and the machinery that this PR adds does not do anything to implement that level of poison tracking.

Implementing this level of poison tracking would be a pretty non-trivial endeavor, since Crucible's simulator doesn't have an built-in notion of poison-ness. We'd have to do something like extending [`LLVMAnnMap`](https://github.com/GaloisInc/crucible/blob/ec039bbf84abdd71ea6a78eca30b68f569b477d9/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs#L120) (which tracks all values that have been tagged with some sort of undefined behavior at simulation time) to also track poisoned values, and then make assertions throughout simulation at times when executing an instruction on a poisoned operand would invoke UB. This would probably be doable, but it would also be considerably more work than what this PR does. I'm already yak-shaving with this PR, so I'm going to intentionally try to leave off this part of the yak's fur for another time :)

All of this is to say: we certainly _could_ add some checks in the relevant instructions that trigger undefined behavior if one of their operands is poisoned, but with the current state of things, these checks would only fire if the operand is a literal `poison` value, not if it is an indirect source of `poison` loaded from a register or memory. As such, I am inclined to put off implementing these instruction-level checks until we commit to tracking poison more thoroughly.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
RyanGlScott: > As far as I can tell, it is completely unused.

Ah, I lied: the `Eq LLVMConst` instance is used in the `crucible-llvm` test suite to check if certain LLVM bitcode constants evaluate to the expected `LLVMConst`s. (Unsurprisingly, none of these tests check for `undef` constants.) As such, we need to have _some_ way to check `LLVMConst`s for equality. Some possible choices:

1. Keep the `Eq LLVMConst` instance as-is, making it return `False` for `undef` and `poison`.
2. Keep the `Eq LLVMConst` instance, but making it panic for `undef` and `poison`.
3. Remove the `Eq LLVMConst` instance, and change the `crucible-llvm` test suite to pattern-match on `LLVMConst`s instead of using `(==)` to check for equality.
4. Remove the `Eq LLVMConst` instance, and instead replace it with a top-level function that checks `LLVMConst`s for syntactic equality (i.e., `undef == undef` and `poison == poison` would return `True`, but `undef == 0` and `poison == 0` would return `False`). Update the `crucible-llvm` test suite to use this new function.

Thoughts on what your preference would be? I have a slight preference towards option (4), but I don't feel strongly about it.

langston-barrett: I would probably do (5), which is close to (4): Move the `(==)` to a top-level function *in the test suite* and use it there. It seems like we don't have much use for this comparison otherwise, and providing it as part of the API seems like a footgun.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
```diff
@@ -345,6 +352,7 @@ isZero sym v =
     LLVMValString bs  -> pure $ Just $ backendPred sym $ not $ isJust $ BS.find (/= 0) bs
     LLVMValZero _     -> pure (Just $ truePred sym)
     LLVMValUndef _    -> pure Nothing
+    LLVMValPoison _   -> pure Nothing
     _                 ->
```
langston-barrett: While we're here, might be nice to make this pattern match explicit for all cases. Feel free to disregard since this was already an issue.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
```diff
@@ -311,9 +331,44 @@ undefExpand _archProxy X86_FP80Type k =
...
+   k proxy# PtrRepr $ BitvectorAsPointerExpr PtrWidth $ bvPoisonExpr PtrWidth
+poisonExpand _archProxy PtrOpaqueType k =
+   k proxy# PtrRepr $ BitvectorAsPointerExpr PtrWidth $ bvPoisonExpr PtrWidth
+poisonExpand _archProxy (StructType si) k =
+   unpackArgs (map PoisonExpr tps) $ \archProxy ctx asgn -> k archProxy (StructRepr ctx) (mkStruct ctx asgn)
+ where tps = map fiType $ toList $ siFields si
+poisonExpand archProxy (ArrayType n tp) k =
+  llvmTypeAsRepr tp $ \tpr -> unpackVec archProxy tpr (replicate (fromIntegral n) (PoisonExpr tp)) $ k proxy# (VectorRepr tpr)
+poisonExpand archProxy (VecType n tp) k =
+  llvmTypeAsRepr tp $ \tpr -> unpackVec archProxy tpr (replicate (fromIntegral n) (PoisonExpr tp)) $ k proxy# (VectorRepr tpr)
```
langston-barrett: Some of these lines are getting pretty long...

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
```diff
@@ -311,9 +331,44 @@ undefExpand _archProxy X86_FP80Type k =
...
 undefExpand _archPrxy tp _ = ?err $ unwords ["cannot undef expand type:", show tp]
 
+poisonExpand :: ( ?err :: String -> a
+                , HasPtrWidth (ArchWidth arch)
+                )
+             => Proxy# arch
+             -> MemType
+             -> (forall tp. Proxy# arch -> TypeRepr tp -> Expr LLVM s tp -> a)
+             -> a
+poisonExpand _archProxy (IntType w) k =
```
langston-barrett: nbd but IMO it would be more readable to define this like:
```
poisonExpand archProxy t k =
  case t of
    IntType w -> _
    -- ...
```
This reduces the repetition of the function and argument names (making it easier to rename in the future), makes it easier to copy/paste the cases, makes it easier to share `where`-clauses among the branches, etc. Then again, it's less consistent with the rest of the file 🤷

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
```diff
@@ -311,9 +331,44 @@ undefExpand _archProxy X86_FP80Type k =
...
+  llvmTypeAsRepr tp $ \tpr -> unpackVec archProxy tpr (replicate (fromIntegral n) (PoisonExpr tp)) $ k proxy# (VectorRepr tpr)
+poisonExpand archProxy (VecType n tp) k =
+  llvmTypeAsRepr tp $ \tpr -> unpackVec archProxy tpr (replicate (fromIntegral n) (PoisonExpr tp)) $ k proxy# (VectorRepr tpr)
+poisonExpand _archProxy FloatType k =
+  k proxy# (FloatRepr SingleFloatRepr) (floatPoisonExpr SingleFloatRepr)
+poisonExpand _archProxy DoubleType k =
+  k proxy# (FloatRepr DoubleFloatRepr) (floatPoisonExpr DoubleFloatRepr)
+poisonExpand _archProxy X86_FP80Type k =
+  k proxy# (FloatRepr X86_80FloatRepr) (floatPoisonExpr X86_80FloatRepr)
+poisonExpand _archPrxy tp _ = ?err $ unwords ["cannot poison expand type:", show tp]
```
langston-barrett: Perhaps worth explicitly enumerating these types?

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
kquick: I agree with Langston here: option 5

crucible-llvm/src/Lang/Crucible/LLVM/Extension/Syntax.hs
```diff
@@ -256,11 +269,16 @@ instance PrettyApp LLVMExtensionExpr where
         pretty "pointerOffset" <+> pp ptr
       LLVM_PointerIte _ cond x y ->
         pretty "pointerIte" <+> pp cond <+> pp x <+> pp y
+      LLVM_BVPoison _ ->
+        pretty "bvPoison"
```
kquick: Minor nitpick but I'd prefer the more regular "adjective-noun" combination: "poisonBV" and "poisonFloat".

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs
```diff
@@ -1009,6 +1024,9 @@ muxLLVMVal sym = merge sym muxval
     muxval _ v1@(LLVMValUndef tp1) (LLVMValUndef tp2)
       | tp1 == tp2 = pure v1
 
+    muxval _ v1@(LLVMValPoison tp1) (LLVMValPoison tp2)
+      | tp1 == tp2 = pure v1
```
kquick, line 1028: Perhaps this gets to the execution support that you were trying to avoid, but my intuition is that muxing a poison with anything results in poison (this has that result only if both sides are poison).  It may be that I misunderstand what's happening here though.

kquick (approved):

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Partial.hs
RyanGlScott: It's not evident to me that `muxval poison x` should always return `poison` for all `x`. I have an example like this (in pseudo-C code) in mind when writing this comment:

```c
int foo(void) {
  uint8_t x = crucible_uint8_t("x");
  if (x) {
    return poison;
  }
  return 0;
}
```

Intuitively, `foo()` should return `poison` in some (but not all) executions of `foo`. If we mux `poison` with `0` and return `poison`, however, then `crucible-llvm` would conclude that `foo()` always returns `poison`, which I believe is incorrect.

This is a bit of a moot point for now, since `crucible-llvm` isn't currently equipped to reason about `poison` across LLVM statement boundaries. If we do expand `crucible-llvm`'s ability to reason about `poison` in the future, however, then we'll have to tackle this problem for real.

crucible-llvm/src/Lang/Crucible/LLVM/Extension/Syntax.hs
RyanGlScott: I've renamed `LLVM_{BV,Float}Poison` to `LLVM_Poison{BV,Float}`.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Constant.hs
RyanGlScott: I've implemented (5), adding a `llvmConstSyntacticEq :: LLVMConst -> LLVMConst -> Bool` function to the `crucible-llvm` test suite for the sake of checking `LLVMConst`s for syntactic equality. As part of this, I needed to refactor some of the code in the `crucible-llvm` test suite to make it possible to apply this function directly on the `LLVMConst`s, instead of indirectly checking for equality by relying on the `Eq` instance for `Map ... (Maybe LLVMConst)`.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Value.hs
RyanGlScott: Done in https://github.com/GaloisInc/crucible/pull/1684/commits/048bb66ac958e4498f95158a46f0a780cd2aef1f.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
RyanGlScott: I've made an effort to wrap some of the long lines in `{undef,poison}Expand` in https://github.com/GaloisInc/crucible/pull/1684/commits/dfafb8a43806f8e3c7bfde599ace17d87b39a435.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
RyanGlScott: Good idea. As it turns out, the only case that this ever matches is `MetadataType`, so avoiding a wildcard pattern here is simple. I've done this in https://github.com/GaloisInc/crucible/pull/1684/commits/42b405781192a613b519beffe6b4fef662a8dc75.

crucible-llvm/src/Lang/Crucible/LLVM/Translation/Expr.hs
RyanGlScott: Good idea. Done in https://github.com/GaloisInc/crucible/pull/1684/commits/4a4472c6757f7d1bc48378afa7aed08018897cc2.

PR #1686, opened 2025-12-18
crux-mir: add coroutine support

spernsteiner:
Adds full support for `TyCoroutine` and `AKCoroutine`.  This includes support for defining and using `async fn`, `async` closures, and `#[coroutine]`s.  See the added tests for examples.

Within crucible-mir, `TyCoroutine` is represented as a struct with several fields:
* A discriminant.  This is a bitvector that indicates the current state of the coroutine's state machine.
* Upvars.  These are ordinary struct fields, meaning they're wrapped in `MaybeType` only if the field type is not default-initializable according to `canInitialize`.
* Saved locals.  These fields are all possibly uninitialized, depending on the current state/discriminant, so they're always wrapped in `MaybeType` regardless of `canInitialize`.

`AKCoroutine` expects a number of operands equal to the number of upvars.  It sets the discriminant to 0, sets the upvars to the provided operand values, and leaves all saved locals uninitialized.

Field access on `TyCoroutine` is a bit unusual.  Types like structs, tuples, and closures only support direct field access (downcasting to a variant is forbidden); enums only support accessing fields of specific variants (downcasting is required).  `TyCoroutine` supports access with or without a downcast, and the two kinds of accesses have different effects.  Field access without a downcast accesses an upvar; for example, `co.0` accesses the first upvar.  Field access with a downcast accesses a saved local, with an indirection through a table that maps each (variant index, field index) pair to an index into the saved locals.  This mapping is not necessarily a bijection, so `(co as variant#1).2` and `(co as variant#3).4` could both refer to the sixth saved local.

### DIFF
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -113,9 +113,9 @@ instance Pretty VariantDiscr where
 instance Pretty CoroutineArgs where
   pretty (CoroutineArgs discrTy upvarTys savedTys fieldMap) =
     pretty discrTy
-    <+> list (map pretty upvarTys)
-    <+> list (map pretty savedTys)
-    <+> list (map viaShow $ Map.toList fieldMap)
+    <+> pretty upvarTys
+    <+> pretty savedTys
+    <+> pretty (Map.toList fieldMap)
 
 
 instance Pretty CtorKind where


crucible-mir/src/Mir/Mir.hs
```diff
@@ -126,6 +122,14 @@ data Ty =
       | TyInterned TyName
       deriving (Eq, Ord, Show, Generic)
 
+data CoroutineArgs = CoroutineArgs
+  { _caDiscrTy :: !Ty
+  , _caUpvarTys :: ![Ty]
+  , _caSavedTys :: ![Ty]
+  , _caFieldMap :: !(Map (Int, Int) Int)
+  }
```
RyanGlScott, line 140: Please add some Haddocks here. It would also be helpful to link each field to the relevant part of `Note [coroutine representation]`.

crucible-mir/src/Mir/ParseTranslate.hs
```diff
@@ -48,7 +48,7 @@ import Debug.Trace
 -- If you update the supported mir-json schema version below, make sure to also
 -- update the crux-mir README accordingly.
 supportedSchemaVersion :: Word64
-supportedSchemaVersion = 7
+supportedSchemaVersion = 8
```
RyanGlScott, line 51: Make sure to also update the schema numbers in the following places:

* https://github.com/GaloisInc/crucible/blob/658975044d77918d6dfb3257364c3e2d14c8c6b4/crux-mir/README.md?plain=1#L51-L53
* https://github.com/GaloisInc/crucible/blob/658975044d77918d6dfb3257364c3e2d14c8c6b4/crux-mir/CHANGELOG.md?plain=1#L3-L5

crucible-mir/src/Mir/PP.hs
```diff
@@ -77,7 +77,7 @@ instance Pretty Ty where
     pretty (TyAdt _ origDefId tys)    = pr_id origDefId <> pretty tys
     pretty (TyFnDef defId)       = pretty "fnDef" <+> pr_id defId
     pretty (TyClosure tys)       = pretty "closure" <+> pretty tys
-    pretty TyCoroutine = pretty "coroutine"
+    pretty (TyCoroutine ca) = pretty "coroutine" <+> viaShow ca
```
RyanGlScott: I do think it would be worth adding a `Pretty CoroutineArgs` instance and using it here (and in the code for pretty-printing `AKCoroutine`s below) rather than relying on the `Show` instance, which will likely look pretty jarring when intermixed with other pretty-printed MIR code.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -311,47 +311,245 @@ tyToReprM ty = do
...
+
+coroutineFieldsM ::
+  TransTyConstraint =>
+  M.CoroutineArgs ->
+  MirGenerator h s ret (Some FieldCtxRepr)
+coroutineFieldsM ca = do
+  col <- use $ cs . collection
+  case coroutineFields col ca of
+    Left err -> mirFail ("coroutineFields: " ++ err)
+    Right x -> return x
```
RyanGlScott, line 366:
```suggestion
coroutineFieldsM ca = do
  col <- use $ cs . collection
  case coroutineFields col ca of
    Left err -> mirFail ("coroutineFieldsM: " ++ err)
    Right x -> return x
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -311,47 +311,245 @@ tyToReprM ty = do
...
+  M.CoroutineArgs ->
+  MirGenerator h s ret (Some C.CtxRepr)
+coroutineFieldTypesM ca = do
+  Some fieldCtx <- coroutineFieldsM ca
+  return $ Some (fieldCtxType fieldCtx)
+
+data CoroutineInfo = forall ctx tp' tp. CoroutineInfo
+  (C.CtxRepr ctx)
+  (Ctx.Index ctx tp')
+  (FieldKind tp tp')
```
RyanGlScott, line 392: Some Haddocks would be helpful here. What do the type parameters and fields represent?

crux-mir/test/conc_eval/async/async.rs
```diff
@@ -1,24 +1,11 @@
...
-// Minimal waker; OK since `f()` is immediately Ready (no .await inside).
-fn dummy_waker() -> Waker {
-    unsafe fn clone(_: *const ()) -> RawWaker { RawWaker::new(std::ptr::null(), &VTABLE) }
-    unsafe fn wake(_: *const ()) {}
-    unsafe fn wake_by_ref(_: *const ()) {}
-    unsafe fn drop(_: *const ()) {}
-    static VTABLE: RawWakerVTable = RawWakerVTable::new(clone, wake, wake_by_ref, drop);
-    unsafe { Waker::from_raw(RawWaker::new(std::ptr::null(), &VTABLE)) }
-}
-
```
RyanGlScott, line 18: I see that you had to delete some code in this test case—was there something problematic in this code that `crucible-mir` couldn't handle? Anything worth recording in a Crucible issue to tackle later?

spernsteiner: `dummy_waker()` works fine, but it does basically the same thing as the standard library's `Waker::noop()`, so I replaced it with a call to the standard library version.

crucible-mir/src/Mir/Mir.hs
spernsteiner: Added doc comments in 76b95d18392bc2d1cf523430758add9802a2371d.

I tried referencing `Note [coroutine representation]` like I saw done for `Note [union representation]`, but neither note shows up in the `cabal haddock` output for `Mir.TransTy`.  Is this the expected behavior?

crucible-mir/src/Mir/ParseTranslate.hs
spernsteiner: Updated README and CHANGELOG in 8f41f5b619a539045a125686e6cadfdf29251546

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Added doc comments in 76b95d18392bc2d1cf523430758add9802a2371d

crucible-mir/src/Mir/Mir.hs
RyanGlScott: Notes aren't natively integrated into Haddock (alas)—they're just a convention we've adopted when writing comments.

crucible-mir/src/Mir/PP.hs
spernsteiner: Added a `Pretty` instance in 4829eefaf, which produces output like this:
```Rust
   let mut _7 : $lang/0::Pin[0]<&mut coroutine u32 [] [ coroutine_closure []
   , u32
   , coroutine u32 [u32] [] []
   , coroutine u32 [u32] [] [] ] [ ((3,0),0)
   , ((3,1),2)
   , ((4,0),0)
   , ((4,1),1)
   , ((4,2),3) ]>;
```
This coroutine has a `u32` discriminant, no upvars, four saved locals, and five entries in the variant field map.

spernsteiner:
I've addressed all of the feedback so far.

crucible-mir/src/Mir/PP.hs
```diff
@@ -110,6 +110,13 @@ instance Pretty VariantDiscr where
   pretty (Explicit a) = pretty_fn1 "Explicit" a
   pretty (Relative a) = pretty_fn1 "Relative" a
 
+instance Pretty CoroutineArgs where
+  pretty (CoroutineArgs discrTy upvarTys savedTys fieldMap) =
+    pretty discrTy
+    <+> list (map pretty upvarTys)
+    <+> list (map pretty savedTys)
+    <+> list (map viaShow $ Map.toList fieldMap)
```
RyanGlScott: Any reason not to just call `pretty` on these lists directly?


```suggestion
    <+> pretty upvarTys
    <+> pretty savedTys
    <+> pretty (Map.toList fieldMap)
```

RyanGlScott (approved):

crucible-mir/src/Mir/PP.hs
spernsteiner: Didn't realize that was possible.  Fixed in b591918

spernsteiner:
I'm planning to merge this and https://github.com/GaloisInc/mir-json/pull/214 once CI passes

PR #1695, opened 2026-01-06
`crux-mir`: support concretizing `repr(transparent)` structs

### LINKED ISSUE #1680: `crux-mir`: `crux::test`s that return `repr(transparent)` structs or enums cause simulation errors
In most circumstances, `crux-mir` is able to simulation `crux::test`s that return structs or enums:

```rs
struct S(u32);

#[crux::test]
pub fn s() -> S {
    S(0)
}

enum E {
    V(u32),
}

#[crux::test]
pub fn e() -> E {
    E::V(0)
}
```
```
test test/9be9a1c7::e[0]: returned V(0), ok
test test/9be9a1c7::s[0]: returned S(0), ok

[Crux] Overall status: Valid.
```

If you mark a struct or enum as `#[repr(transparent)]`, however:

```rs
#[repr(transparent)]
struct S(u32);

#[crux::test]
pub fn s() -> S {
    S(0)
}

#[repr(transparent)]
enum E {
    V(u32),
}

#[crux::test]
pub fn e() -> E {
    E::V(0)
}
```

Then `crux-mir` will fail to simulate them:

```
test test/07ae6d14::e[0]: [Crux] Attempting to prove verification conditions.
FAILED
test test/07ae6d14::s[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/07ae6d14::e[0] counterexamples ----
[Crux]   test.rs:17:2: 17:2: error: in test/07ae6d14::e[0]
[Crux]   expected enum to have typeStructRepr [BVRepr 64, VariantRepr [StructRepr [BVRepr 32]]], but got BVRepr 32

---- test/07ae6d14::s[0] counterexamples ----
[Crux]   test.rs:7:2: 7:2: error: in test/07ae6d14::s[0]
[Crux]   expected struct to have typeStructRepr [BVRepr 32], but got BVRepr 32

[Crux] Goal status:
[Crux]   Total: 2
[Crux]   Proved: 0
[Crux]   Disproved: 2
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

samcowger:
Fixes #1680, and adds a few more test cases for `concretize` applied to other types.

### DIFF
--- a/crux-mir/src/Mir/Language.hs
+++ b/crux-mir/src/Mir/Language.hs
@@ -698,7 +698,9 @@ showRegEntry col mty entry@(C.RegEntry tp rv) =
            in return $ varName ++ " { " ++ List.intercalate ", " fieldStrs' ++ " }"
 
     -- Render the value that would inhabit the given zero-sized type, failing if
-    -- the type is not zero-sized.
+    -- the type is not zero-sized, or if the type is uninhabitable. As such,
+    -- note that this only succeeds on a subset of the types that satisfy
+    -- `isZeroSized`.
     showZSTValue :: Ty -> C.OverrideSim p sym MIR rtp args ret String
     showZSTValue ty =
       case ty of


crux-mir/src/Mir/Language.hs
```diff
@@ -552,6 +552,30 @@ showRegEntry col mty (C.RegEntry tp rv) =
 
     -- Tagged union type
     (TyAdt name _ _, _)
+      | Just adt <- List.find (\(Adt n _ _ _ _ _ _) -> name == n) (col ^. adts)
```
RyanGlScott: It might be worth factoring out this `List.find (\(Adt n _ _ _ _ _ _) -> name == n) (col ^. adts)` subexpression into its own local definition, especially since it's now used multiple times in this function.

crux-mir/src/Mir/Language.hs
```diff
@@ -643,26 +667,56 @@ showRegEntry col mty (C.RegEntry tp rv) =
...
+          vs <- mapM showZSTValue ts
+          pure $ "(" <> List.intercalate ", " vs <> ")"
+        TyAdt monoName _ _
+          | Just adt <- List.find (\(Adt n _ _ _ _ _ _) -> monoName == n) (col ^. adts) -> do
+              let variant = onlyVariant adt
+              let fieldTys = variant ^.. vfields . each . fty
+              fieldStrs <- mapM showZSTValue fieldTys
+              showVariant' variant fieldStrs
+          | otherwise ->
+              fail $ "showZST: unknown ADT"
```
RyanGlScott: Should `showZST` be `showZSTValue` instead? (And similarly for the catch-all case directly below this one.)

crux-mir/src/Mir/Language.hs
```diff
@@ -643,26 +667,56 @@ showRegEntry col mty (C.RegEntry tp rv) =
...
+          pure "[]"
+        TyTuple ts -> do
+          vs <- mapM showZSTValue ts
+          pure $ "(" <> List.intercalate ", " vs <> ")"
+        TyAdt monoName _ _
+          | Just adt <- List.find (\(Adt n _ _ _ _ _ _) -> monoName == n) (col ^. adts) -> do
+              let variant = onlyVariant adt
+              let fieldTys = variant ^.. vfields . each . fty
+              fieldStrs <- mapM showZSTValue fieldTys
+              showVariant' variant fieldStrs
```
RyanGlScott, line 717: Shouldn't there also be a case for structs with no fields (e.g., `struct S {}`)?

I think enums with no variants (as well as the `!` type) are also represented as zero-sized types, but since you can't construct values of these types, I think it's fine to exclude those. (Perhaps we should mention in the comments that the cases here are a proper subset of the cases for which `isZeroSized` returns `True`.)

samcowger: > Shouldn't there also be a case for structs with no fields (e.g., `struct S {}`)?

I think this works appropriately as-is for empty structs - what kind of special treatment did you think they needed, exactly?

> I think enums with no variants (as well as the `!` type) are also represented as zero-sized types, but since you can't construct values of these types, I think it's fine to exclude those. (Perhaps we should mention in the comments that the cases here are a proper subset of the cases for which `isZeroSized` returns `True`.)

Yes, I agree. I'll clarify the relationship with `isZeroSized`, and also mention that it will fail on uninhabitable types.

RyanGlScott: > I think this works appropriately as-is for empty structs

You are correct, my apologies! I think I confused myself when reading this code into thinking that this required there to be at least one field, but clearly that's not the case. (There's even an empty `ZST0` struct in your test case which proves this.)

crux-mir/src/Mir/Language.hs
samcowger: Yes, it should, I'll fix that.

crux-mir/src/Mir/Language.hs
samcowger: Yes, it probably is. It might even be a reimplementation of `Mir.Generator.findAdt` - if so, I'll try to use that instead.

Issue #1696, opened 2026-01-07
`crucible-mir`: Simplify `ctlz`/`ctlz_nonzero` override implementations


RyanGlScott:
`crucible-mir`'s custom overrides for `ctlz` and `ctlz_nonzero` are implemented like so:

https://github.com/GaloisInc/crucible/blob/496e4ce9a917691718fd3b7263164683b9e8832b/crucible-mir/src/Mir/TransCustom.hs#L1031-L1073

There are a couple of aspects of this implementation that I think are more complicated than they need to be:

1. `ctlz_impl` is parameterized by a `Maybe (Some NatRepr)` argument to allow picking whether to use a fixed output width (`Just`) or to make the output width match the input width (`Nothing`). In practice, however, all instantiations of `ctlz_impl` instantiate this argument with `Just (knownNat @32)`. As a result, this added generality goes unused. We should specialize `ctlz_impl` to always return `u32`.
2. The bulk of `ctlz_impl`'s code is dedicated to counting the number of leading zeros. This is wasted effort, however, because `crucible` already includes a way to count leading zeroes via the [`BVCountLeadingZeros`](https://github.com/GaloisInc/crucible/blob/496e4ce9a917691718fd3b7263164683b9e8832b/crucible/src/Lang/Crucible/CFG/Expr.hs#L824-L830) expression. We could (and should) delete much of this zero-counting logic in favor of `BVCountLeadingZeros`, which would save quite a bit of code.

   One minor complication is that `BVCountLeadingZeros` always uses the same output width as the input width, unlike what we want in the `ctlz` override. It would be fairly straightforward to work around this by zero-extending or truncating the result of `BVCountLeadingZeros` as needed, however. In fact, we already do such a thing in the [override for `crucible::bitvector::convert`](https://github.com/GaloisInc/crucible/blob/496e4ce9a917691718fd3b7263164683b9e8832b/crucible-mir/src/Mir/TransCustom.hs#L1689-L1694), so we could factor out this code and use it in `ctlz_impl` as well.

PR #1697, opened 2026-01-07
`crucible-mir`: Clean up `ctlz` overrides, add `cttz` overrides

### LINKED ISSUE #1689: `crucible-mir`: Add custom overrides for `cttz` and `cttz_nonzero`
Rust's primitive integer types feature [`trailing_zeroes`](https://doc.rust-lang.org/std/primitive.u32.html#method.trailing_zeros) and [`trailing_ones`](https://doc.rust-lang.org/std/primitive.u32.html#method.trailing_ones) methods, both of which are defined in terms of the [`cttz` intrinsic](https://doc.rust-lang.org/std/intrinsics/fn.cttz.html):

```rs
pub const fn cttz<T: Copy>(x: T) -> u32;
```

This intrinsic counts the number of trailing unset bits (zeroes) in an integer value. Currently, `crucible-mir` lacks a custom override for this intrinsic, which means that it cannot simulate code involving `trailing_{zeroes,ones}`. Adding one would be pretty straightforward, however, as we could use `what4`'s [`BVCountTrailingZeros`](https://hackage-content.haskell.org/package/what4-1.7.2/docs/What4-Expr.html#v:BVCountTrailingZeros) operation.

By a similar token, we should also implement a custom override for the [`cttz_nonzero`](https://doc.rust-lang.org/std/intrinsics/fn.cttz_nonzero.html) intrinsic. This behaves like `cttz`, but it returns an undefined value when its argument is zero. See the [`crucible-llvm` override for `llvm.cttz.i*`](https://github.com/GaloisInc/crucible/blob/109f1627fa573aa39bd9bba4ab92ef8d153063a3/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/LLVM.hs#L2093-L2109) (which also requires the argument to be non-zero) for prior art.

RyanGlScott:
Simplify the implementations of the custom overrides for `ctlz` and `ctlz_nonzero` to be defined in terms of `crucible`'s existing `BVCountLeadingZeros` operation, and then add new overrides for `cttz` and `cttz_nonzero` defined in terms of the `BVCountTrailingZeros` operation.

Fixes https://github.com/GaloisInc/crucible/issues/1696. Fixes #1689.

### DIFF
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -109,6 +109,8 @@ customOpDefs = Map.fromList $ [
                          , unchecked_shr
                          , ctlz
                          , ctlz_nonzero
+                         , cttz
+                         , cttz_nonzero
                          , rotate_left
                          , rotate_right
                          , size_of
@@ -1052,6 +1054,31 @@ ctlz =
 ctlz_nonzero :: (ExplodedDefId, CustomRHS)
 ctlz_nonzero =
     (["core", "intrinsics", "ctlz_nonzero"], ctlz_impl "ctlz_nonzero")
+
+-- Build a \"count trailing zeros\" implementation. The function will be
+-- polymorphic, accepting bitvectors of any width. The function will return
+-- a width-32 bitvector, as everything that makes use of this functionality
+-- hardcodes this width.
+cttz_impl :: Text -> CustomRHS
+cttz_impl name _substs = Just $ CustomOp $ \_optys ops -> case ops of
+    [MirExp (C.BVRepr w) v] -> do
+        let trailingZeros = S.app $ E.BVCountTrailingZeros w v
+        let w32 = knownNat @32
+        -- Because BVCountTrailingZeros returns a bitvector with the same width
+        -- as the input, we may need to extend or truncate it to make it have a
+        -- width of 32. There is no risk of information loss here due to
+        -- truncation, as 32 bits is more than enough to represent the number
+        -- of digits for all supported primitive integer and Bv types.
+        return $ MirExp (C.BVRepr w32) $ bv_convert_impl w trailingZeros w32
+    _ -> mirFail $ "BUG: invalid arguments to " ++ Text.unpack name ++ ": " ++ show ops
+
+cttz :: (ExplodedDefId, CustomRHS)
+cttz =
+    (["core", "intrinsics", "cttz"], cttz_impl "cttz")
+
+cttz_nonzero :: (ExplodedDefId, CustomRHS)
+cttz_nonzero =
+    (["core", "intrinsics", "cttz_nonzero"], cttz_impl "cttz_nonzero")
 
 rotate_left :: (ExplodedDefId, CustomRHS)
 rotate_left = ( ["core","intrinsics", "rotate_left"],
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -30,6 +30,10 @@ This release supports [version
   pointer is greater than or equal to the second.  If the offset would be
   negative (i.e., the first pointer is less than the second), this is undefined
   behavior.
+* Add intrinsics for
+  [`cttz`](https://doc.rust-lang.org/std/intrinsics/fn.cttz.html) and
+  [`cttz_nonzero`](https://doc.rust-lang.org/std/intrinsics/fn.cttz_nonzero.html),
+  which compute the number of trailing zeros in an integer value.
 * Replace the `MirVector` type with `MirAggregate`. As part of this change, the
   following have been removed:
 
--- a/crux-mir/test/conc_eval/prim/trailing_zeros.rs
+++ b/crux-mir/test/conc_eval/prim/trailing_zeros.rs
@@ -1,0 +1,30 @@
+// A regression test for #1689.
+//
+// `trailing_zeros` (implemented for primitive integer types) is implemented in
+// terms of the `cttz` intrinsic, so this test case does some basic validation
+// of crucible-mir's custom override for this intrinsic.
+macro_rules! trailing_zeros_tests {
+    ($($ty:ty,)*) => {
+        $(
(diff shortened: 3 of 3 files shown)


crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1028,59 +1030,55 @@ exact_div =
...
+ctlz_impl :: Text -> CustomRHS
+ctlz_impl name _substs = Just $ CustomOp $ \_optys ops -> case ops of
+    [MirExp (C.BVRepr w) v] -> do
+        let leadingZeros = S.app $ E.BVCountLeadingZeros w v
+        let w32 = knownNat @32
+        -- Because BVCountLeadingZeros returns a bitvector with the same width
+        -- as the input, we may need to extend or truncate it to make it have a
+        -- width of 32. There is no risk of information loss here due to
+        -- truncation, as 32 bits is more than enough to represent the number
+        -- of digits for primitive integer types up to 128.
```
spernsteiner: It's also large enough for types like `crucible::Bv512`, which use `BVRepr` and thus will be accepted by this function.

RyanGlScott: Good catch. I've changed this phrasing to "all supported primitive integer and Bv types".

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1028,59 +1030,55 @@ exact_div =
...
-ctlz_impl :: Text -> Maybe (Some NatRepr) -> CustomRHS
-ctlz_impl name optFixedWidth _substs = Just $ CustomOp $ \_optys ops -> case ops of
-    [MirExp (C.BVRepr w) v] -> case optFixedWidth of
-        Nothing ->
-            return $ MirExp (C.BVRepr w) $ S.app $ buildMux w w w v
-        Just (Some w')
-          | Just LeqProof <- isPosNat w' ->
-            return $ MirExp (C.BVRepr w') $ S.app $ buildMux w w w' v
-          | otherwise -> error $ "bad output width " ++ show w' ++ " for ctlz_impl"
+-- Build a \"count leading zeros\" implementation. The function will be
```
samcowger, line 1033:
```suggestion
-- | Build a \"count leading zeros\" implementation. The function will be
```

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1028,59 +1030,55 @@ exact_div =
...
-    , ctlz_impl "ctlz" (Just $ Some $ knownNat @32) )
+    (["core", "intrinsics", "ctlz"], ctlz_impl "ctlz")
 
 ctlz_nonzero :: (ExplodedDefId, CustomRHS)
 ctlz_nonzero =
-    ( ["core","intrinsics", "ctlz_nonzero"]
-    , ctlz_impl "ctlz_nonzero" (Just $ Some $ knownNat @32) )
+    (["core", "intrinsics", "ctlz_nonzero"], ctlz_impl "ctlz_nonzero")
+
+-- Build a \"count trailing zeros\" implementation. The function will be
```
samcowger, line 1058:
```suggestion
-- | Build a \"count trailing zeros\" implementation. The function will be
```

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1686,15 +1684,26 @@ bv_convert = (["crucible", "bitvector", "convert"], \substs -> case substs of
...
+  (1 <= inputW, 1 <= outputW) =>
+  NatRepr inputW ->
+  R.Expr MIR s (C.BVType inputW) ->
+  NatRepr outputW ->
+  R.Expr MIR s (C.BVType outputW)
+bv_convert_impl w1 v w2 =
+  case compareNat w1 w2 of
+      NatLT _ -> S.app $ E.BVZext w2 w1 v
+      NatGT _ -> S.app $ E.BVTrunc w2 w1 v
+      NatEQ -> v
```
samcowger, line 1705: This is somewhat duplicative of `Mir.Trans.extendUnsignedBV`, which is used for this very purpose in our `ctpop` intrinsic (as well as in other places, though mysteriously not in `bv_convert`): https://github.com/GaloisInc/crucible/blob/496e4ce9a917691718fd3b7263164683b9e8832b/crucible-mir/src/Mir/TransCustom.hs#L2130-L2139

You might consider using that instead in your intrinsics, as well as in `bv_convert`.

crux-mir/test/conc_eval/prim/leading_zeros.rs
```diff
@@ -0,0 +1,28 @@
+// `leading_zeros` (implemented for primitive integer types) is implemented in
+// terms of the `ctlz` intrinsic, so this test case does some basic validation
+// of crucible-mir's custom override for this intrinsic.
+macro_rules! leading_zeros_tests {
+    ($($ty:ty,)*) => {
+        $(
+            let n = <$ty>::MAX >> 2;
+            assert_eq!(n.leading_zeros(), 2);
```
samcowger, line 8: We could make this test a little bit more focused by adding `#![feature(core_intrinsics)]` and calling the intrinsic directly, as in:
```suggestion
            assert_eq!(core::intrinsics::ctlz(n), 2);
```

Ditto for the trailing zeros test.

crux-mir/test/conc_eval/prim/leading_zeros.rs
```diff
@@ -0,0 +1,28 @@
...
+            let max = <$ty>::MAX;
+            assert_eq!(max.leading_zeros(), 0);
+        )*
+    }
+}
+
+#[cfg_attr(crux, crux::test)]
+pub fn crux_test() {
+    leading_zeros_tests! {
+        u8, u16, u32, u64, u128, usize,
```
samcowger, line 22: We might also want to test this on signed types.

samcowger (commented):
I see I wasn't quite quick enough with this review, but I'll offer it post-merge anyway for posterity. I don't think any of it is pressing enough to warrant special follow-up.

PR #1699, opened 2026-01-07
Release pre-phase submodule update.

aschwerdfeger-galois:


RyanGlScott:
The test suite failures are due to line numbers in `mir-json`'s standard libraries changing as a result of https://github.com/GaloisInc/mir-json/pull/217. These changes are benign, so you can accept the new line numbers by running `cabal run crux-mir-test -- --accept`.

Issue #1666, opened 2025-12-08
`crucible-mir`: more realistic `MirAggregate` layout


samcowger:
As implemented, `MirAggregate`s (see #1499) support storing elements of any byte width at any offset, but in practice, we've been hardcoding elements as being one byte wide and adjacent to one another. In most places, this has been denoted with a comment like `TODO: hardcoded size=1` or similar. We should try to give aggregates more realistic layout information.

This is easier with `MirAggregate`s representing Rust arrays, since array layout rules are [straightforward](https://doc.rust-lang.org/reference/type-layout.html#array-layout) and we already have access to the type size/alignment information needed to implement them. It's more challenging for other structures like ADTs, since Rust makes [few layout guarantees](https://doc.rust-lang.org/reference/type-layout.html#the-rust-representation) for structures without `#[repr(...)]` pragmas. We'd need access to the particular layout chosen by `rustc`, which we don't yet have - see https://github.com/GaloisInc/mir-json/issues/184.

sauclovian-g:
Isn't it better to _not_ model things the compiler doesn't guarantee?

spernsteiner:
> Isn't it better to _not_ model things the compiler doesn't guarantee?

Unfortunately it's possible in Rust to write code that uses layout information during const evaluation, which is difficult for us to intercept ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=110561e006077a7007317c66cfb53675)):

```Rust
struct S {
    x: u8,
    y: u16,
    z: u32,
}

static OFFSETS: &[usize] = &[
    offset_of!(S, x),
    offset_of!(S, y),
    offset_of!(S, z),
];

fn get_y(s: &S) -> &u16 {
    unsafe {
        &*(s as *const S).byte_add(OFFSETS[1]).cast::<u16>()
    }
}
```

The MIR output includes the offset of `y`, which happens to be 4 under the current `#[repr(Rust)]` layout algorithm:

```
alloc3 (size: 24, align: 8) {
    0x00 │ 06 00 00 00 00 00 00 00 04 00 00 00 00 00 00 00 │ ................
    0x10 │ 00 00 00 00 00 00 00 00                         │ ........
}
```

Ideally we'd want the value to be some abstract representation of "the offset of field `y`", but we don't have a good way to do that without invasive changes to rustc.

[...]

PR #1700, opened 2026-01-07
`crux-mir`: discriminate on MIR type during concretization

samcowger:
This refactors `crux-mir`'s `Mir.Overrides.regEval` to give access to the MIR `Type` of a value being concretized, in addition to its `TypeRepr`.

This is largely meant to be a refactoring and functionality-preserving change to help give access to more type information at concretization time, to support work on #1666. However, it does restrict concretization a bit, as `regEval` conducts stricter typechecking of values being concretized. I've included an xfail test demonstrates one such bit of restriction on slices.

Note to reviewers: this includes a fair amount of whitespace-only changes, so it's probably easier to review it commit by commit.

### DIFF
--- a/crux-mir/src/Mir/Overrides.hs
+++ b/crux-mir/src/Mir/Overrides.hs
@@ -78,7 +78,11 @@ import Mir.FancyMuxTree
 import Mir.Generator (CollectionState, collection, handleMap, MirHandle(..))
 import Mir.Intrinsics
 import qualified Mir.Mir as M
-import Mir.TransTy (tyToRepr, pattern CTyVector)
+import Mir.TransTy
+  ( reprTransparentFieldTy,
+    tyToRepr,
+    pattern CTyVector
+  )
 
 
 getString :: forall sym rtp args ret p. (IsSymInterface sym) =>
@@ -243,7 +247,9 @@ regEval bak baseEval col ty = go (Just ty)
     go Nothing (VectorRepr tpr) vec =
       traverse (go Nothing tpr) vec
 
-    go _tyM (StructRepr ctx) v = Ctx.zipWithM (go' Nothing) ctx v
+    go (Just (M.TyAdt adtName _ _)) tpr v = goAdt adtName tpr v
+    go Nothing (StructRepr ctx) v = Ctx.zipWithM (go' Nothing) ctx v
+
     go _tyM (VariantRepr ctx) v = Ctx.zipWithM (goVariantBranch Nothing) ctx v
     go tyM tpr@(ReferenceRepr _tpr) v = do
       -- Can't use `collapseMuxTree` here since it's in the IO monad, not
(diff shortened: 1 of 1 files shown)


crux-mir/src/Mir/Overrides.hs
```diff
@@ -461,7 +614,10 @@ bindFn symOnline cs name cfg
   | hasInstPrefix ["crucible", "concretize"] explodedName
   , Empty :> tpr <- cfgArgTypes cfg
   , Just Refl <- testEquality tpr (cfgReturnType cfg)
-  = bindFnHandle (cfgHandle cfg) $ UseOverride $ mkOverride' "concretize" tpr $ concretize symOnline
+  = let fns = cs ^. collection . M.functions
+        fn = fns Map.! textId name
```
RyanGlScott, line 618: Worth leaving a comment here about why this use of `Map.!` (a partial function) is expected to succeed. It might also be worth making this panic in the event that the lookup fails, as `Map.!` won't give a very helpful error message if it does fail.

crux-mir/src/Mir/Overrides.hs
```diff
@@ -177,239 +185,384 @@ concretize (Just (SomeOnlineSolver bak)) = do
...
-        OverrideSim p sym MIR rtp args ret (RegValue sym tp')
-    go tpr v | AsBaseType btr <- asBaseType tpr = baseEval btr v
+    -- Note: the `Maybe M.Ty` arguments here, and in the various recursors
+    -- below, represent the original MIR type of the value being
+    -- evaluated/concretized, if it's known/applicable. In other words, if a
+    -- function is provided `Just ty` and a `tpr`, then `tyToRepr ty` should
+    -- produce that same `tpr`. The type is in `Maybe` because there are cases
+    -- in which it's unknowable/inapplicable - for instance, in `goMirRef`, we
+    -- can't recover the `M.Ty` of the reference's `RefRoot`, only its
+    -- `TypeRepr`.
```
RyanGlScott, line 214: When we implement a solution for #1666, what will happen in the cases where the `Maybe Ty` argument is `Nothing`? Will we not be able to deduce type widths for array elements in such cases? If so, what are the ramifications of not being able to do so?

crux-mir/src/Mir/Overrides.hs
```diff
@@ -177,239 +185,384 @@ concretize (Just (SomeOnlineSolver bak)) = do
...
+    go (Just (M.TyRef M.TyStr _mut)) MirSliceRepr (Empty :> RV ptr :> RV len) =
+      goSlice (M.TyUint M.B8) ptr len
+    go (Just (M.TyRawPtr (M.TySlice elemTy) _mut)) MirSliceRepr (Empty :> RV ptr :> RV len) =
+      goSlice elemTy ptr len
+    go (Just (M.TyRawPtr M.TyStr _mut)) MirSliceRepr (Empty :> RV ptr :> RV len) =
+      goSlice (M.TyUint M.B8) ptr len
+    go (Just t) MirSliceRepr _ =
+      throwUnsupported sym $ "slice-like value of non-slice MIR type: " <> show t
+    go Nothing MirSliceRepr _ =
+      throwUnsupported sym "slice-like value of unknown MIR type"
```
RyanGlScott, line 237: Unlike other `go Nothing` cases, this one actually throws an error outright. Does this mean that concretization of slices can throw this error message in certain cases? Would it be worth adding a test for this?

samcowger:
Superseded by #1704, as mentioned in https://github.com/GaloisInc/crucible/pull/1704#issue-3810865124.

PR #1703, opened 2026-01-12
crux-mir: support calling `Box<dyn FnOnce>`

### LINKED ISSUE #1688: `crucible-mir`: Support simulating `Box<dyn FnOnce() -> ...>`
Although `crucible-mir` generally supports trait objects, one notable exception are trait objects of the form `Box<dyn FnOnce() -> ...>`. This can be seen in the [`conc_eval/clos/fn_dyn_box.rs`](https://github.com/GaloisInc/crucible/blob/bcdceb07fb5db591379600a5fc7684f8a748534b/crux-mir/test/conc_eval/clos/fn_dyn_box.rs) test case in `crux-mir`:

```
$ cabal -v0 run exe:crux-mir -- test/conc_eval/clos/fn_dyn_box.rs
user error (mkShimHandleMap: standalone use of `dyn` is not supported: TyDynamic core/729aed48::ops[0]::function[0]::FnOnce[0]::_trait2a445b904acf26bd[0])
```

The reason this happens is because of how the [`FnOnce` trait is defined](https://doc.rust-lang.org/src/core/ops/function.rs.html#242-251):

```rs
pub const trait FnOnce<Args: Tuple> {
    type Output;
    extern "rust-call" fn call_once(self, args: Args) -> Self::Output;
}
```

Invoking `call_once` takes ownership of `self`, so when you invoke `call_once` on something of type `Box<dyn FnOnce() -> ...>`, it invokes `call_once` on a value of type `dyn FnOnce() -> ...`, which is unsized. `crucible-mir` doesn't support translating unsized types that don't appear underneath a reference or pointer, however, so this fails. (Contrast this with `Fn` or `FnMut`, both of which borrow `self` instead of taking ownership.)

Strictly speaking, there is nothing special about the `FnOnce` trait here, as any trait that takes ownership of `self` would exhibit the same problem. Note, however, that `Box<dyn FnOnce() -> ...>` is a reasonably common idiom (for instance, it is used in the implementation of the `tokio` crate). Supporting unsized types in full generality in `crucible-mir` would be fairly challenging, but as an alternative, it may suffice to special-case invocations of `FnOnce::call_once`. This issue exists to track this idea.

One idea would be to implement a `TransCustom`-style override that replaces an entire function call when the callee is `IkVirtual` for `FnOnce::call_once`. This would avoid needing to create an intermediate value of type `dyn FnOnce() -> ...`.

spernsteiner:
Currently, calling `Box<dyn FnOnce>` fails because the implementation in libcore depends on `#[feature(unsized_fn_params)]`, which we don't support.  Adding full support for this feature seemed complex, and AFAIK it's mostly unused aside from this, so this branch adds some special cases around `FnOnce::call_once` instead of adding support for `unsized_fn_params`.

See [this comment](https://github.com/GaloisInc/crucible/blob/9206802089ecb06db10639b42a06ed7b69d93030/crucible-mir/src/Mir/TransCustom.hs#L2223) for an explanation of the design.

Fixes #1688

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -2646,8 +2646,8 @@ data AssignUncons (f :: k -> Type) :: Ctx.Ctx k -> Type where
 
 -- | Un-cons an element from an assignment.  Returns the first (leftmost)
 -- element and the remaining elements, bundled into a GADT to establish that
--- @ctx ~ (Ctx.SingleCtx first Ctx.<+> rest)@.  Returns `Nothing` if the input
--- is empty.
+-- @ctx ~ (Ctx.SingleCtx first Ctx.<+> rest)@.  Returns `Left Refl` if the
+-- input is empty.
 --
 -- Since `Ctx.Assignment` is a snoc-list rather than a cons-list, this is an
 -- O(n) operation.


crux-mir/test/conc_eval/clos/fn_dyn_box_of_box.rs
```diff
@@ -0,0 +1,16 @@
+// Test that calling `Box<dyn FnOnce>` works even when the erased type is not a closure but
+// something else (in this case, `Box<F>`).
```
RyanGlScott: What is `F` in this example?

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2861,17 +2920,24 @@ doVirtCall col dynTraitName methodIndex recvTy recvExpr argTys argExprs retTy =
...
--- | Translate a virtual call. The logic for looking up the right method in the
--- vtable is implemented in 'mkVirtCall'.
+-- | Generate Crucible code for an `IkVirtual` virtual-call shim. The logic for
+-- looking up the right method in the vtable is implemented in 'mkVirtCall'.
+--
+-- This will return `Nothing` for `IkVirtual` shims that were skipped by
+-- `mkShimHandleMap` and thus `have no `MethodHandle` available.  This is
+-- notably the case for `FnOnce::call_once`, which is skipped because it takes
+-- accesses `self` by value (and thus would require support for the unstable
+-- `unsized_fn_params` feature).
```
RyanGlScott: Minor Haddock nit picks: backticks are for linking Haskell identifiers, but `@...@` is meant for things that should only be rendered in monospace font (but not hyperlinked):


```suggestion
-- | Generate Crucible code for an `IkVirtual` virtual-call shim. The logic for
-- looking up the right method in the vtable is implemented in 'mkVirtCall'.
--
-- This will return `Nothing` for `IkVirtual` shims that were skipped by
-- `mkShimHandleMap` and thus `have no `MethodHandle` available.  This is
-- notably the case for @FnOnce::call_once@, which is skipped because it takes
-- accesses @self@ by value (and thus would require support for the unstable
-- @unsized_fn_params@ feature).
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2612,6 +2639,28 @@ type (x :: k) :<: (xs :: Ctx.Ctx k) = Ctx.SingleCtx x Ctx.<+> xs
...
+  AssignUncons :: f x -> Ctx.Assignment f y -> AssignUncons f (Ctx.SingleCtx x Ctx.<+> y)
+
+assignUncons :: Ctx.Assignment f ctx -> Maybe (AssignUncons f ctx)
+assignUncons Ctx.Empty = Nothing
+assignUncons (Ctx.Empty Ctx.:> x) = Just $ AssignUncons x Ctx.Empty
+assignUncons (xs Ctx.:> x) =
+  case assignUncons xs of
+    Just (AssignUncons y ys) -> Just $ AssignUncons y (ys Ctx.:> x)
+    Nothing -> panic "assignUncons"
+      ["impossible: assignUncons returned Nothing for nonempty assignment?"]
```
RyanGlScott: I think this can be refactored to avoid the panic:


```suggestion
assignUncons (xs Ctx.:> x) =
  case assignUncons xs of
    Just (AssignUncons y ys) -> Just $ AssignUncons y (ys Ctx.:> x)
    Nothing -> Just $ AssignUncons x Ctx.Empty
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2612,6 +2639,28 @@ type (x :: k) :<: (xs :: Ctx.Ctx k) = Ctx.SingleCtx x Ctx.<+> xs
 (<:) :: forall f tp ctx. f tp -> Ctx.Assignment f ctx -> Ctx.Assignment f (tp :<: ctx)
 x <: xs = Ctx.singleton x Ctx.<++> xs
 
+data AssignUncons (f :: k -> Type) :: Ctx.Ctx k -> Type where
+  AssignUncons :: f x -> Ctx.Assignment f y -> AssignUncons f (Ctx.SingleCtx x Ctx.<+> y)
+
+assignUncons :: Ctx.Assignment f ctx -> Maybe (AssignUncons f ctx)
```
RyanGlScott: It might be worth a comment here that `assignUncons` is an O(_n_) operation.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2580,9 +2581,14 @@ transVtableShim colState vtableName (VtableItem fnName defName)
     buildShim recvMirTy recvTy argTys _retTy implFH
       | M.TyRef    _recvMirTy' _ <- recvMirTy = buildShimForRef recvTy argTys implFH
       | M.TyRawPtr _recvMirTy' _ <- recvMirTy = buildShimForRef recvTy argTys implFH
+      | M.idKey defName == ["core", "ops", "function", "FnOnce", "call_once"] =
+          buildShimForByValue recvTy argTys implFH
```
RyanGlScott: It would be worth leaving a comment here explaining why we include a special case for `call_once`, pointing to `callOnceVirtShimDef` for more details.

crucible-mir/src/Mir/Generator.hs
```diff
@@ -203,6 +203,10 @@ data CustomOpMap = CustomOpMap
     { _opDefs :: Map ExplodedDefId CustomRHS
     , _cloneShimOp :: Ty -> [DefId] -> CustomOp
     , _cloneFromShimOp :: Ty -> [DefId] -> CustomOp
+    , _callOnceVirtShimOp :: Integer -> CustomOp
+    -- ^ Implementation of the `IkVirtual` shim for @FnOnce::call_once@.  See
+    -- the definition in @TransCustom.hs@ for more details.  The `Integer` is
```
RyanGlScott: Perhaps instead of saying "See the definition in `TransCustom.hs`", we should just say "See `callOnceVirtShimDef`".

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -2213,6 +2218,148 @@ cloneFromShimDef :: Ty -> [M.DefId] -> CustomOp
 cloneFromShimDef ty _parts = CustomOp $ \_ _ -> mirFail $ "cloneFromShimDef not implemented for " ++ show ty
 
 
+{- | Replacement for an `IkVirtual` shim for `FnOnce::call_once`.
+
+To understand the design, first, consider the case of `Box<dyn Fn>`.  Suppose
```
RyanGlScott: Similarly, there are lots of places in this Haddock comment that should use `@...@` instead of backticks, such as in this spot:


```suggestion
To understand the design, first, consider the case of @Box<dyn Fn>@.  Suppose
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2687,15 +2736,26 @@ mkShimHandleMap col halloc = mconcat <$> mapM mkHandle (Map.toList $ col ^. M.in
...
+              -- If the signature is not representable, we skip generating a
+              -- `MethodHandle` for this shim.  This will also cause
+              -- `transInstance` to skip generating code for it.  This will
+              -- result in a "don't know how to call" error if the shim is ever
+              -- used.
+              --
+              -- This notably occurs for `FnOnce::call_once` shims, which take
+              -- `dyn FnOnce` by value (using `#[feature(unsized_fn_params)]`).
+              when (?debug > 0) $ do
+                traceM $ "mkShimHandleMap: error translating " ++ show intr ++ ": " ++ err
```
RyanGlScott: I think this message should only be printed on a higher verbosity level. Currently, this prints even on bare invocations of `crux-mir`, e.g.,

```
$ cabal -v0 run exe:crux-mir -- test/conc_eval/clos/fn_dyn_box.rs
mkShimHandleMap: error translating Intrinsic {_intrName = core/729aed48::ops[0]::function[0]::FnOnce[0]::call_once[0]::_virt3_d5f63670a72e0fbc[0], _intrInst = Instance {_inKind = IkVirtual core/729aed48::ops[0]::function[0]::FnOnce[0]::_trait2a445b904acf26bd[0] 1, _inDefId = core/729aed48::ops[0]::function[0]::FnOnce[0]::call_once[0], _inSubsts = Substs []}}: standalone use of `dyn` is not supported: TyDynamic core/729aed48::ops[0]::function[0]::FnOnce[0]::_trait2a445b904acf26bd[0]
test fn_dyn_box/04f2d9fa::crux_test[0]: returned 6, ok

[Crux] Overall status: Valid.
```

RyanGlScott:
It would be worth stating "fixes https://github.com/GaloisInc/crucible/issues/1688" somewhere in a commit message (or the PR description).

crucible-mir/src/Mir/Trans.hs
spernsteiner: Fixed in 961c0a531c1f117967cfd5d6eb3e9e135aa03f1a.  This wound up being a bit more complex because in the `Nothing` case, you need a proof that `ctx ~ Ctx.EmptyCtx`.  I wound up changing the return type from `Maybe (AssignUncons f ctx)` to `Either (ctx :~: Ctx.EmptyCtx) (AssignUncons f ctx)`.  I did it this way instead of adding a second case to the GADT for empty inputs because I think it would break code like this:
```Haskell
do
  ??? <- case assignUncons ctxRepr of
    AssignUncons xRepr ysRepr -> return ???
    AssignUnconsEmpty -> mirFail "..."
```
I don't think there's a way to fill in the `???`s that will propagate the existential variables and the new constraint `ctx ~ SingleCtx x <+> ys` out of the `case`, without making the outer pattern match partial (`do AssignUncons ... <- case ...`).

crucible-mir/src/Mir/Trans.hs
spernsteiner: Added O(n) comment in 961c0a531c1f117967cfd5d6eb3e9e135aa03f1a

crucible-mir/src/Mir/Trans.hs
spernsteiner: Added comment in f78eba915f0e455f324c04ab4bfe8734caa811b0

crucible-mir/src/Mir/Generator.hs
spernsteiner: Updated comment in f78eba915f0e455f324c04ab4bfe8734caa811b0

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Fixed comment in f78eba915f0e455f324c04ab4bfe8734caa811b0.  (I originally wrote this text in Markdown as part of the PR description, but I decided it would work better as a comment in the code.)

crucible-mir/src/Mir/Trans.hs
spernsteiner: Fixed comment in f78eba915f0e455f324c04ab4bfe8734caa811b0

crux-mir/test/conc_eval/clos/fn_dyn_box_of_box.rs
spernsteiner: Updated comment in f78eba915f0e455f324c04ab4bfe8734caa811b0.  (`F` is a closure type.)

crucible-mir/src/Mir/Trans.hs
spernsteiner: Increased debug level in efd854bfef0e17d2666e4c462d5d6012d0d4392d

crucible-mir/src/Mir/Trans.hs
RyanGlScott: Great! And I think that `Either (ctx :~: Ctx.EmptyCtx) (AssignUncons f ctx)` is a reasonable return type to use here.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2612,6 +2641,33 @@ type (x :: k) :<: (xs :: Ctx.Ctx k) = Ctx.SingleCtx x Ctx.<+> xs
 (<:) :: forall f tp ctx. f tp -> Ctx.Assignment f ctx -> Ctx.Assignment f (tp :<: ctx)
 x <: xs = Ctx.singleton x Ctx.<++> xs
 
+data AssignUncons (f :: k -> Type) :: Ctx.Ctx k -> Type where
+  AssignUncons :: f x -> Ctx.Assignment f y -> AssignUncons f (Ctx.SingleCtx x Ctx.<+> y)
+
+-- | Un-cons an element from an assignment.  Returns the first (leftmost)
+-- element and the remaining elements, bundled into a GADT to establish that
+-- @ctx ~ (Ctx.SingleCtx first Ctx.<+> rest)@.  Returns `Nothing` if the input
+-- is empty.
```
RyanGlScott: I think "Returns `Nothing`" should be something like "Returns `Left Refl`" now.

spernsteiner: Fixed in ebce4e2eb

PR #1704, opened 2026-01-13
`crucible-mir`: use real element sizes in arrays

samcowger:
Towards #1666, this makes `crucible-mir` construct and eliminate `MirAggregate`s representing Rust arrays with offsets and sizes that match those used in Rust, rather than hardcoding all element sizes as 1. (This does not change that same hardcoding in structs or tuples.)

The first two commits in this essentially supersede #1700. A major motivation of that PR was to give access to slice element types in order to be able to provide their proper width when accessing them, but it turns out that we can avoid accessing individual slice elements in a way that requires their width.

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -1334,9 +1334,7 @@ mirAggregate_fromChunks sym chunkedAg@(MirAggregate chunkedTotalSize _) = runExc
       when (off1 > off2 || off2 - off1 < sz1) $ panic "mirAggregate_fromChunks"
         [ "overlapping chunks"
         , "at " ++ show off1 ++ " .. " ++ show (off1 + sz1)
-          ++ " (from " ++ show off1 ++ ")"
         , "and " ++ show off2 ++ " .. " ++ show (off2 + sz2)
-          ++ " (from " ++ show off2 ++ ")"
         ]
 
   ms <- forM chunkParts $ \(off, sz, outerPred, m) -> do


crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2233,30 +2267,41 @@ mirRef_offsetWrapLeaf ::
...
+    -- `offset` on the latter, then casting back to the former. The cast isn't
+    -- (currently) implemented to change the element size in the
+    -- `AgElem_RefPath`, so to use that size in that case would have us
+    -- improperly offset by 4 bytes (i.e. the size of a `u32`) at a time.
     let sym = backendGetSym bak
     -- `wrapping_offset` puts no restrictions on the arithmetic performed.
-    idx' <- liftIO $ bvAdd sym idx offset
-    return $ MirReference tpr root $ AgElem_RefPath idx' sz tpr' path
-mirRef_offsetWrapLeaf bak ref@(MirReference _ _ _) offset = do
+    extraOff <- liftIO $ bvMul sym numElems =<< bvLit sym knownRepr (BV.mkBV knownRepr (toInteger elemSize))
```
spernsteiner: Can this use `wordLit` for the conversion of `elemSize`?

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2286,13 +2331,12 @@ mirRef_tryOffsetFromLeaf sym (MirReference _ root1 path1) (MirReference _ root2
...
-        (AgElem_RefPath off1 _ _ path1', AgElem_RefPath off2 _ _ path2') -> do
+        (AgElem_RefPath off1 elemSz _ path1', AgElem_RefPath off2 _ _ path2') -> do
             pathEq <- refPathEq sym path1' path2'
             similar <- liftIO $ andPred sym rootEq pathEq
-            -- TODO: divide by `sz`?  This implements `byte_offset_from`, which
-            -- is the same as `offset_from` only when size=1
-            offset <- liftIO $ bvSub sym off1 off2
-            return $ mkPE similar offset
+            byteOffset <- liftIO $ bvSub sym off1 off2
+            elemOffset <- liftIO $ bvSdiv sym byteOffset =<< wordLit sym elemSz
```
spernsteiner: `elemSz` is potentially the wrong size to use in the division here.  `offset_from<T>` is defined to take the byte offset and divide it by the size of `T`, but `elemSz` is not necessarily the size of `T`.  I think we need to do the same thing here as in `offset`: have the intrinsic take the size of `T` as an argument, and ignore it in non-`AgElem` cases.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2333,8 +2377,8 @@ mirRef_peelIndexLeaf sym (MirReference tpr root (VectorIndex_RefPath _tpr' path
 mirRef_peelIndexLeaf sym (MirReference _tpr root (ArrayIndex_RefPath btpr path idx)) = do
     let ref = MirReferenceMux $ toFancyMuxTree sym $ MirReference (UsizeArrayRepr btpr) root path
     return $ Empty :> RV ref :> RV idx
-mirRef_peelIndexLeaf sym (MirReference _tpr root (AgElem_RefPath idx _sz _tpr' path)) = do
-    -- TODO: assumes hardcoded size=1
+mirRef_peelIndexLeaf sym (MirReference _tpr root (AgElem_RefPath off sz _tpr' path)) = do
+    idx <- liftIO $ bvUdiv sym off =<< wordLit sym sz
```
spernsteiner: This is another place where we need the size passed in from a higher level.  It looks like this gets called from `intrinsics::copy` and from `reallocate`, both of which take a type `T` whose size we can use.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2286,13 +2331,12 @@ mirRef_tryOffsetFromLeaf sym (MirReference _ root1 path1) (MirReference _ root2
...
+        (AgElem_RefPath off1 elemSz _ path1', AgElem_RefPath off2 _ _ path2') -> do
             pathEq <- refPathEq sym path1' path2'
             similar <- liftIO $ andPred sym rootEq pathEq
-            -- TODO: divide by `sz`?  This implements `byte_offset_from`, which
-            -- is the same as `offset_from` only when size=1
-            offset <- liftIO $ bvSub sym off1 off2
-            return $ mkPE similar offset
+            byteOffset <- liftIO $ bvSub sym off1 off2
+            elemOffset <- liftIO $ bvSdiv sym byteOffset =<< wordLit sym elemSz
+            return $ mkPE similar elemOffset
```
spernsteiner, line 2701: `offset_from` is UB if the offset between the two pointers is not a multiple of `size_of::<T>()`; we should either check that here, or else add a comment noting the unsoundness.  (Previously this wasn't an issue because it was impossible to have a pointer to a location that's in between two adjacent elements of an array.)

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2385,22 +2429,27 @@ mirRef_indexAndLenLeaf bak gs iTypes (MirReference tpr root (VectorIndex_RefPath
...
-    --when (totalSize `mod` sz /= 0) $
-    --    leafAbort $ Unsupported callStack $
-    --        "exepcted aggregate size (" ++ show totalSize ++ ") to be a multiple of "
-    --            ++ "element size (" ++ show sz ++ ")"
-    --let lenWord = totalSize `div` sz
+    when (totalSize `mod` elemSize /= 0) $
+       leafAbort $ Unsupported callStack $
+           "expected aggregate size (" ++ show totalSize ++ ") to be a multiple of "
+               ++ "element size (" ++ show elemSize ++ ")"
+    let lenWord = totalSize `div` elemSize
```
spernsteiner, line 2910: Another place where we need a proper size passed in, instead of using the one from `AgElem_RefPath`.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2385,22 +2429,27 @@ mirRef_indexAndLenLeaf bak gs iTypes (MirReference tpr root (VectorIndex_RefPath
...
-    -- TODO: also divide `idx` by `sz`, and assert that it's divisible
-    return (idx, len)
+
+    elemSizeBV <- liftIO $ bvLit sym knownNat $ BV.mkBV knownNat $ fromIntegral elemSize
+    offModSz <- liftIO $ bvUrem sym elemSizeBV elemOff
+    offModSzIsZero <- liftIO $ bvEq sym offModSz =<< wordLit sym 0
+    leafAssert bak offModSzIsZero $ Unsupported callStack $
+        "expected element offset to be a multiple of element size (" ++ show elemSize ++ ")"
+
+    offDivSz <- liftIO $ bvUdiv sym elemSizeBV elemOff
```
spernsteiner: Is the argument order backwards on these `bvUrem` and `bvUdiv` calls?  It looks like it's doing `size / offset` rather than `offset / size`.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2385,22 +2429,27 @@ mirRef_indexAndLenLeaf bak gs iTypes (MirReference tpr root (VectorIndex_RefPath
...
+    when (totalSize `mod` elemSize /= 0) $
+       leafAbort $ Unsupported callStack $
+           "expected aggregate size (" ++ show totalSize ++ ") to be a multiple of "
+               ++ "element size (" ++ show elemSize ++ ")"
+    let lenWord = totalSize `div` elemSize
     len <- liftIO $ bvLit sym knownNat $ BV.mkBV knownNat $ fromIntegral lenWord
-    -- TODO: also divide `idx` by `sz`, and assert that it's divisible
-    return (idx, len)
+
+    elemSizeBV <- liftIO $ bvLit sym knownNat $ BV.mkBV knownNat $ fromIntegral elemSize
```
spernsteiner: Can this use `wordLit`?

spernsteiner (commented):
Just reviewed `Intrinsics.hs` so far

crux-mir/test/symb_eval/mux/array.good
```diff
@@ -3,8 +3,8 @@ ok
 
 [Crux-MIR] ---- FINAL RESULTS ----
 [Crux] Goal status:
-[Crux]   Total: 6
-[Crux]   Proved: 6
+[Crux]   Total: 9
+[Crux]   Proved: 9
```
spernsteiner, line 7: Why is this test producing extra subgoals?  It would be good to know if we can eliminate those, to reduce the chances of introducing a new performance regression

crux-mir/test/symb_eval/concretize/array.rs
```diff
@@ -14,8 +14,10 @@ fn crux_test() -> (i32, i32, i32) {
     }
     crucible_assume!(s[0] + s[1] + s[2] == 6);
 
-    let s = concretize(s);
-    (s[0], s[1], s[2])
+    let s0 = concretize(s[0]);
+    let s1 = concretize(s[1]);
+    let s2 = concretize(s[2]);
+    (s0, s1, s2)
```
spernsteiner, line 23: Can you add a comment explaining why `concretize(s)` doesn't work here?

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1865,16 +1921,23 @@ mkZero tpr = mirFail $ "don't know how to zero-initialize " ++ show tpr
...
                 S.litExpr "bad pointer in reallocate: not the start of an allocation"
 
             oldAg <- readMirRef MirAggregateRepr agPtr
-            newAg <- mirAggregate_resize oldAg newSz
+
+            col <- use $ cs . collection
+            elemSize <- case tySizedness col elemTy of
+                Sized s -> pure s
+                Unsized ->
+                    mirFail $ "reallocate: unsized array element type: " <> show elemTy
```
spernsteiner: This pattern is repeated in a few places—maybe it's worth having a `tySizeM` helper, analogous to `tyToReprM`?

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -342,7 +342,8 @@ vector_as_slice_impl (Substs [t]) =
             -- This is similar to `&mut [T; n] -> &mut [T]` unsizing.
             v <- readMirRef (C.VectorRepr tpr) e
             let end = R.App $ vectorSizeUsize R.App v
-            e' <- subindexRef tpr e (R.App $ usizeLit 0)
+            let elemSize = 1 -- TODO: hardcoded size=1
```
spernsteiner: There's a real type `t` available, so this could use its actual size instead.  Either that or add a comment explaining that size=1 is okay because we know the size won't be used.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -438,7 +444,8 @@ array_as_slice_impl (Substs [t]) =
           MirExp UsizeRepr start,
           MirExp UsizeRepr len ] -> do
             Some tpr <- tyToReprM t
-            ptr <- subindexRef tpr e start
+            let elemSize = 1 -- TODO: hardcoded size=1
```
spernsteiner: This also has an actual type `t` available

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Yeah, sure seems to be...

crux-mir/src/Mir/Overrides.hs
```diff
@@ -371,18 +337,18 @@ regEval bak baseEval = go
...
 
-        return gv'
+        return rc
 
     goMirReferenceRoot :: forall tp' .
         MirReferenceRoot sym tp' ->
         OverrideSim p sym MIR rtp args ret (MirReferenceRoot sym tp')
     goMirReferenceRoot (RefCell_RefRoot rc) = RefCell_RefRoot <$> goRefCell rc
-    goMirReferenceRoot (GlobalVar_RefRoot gv) = GlobalVar_RefRoot <$> goGlobalVar gv
+    goMirReferenceRoot (GlobalVar_RefRoot gv) = RefCell_RefRoot <$> goGlobalVar gv
```
RyanGlScott: It would be worth leaving a comment here pointing to `goGlobalVar` to justify why this case returns `RefCell_RefRoot` instead of `GlobalVar_RefRoot`.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2670,14 +2672,15 @@ writeMirRefIO bak gs iTypes tpr (MirReferenceMux ref) x =
...
     TypeRepr tp ->
     MirReferenceMux sym ->
     RegValue sym UsizeType ->
     -- | Size of the element, in bytes
     Word ->
     OverrideSim m sym MIR rtp args ret (MirReferenceMux sym)
-subindexMirRefSim tpr ref idx elemSize = do
-    modifyRefMuxSim (\ref' -> subindexMirRefLeaf tpr ref' idx elemSize) ref
+subindexMirRefSim sym tpr ref idx elemSize = do
+    modifyRefMuxSim (\ref' -> subindexMirRefLeaf sym tpr ref' idx elemSize) ref
```
RyanGlScott, line 3240: As an FYI, adding a `sym` argument will require some minor changes on the SAW and `crucible-mir-comp` side to accommodate.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: I've opened https://github.com/GaloisInc/mir-json/issues/236 for laying the groundwork to test this functionality.

crux-mir/test/symb_eval/mux/array.good
samcowger: The excess goals come from checking the validity of an offset into an aggregate, arising in this case from indexing the 1-length slice. The statement to be proved is that the index (`if c { 1 } else { 0 }`) is valid, and in this case, that means checking it for equality against the only valid offset, 0. So, this ended up in `what4` as roughly `if c then 1 else 0 == 0`, and `what4` then recognized that this could reduce to `not c`. (I learned this from printing out several intermediate expressions within `Mir.Intrinsics.offsetInSpans`.) I surmise that `not c` was simple enough not to show up as a proved goal. With my changes, we're still checking for equality with 0, but because the offset needs to account for the elements' width, the goal statement looks more like `4 * (if c then 1 else 0) == 0`, which `what4` doesn't seem able to simplify.

I don't know that there's much we can do about this (though perhaps we could to try to patch `what4` to recognize and simplify expressions like the one we're producing here). On the bright side, these goals shouldn't grow in size with the number of array elements, because with any length above 1 we should be generating validity expressions using the "run" recognition functionality of #1625 instead.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: It can, and now does.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Fixed in f7a5f6e5476011a8d34b2ccec3e8f7b5c3e68902.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Fixed in d2e58da16fa4670fd6b50be6c7c4827d657a57e7.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Fixed the order in 32ec0dd2dc98e109f40ffeeba7510d793a7283ff.

crux-mir/test/symb_eval/concretize/array.rs
samcowger: Included a comment in 2c13cb619648d74e589f8be7d1690edcf31c3670.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Done in c59f9590757a445443533857c26beb3057b59dcb.

crucible-mir/src/Mir/TransCustom.hs
samcowger: This was a straggler; I went back and used the proper size in abb56cd8e8bc53efaeb502ceb5788aa617cf3a1f.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Likewise a straggler, likewise fixed in abb56cd8e8bc53efaeb502ceb5788aa617cf3a1f.

crux-mir/src/Mir/Overrides.hs
samcowger: Done

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Ah, indeed. I see two calls in `crucible-mir-comp`'s `Mir.Compositional.Override` to `subindexMirRefSim`, and both appear to have the necessary `sym` parameter in scope at the call site already, so this should be fine.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Done in 9d743b028bce441042732b79fe36cb113245123b.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: It can, and now does.

crucible-mir/src/Mir/Intrinsics.hs
RyanGlScott: Indeed, I don't expect it to be difficult to adapt to these changes. This is more of a "please hold off on merging this until the Crux/SAW releases are done" warning.

samcowger: Right, I certainly will.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Done in df31f3b1744fe0129a889d4ca31af92e3759879a.

crux-mir/test/symb_eval/mux/array.good
RyanGlScott: It does seem strange that `what4` wouldn't push multiplications into if-then-else expressions. At the very least, it might be worth opening a `what4` issue about this.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1062,7 +1069,20 @@ tySizedness col ty =
...
+        Just Nothing -> Unsized
+        Just (Just lay) -> Sized (fromIntegral (lay ^. M.laySize))
+
+-- | Get a type's size, failing (via `mirFail`) if it's unsized.
+tySizeM :: HasCallStack => String -> M.Ty -> MirGenerator h s ret Word
+tySizeM site ty = do
+  col <- use $ cs . collection
+  case tySizedness col ty of
+    Sized s -> pure s
+    Unsized -> mirFail $ site <> ": unsized type: " <> show ty
```
spernsteiner: Seems like it should be possible to get the name of the call site from `HasCallStack`, so we don't have to pass in `site` explicitly.   (And as a longer-term improvement, maybe `mirFail` should include some/all of the Haskell callstack in its assertion message.)

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2313,34 +2379,38 @@ mirRef_tryOffsetFromIO ::
...
     let ref = MirReferenceMux $ toFancyMuxTree sym $ MirReference (VectorRepr tpr) root path
     return $ Empty :> RV ref :> RV idx
-mirRef_peelIndexLeaf sym (MirReference _tpr root (ArrayIndex_RefPath btpr path idx)) = do
+mirRef_peelIndexLeaf sym _elemSize (MirReference _tpr root (ArrayIndex_RefPath btpr path idx)) = do
     let ref = MirReferenceMux $ toFancyMuxTree sym $ MirReference (UsizeArrayRepr btpr) root path
     return $ Empty :> RV ref :> RV idx
-mirRef_peelIndexLeaf sym (MirReference _tpr root (AgElem_RefPath idx _sz _tpr' path)) = do
-    -- TODO: assumes hardcoded size=1
+mirRef_peelIndexLeaf sym elemSize (MirReference _tpr root (AgElem_RefPath off _sz _tpr' path)) = do
+    idx <- liftIO $ bvUdiv sym off =<< wordLit sym elemSize
```
spernsteiner: What should this do if `off` isn't a multiple of `elemSize`?

crucible-mir/src/Mir/Trans.hs
```diff
@@ -474,6 +476,16 @@ transBinOp bop op1 op2 = do
         WithOverflow bop' -> do
             (res, overflow) <- evalBinOp bop' mat me1 me2
             buildTupleM [typeOf op1, TyBool] [res, MirExp C.BoolRepr overflow]
+        Offset
+          | MirExp MirReferenceRepr e1 <- me1
+          , MirExp UsizeRepr e2 <- me2 -> do
```
spernsteiner, line 505: What's this new binop case needed for?

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1062,7 +1069,20 @@ tySizedness col ty =
       case col ^? M.adts . ix adtName of
         Nothing -> error $ "tySizedness: unknown ADT: " <> show adtName
         Just adt -> adtSizedness col adt
-    _ -> Sized
+    _ ->
+      -- TODO: can the whole function be implemented via this stanza?
```
spernsteiner: The caveat here is that only types that are mentioned somewhere in the crate will show up in `layouts`.  So we would need to be careful not to pass made-up types to this function.  For example, if we see type `T` somewhere in the crate, and we want to know the sizedness of `[T]`, we would need to have some evidence that `[T]` must also appear in the crate, otherwise the lookup into `M.layouts` could fail.

spernsteiner (approved):
I think this is in pretty good shape overall.  We may want to warn people to watch for performance regressions around symbolic indexing after upgrading.  If we get reports of actual problems then we can look into those valid-index assertions more thoroughly.

crucible-mir/src/Mir/Trans.hs
samcowger: This case existed before this PR; I only moved it (in 4a55413fb88471675d1cffb817dd1c8ca5bed24a) to grant access to the type of `e1` (technically `op1`), whose size I needed to pass to `mirRef_offsetWrap` (which I did in 2045fd1d807d2cffbba487d85803dda3f012b0c3).

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Well, I guess I think it would be appropriate to fail. If we didn't fail, we'd have to be able to treat a byte offset of 1 as a well-defined index into an allocation of 2-width elements, and I don't think the memory model can support that. I'll add a divisibility check and hope that it doesn't lead to a bunch of extra goals.

crucible-mir/src/Mir/TransTy.hs
samcowger: Ah, thanks! I'll remove the TODO and mention this rationale in a comment.

samcowger: Done in f3698efe67bfe8b4e69baff3d7f44bcce07c5dac.

crucible-mir/src/Mir/TransTy.hs
samcowger: I've switched to `HasCallStack`-derived information in 34d3503c99cf8751501716948b297bea6ffb43a2, and I've opened #1732 for augmenting `mirFail` messages.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1041,11 +1048,11 @@ data Sizedness = Sized | Unsized
...
   case adt ^. M.adtkind of
-    M.Enum _ -> Sized
-    M.Union -> Sized
+    M.Enum _ -> Sized (adt ^. M.adtSize)
+    M.Union -> Sized (adt ^. M.adtSize)
     M.Struct ->
       case M.onlyVariant adt ^. M.vfields of
-        [] -> Sized  -- size 0 is still sized
+        [] -> Sized 0  -- size 0 is still sized
         fields -> tySizedness col (last fields ^. M.fty)
```
spernsteiner: Now that `Sizedness` contains the actual size and not just boolean `Sized`/`Unsized`, this last case wrongly reports the size of the struct to be the size of its last field, ignoring all other fields.
```Rust
struct S(i32, ());

#[cfg_attr(crux, crux::test)]
fn crux_test() -> i32 {
    let arr = [S(1, ()), S(2, ()), S(3, ())];
    arr[0].0 + arr[1].0 + arr[2].0
}

pub fn main() {
}
```
The correct answer from this test is 6, but crux-mir on this branch instead outputs 9.  The array handling thinks the size of `S` is zero, so all three writes in the initializer go to the same place (offset 0), and all three reads access that same location and get the last value written (`S(3, ())`).

samcowger: Ah, good catch. I've fixed this in 8bd5f465951bc0f169e79c54679bff36a7cf14d7, and have included your example as a testcase.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: I've implemented a divisibility check in f479c4f72740b20496c227aac8ff40d870f70fb2, and it doesn't seem to have evinced any new goals.

samcowger:
Update: I believe this PR is good to go in isolation, but I've encountered some issues in updating SAW to account for its changes. In particular, there are places in SAW where we'll need to determine the size of a `Ty`, but we only have access to a version of the `Ty` that's been translated to and from a Cryptol type, and the translated version doesn't appear in the original `Collection`. (The offending type occurs in SAW's [`test2032`](https://github.com/GaloisInc/saw-script/blob/10d32a5a9d69aeceaa36e1d6200994afebf8b514/intTests/test2032/test.rs), which uses the type `(u8, i64, u128)`. After round-trip translation, that type becomes `(u8, u64, u128)`, a type which doesn't appear in the original program and for which we have no layout information.)

@spernsteiner and I spoke about this and see two possible paths forward:
- Modify `mir-json` to include a table that allows looking up tuple layout based on a list of size-alignment pairs.
- If it's sufficiently simple, reimplement `rustc`'s internal tuple layout algorithm ourselves as a stopgap, and defer a more durable solution to an issue. (This solution is non-durable because `rustc` makes few guarantees about tuple layout.)

SAW only supports translating [relatively simple](https://github.com/GaloisInc/saw-script/blob/10d32a5a9d69aeceaa36e1d6200994afebf8b514/saw-central/src/SAWCentral/Crucible/MIR/Builtins.hs#L1769-L1780) MIR `Ty`s into Cryptol types, which means we'd only need to support relatively simple types in a layout algorithm implementation, which I hope would greatly simplify that implementation. So, I'm minded to explore the second approach.

spernsteiner:
One thing to consider for the second option is how we would detect that the algorithm has changed.  I guess we'd want to have some asserts in crucible-mir that the tuple layouts loaded from the JSON match the ones computed by the algorithm.

Also, I did some testing and it looks like (as of rustc 1.89.0) tuple layouts are not affected by `-Z randomize-layout`, but they are affected by the "last field remains last" rule (`(u8, u16, u16)` has offsets 2, 0, 4).

RyanGlScott:
Sigh. That round-trip code is a bit wonky, as we really ought not to be assuming that Cryptol's `[32]` type always corresponds to `u32` instead of `i32`. Fixing that properly would probably require something like https://github.com/GaloisInc/saw-script/issues/1976, however, which is not a simple feat.

I agree that re-implementing a small subset of Rust's layout algorithm is probably the best bet for now.

samcowger:
On reflection, reimplementing the layout algorithm seems [a bit daunting](https://github.com/rust-lang/rust/blob/71e00273c0921e1bc850ae8cc4161fbb44cfa848/compiler/rustc_abi/src/layout.rs#L1097-L1436), and I worry that it'll be both tricky to translate correctly and potentially tricky to keep up to date. (Unless there is a nice, high-level summary of the algorithm somewhere, and I'm missing it.)

I'm wondering about a different approach, along the lines of the first thing I mentioned. Instead of creating a map from size-alignment pairs to layouts, we could evince a map from _Cryptol_ types to layouts, and do so entirely in SAW. It seems like this sort of map could maybe live in `MIRCrucibleContext`, if I understand that type's role correctly. To me, this feels preferable to a change that extends all the way down into `mir-json`. Do y'all think this would be appropriate and feasible?

RyanGlScott:
I think that could potentially work. How many types would need to live in this map? I ask since there are conceptually an infinite number of Cryptol tuple types, but perhaps we can do something clever to represent that in a finite way.

samcowger:
I was imagining that we'd have functions that are responsible for MIR-Cryptol type translation add entries to this map for the types they actually translate. For instance, I'd imagine [`mir_fresh_var`](https://github.com/GaloisInc/saw-script/blob/8dae9bde8c21258e05fa67e5010c8af0d93d7b68/saw-central/src/SAWCentral/Crucible/MIR/Builtins.hs#L658-L667) would want to add to the map. I guess the more completionist approach would be to add entries for every MIR type in the MIR `Collection`, but I hope we wouldn't need to go that far.

RyanGlScott:
That sounds reasonable. Perhaps the [`cryptolTypeOfActual`](https://github.com/GaloisInc/saw-script/blob/10d32a5a9d69aeceaa36e1d6200994afebf8b514/saw-central/src/SAWCentral/Crucible/MIR/Builtins.hs#L1769-L1780) function could return a layout entry alongside its Cryptol type, and then call sites could add this layout entry to the map as needed.

spernsteiner:
> Instead of creating a map from size-alignment pairs to layouts, we could evince a map from _Cryptol_ types to layouts, and do so entirely in SAW.

One option is to have a pass in SAW (or in Crucible) that runs just after loading the `Collection` to build the table of layouts for all the tuple types mentioned in the crate.  Or, write a function for normalizing Cryptol-compatible Rust types (take a `Mir.Ty` and produce a new `Mir.Ty` where all ints are replace by uints, etc), and use that to extend the existing layout table with the normalized version of each type.

I don't know the details of the SAW integration, but IIRC crux-mir-comp considers the Cryptol type `[32]` to correspond not only to `u32` and `i32`, but also to `[bool; 32]`, which has a completely different size and layout.  Does this mean `[bool; 32]` might get mapped to `u32` through translation to and from Cryptol?

> Unless there is a nice, high-level summary of the algorithm somewhere, and I'm missing it.

The general idea is that it sorts fields by alignment, then applies the standard C layout rules.  However, looking at that code, it seems there are some special cases around the sorting of types with niches (which includes `bool`) that we probably don't want to try to duplicate.

samcowger:
> I don't know the details of the SAW integration, but IIRC crux-mir-comp considers the Cryptol type `[32]` to correspond not only to `u32` and `i32`, but also to `[bool; 32]`, which has a completely different size and layout. Does this mean `[bool; 32]` might get mapped to `u32` through translation to and from Cryptol?

Unfortunately, I think that's correct - `u32` and `[bool; 32]` both have the same Cryptol type according to `cryptolTypeOfActual`, so I'm sure they'd produce the same MIR type (`u32`) when translated in the other direction. I fear that means it will not be possible to unambiguously determine the Rust layout of a Cryptol type on its own.

spernsteiner:
@samcowger Can you remind me where the current Cryptol-to-Rust type conversion is being used?  I looked through the thread but didn't see it.

At this point it sounds like our major options are:
1. Get rid of the conversion entirely, either by obtaining the `Mir.Ty` from some other source or by changing the consumer(s) not to need it.  IIRC this sounded like it would be quite difficult last time you explained it.
2. Remove the Cryptol `[32]` <-> Rust `[bool; 32]` conversion, leaving only `[32]` <-> `u32`/`i32`.  I'm not sure, but I think that conversion is rarely used.

samcowger:
> Can you remind me where the current Cryptol-to-Rust type conversion is being used? I looked through the thread but didn't see it.

Is [`SAWCentral.Crucible.MIR.ResolveSetupValue.toMIRType`](https://github.com/GaloisInc/saw-script/blob/master/saw-central/src/SAWCentral/Crucible/MIR/ResolveSetupValue.hs#L1203-L1232) what you're looking for?

I think those two options sound right. There may be others, but I think it's worth trying to pursue #2, along with a maintaining a Cryptol-type-to-Rust-layout map or augmenting a `Collection` with layouts of normalized types, to see how far it gets us.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1283,78 +1297,61 @@ mirAggregate_fromChunks ::
...
-          ++ show mappedOff2 ++ " .. " ++ show (mappedOff2 + totalSize2)
+  forM_ (zip chunkParts (Prelude.drop 1 chunkParts)) $
+    \((off1, sz1, _, _), (off2, sz2, _, _)) -> do
+      when (off1 > off2 || off2 - off1 < sz1) $ panic "mirAggregate_fromChunks"
+        [ "overlapping chunks"
+        , "at " ++ show off1 ++ " .. " ++ show (off1 + sz1)
+          ++ " (from " ++ show off1 ++ ")"
+        , "and " ++ show off2 ++ " .. " ++ show (off2 + sz2)
           ++ " (from " ++ show off2 ++ ")"
+        ]
```
spernsteiner, line 1339:
```suggestion
        [ "overlapping chunks"
        , "at " ++ show off1 ++ " .. " ++ show (off1 + sz1)
        , "and " ++ show off2 ++ " .. " ++ show (off2 + sz2)
        ]

```
Previous version included both `off` and `mappedOff` since they might be different.  Now only `off` is relevant.

samcowger: Whoops, thanks!

PR #1705, opened 2026-01-14
llvm: Remove LLVM `declare`s from LLVM overrides

### LINKED ISSUE #1138: llvm: Remove the LLVM declaration from `LLVMOverride`

langston-barrett:
Fixes #1138.

While this PR adds lines overall, a large chunk of that is identity cases in pattern matches that are included for the sake of future-proofing those `case`s. Fundamentally, this is a significant simplification.

Some improvements to the `Cast` module that were enabled by this change:

- A type family that specifies the sorts of casts that are performed
- Make all functions total
- Separate casts to and from the Crucible-LLVM ABI

One breaking change: Crucible-LLVM S-expression files now have to use the *exact* signature of the underlying override. See the changes to tests for examples of this breakage. Not a huge deal, as the S-expression syntax is mostly developer-facing.

After this change, it would be feasible to completely remove the quasiquoters. I haven't done that in this PR, but would be happy to do so if requested. It might be nice to instead move them to the type level as quasiquoters for `LLVMOverride p sym ext args ret` where `args` and `ret` are determined by their LLVM types.

### DIFF
--- a/crucible-llvm/CHANGELOG.md
+++ b/crucible-llvm/CHANGELOG.md
@@ -1,5 +1,23 @@
 # next
 
+* Remove `llvmOverride_declare :: Text.LLVM.AST.Declare` from `LLVMOverride`.
+
+  * Add `Lang.Crucible.LLVM.Intrinsics.Declare` module.
+  * Change functions in `Lang.Crucible.LLVM.Intrinsics` to work
+    with `Lang.Crucible.LLVM.Intrinsics.Declare.Declare`s. To
+    migrate, use `Lang.Crucible.LLVM.Intrinsics.Declare.fromLLVM`
+    to translate `Text.LLVM.AST.Declare`s into
+    `Lang.Crucible.LLVM.Intrinsics.Declare.Declare`s
+  * `do_register_llvm_override` no longer does any mapping nor adaptation of
+    types, use `Lang.Crucible.LLVM.Intrinsics.Cast.lowerLLVMOverride` for that.
+  * Replace `build_llvm_override` with
+    `Lang.Crucible.LLVM.Intrinsics.Cast.lowerLLVMOverride`.
+  * Overhaul the API of `Lang.Crucible.LLVM.Intrinsics.Cast`.
+  * Replace various fields of `LLVMOverride` with a `Declare`. To migrate:
+    * Replace `llvmOverride_name` with `llvmOvSymbol`
+    * Replace `llvmOverride_args` with `llvmOvArgs`
+    * Replace `llvmOverride_ret` with `llvmOvRet`
+
 # 0.9 -- 2026-01-29
 
 * The `LLVM_Debug` data constructor for `LLVMStmt`, as well as the related
@@ -44,23 +62,6 @@
   * `Lang.Crucible.LLVM.MemModel.Pointer.ppLLVMPointerIntrinsicType`
 * Overrides for `memcmp`, `strcmp`, `strncmp`, `strnlen`, `strcpy`, `strdup`,
   and `strndup`, supported by new APIs in `Lang.Crucible.LLVM.MemModel.Strings`.
-* Remove `llvmOverride_declare :: Text.LLVM.AST.Declare` from `LLVMOverride`.
-
-  * Add `Lang.Crucible.LLVM.Intrinsics.Declare` module.
-  * Change functions in `Lang.Crucible.LLVM.Intrinsics` to work
-    with `Lang.Crucible.LLVM.Intrinsics.Declare.Declare`s. To
-    migrate, use `Lang.Crucible.LLVM.Intrinsics.Declare.fromLLVM`
-    to translate `Text.LLVM.AST.Declare`s into
-    `Lang.Crucible.LLVM.Intrinsics.Declare.Declare`s
-  * `do_register_llvm_override` no longer does any mapping nor adaptation of
-    types, use `Lang.Crucible.LLVM.Intrinsics.Cast.lowerLLVMOverride` for that.
-  * Replace `build_llvm_override` with
-    `Lang.Crucible.LLVM.Intrinsics.Cast.lowerLLVMOverride`.
-  * Overhaul the API of `Lang.Crucible.LLVM.Intrinsics.Cast`.
-  * Replace various fields of `LLVMOverride` with a `Declare`. To migrate:
-    * Replace `llvmOverride_name` with `llvmOvSymbol`
-    * Replace `llvmOverride_args` with `llvmOvArgs`
-    * Replace `llvmOverride_ret` with `llvmOvRet`
 
 # 0.8.0 -- 2025-11-09
 


langston-barrett:
[EDIT]: This is done.

I realized that in a downstream codebase the `L.Declare`s from `LLVMOverride` were fed back into `register_llvm_overrides_`, so I'll need to adjust that function (and its friends) to not require `L.Declare`s.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -291,44 +319,55 @@ basic_llvm_override ovr = OverrideTemplate matcher regOvr
...
-              , ""
+              , " *** found args: " ++ show (llvmOverride_args llvmOverride)
+              , " *** found ret: " ++ show (llvmOverride_ret llvmOverride)
               ]
   else do_register_llvm_override llvmctx llvmOverride
 
 -- | Low-level function to register LLVM overrides.
 --
--- Type-checks the LLVM override against the 'L.Declare' it contains, adapting
+-- TODO Type-checks the LLVM override against the 'L.Declare' it contains, adapting
```
langston-barrett: NTS: Update this comment

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -291,47 +345,56 @@ basic_llvm_override ovr = OverrideTemplate matcher regOvr
...
+    Ctx.Assignment TypeRepr ctx2 ->
+    Bool
+  -- Ignore varargs as long as the rest of the arguments match
+  matchingArgList (rest Ctx.:> VectorRepr AnyRepr) ys = matchingArgList rest ys
+  matchingArgList xs (rest Ctx.:> VectorRepr AnyRepr) = matchingArgList xs rest
+  matchingArgList Ctx.Empty Ctx.Empty = True
+  matchingArgList Ctx.Empty _ = False
+  matchingArgList _ Ctx.Empty = False
+  matchingArgList (xs Ctx.:> x) (ys Ctx.:> y) =
+    Maybe.isJust (testEquality x y) && matchingArgList xs ys
```
langston-barrett: NTS: Replace these with `isJust . testEquality`

langston-barrett:
Status: I'm adapting [GREASE](https://github.com/GaloisInc/grease) to these changes to make sure they work in practice.

[EDIT]: See https://github.com/GaloisInc/grease/pull/525

crux-llvm/test-data/golden/golden/double_free.pre-clang19.z3.good
```diff
@@ -4,7 +4,7 @@
 [Crux]   Undefined behavior encountered
 [Crux]   Details:
 [Crux]     `free` called on a pointer to already-freed memory
-[Crux]     Pointer: (10, 0x0:[64])
+[Crux]     Pointer: (9, 0x0:[64])
```
RyanGlScott: Does `crux-llvm/test-data/golden/golden/double_free.z3.good` (for LLVM 19 or later) also need to be updated?

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -291,21 +316,34 @@ basic_llvm_override ovr = OverrideTemplate matcher regOvr
...
+      , matchingArgList args (Cast.ctxToLLVMType (llvmOverride_args provided))
+      , Maybe.isJust (testEquality ret (Cast.toLLVMType (llvmOverride_ret provided)))
+      ]
+
+  where
+  matchingArgList ::
+    Ctx.Assignment TypeRepr ctx1 ->
+    Ctx.Assignment TypeRepr ctx2 ->
+    Bool
+  -- Ignore varargs as long as the rest of the arguments match
```
RyanGlScott: It would be worth clarifying that `VectorRepr AnyRepr` is the Crucible type representation for varargs.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -291,21 +316,34 @@ basic_llvm_override ovr = OverrideTemplate matcher regOvr
...
+  matchingArgList (rest Ctx.:> VectorRepr AnyRepr) ys = matchingArgList rest ys
+  matchingArgList xs (rest Ctx.:> VectorRepr AnyRepr) = matchingArgList xs rest
+  matchingArgList Ctx.Empty Ctx.Empty = True
+  matchingArgList Ctx.Empty _ = False
+  matchingArgList _ Ctx.Empty = False
+  matchingArgList (xs Ctx.:> x) (ys Ctx.:> y) =
+    Maybe.isJust (testEquality x y) && matchingArgList xs ys
```
RyanGlScott: Doesn't `TypeRepr` have an `Eq` instance? If so, then it might suffice to say `x == y` instead of `isJust (testEquality x y)`.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -291,21 +316,34 @@ basic_llvm_override ovr = OverrideTemplate matcher regOvr
...
- matchingArgList (x:xs) (y:ys) = x `L.eqTypeModuloOpaquePtrs` y && matchingArgList xs ys
+  LLVMOverride p sym ext args ret ->
+  m Bool
+isMatchingDeclaration requested provided =
+  llvmDeclToFunHandleRepr' requested $ \args ret ->
+    pure $
+      and
+      [ L.decName requested == llvmOverride_name provided
+      , matchingArgList args (Cast.ctxToLLVMType (llvmOverride_args provided))
+      , Maybe.isJust (testEquality ret (Cast.toLLVMType (llvmOverride_ret provided)))
```
RyanGlScott: Similarly, I think you could just use `(==)` here instead of `isJust . testEquality`.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Cast.hs
```diff
@@ -455,3 +294,33 @@ regValueFromLLVM' fNm wanted ty bak val = do
...
     (CT.SymbolicStructRepr {}, _) -> pure val
+
+-- | Map 'regValueFromLLVM' over an 'Ctx.Assignment' of 'CRM.RegEntry's.
+regEntriesFromLLVM ::
+  CB.IsSymBackend sym bak =>
+  -- | Only used in error messages
+  WFN.FunctionName ->
+  Ctx.Assignment CT.TypeRepr tys ->
+  Ctx.Assignment CT.TypeRepr (CtxToLLVMType tys) ->
+  bak ->
```
RyanGlScott: Minor gripe: `bak` is usually one of the first (if not _the_ first) argument in Crucible-oriented type signatures, and I think it would be good to stick to that convention. (Similarly for `regMapFromLLVM` as well.)

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
```diff
@@ -81,9 +82,9 @@ import           Lang.Crucible.LLVM.Translation.Types
 -- LLVM memory model, such as Macaw.
 data LLVMOverride p sym ext args ret =
   LLVMOverride
-  { llvmOverride_declare :: L.Declare    -- ^ An LLVM name and signature for this intrinsic
-  , llvmOverride_args    :: CtxRepr args -- ^ A representation of the argument types
-  , llvmOverride_ret     :: TypeRepr ret -- ^ A representation of the return type
+  { llvmOverride_name :: L.Symbol
+  , llvmOverride_args :: CtxRepr args -- ^ A representation of the argument types
+  , llvmOverride_ret :: TypeRepr ret -- ^ A representation of the return type
```
RyanGlScott: Would it be worthwhile to replace these first three fields with a single `Declare` field? (Either in this PR or as future work)

langston-barrett:
> Do you envision these changes being breaking to most downstream consumers?

Only the ones (like crucible-llvm-syntax or GREASE) that use the S-expression syntax. I would expect that SAW, like Crux-LLVM, will require no changes. That being said, I'd prefer to delay merging until the SAW release is complete just in case.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Cast.hs
langston-barrett: Ah yeah this is mostly a result of how these signatures evolved while they were in development. Will fix.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
langston-barrett: Yeah, that's a good idea! Not sure why it didn't occur to me haha.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
langston-barrett: Wouldn't `==` only work if we knew that the type variables were equal already?

langston-barrett: For instance, with the return types:
```
src/Lang/Crucible/LLVM/Intrinsics/Common.hs:302:29-36: error: [GHC-25897]
    • Couldn't match type ‘ret’ with ‘ret'’
      Expected: LLVMOverride p sym ext args ret'
        Actual: LLVMOverride p sym ext args ret
      ‘ret’ is a rigid type variable bound by
        the type signature for:
          isMatchingDeclaration :: forall (args' :: Ctx CrucibleType)
                                          (ret' :: CrucibleType) p sym ext
                                          (args :: Ctx CrucibleType) (ret ::
CrucibleType).
                                   Decl.Declare args' ret'
                                   -> LLVMOverride p sym ext args ret -> Bool
        at src/Lang/Crucible/LLVM/Intrinsics/Common.hs:(292,1)-(295,6)
      ‘ret'’ is a rigid type variable bound by
        the type signature for:
          isMatchingDeclaration :: forall (args' :: Ctx CrucibleType)
                                          (ret' :: CrucibleType) p sym ext
                                          (args :: Ctx CrucibleType) (ret ::
CrucibleType).
                                   Decl.Declare args' ret'
                                   -> LLVMOverride p sym ext args ret -> Bool
        at src/Lang/Crucible/LLVM/Intrinsics/Common.hs:(292,1)-(295,6)
    • In the first argument of ‘llvmOverride_ret’, namely ‘provided’
      In the second argument of ‘(==)’, namely
        ‘llvmOverride_ret provided’
      In the expression: ret == llvmOverride_ret provided
```

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
langston-barrett: I thnk not: https://github.com/GaloisInc/crucible/pull/1705#discussion_r2700220816

crux-llvm/test-data/golden/golden/double_free.pre-clang19.z3.good
langston-barrett: Wouldn't CI tell is if that test failed, given that it now uses a recent LLVM?

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics/Common.hs
RyanGlScott: Oops, you are quite right. I suppose what I actually had in mind was `eqF`, but that in turns requires the changes from https://github.com/GaloisInc/parameterized-utils/pull/132 (which haven't landed yet). Never mind!

crux-llvm/test-data/golden/golden/double_free.pre-clang19.z3.good
RyanGlScott: Ah, you're right.

In that case, I wonder why only this particular golden output needed to be changed? It's not obvious to me what would have caused the block number to change here.

langston-barrett: I'm also not sure why the block number would have changed. One thing that could cause this would be registering one fewer override, but I'm not sure why we would do that. It's possible that `isMatchingDeclaration` is a bit more strict now? The fact that it doesn't cause any real problems for Crux-LLVM or GREASE seems encouraging, but it would be nice to understand this better.

For what it's worth the output is now the same regardless of LLVM version:
```
diff -u crux-llvm/test-data/golden/golden/double_free.pre-clang19.z3.good crux-llvm/test-data/golden/golden/double_free.z3.good

<no output>
```

RyanGlScott: Gotcha. It's not worth spending a lot of time trying to figure this out, since the total number of block numbers is a somewhat fickle property that can change easily.

Now that the two golden outputs are the same, we should merge them back into a single output file.

langston-barrett: > Gotcha. It's not worth spending a lot of time trying to figure this out, since the total number of block numbers is a somewhat fickle property that can change easily.

Fair enough.

> Now that the two golden outputs are the same, we should merge them back into a single output file.

Will do. :+1:

langston-barrett:
SAW 1.5 was released, so I rebased and this is ready to go when CI passes.

PR #1707, opened 2026-01-16
llvm: Add a pretty-printing function for LLVM pointer types

langston-barrett:


langston-barrett:
> Besides pointers ("LLVM_symbol"), crucible-llvm also defines another Crucible intrinsic for memory ("LLVM_memory"). Would it be worth adding an intrinsic printer for the latter for completeness?

Sure, done!

> If we have two crucible-llvm intrinsic printers, what is the best way to compose them? Would it be worth offering a function that composes the two functions as a convenience?

Well, these functions are set up to take "fallback" callbacks as arguments, so you could compose them with each other (and the "default" intrinsic pretty-printer) that way. I don't know that there's a good super general way to compose these functions, but if you choose an `Alternative f`, you could compose two fallible intrinsic pretty-printers with `<|>`.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Pointer.hs
```diff
@@ -425,6 +431,33 @@ ppPtr (llvmPointerView -> (blk, bv))
...
 
+-- | An intrinsic-printing function for use with
+-- 'Lang.Crucible.Types.ppTypeRepr'.
+ppLLVMPointerIntrinsicType ::
+  Applicative f =>
+  -- | Fallback for other instrinsics, can be
+  -- 'Lang.Crucible.Types.ppIntrinsicDefault'.
+  (forall s ctx'. SymbolRepr s -> CtxRepr ctx' -> f (PP.Doc ann)) ->
+  SymbolRepr symb ->
+  Ctx.Assignment TypeRepr ctx ->
```
RyanGlScott: A very minor thing, but it might be nice to use `CtXRepr` consistently throughout this type signature. Currently, we use `CtxRepr` in one place and `Assignment TypeRepr` in another.

Similarly for `ppLLVMMemIntrinsicType`.

langston-barrett:
Done and done!

Issue #1614, opened 2025-11-04
`crucible-mir`: Allow `size_of_val` override to be called on trait objects and custom DSTs


RyanGlScott:
https://github.com/GaloisInc/crucible/pull/1613 adds a `crucible-mir` override for the [`size_of_val`](https://doc.rust-lang.org/std/intrinsics/fn.size_of_val.html) intrinsic, but stops short of making it support trait objects and custom DSTs like the intrinsic does. In order to do so, we would need to get the size (and maybe alignment) of the last field, and then we would need to add the sizes of the remaining fields. This would require us to augment `mir-json` to record the layout information of struct fields, which is not currently available (see https://github.com/GaloisInc/mir-json/issues/184).

RyanGlScott:
See https://github.com/GaloisInc/mir-json/issues/184#issuecomment-3487609992 for more specifics on how this would work.

RyanGlScott:
Similar considerations also apply to the [`align_of_val`](https://doc.rust-lang.org/std/intrinsics/fn.align_of_val.html) intrinsic (https://github.com/GaloisInc/crucible/issues/1708).

RyanGlScott:
https://github.com/GaloisInc/crucible/pull/1817 implemented `{size,align}_of_val` support for trait objects, so custom DST support is the last remaining thing to be done.

PR #1709, opened 2026-01-20
`crucible-mir`: Add basic `align_of_val` override

### LINKED ISSUE #1708: `crucible-mir`: Support the `align_of_val` intrinsic
It would be convenient to be able to simulate the [`align_of_val`](https://doc.rust-lang.org/std/intrinsics/fn.align_of_val.html) intrinsic:

```rs
pub const unsafe fn align_of_val<T: ?Sized>(ptr: *const T) -> usize;
```

For instance, `align_of_val` is invoked when dropping `Rc` values.

As the comments for the higher-level [`std::mem::align_of_val`](https://doc.rust-lang.org/std/mem/fn.align_of_val.html) and [`std::mem::align_of_val_raw`](https://doc.rust-lang.org/std/mem/fn.align_of_val_raw.html) functions indicate, the semantics of `align_of_val::<T>(x)` are similar to that of `align_of::<T>()`, except that `align_of_val` also works for unsized types such as slices and trait objects, e.g.,

```rs
let x: [u16; 13] = [0; 13];
let y: &[u16] = &x;
assert_eq!(2, align_of_val(y));
```

Supporting slice types `[T]` would be straightforward, as all that would be required is to `align_of::<T>()`. Supporting trait objects (e.g., `dyn Trait`) or custom DSTs would be trickier, however, as we would need to obtain the alignment of the last field at runtime. (Note that for the `Rc` use case above, it would suffice to only support sized types.)

RyanGlScott:
For now, this override does not support trait objects or custom DSTs. This is still enough to be useful in practice despite these limitations. (See #1614 for the task of supporting trait objects/custom DSTs in this override.)

This also bumps the `mir-json` submodule to bring in the changes from https://github.com/GaloisInc/mir-json/pull/221. Together with the `align_of_val` override, this is enough to finally make the `test/conc_eval/rc/rc_basic.rc` and `test/conc_eval/clos/fn_dyn_rc.rs` test cases pass.

Fixes #1708.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -5,6 +5,9 @@ This release supports [version
 `mir-json`'s schema.
 
 * Fix a translation error when translating constant slice raw pointers.
+* Add an intrinsic for [`align_of_val`](https://doc.rust-lang.org/std/intrinsics/fn.align_of_val.html),
+  which computes the alignment of a value in bytes. This works for all sized types
+  as well as a limited number of unsized types (currently, only slices).
 
 # 0.6 -- 2026-01-29
 
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -121,6 +121,7 @@ customOpDefs = Map.fromList $ [
                          , size_of
                          , size_of_val
                          , min_align_of
+                         , align_of_val
                          , intrinsics_assume
                          , assert_inhabited
                          , unlikely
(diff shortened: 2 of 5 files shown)


crucible-mir/CHANGELOG.md
```diff
@@ -57,6 +57,9 @@ This release supports [version
   values to instead return `MirAggregate` values explicitly. The
   `mirAggregate_zst`, `mirAggregate_zstIO`, and `mirAggregate_zstSim` functions
   have been introduced to make this process easier.
+* Add an intrinsic for [`align_of_val`](https://doc.rust-lang.org/std/intrinsics/fn.align_of_val.html),
+  which computes the alignment of a value in bytes. This works for all sized types
+  as well as a limited number of unsized types (currently, only slices).
```
RyanGlScott: This changelog entry may need to be moved depending on the timing of the upcoming Crux release.

RyanGlScott:
Yes, good point. I think I will park this PR until https://github.com/GaloisInc/mir-json/pull/221 lands first and then incorporate that as a submodule bump, which will allow us to remove the `FAIL` message for good.

RyanGlScott:
I've now incorporated the changes from https://github.com/GaloisInc/mir-json/pull/221 via a submodule bump, which finally allow the `rc/rc_basic` and `clos/fn_dyn_rc` test cases to pass.

PR #1711, opened 2026-01-20
crux-llvm: Add .gitignore for files generated by test suite

langston-barrett:


crux-llvm/.gitignore
```diff
@@ -0,0 +1,2 @@
+crux-build/
```
RyanGlScott: Technically speaking, `crux-build` is used for all Crux backends, not just `crux-llvm`. (`crux-mir` does create a `crux-build` subdirectory, although it doesn't yet populate it with anything.) It might be better to put `crux-build` in a top-level `.gitignore` file just to be on the safe side.

langston-barrett: Good call!

PR #1717, opened 2026-01-23
`crucible-mir`: Handle `#[repr(transparent)]` pointer-to-pointer casts

### LINKED ISSUE #1716: `crucible-mir`: Support pointer-to-pointer casts where the source is a pointer to a `#[repr(transparent)]` type
Some pointer-to-pointer casts in `crucible-mir` internally transform the pointer in some way, like the cast from `*const [T; N]` to `*const T` which performs a `subindexRef` on the pointer. If the source pointer is wrapped with a `#[repr(transparent)]` type, however, `crucible-mir` does not transform the pointer at all, and instead treats it as a generic pointer-to-pointer cast. This leads to type mismatches when the result is subsequently dereferenced. For instance:
```rust
#[repr(transparent)]
struct Wrapper {
    array: [i32; 5],
}

#[crux::test]
fn crux_test() -> i32 {
    let array = [1, 2, 3, 4, 5];
    let wrapper = Wrapper { array };
    let ptr_to_first = &raw const wrapper as *const i32;
    unsafe { *ptr_to_first }
}
```
```
$ crux-mir cast_from_ptr_repr_transparent.rs
test cast_from_ptr_repr_transparent/9f037b68::crux_test[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- cast_from_ptr_repr_transparent/9f037b68::crux_test[0] counterexamples ----
[Crux]   cast_from_ptr_repr_transparent.rs:11:14: 11:27: error: in cast_from_ptr_repr_transparent/9f037b68::crux_test[0]
[Crux]   read requires a reference to BVRepr 32, but got a reference to IntrinsicRepr MirAggregate []
[Crux]   cast_from_ptr_repr_transparent.rs:11:14: 11:27: error: in cast_from_ptr_repr_transparent/9f037b68::crux_test[0]

[Crux] Goal status:
[Crux]   Total: 2
[Crux]   Proved: 0
[Crux]   Disproved: 2
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```
As the error message indicates, `crucible-mir` thinks `ptr_to_first` is a pointer to a `MirAggregate` rather than a pointer to a bitvector.

We should apply the pointer transformations in casts where the source pointer is wrapped with a `#[repr(transparent)]` type as well.

qsctr:
Fixes #1716.

Also improves the comment for the general pointer-to-pointer cast case where the pointer is unchanged.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -943,7 +943,13 @@ evalCast' ck ty1 e ty2  = do
       (M.Misc, M.TyRawPtr M.TyStr m1, M.TyRawPtr (M.TySlice (M.TyUint M.B8)) m2)
         | m1 == m2 -> return e
 
-      -- repr(transparent) pointer-to-pointer cast
+      -- repr(transparent) pointer-to-pointer cast. Some of the
+      -- pointer-to-pointer cases above match on the specific Ty inside the
+      -- TyRawPtr. We want this to also work when the inside Ty is wrapped in a
+      -- repr(transparent) newtype, so we check for that here and recurse with
+      -- the unwrapped type. It's important that this case comes before the
+      -- general pointer-to-pointer case, which would just leave the pointer
+      -- unmodified.
       (M.Misc, M.TyRawPtr (M.TyAdt an1 _ _) m1, M.TyRawPtr _ _)
         | Just adt1 <- findAdt' col an1
         , Just fieldTy1 <- reprTransparentFieldTy col adt1


crucible-mir/src/Mir/Trans.hs
```diff
@@ -943,9 +943,16 @@ evalCast' ck ty1 e ty2  = do
       (M.Misc, M.TyRawPtr M.TyStr m1, M.TyRawPtr (M.TySlice (M.TyUint M.B8)) m2)
         | m1 == m2 -> return e
 
+      -- repr(transparent) pointer-to-pointer cast
+      (M.Misc, M.TyRawPtr (M.TyAdt an1 _ _) m1, M.TyRawPtr _ _)
+        | Just adt1 <- findAdt' col an1
+        , Just fieldTy1 <- reprTransparentFieldTy col adt1
+        -> evalCast' M.Misc (M.TyRawPtr fieldTy1 m1) e ty2
```
RyanGlScott: I wonder if we should have a similar case in `transmuteExp`? At the moment, `crucible-mir` errors at simulation time if you rewrite the `cast_from_ptr_repr_transparent` test case to use `transmute` instead of a cast, and I imagine that it's for largely the same reasons that this case fixes.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -943,9 +943,16 @@ evalCast' ck ty1 e ty2  = do
       (M.Misc, M.TyRawPtr M.TyStr m1, M.TyRawPtr (M.TySlice (M.TyUint M.B8)) m2)
         | m1 == m2 -> return e
 
+      -- repr(transparent) pointer-to-pointer cast
```
spernsteiner: Maybe worth noting that this is specifically meant to allow some of the `Ty`-based special cases above to "see through" `repr(transparent)`, and that it needs to come before the general case because the two often overlap.

crucible-mir/src/Mir/Trans.hs
spernsteiner: I'd say the general expectation right now is that `transmute` mostly doesn't work, aside from a short list of special cases.  I think it's fine to leave this case out until we come across something that needs it.

qsctr:
> I wonder i

Was that just leftover text or did you want me to do something else before merging?

RyanGlScott:
I can only speculate that a demogorgon was inhabiting my text buffer when I last reviewed this. Please ignore :)

Issue #1713, opened 2026-01-21
llvm: Additional `string.h` overrides


langston-barrett:
Overrides for the following functions from [string.h](https://en.cppreference.com/w/c/header/string) would be possible using the machinery from #1468.

- [ ] `strlen` family
  - [ ] `strlen`
  - [ ] `strnlen` (#1468)
- [ ] `strcpy` family
  - [ ] `strcpy` (#1468)
  - [ ] `strncpy`
  - [ ] `strlcpy`
  - [ ] `stpcpy`
  - [ ] `stpncpy`
  - [ ] `strcpy_s`
  - [ ] `strncpy_s`
- [ ] `stdup` family
  - [ ] `strdup`
  - [ ] `strndup`
- [ ] `strchr` family
  - [ ] `strchr`
  - [ ] `strnchr`
  - [ ] `strstr`
- [ ] `strcat` family
  - [ ] `strcat`
  - [ ] `strncat`
  - [ ] `strcat_s`
  - [ ] `strncat_s`

langston-barrett:
It is worth noting that macaw-symbolic provides a similar API for its own memory model, which is based on the code in Crucible-LLVM's `Strings` module. Any additional APIs/overrides added to Crucible-LLVM should likely be mirrored in [Data.Macaw.Symbolic.Memory.Strings](https://github.com/GaloisInc/macaw/blob/43eec89132775c5c54219f869aff5a18bed177b0/symbolic/src/Data/Macaw/Symbolic/Memory/Strings.hs).

sauclovian-g:
`strncat` is accursed and tbh my recommendation would be to just reject any attempt to verify code using it.

There is however `strlcat`

langston-barrett:
> strncat is accursed and tbh my recommendation would be to just reject any attempt to verify code using it.

There are use-cases for Crucible-LLVM beyond verification, e.g., [GREASE](https://galoisinc.github.io/grease/) is more of a bug-finding tool. In that context, a model of a cursed function could be especially useful 😄

sauclovian-g:
That's true; on the other hand, maybe "this code uses `strncat`" is itself the bug report :-)

RyanGlScott:
I think it is worth having _some_ model of `strncat`, even if it has limitations (which we should document).

sauclovian-g:
No objection, was just proposing justification for cutting a corner :-)

Anyway, I added `strlcat` to the checklist since we should have that if we have `strlcpy`.

[...]

PR #1719, opened 2026-01-27
llvm: `strdup`, `strndup`

langston-barrett:
More work on #1713. Mostly as an exercise to try out Claude code.

### DIFF
--- a/crucible-llvm-cli/test-data/strndup.cbl
+++ b/crucible-llvm-cli/test-data/strndup.cbl
@@ -1,0 +1,44 @@
+(declare @strndup ((src (Ptr 64)) (n (Bitvector 64))) (Ptr 64))
+(declare @free ((ptr (Ptr 64))) Unit)
+(declare @read-c-string ((x Pointer)) (String Unicode))
+(declare @write-c-string ((dst Pointer) (src (String Unicode))) Unit)
+
+(defun @main () Unit
+  (start start:
+    ;; Test 1: strndup("", 0)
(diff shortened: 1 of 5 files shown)


crucible-llvm-cli/test-data/strdup.cbl
```diff
@@ -0,0 +1,32 @@
...
+(declare @write-c-string ((dst Pointer) (src (String Unicode))) Unit)
+
+(defun @main () Unit
+  (start start:
+    ;; Test 1: strdup("")
+    (let src0 (alloca none (bv 64 1)))
+    (funcall @write-c-string src0 "")
+    (let dup0 (funcall @strdup src0))
+    (let d0 (funcall @read-c-string dup0))
+    (assert! (equal? d0 "") "strdup('') produces ''")
```
RyanGlScott, line 13: A minor thing, but the comment above (`;; Test 1: strdup("")`) and the text here (`"strdup('') produces ''"`) use two different conventions for string quotes (double vs. single quotes). I would favor using double quotes consistently, given that this code is meant to evoke C.

Similarly in other assertions' text here and in `strndup.cbl`.

langston-barrett: We could, but then we'd have to escape them as they appear in Crucible strings.

RyanGlScott: Fair enough. This is a very minor thing, so feel free to ignore it :)

PR #1720, opened 2026-01-28
llvm: `memcmp`, `strcmp`, `strncmp`

langston-barrett:
Towards #1713. Written with Claude code.

### DIFF
--- a/crucible-llvm/test/behavior/README.md
+++ b/crucible-llvm/test/behavior/README.md
@@ -46,8 +46,10 @@ starting with `/// `. Generally speaking, these comments closely follow the
 `printf` invocation that produces the output.
 
 **WARNING**: The tests should use `argc` to ensure that the compiler can't
-compile away calls to any overrides under test. Use `llvm-dis` to disassemble
-the bitcode to ensure it has the structure you want.
+compile away calls to any overrides under test. The test harness guarantees that
+`argc` is always 1. Use `llvm-dis` to disassemble the bitcode to ensure it has
+the structure you want. See the example below (or the existing tests) for how to
+do this.
 
 The tests have a timeout of 5s, but they should generally complete in under 1s
 for the sake of a reasonably snappy test-suite and CI.


crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -952,3 +969,319 @@ loadBytes bak mem = go
...
+
+-- | Load a byte from each of two memory locations.
+--
+-- The loader can optionally add assumptions after loading bytes when iteration
+-- continues. This is used for null-terminated string operations where we need
+-- to assume loaded bytes are non-null for termination proofs.
+data BytesLoader m sym bak wptr
+  = BytesLoader
+      { runBytesLoader :: bak -> Mem.LLVMPtr sym wptr -> Mem.LLVMPtr sym wptr -> m (Mem.LLVMPtr sym 8, Mem.LLVMPtr sym 8)
+      -- | Called when the checker returns 'Continue', allowing the loader to
```
RyanGlScott: By "the checker", do you mean `BytesChecker`? What is the relationship between `BytesChecker` and this data type?

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -952,3 +969,319 @@ loadBytes bak mem = go
 
           ptr' <- liftIO (Mem.doPtrAddOffset bak mem ptr =<< WI.bvOne sym Mem.PtrWidth)
           go acc' ptr' loader checker
+
+---------------------------------------------------------------------
+-- * Loading and checking two byte streams
+
+-- ** 'BytesLoader'
+
+-- | Load a byte from each of two memory locations.
```
RyanGlScott, line 1117: Perhaps worth leaving a reference to `ByteLoader`, which is closely related.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -952,3 +969,319 @@ loadBytes bak mem = go
...
+    }
+
+---------------------------------------------------------------------
+-- ** 'BytesChecker'
+
+-- | Compute a result from two symbolic bytes, and check if loading should
+-- continue to the next pair of bytes.
+--
+-- Used to compare two byte streams simultaneously, e.g., for 'strcmp'.
+newtype BytesChecker m sym bak a b
```
RyanGlScott, line 1204: Perhaps worth leaving a reference to `ByteChecker`, which is closely related.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -952,3 +969,319 @@ loadBytes bak mem = go
...
+  WI.SymBV sym 8 ->
+  m (Maybe (WI.SymBV sym 32))
+compareBytesForStrings bak byte1 byte2 = do
+  let sym = LCB.backendGetSym bak
+  let i32 = DPN.knownNat @32
+  eq <- liftIO (WI.bvEq sym byte1 byte2)
+  case WI.asConstantPred eq of
+    Just False -> do
+      lt <- liftIO (WI.bvUlt sym byte1 byte2)
+      negOne <- liftIO (WI.bvLit sym i32 (BV.mkBV i32 (- 1)))
```
RyanGlScott: I think it would be worth enabling the `NegativeLiterals` extension so that we can simply write `-1` without any parentheses. (Similarly for other uses of `(- 1)`.)

crucible-llvm/src/Lang/Crucible/LLVM/MemModel/Strings.hs
```diff
@@ -952,3 +969,319 @@ loadBytes bak mem = go
...
+  GHC.HasCallStack =>
+  LCB.IsSymBackend sym bak =>
+  BytesChecker m sym bak (WI.SymBV sym 32) (WI.SymBV sym 32)
+concretelyNullTerminatedStrings =
+  BytesChecker $ \bak acc byte1Ptr byte2Ptr -> do
+    byte1 <- liftIO (ptrToBv8 bak byte1Ptr)
+    byte2 <- liftIO (ptrToBv8 bak byte2Ptr)
+
+    let isNull1 = case BV.asUnsigned <$> WI.asBV byte1 of { Just 0 -> True; _ -> False }
+    let isNull2 = case BV.asUnsigned <$> WI.asBV byte2 of { Just 0 -> True; _ -> False }
```
RyanGlScott: Perhaps more legible:

```hs
    let isNull1 = (BV.asUnsigned <$> WI.asBV byte1) == Just 0
    let isNull2 = (BV.asUnsigned <$> WI.asBV byte2) == Just 0
```

langston-barrett:
I'll likely:

- [x] wait on #1725
- [x] rebase this on that
- [x] add a few behavior tests (including from the GCC suite) [EDIT: sadly, they all timed out]
- [x] update the "adding an override" doc added in this PR to mention the new test suite

langston-barrett:
Okay, done. While porting these to behavior tests, I noticed an issue described in #1729. That's the reason for all the `argc`-related changes.

crucible-llvm/doc/dev.md
```diff
@@ -0,0 +1,31 @@
...
+- Add the declaration and definition to `Intrinsics.Libc` or `Intrinsics.LLVM`.
+- Add the declaration to the appropriate list, e.g., `libc_overrides` if it
+  appears in libc or `basic_llvm_overrides` if it is an LLVM override (usually
+  starting with `@llvm.*`).
+- Add tests.
+  - Consider adding a test in `crucible-llvm/test/behavior`. See the README
+    there for further discussion.
+  - Consider adding a test in `crucible-llvm-cli/test-data/simulate`. See
+    `strdup.cbl` for an example. If the override can exhibit undefined behavior,
+    that should be covered by the test.
```
RyanGlScott: Are there guidelines to follow to figure out whether an override deserves a test in `crucible-llvm/test/behavior` or `crucible-llvm-cli/test-data/simulate` (or both)?

crucible-llvm/test/behavior/README.md
```diff
@@ -45,6 +45,10 @@ doc](../../doc/limitations.md)). The tests embed the expected output in comments
 starting with `/// `. Generally speaking, these comments closely follow the
 `printf` invocation that produces the output.
 
+**WARNING**: The tests should use `argc` to ensure that the compiler can't
+compile away calls to any overrides under test. Use `llvm-dis` to disassemble
+the bitcode to ensure it has the structure you want.
```
RyanGlScott: Give an example here of how you might use `argc`. Also, if it is guaranteed to have the value `1`, it would be worth saying as much here.

(FWIW, an alternative to using `argc` would be to call a function that is marked as `__attribute__((noinline))`.)

crucible-llvm/test/behavior/strcmp.c
```diff
@@ -0,0 +1,51 @@
+#include <string.h>
+#include <stdio.h>
+
+int main(int one, char** argv) {
```
RyanGlScott, line 4: A very minor thing, but it's a bit odd to see this argument called `one` instead of `argc` here (and in the other test cases).

crucible-llvm/doc/dev.md
langston-barrett: Added guidelines!

crucible-llvm/test/behavior/README.md
langston-barrett: There is an example in the next section, I referred the reader to that.

> (FWIW, an alternative to using argc would be to call a function that is marked as __attribute__((noinline)).)

My thought is that that's essentially asking the compiler nicely not to inline, whereas using dynamic data makes it a straight-up miscompilation to erase the call. My estimation is that it's more likely that (some version of some) compiler ends up inlining a `noinline` call than doing a flagrant miscompilation.

crucible-llvm/test/behavior/strcmp.c
langston-barrett: I agree that it's unconventional, but I think it accurately conveys the semantics in this particular case. Without this name, I think the test cases would be more confusing at first glance. I could make an `int one = argc;` declaration in each test, but that feels verbose for the same effect.

PR #1725, opened 2026-01-29
llvm: Behavior tests

### LINKED ISSUE #1721: `llvm`: Test suite for libc overrides
We often add tests for new overrides to the crucible-llvm-cli test suite. This is nice, because we can test the overrides directly without depending on a compiler or the LLVM -> Crucible translation. These tests are also quite fast. However, such tests don't (directly) check the *fidelity* of these overrides as compared to an actual implementation.

I'd like to propose that we add a new group of tests to Crucible-LLVM. These tests would consist of a C file that performs various calls to libc functions and asserts things about the result. They would be run both (1) by being compiled to a binary with a fixed version of Clang and (2) by being compiled to LLVM bitcode by that same version of Clang and simulated by Crucible-LLVM. The test would check that all asserts pass in both scenarios. To avoid Clang simply optimizing all of the code away, the C files should utilize runtime data. Specifically, they should make use of `argc` and `argv`, as these are available without function calls.

langston-barrett:
Adds a new test suite to crucible-llvm that compares the output of a C program when executed natively vs. when run in the Crucible simluator.

These tests will be helpful for checking the fidelity of our symbolic execution against a real execution. This is unlike the existing Crux-LLVM or Crucible-LLVM-CLI test suites which check hand-coded assertions. In those test-suites, there's no guarantee that the assertions *themselves* are correct.

As examples, I added a few tests for existing libc overrides.

Fixes #1721.

Potential follow-ups:

- Support `.ll` files, potentially import [lli tests](https://github.com/llvm/llvm-project/tree/d698b3a0670b2c4c3ac4039b5f42d4eef6acc53f/llvm/test/ExecutionEngine/Interpreter)

### DIFF
--- a/crucible-llvm/test/TestBehavior.hs
+++ b/crucible-llvm/test/TestBehavior.hs
@@ -109,8 +109,9 @@ compileFile outputExt cFile additionalArgs = do
     fail $ unlines $
       [ "Compilation failed!"
       , "clang " ++ unwords args
-      , "Output:"
+      , "stdout:"
       , stdout
+      , "stderr:"
       , stderr
       ]
   return outPath
@@ -130,8 +131,9 @@ runExe exePath = do
     fail $ unlines $
       [ "Native execution failed!"
       , exePath
-      , "Output:"
+      , "stdout:"
       , stdout
+      , "stderr:"
       , stderr
       ]
   return stdout


crucible-llvm/test/TestBehavior.hs
```diff
@@ -0,0 +1,208 @@
...
+compileToExecutable :: FilePath -> IO FilePath
+compileToExecutable cFile = do
+  let exePath = cFile -<.> ".exe"
+  (exitCode, stdout, stderr) <- Proc.readProcessWithExitCode "clang"
+    ["-O1", "-fsanitize=undefined", "-o", exePath, cFile] ""
+  when (exitCode /= ExitSuccess) $ do
+    putStrLn $ "Failed to compile " ++ cFile ++ " to executable"
+    putStrLn stdout
+    putStrLn stderr
+    fail "Compilation to executable failed"
```
RyanGlScott: The logic for printing the context upon a non-zero exit code is repeated quite a bit. Can you factor this bit of code out?

crucible-llvm/test/TestBehavior.hs
```diff
@@ -0,0 +1,208 @@
...
+-- | Compile a C file to a native executable
+compileToExecutable :: FilePath -> IO FilePath
+compileToExecutable cFile = do
+  let exePath = cFile -<.> ".exe"
+  (exitCode, stdout, stderr) <- Proc.readProcessWithExitCode "clang"
+    ["-O1", "-fsanitize=undefined", "-o", exePath, cFile] ""
+  when (exitCode /= ExitSuccess) $ do
+    putStrLn $ "Failed to compile " ++ cFile ++ " to executable"
+    putStrLn stdout
+    putStrLn stderr
```
RyanGlScott: It might also be worth printing a brief message beforehand (akin to "Stdout:" and "Stderr:") before each of these lines to make it more visually obvious where the stdout ends and where the stderr begins.

crucible-llvm/test/TestBehavior.hs
langston-barrett: I factored out compilation as a helper, so this pattern now only appears twice. I think it's fine to repeat it twice and might not be super clear when factored out. Take another look and let me know what you think!

crucible-llvm/test/TestBehavior.hs
langston-barrett: Good call, done

langston-barrett:
> I also think it would be worth adding a README to crucible-llvm/test/behavior which:

Yes, very good call. Done, let me know what you think!

PR #1739, opened 2026-01-30
Require exports lists in `crucible`

langston-barrett:
See the `Note` for motivation. Based on #1738, as `ForwardDataflow` was the only module actually missing an export list.

### DIFF
--- a/crucible/src/Lang/Crucible/README.hs
+++ b/crucible/src/Lang/Crucible/README.hs
@@ -1,6 +1,7 @@
 {- | This module is only for documentation purposes, and provides a high
 level overview of Crucible aimed at developers. -}
 {-# OPTIONS_GHC -Wno-unused-imports #-}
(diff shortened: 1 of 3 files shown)


crucible/src/Lang/Crucible/README.hs
```diff
@@ -1,6 +1,7 @@
 {- | This module is only for documentation purposes, and provides a high
 level overview of Crucible aimed at developers. -}
 {-# OPTIONS_GHC -Wno-unused-imports #-}
+{-# OPTIONS_GHC -Wno-missing-export-lists #-}
 module Lang.Crucible.README where
```
RyanGlScott, line 5: Could we also fix this warning by just writing `module Lang.Crucible.README () where`?

langston-barrett: See discussion on https://github.com/GaloisInc/crucible/pull/1633#discussion_r2524341055

PR #1741, opened 2026-01-30
crucible: Turn more `error`s into `panic`s

langston-barrett:
Progress towards #245.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Arch/X86.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Arch/X86.hs
@@ -137,7 +137,7 @@ $([d| {- New TH Scope -} |])
 
 -- This is going to go away
 instance ShowFC ExtX86 where
-  showFC _ _ = error "[ShowFC ExtX86] Not implmented."
+  showFC _ _ = panic "ExtX86.showFC" ["Not implemented"]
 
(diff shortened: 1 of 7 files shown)


sauclovian-g:
One of the things that came up when fixing these in SAW is that "unimplemented feature" is not really properly either a panic or a user error; it has some properties of each. The eventual conclusion was to add an additional set of error infrastructure. Dunno if you want to get into that sort of thing in Crucible just yet, but it'd probably desirable in the long run.

(unless we can eliminate all unimplemented features! but that seems unlikely)

langston-barrett:
Yeah, totally agreed. We don't really have a complete error-handling plan here, I'm just trying to improve the situation incrementally. This PR takes the user from "Oh, the tool emitted an opaque, obscure message with no further information, and I'm not sure whether I'm holding it wrong or what" to "Oh, the tool is telling me that it's an internal problem and I should file a bug". To me, this seems like a conservative change - it's strictly better, because the user at least knows that it's not their fault and they should file an issue to get support. Further triage might show it to be an unimplemented feature or a bug or something else - but at least that triage will happen as needed.

Would love to get dedicated support for unsupported features in a follow-up.

sauclovian-g:
Right, just mentioning it. And I have a feeling I keep mentioning the same things, so if I start to sound like a stuck record, go ahead and say so :-)

Issue #188, opened 2019-04-09
Crux-llvm counterexamples involving NaN


robdockins:
When constructing counterexamples involving `NaN`, `crux-llvm` outputs incorrect C syntax, (example below) which fails to compile.

```
float const crucible_values_float[] = { 3.746995e19,3.6894614e19,NaN,NaN,NaN,0.0,0.0,0.0,0.0,0.0,0.0 };
```

Some (but maybe not all?) versions of `math.h` include the `NAN` constant.  We should figure out what to do here.

atomb:
I can think of two ways to potentially handle this. The simplest might be to include the following at the beginning of each counterexample file:

````
#include <math.h>
#define NaN NAN
````

A separate possibility could be to use `isNaN` in Haskell to test values before printing and, if a value is NaN, print out the actual bit representation. This seems more accurate if we can manage it, since it would allow us to distinguish different NaN values, but trickier to make work correctly. I ran a simple experiment where I constructed a Haskell `Float` value that passed `isNaN` in Haskell, used `unsafeCoerce` to cast it to a `Word32`, printed its hex value, pasted it  into a C file, and cast it to a `float`. The resulting value caused `isnan` in C to return false, so something in that pipeline must have changed its bit pattern.

robdockins:
In C, casting from an integer type to a float will attempt to cast the numeric value, not the bitpattern.  AFAIK, directly casting the bitpattern requires either intrinsics of some sort, or using a `union` type.  I don't believe either can appear in an initializer list, unfortunately.

atomb:
Given that, the first, simpler approach seems like the way to go. The worst thing that could happen is that it'd fail to compile on some strange platform that doesn't support "quiet" NaN values, and then we could figure out how to handle things on that platform.

In my Googling, I haven't come across a single mention of a specific platform that doesn't support them, so I don't know how rare that is. And [this page](http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1011.htm) suggests making signaling NaNs optional, and quiet the default.

atomb:
Going in to try and fix this, I noticed that it's already fixed, as of 4ec946c1927.

robdockins:
Oh, oops, I really should have closed this at the time...

Issue #186, opened 2019-04-08
Crucible AST representation for What4 uninterpreted functions


travitch:
There is currently no way to represent what4 uninterpreted functions (or defined functions) in the Crucible AST.

I recently worked around this by adding a syntax extension to let me create my uninterpreted functions (as `SymExpr`s) when necessary, but being able to directly form them in Crucible would be convenient.

kquick:
Might be obsoleted if crucible-server is removed.

travitch:
Redirect to #189

PR #1742, opened 2026-02-02
Bump llvm-pretty and llvm-pretty-bc-parser

sauclovian-g:
This brings in [llvm-pretty #186](https://github.com/GaloisInc/llvm-pretty/pull/186) and [#188](https://github.com/GaloisInc/llvm-pretty/pull/188), and [llvm-pretty-bc-parser #345](https://github.com/GaloisInc/llvm-pretty-bc-parser/pull/345), which add missing support for assorted less-common floating point types.

Update a few float references to use the new stuff.

No support for bfloats yet; that requires adding them in what4. That is in turn blocked, or at least complicated, by https://github.com/GaloisInc/what4/issues/342.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/QQ.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/QQ.hs
@@ -85,6 +85,7 @@ parseSymbol = L.Symbol <$> (AT.char '@' *>
 parseFloatType :: AT.Parser L.FloatType
 parseFloatType = AT.choice
   [ pure L.Half      <* AT.string "half"
+  , pure L.BFloat    <* AT.string "bfloat"
   , pure L.Float     <* AT.string "float"
   , pure L.Double    <* AT.string "double"
   , pure L.Fp128     <* AT.string "fp128"


crucible-llvm/src/Lang/Crucible/LLVM/QQ.hs
```diff
@@ -301,6 +301,7 @@ liftTypeRepr t = case t of
   liftFloatType ft = case ft of
     L.Half      -> [| HalfFloatRepr |]
     L.Float     -> [| SingleFloatRepr |]
+    L.BFloat    -> fail "No support for Brain floats / bfloats yet"
```
RyanGlScott, line 305: It might also be worth adding a TODO to the [`parseFloatType`](https://github.com/GaloisInc/crucible/blob/73b5b9333c4dc97f0360303bec73ccdd9c32747a/crucible-llvm/src/Lang/Crucible/LLVM/QQ.hs#L85-L93) function above about needing a case for `bfloat`.

sauclovian-g: I think I can just insert the case; `L.BFloat` exists and the code itself is trivial.

PR #1743, opened 2026-02-03
crux-mir: use layout info for tuple aggregates

spernsteiner:
Currently, `TyTuple` use `MirAggregateRepr` (as do the related types `TyClosure` and `TyCoroutineClosure`), but fields are laid out sequentially, with each `MirAggregateEntry` having size 1.  This branch changes tuple handling to use real layout information emitted by mir-json.  Each field of the tuple is written to the `MirAggregate` using the actual size and alignment computed by rustc.

Using real offsets introduces a problem related to ZSTs: a zero-sized field can have the same offset as another field.  For example, in the type `((), i32)`, both fields have offset 0.  Since `MirAggregate` can only hold at most one entry for a given offset, writing both fields to offset zero will overwrite one with the other, causing a conflict (this usually appears as a type or size mismatch in `mirAggregate_insert`).  The approach taken here is to omit zero-sized fields from the `MirAggregate`.  For example, `((), i32)` is represented by a `MirAggregate` with only one entry, a `BVRepr 32` at offset 0.  ZSTs carry no data, so it's always possible to reconstitute them on the fly when read from memory.  However, often this is not even needed, since rustc omits loads and stores of zero-sized values.

---

This builds upon #1704 (that PR removed most of the hardcoded `size=1` assumptions; this one removes the rest).  I also plan to hold off merging this until I have a matching saw-script PR ready to go.

### DIFF
--- a/crucible-mir/src/Mir/TransTy.hs
+++ b/crucible-mir/src/Mir/TransTy.hs
@@ -1189,6 +1189,10 @@ tyLayoutM ty = do
 -- | Get the offset and type of each field of a type.  Returns `Nothing` if the
 -- type is not present in the `M.Collection` (as described in Note [present]).
 -- Results are given in declaration order.
+--
+-- This panics if it's called on a type that can't have fields or is otherwise
+-- unsupported.  It also panics on certain kinds of inconsistencies in the
+-- layout data retrieved from the `M.Collection`.
 tyFields :: M.Collection -> M.Ty -> Maybe [(Word, M.Ty)]
 tyFields col ty = do
     layout <- tyLayout col ty


crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1092,6 +1142,40 @@ tySizeM ty = do
...
+-- | Get the offset and type of each field of a type.  The type must be present
+-- in the `M.Collection`, as described in Note [present], or this will
+-- `mirFail`.
+tyFieldsM :: M.Ty -> MirGenerator h s ret [(Word, M.Ty)]
+tyFieldsM ty = do
+    tys <- case ty of
+        M.TyTuple tys -> return tys
+        M.TyClosure tys -> return tys
+        M.TyCoroutineClosure tys -> return tys
+        M.TyAdt {} -> mirFail "TODO: tyFieldsM struct case"
```
RyanGlScott: Are we unlikely to hit this failure in practice because we are only using `tyFieldsM` in tuple-related code paths?

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1092,6 +1142,40 @@ tySizeM ty = do
...
+tyLayoutM ty = do
+    col <- use $ cs . collection
+    case tyLayout col ty of
+        Just x -> pure x
+        Nothing -> mirFail $ "tyLayoutM: layout not found for type: " <> show ty
+
+-- | Get the offset and type of each field of a type.  The type must be present
+-- in the `M.Collection`, as described in Note [present], or this will
+-- `mirFail`.
+tyFieldsM :: M.Ty -> MirGenerator h s ret [(Word, M.Ty)]
```
RyanGlScott, line 1216: Is it guaranteed that the pairs will be sorted in ascending order of offset size? If so, that would be worth documenting here.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1634,19 +1718,21 @@ enumDiscriminant adt (MirExp enumTpr v) = do
...
-    ty <- case drop i tys of
-        x : _ -> return x
-        [] -> mirFail $ "field index " ++ show i ++
-            " is out of range for tuple " ++ show tys
+        " for tuple type " ++ show tupleTy ++ ": expected " ++ show tpr'
+    fields <- tyFieldsM tupleTy
+    (off, ty) <- case fields ^? ix i of
+        Just x -> return x
+        Nothing -> mirFail $ "tupleFieldRef: field index " ++ show i ++
+            " is out of range for tuple " ++ show tupleTy
```
RyanGlScott, line 1825: Is this failure ever reachable? I'm wondering if this should panic instead.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1065,16 +1117,14 @@ tySizedness col ty =
...
     M.TySlice _ -> Unsized
     M.TyStr -> Unsized
-    M.TyAdt adtName _ _ ->
-      case col ^? M.adts . ix adtName of
-        Nothing -> error $ "tySizedness: unknown ADT: " <> show adtName
-        Just adt -> adtSizedness col adt
+--    M.TyAdt adtName _ _ ->
+--      case col ^? M.adts . ix adtName of
+--        Nothing -> error $ "tySizedness: unknown ADT: " <> show adtName
+--        Just adt -> adtSizedness col adt
```
RyanGlScott: Should this commented-out code be removed?

crux-mir/src/Mir/Language.hs
```diff
@@ -540,13 +540,19 @@ showRegEntry col mty entry@(C.RegEntry tp rv) =
...
-          Just (MirAggregateEntry _ elemTpr elemRvPart) ->
-            goMaybe ty elemTpr elemRvPart
-          Nothing -> return "<uninit>"
+      fieldOffsets <- case col ^? layouts . ix tupleTy . _Just . layFieldOffsets . _Just of
+        Just x -> return x
+        Nothing -> fail $ "missing field offsets for " ++ show tupleTy
+      strs <- forM (zip fieldOffsets tys) $ \(off, ty) -> do
+        case col ^? layouts . ix ty . _Just . laySize of
+          Just 0 -> showZSTValue ty
+          _ -> do
```
RyanGlScott: Surely this code should distinguish between the `Just x` case (where `x` is non-zero) and the `Nothing` case? The `Nothing` case feels like it should fail or panic, given that `mir-json` really should be filling out layout information for `TyTuple`s now.

crux-mir/src/Mir/Language.hs
```diff
@@ -540,13 +540,19 @@ showRegEntry col mty entry@(C.RegEntry tp rv) =
...
+    (tupleTy@(TyTuple tys), MirAggregateRepr) -> do
       let MirAggregate _ m = rv
-      strs <- forM (zip [0..] tys) $ \(off, ty) -> do
-        case IntMap.lookup off m of
-          Just (MirAggregateEntry _ elemTpr elemRvPart) ->
-            goMaybe ty elemTpr elemRvPart
-          Nothing -> return "<uninit>"
+      fieldOffsets <- case col ^? layouts . ix tupleTy . _Just . layFieldOffsets . _Just of
+        Just x -> return x
+        Nothing -> fail $ "missing field offsets for " ++ show tupleTy
```
RyanGlScott: The expression `col ^? layouts . ix tupleTy . _Just . layFieldOffsets . _Just` combines two different failure modes: one where `tupleTy` isn't present in the `Collection`, and another where it _is_ present, but its field offsets aren't. This seems potentially confusing to me—I wonder if these should be split into two separate checks.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -835,62 +835,114 @@ buildArrayLit elemTy exps = do
...
+                  -- Omit zero-sized elements.  Rustc doesn't generate loads or
+                  -- stores for ZSTs when building MIR, so these elements will
+                  -- never be accessed.  A ZST field can have the same offset
+                  -- as another field; omitting the ZST field prevents one from
+                  -- overwriting the other.
+                  | sz /= 0 -> mirAggregate_set (fromIntegral off) sz tpr rv ag
+                  | otherwise -> return ag
                 Nothing -> return ag)
-        ag0 (zip [0..] xs)
+        ag0 (zip3 fieldTys fieldOffsets xs)
```
samcowger, line 926: It might be worth a sanity check here that these three lists are of equal length.

crucible-mir/src/Mir/TransTy.hs
samcowger: I would hazard a guess that these pairs are actually sorted in declaration order - though whichever it is is probably worth documenting.

crux-mir/src/Mir/Language.hs
```diff
@@ -540,13 +540,19 @@ showRegEntry col mty entry@(C.RegEntry tp rv) =
...
+      strs <- forM (zip fieldOffsets tys) $ \(off, ty) -> do
+        case col ^? layouts . ix ty . _Just . laySize of
+          Just 0 -> showZSTValue ty
+          _ -> do
+            case IntMap.lookup (fromIntegral off) m of
+              Just (MirAggregateEntry _ elemTpr elemRvPart) ->
+                goMaybe ty elemTpr elemRvPart
+              Nothing -> return "<uninit>"
```
samcowger, line 566: Is it worth panicking if `off` somehow exceeds the size of the aggregate?

crux-mir/test/conc_eval/drop/zst.rs
```diff
@@ -0,0 +1,27 @@
...
+}
+
+fn f<T>(x: T) {}
+
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> usize {
+    {
+        let s = S;
+        let s2 = (S,);
+        f(s2);
```
samcowger: Perhaps:
```suggestion
        std::mem::drop(s2);
```

crucible-mir/src/Mir/Mir.hs
```diff
@@ -156,6 +156,10 @@ data Layout = Layout
     _layAlign :: Word64
     -- | Size in bytes
   , _laySize :: Word64
+    -- | Offset in bytes of each field.  This is always `Just` for struct-like
+    -- types, such as `TyTuple`, `TyClosure`, and `TyAdt` with kind `Struct`.
+    -- It may be `Nothing` for other types.
+  , _layFieldOffsets :: Maybe [Word64]
```
samcowger: If these are meant to appear in declaration order, perhaps mention that here?

crucible-mir/src/Mir/TransTy.hs
samcowger: I suspect this is commented out to prevent the issue described in https://github.com/GaloisInc/crucible/pull/1704#discussion_r2748562562 from occurring. I fixed that issue in that branch, so I imagine this will resolve itself with a merge/rebase.

spernsteiner: Correct.  I've rebased now and removed this WIP commit.

crux-mir/src/Mir/Language.hs
spernsteiner: I think it's reasonable for this code to assume that the `RegValue` is valid for the type.  Ideally we would do validity checks much earlier (such as at the point of assignment), like MIRI does.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Added some comments in 2356ec8516369a1ea43fb601807056104d269ec6 noting that all lists of fields coming from mir-json are in declaration order.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Correct—added a comment in 99f127d5caa9e61349c068cbbf5daa22fbf326eb

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Changed to `panic` in 99f127d5caa9e61349c068cbbf5daa22fbf326eb.  I'm not sure if we have a consistent convention as to whether code should panic or `mirFail` on malformed input from mir-json, but there's nearby code that panics for a similar reason, so it seems fine here.

crux-mir/src/Mir/Language.hs
spernsteiner: Changed in 99f127d5caa9e61349c068cbbf5daa22fbf326eb to `fail` on `Nothing`

crux-mir/src/Mir/Language.hs
spernsteiner: There's actually a third failure mode, where `tupleTy` is present in `layouts` but maps to `Nothing`, meaning it's unsized.  I split these apart in 99f127d5caa9e61349c068cbbf5daa22fbf326eb to fail with three different messages.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Added a length check in 99f127d5caa9e61349c068cbbf5daa22fbf326eb

crux-mir/test/conc_eval/drop/zst.rs
spernsteiner: Changed to call `drop(s2)` in 99f127d5caa9e61349c068cbbf5daa22fbf326eb

crucible-mir/src/Mir/Mir.hs
spernsteiner: Added a comment in 2356ec8516369a1ea43fb601807056104d269ec6

spernsteiner:
Added a new non-monadic `tyFields` helper in 9b4cbcdc799bf29b521823dc83a05640006f9263 to support https://github.com/GaloisInc/saw-script/pull/3119

crucible-mir/src/Mir/ParseTranslate.hs
```diff
@@ -48,7 +48,7 @@ import Debug.Trace
 -- If you update the supported mir-json schema version below, make sure to also
 -- update the crux-mir README accordingly.
 supportedSchemaVersion :: Word64
-supportedSchemaVersion = 9
+supportedSchemaVersion = 10
```
RyanGlScott, line 51: Make sure to also update the `crucible-mir` and `crux-mir` changelogs (as well as the `crux-mir` README) to state that schema version 10 is supported.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -847,62 +847,117 @@ buildArrayLit elemTy exps = do
...
+
+tupleLikeFieldTysM :: M.Ty -> MirGenerator h s ret [M.Ty]
+tupleLikeFieldTysM ty =
+    case tupleLikeFieldTys ty of
+        Just x -> return x
+        Nothing -> mirFail $ "tupleLikeFieldTysM: expected tuple-like type, but got " ++ show ty
+
+{-
+Note [present]
+----------------------------------------
```
RyanGlScott: Minor formatting nit pick: I would prefer that these lines be the same length:


```suggestion
Note [present]
--------------
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1120,6 +1172,50 @@ tySizeM ty = do
...
+    layout <- tyLayout col ty
+    let offsets = case layout ^. M.layFieldOffsets of
+            Just x -> x
+            Nothing -> panic "tyFields" ["missing field offsets for type", show ty]
+    let tys = case ty of
+            M.TyTuple x -> x
+            M.TyClosure x -> x
+            M.TyCoroutineClosure x -> x
+            M.TyAdt {} -> panic "tyFields" ["TODO: tyFields struct case"]  -- For now, we only call this on tuples
+            _ -> panic "tyFields" ["tyFieldsM: unsupported type:", show ty]
```
RyanGlScott:
```suggestion
            _ -> panic "tyFields" ["tyFields: unsupported type:", show ty]
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1120,6 +1172,50 @@ tySizeM ty = do
...
+    when (length tys /= length offsets) $
+        panic "tyFieldsM" ["field count vs offset count mismatch",
```
RyanGlScott:
```suggestion
        panic "tyFields" ["field count vs offset count mismatch",
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1120,6 +1172,50 @@ tySizeM ty = do
...
+    col <- use $ cs . collection
+    case tyLayout col ty of
+        Just x -> pure x
+        Nothing -> mirFail $ "tyLayoutM: layout not found for type: " <> show ty
+
+-- | Get the offset and type of each field of a type.  Returns `Nothing` if the
+-- type is not present in the `M.Collection` (as described in Note [present]).
+-- Results are given in declaration order.
+tyFields :: M.Collection -> M.Ty -> Maybe [(Word, M.Ty)]
+tyFields col ty = do
```
RyanGlScott, line 1197: This function can also panic in addition to returning `Nothing`, so it may be worth documenting this possibility.

crucible-mir/src/Mir/ParseTranslate.hs
spernsteiner: Updated changelogs and readmes in cc0b0b3137084b99fe8ed323f871f0e7b89659da

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Added a note about the panics in bdf083d738d5fa73fb45da463824c0f4b9e16e97

PR #1745, opened 2026-02-04
llvm: Use Oughta in behavior tests to verify LLVM IR structure

### LINKED ISSUE #1729: llvm: Migrate behavior tests to Oughta
When working on #1720, I realized that my tests were compiling to a series of `printf` calls with literals. This is because they didn't depend on runtime data, so the compiler could just aggressively optimize away the very calls I was hoping to test. I fixed this by making the tests use `argc`, which is always set to 1 in the test harness. Since the compiler is unaware that `argc == 1`, it must output code that handles any value, thus including the calls of interest.

However, this is a concerning and delicate situation. Manual inspection of the LLVM assembly when the test is created are helpful, but ultimately we want to enforce such checks automatically and in perpetuity.

I propose that we migrate the test suite to use [Oughta](https://github.com/GaloisInc/oughta). Then tests could embed not only the expected output, but also the expected structure of the LLVM assembly (much like LLVM's own FileCheck-based tests do). Such a test might look like this:
```c
#include <string.h>
#include <stdio.h>

int main(int one, char** argv) {
    int zero = argc - 1;
    char src[6] = "hello";
    char dst[6] = "XXXXX";

    memcpy(&dst[zero], &src[zero], zero);
    //- check "call ptr @__memcpy_check"
    printf("memcpy zero-length: %c\n", dst[zero]);
    /// checkln "memcpy zero-length: X"

    memcpy(&dst[zero], &src[zero], one);
    //- check "call ptr @__memcpy_check"
    printf("memcpy single byte: %c\n", dst[zero]);
    /// checkln "memcpy single byte: h"
}
```
The `//-` comments are Oughta directives for checking the LLVM assembly, whereas the `///` comments are for checking the program output.

In addition to making sure that these tests are testing exactly what we want, it would give us the flexibility to expect different output and/or LLVM assembly for different versions of Clang, if the need arises.

langston-barrett:
Fixes #1729


crucible-llvm/test/TestBehavior.hs
```diff
@@ -138,23 +151,23 @@ runExe exePath = do
...
+extractLuaProgs cFile = do
+  content <- TextIO.readFile cFile
+  let contentLines = lines (Text.unpack content)
+      hasOutputChecks = any ("/// " `List.isInfixOf`) contentLines
+      hasLlvmChecks = any ("//- " `List.isInfixOf`) contentLines
+
+  unless hasOutputChecks $
+    fail $ "Test file " ++ cFile ++ " must have output checks (/// comments)"
+  unless hasLlvmChecks $
+    fail $ "Test file " ++ cFile ++ " must have LLVM IR checks (//- comments)"
```
RyanGlScott, line 165: If I'm reading this correctly, all behavior tests must have both output checks _and_ LLVM IR checks in order to be valid. Is this correct? I am a bit surprised by this, since I would have expected that just having one or the other would suffice.

RyanGlScott (approved):
Looks reasonable, but one design question inline.

crucible-llvm/test/TestBehavior.hs
langston-barrett: That's right! This is important, because they check different things. The output checks test the semantics of the overrides, whereas the LLVM IR checks test that the overrides are even being called (i.e., not optimized away by the compiler).

langston-barrett: Let me know if there's a way I could make this more clear in the README!

RyanGlScott: OK. Perhaps it would be worth saying something akin to "there must be at least one output check and at least one LLVM IR check in each behavior test" to make this clear.

langston-barrett: Fixed!

Issue #1087, opened 2023-05-05
`crucible-mir`: Include stack traces in error messages


RyanGlScott:
Currently, a typical `crux-mir` error message looks something like this:

```
Found counterexample for verification goal
  /home/rscott/Documents/Hacking/Haskell/sandbox/crucible/crux-mir/lib/core/src/ptr/alignment.rs:89:53: 89:58: error: in core/5a1f6960::ptr[0]::alignment[0]::{impl#0}[0]::new_unchecked[0]
  Translation error in core/5a1f6960::ptr[0]::alignment[0]::{impl#0}[0]::new_unchecked[0]: callExp: Don't know how to call core/5a1f6960::intrinsics[0]::{extern#0}[0]::transmute[0]::_instfab2666815496a23[0]
```

The immediate problem is that `crucible-mir` has very limited support for the `transmute` function, but such an error often masks the _actual_ source of the problem. For instance, it is common for low-level Rust code (e.g., the code that manages allocations in the implementation of Rust vectors) to use `transmute`, and even if we supported more uses of `transmute`, it is likely that you would run into other issues elsewhere in the low-level code that uses `transmute`. But this error message does not make this obvious, and you would have to do a significant amount of digging into the code before coming to this realization. This is confusing both for `crux-mir` users and for developers who are trying to debug it.

One significant step that we could perform to make things better is to include stack traces upon translation errors like the one above. `crux-llvm` already does something similar [here](https://github.com/GaloisInc/crucible/blob/4f81a72d3f0aaabcfa67e7bd69bcec00d6e40b3c/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/CallStack.hs), and it is likely that we could port the stack trace machinery over to `crux-mir` as well.

m10f:
Any idea of how to trigger this kind of error?  I think I have something that might work but I don't know what might be a good way to see if the machinery I have added works?

RyanGlScott:
Sadly, I have lost the original context for the error above. That being said, we wouldn't want to add a test case like that one anyway, as it's possible that a future version of `crucible-mir` might become powerful enough to support it. Here is a test case that more reliably demonstrates the issue:

```rs
// test.rs
extern crate crucible;
use crucible::*;

pub fn f(idx: usize) -> u32 {
    let a: [u32; 1] = [0];
    a[idx]
}

pub fn g(idx: usize) -> u32 {
    f(idx)
}

#[crux::test]
pub fn main() -> u32 {
[...]

PR #1746, opened 2026-02-09
Issue #1087 - add stack trace infrastructure to `crucible`

m10f:
Fix for issue #1087

This fix involves:
- Adding an optional dynamic exception context (like a program stack trace) to `IsSymBackend` via the `getExceptionContext` and `withExceptionContext`
- Adding a new constructor for `SimError` that includes a program stack
- Modifying the functions that make `SimError` in `Backend` to use the stack trace if it is present
- Adding a `withBackend'` function that uses a `SimState` (instead of a `SimContext`) to build instances of this backend with the context set, which might serve as a drop-in replacement for `withBackend` in many cases
- Modifying calls to `withBackend` to use `withBackend'` instead in cases relevant to the above error in `crucible-mir` (`Intrinsics`) and `crucible` (`EvalStmt`) - also `ovrWithBackend` (`OverrideSim`) now uses `withBackend'`

For any reviewers, some questions to consider:
- Does this overall approach make sense?  I think it does if we are not interested in modifying `what4` - my reasoning is that `IsSymBackend` is already expected to contextualize errors with functions like `assert`
- Iavor suggested that the `ProgramLoc` (which we get from `what4` via `IsSymInterface`)  already present in `SimError` would be different than just being the top of a program stack (if any) - do we expect this to be the case?  If not, does it make sense to collapse the two constructors of `SimError` making the location into a `NonEmpty`?
- Not much attempt has been able to migrate other uses of `withBackend` or naked calls to the `SimError` constructor - does it make sense to make a new issue for that assuming that this solution is satisfactory?

Note that despite the fact I didn't fully incorporate this in many places, call contexts have started to pop up everywhere causing many golden tests to fail - I've fixed the cases for `crux-mir` but I will continue to work through all such test cases while this gets reviewed.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/SimError.hs
+++ b/crucible/src/Lang/Crucible/Simulator/SimError.hs
@@ -60,6 +60,12 @@ data SimErrorReason
 data SimError 
    = SimErrorWithContext !ProgramLoc !SimErrorReason !(Maybe ProgramStack)
  
+-- | This pattern synonym constructs SimErrors without a program stack context when used 
+-- as an expression and ignores the program stack when used as a pattern.  It exists
+-- because SimError did not used to have a `ProgramStack`, and there are many usages
+-- in the code of the previous constructor which is approximated by this pattern.
+--
+-- Using SimErrorWithContext should be preferred.
 pattern SimError :: ProgramLoc -> SimErrorReason -> SimError
 pattern SimError { simErrorLoc, simErrorReason } <- SimErrorWithContext simErrorLoc simErrorReason _
   where SimError loc reason = SimErrorWithContext loc reason Nothing


RyanGlScott:
Adding myself as a review to remind myself to review this more closely later.

RyanGlScott:
I've taken a very cursory look at the code. Here are my initial impressions:

* This is actually quite a bit more involved than adding stack traces just for `crucible-mir`—this adds a stack trace capability for _all_ Crucible backends! It would perhaps be worth updating the title of this PR to reflect this.
* Do we want to print stack traces for _all_ Crucible backends? I think that there is merit in doing so, but at the same time, I think that the output could potentially add a lot of noise for large programs that involve deeply nested function calls. For that reason, I wonder if we should limit the number of function calls that we print in a stack trace by default, and require the user to explicitly opt in to printing a more verbose stack trace if they really want it.
* Speaking of verbosity, I do think that some of the stack traces could stand to have some information pruned. For example, in this stack trace from `crux-mir/test/symb_eval/sym_bytes/construct.good`:

  ```
  [Crux]   Context:
  [Crux]     internal: _start
  ```

  This only reports an internal location, and it's not a terribly informative one. I wonder if we should omit printing internal locations by default, instead requiring users to opt in to displaying them. If we did, then this program wouldn't need to display a context at all.
* `crucible-llvm` also has its own stack trace mechanism ([here](https://github.com/GaloisInc/crucible/blob/2ec288895afbf2a7ad72593ef5716600512ef63d/crucible-llvm/src/Lang/Crucible/LLVM/MemModel/CallStack/Internal.hs#L50-L53)), and I wonder if that can be made obsolete by using this more general stack trace mechanism that you've added.
* > Iavor suggested that the `ProgramLoc` (which we get from `what4` via `IsSymInterface`) already present in `SimError` would be different than just being the top of a program stack (if any) - do we expect this to be the case?

  I am not sure. If I understand correctly, however, when Crucible presents a `Found counterexample for verification goal` error message, it is already displaying the `ProgramLoc` from a `SimError`. We could certainly record this `ProgramLoc` as part of the overall stack trace (which gets displayed afterwards), but it would be slightly redundant to do so. For that reason, I am inclined not to put this `ProgramLoc` in the stack trace.

* > If not, does it make sense to collapse the two constructors of `SimError` making the location into a `NonEmpty`?

  I was going to suggest collapsing the two constructors of `SimError` regardless, as I think we should just always have a field representing the stack trace. It's conceivable that the stack trace might not be very interesting (or even empty), but I think that's fine—we should encode this anyway.

* > Not much attempt has been able to migrate other uses of `withBackend` or naked calls to the `SimError` constructor - does it make sense to make a new issue for that assuming that this solution is satisfactory?

  I think it could make sense to migrate these gradually, assuming that we can find a graceful way to ensure that other Crucible backends have to explicitly opt in to the stack trace machinery. In the current state of the PR, other Crucible backends appear to be already using it, but perhaps that is something that we could change.

m10f:
Thanks for the feedback, what you've said makes sense and I'll do a little work to incorporate your ideas.

RyanGlScott:
Thanks, @m10f! I realize that it is somewhat tricky to get the design right here, so let me know if you'd find value in discussing some of the finer details synchronously.

RyanGlScott:
I talked with @m10f about this. Some conclusions we reached:

* We should only enable stack traces in `crucible-mir` for now, and we should open an issue to track the task of enabling it at a later date in `crucible-llvm`. (This would be a good place to discuss how to reconcile stack traces with `crucible-llvm`'s `CallStack`s.) Doing so should make the CI pass.
* @m10f added a Crux flag to only display a subset of the stack trace by default, which can be overridden by the user if desired. He also omits the internal `_start` function from being displayed in stack traces, as this adds more noise than signal.
* @m10f will introduce a pattern synonym to represent the old definition of `SimError` to make it easier to gradually migrate bare calls to `SimError` (of which there are quite a few in the `crucible` and `saw-script` repos). Longer-term, we may want to deprecate and remove this pattern synonym, but that can be deferred to a separate issue.

m10f:
I have added the option to `crux-mir` called `--exception-context` (default 10) as well as fixing up a few other display issues.

However, the pattern synonym changed is a little confused about what counts as exported, so maybe we should defer that.

Stack traces are only visible in `crux-mir` for now, should be ready for a through review again when you have time @RyanGlScott

RyanGlScott:
> However, the pattern synonym changed is a little confused about what counts as exported, so maybe we should defer that.

Can you clarify what you mean by this? It should be possible to export a pattern synonym like any other definition.

crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
```diff
@@ -1321,6 +1349,22 @@ withBackend ::
   a
 withBackend ctx f = case _ctxBackend ctx of SomeBackend bak -> f bak
 
+-- | Get a backend from a SimState and populate the error context
+-- from the current simulation state.
+withBackend' ::
```
RyanGlScott: I'm not super fond of the name `withBackend'`, if only because it's not terribly obvious from the name alone how it differs from `withBackend`. I would suggest naming this something like `withStateBackend` instead.

Also, it would be worth stating how this function compares to `withBackend` in the Haddocks, and that we should encourage new code to be using this function over `withBackend` so that we get the expected behavior vis-a-vis stack traces.

crucible/src/Lang/Crucible/Simulator/SimError.hs
```diff
@@ -80,13 +105,34 @@ ppSimError :: SimError -> Doc ann
...
+                            ])
  where loc = simErrorLoc er
        details = simErrorDetailsMsg rsn
-       rsn = simErrorReason er
+       rsn = simErrorReason er          
+
+data ProgramStack = ProgramStack 
+  { psFrameOmitCount :: Int
+  , psFrames :: [ProgramLoc]
+  }
```
RyanGlScott, line 126: Given how important this data type is to making stack frames work, I think this data type deserves some Haddocks.

m10f:
> > However, the pattern synonym changed is a little confused about what counts as exported, so maybe we should defer that.
> 
> Can you clarify what you mean by this? It should be possible to export a pattern synonym like any other definition.

The problem is that explicitly importing the type `SimError` and its constructors (e.g. `SimError(..)`) doesn't also import the pattern `SimError` by default - so such imports need to change to something like `import Lang.Crucible.Simulator.SimError(SimError(..), pattern SimError, ...)` and enable `PatternSynonyms` if it wasn't already enabled.

While it's not a big deal to change all such imports, I think it's better to try to limit this change to the minimum scope needed.

RyanGlScott:
Ah, that's easily fixable. You can write the following in the export list:

```hs
SimError(.., SimError)
```

This will export the `SimError` pattern synonym along with `SimError`'s native data constructor. That way, when you import `SimError(..)` elsewhere, it will bring both the data constructor and the pattern synonym into scope.

m10f:
I've incorporated your feedback and I think this change might be done unless you want to pull the pattern synonym into this change

RyanGlScott:
I do think it would be worth trying out the pattern synonym suggestion—that should strike a nice balance of keeping the definition of `SimError` relatively compact while not disrupting the Crucible ecosystem too much. If you try this out and discover that it would still require invasive changes, however, let me know and we can revisit this.

crucible/src/Lang/Crucible/Simulator/SimError.hs
```diff
@@ -49,11 +57,15 @@ data SimErrorReason
...
-   = SimError
-   { simErrorLoc :: !ProgramLoc
-   , simErrorReason :: !SimErrorReason
-   }
+data SimError 
+   = SimErrorWithContext !ProgramLoc !SimErrorReason !(Maybe ProgramStack)
+ 
+pattern SimError :: ProgramLoc -> SimErrorReason -> SimError
+pattern SimError { simErrorLoc, simErrorReason } <- SimErrorWithContext simErrorLoc simErrorReason _
+  where SimError loc reason = SimErrorWithContext loc reason Nothing
```
RyanGlScott, line 71: It would be worth adding a Haddock comment here that states:

* This ignores the `ProgramStack` when used as a pattern, and constructs a trivial `ProgramStack` when used as an expression.
* Newer code is encourages to construct a program stack using `SimErrorWithContext` whenever possible.

crucible/src/Lang/Crucible/Simulator/SimError.hs
```diff
@@ -67,6 +79,9 @@ simErrorDetailsMsg (AssertFailureSimError _ msg) = msg
 simErrorDetailsMsg (Unsupported stk _) = show stk
 simErrorDetailsMsg _ = ""
 
+mkSimError :: ProgramLoc -> SimErrorReason -> Maybe ProgramStack -> SimError
+mkSimError loc reason mbCtx = SimErrorWithContext loc reason mbCtx
```
RyanGlScott, line 89: `mkSimError` is just a simple wrapper around `SimErrorWithContext`. Do we need both?

m10f: We don't, but I think we want to move away from using raw constructors for `SimError`

PR #1748, opened 2026-02-13
`crucible-mir`: Don't use `Any` to represent vtables in `DynRefType`

### LINKED ISSUE #1747: `crucible-mir`: Don't use `Any` to represent vtables in dynamic references
_(Spun off from https://github.com/GaloisInc/crucible/pull/1397#discussion_r2105008541)_

Currently, `crucible-mir` represents vtables in dynamic references (e.g., `&dyn Trait`) using Crucible's `Any` type:

https://github.com/GaloisInc/crucible/blob/2ec288895afbf2a7ad72593ef5716600512ef63d/crucible-mir/src/Mir/Intrinsics.hs#L1210-L1212

This results in some additional complexity when we retrieve the vtable in `mkVirtCall`, as we have to check at simulation time whether the `Any` type has the correct underlying type or not:

https://github.com/GaloisInc/crucible/blob/2ec288895afbf2a7ad72593ef5716600512ef63d/crucible-mir/src/Mir/Trans.hs#L2902-L2905

I claim that this complexity is unnecessary, as we can just as well encode the type of the vtable's fields directly. To do so, we would need to do the following:

1. Change the definition of `DynRefType` to something like the following:

   ```hs
   -- | The 'MirReferenceType' is the data pointer - either an immutable or mutable
   -- reference. @vtableCtx@ consists of the types of the vtable fields.
   type DynRefType vtableCtx = StructType (EmptyCtx ::> MirReferenceType ::> StructType vtableCtx)
   ```

   Also make similar changes to `DynRefRepr` and friends.
2. Add a field to [`DynMeta`](https://github.com/GaloisInc/crucible/blob/2ec288895afbf2a7ad72593ef5716600512ef63d/crucible-mir/src/Mir/Generator.hs#L139) to capture the types of the vtable's fields (of type `CtxRepr vtableCtx`, where `vtableCtx` is existentially quantified).
3. In places that construct a `DynRefRepr` or `DynMeta` value, use [`traitVtableType`](https://github.com/GaloisInc/crucible/blob/2ec288895afbf2a7ad72593ef5716600512ef63d/crucible-mir/src/Mir/TransTy.hs#L1695-L1710) to retrieve the vtable's field types.

RyanGlScott:
We now store the Crucible types of vtables directly, allowing us to avoid some simulation-time checks to unpack `Any` values when retrieving vtables.

While this is a backwards-incompatible change, it is one that appears unlikely to break much code in practice. (For instance, SAW continues to build with needing any modifications.)

Fixes #1747.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -11,6 +11,8 @@ This release supports [version
 * Extend the override for the `atomic_xchg` intrinsic to support storing
   pointer values in addition to integer values.
 * Support translating constant trait object values.
+* Change `DynRefType` and `DynRefRepr` to store the types of vtable fields
+  directly instead of erasing their types using Crucible's `Any` type.
 
 # 0.6 -- 2026-01-29
 
--- a/crucible-mir/src/Mir/Generator.hs
+++ b/crucible-mir/src/Mir/Generator.hs
@@ -12,6 +12,7 @@
 {-# LANGUAGE PolyKinds #-}
 {-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE StandaloneDeriving #-}
 {-# LANGUAGE TemplateHaskell #-}
 {-# LANGUAGE TupleSections #-}
 {-# LANGUAGE TypeApplications #-}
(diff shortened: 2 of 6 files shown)


RyanGlScott:
Unfortunately, my local testing was insufficient, as the `cursor_read` test case reveals that this new code can go into an infinite loop. Here is a minimized example:

```rs
// test.rs
pub trait Err {
    fn cause(&self) -> Option<&dyn Err> {
        None
    }
}

struct S;

impl Err for S {}

#[crux::test]
pub fn f() -> Option<&'static dyn Err> {
    S.cause()
}
```

Specifically, when computing the Crucible type representation for `dyn Err`, we have to compute the types of the vtable fields:

https://github.com/GaloisInc/crucible/blob/7d0476d3ba86636dfcd08061a78868803835e781/crucible-mir/src/Mir/TransTy.hs#L286-L290

 But note that one of the fields of the vtable for `Err` has the type `fn(&self) -> Option<&dyn Err>`, which recursively mentions `dyn Err`! As such, this code will go into an infinite loop as is.

I'll mark this PR as a draft until I've figured out a satisfactory way to break the infinite recursion.

RyanGlScott:
Thinking about this some more, I'm not sure if it is straightforward to overcome this limitation. The problem is that the Crucible type representation for `Err` is genuinely recursive, so if we don't use a type like `Any` to break the recursion, then we'd have to use something else, such as [`RecursiveType`](https://github.com/GaloisInc/crucible/blob/b99c141bcd04de0cb648dee0dcc796d2139cbfe6/crucible/src/Lang/Crucible/Types.hs#L191-L198). This would arguably require just as much extra boilerplate to achieve as using `Any`, so it's doubtful that this would result in easier-to-maintain code.

In light of this, I'm thinking of closing this PR and opening a separate one that just adds more documentation on why we opt to use `Any` to represent vtables.

spernsteiner:
Yeah, I don't see a good way to break the recursion here other than using `AnyType`.

PR #1750, opened 2026-02-16
`crucible-mir`: Document why vtable types are erased

### LINKED ISSUE #1747: `crucible-mir`: Don't use `Any` to represent vtables in dynamic references

RyanGlScott:
Resolves #1747.

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -1208,7 +1208,7 @@ type instance ExprExtension MIR = EmptyExprExtension
 type instance StmtExtension MIR = MirStmt
 
 -- | The 'MirReferenceType' is the data pointer - either an immutable or mutable
--- reference. The 'AnyType' is the vtable.
+-- reference. The 'AnyType' is the vtable. See @Note [Erase vtable types]@.
 type DynRefType = StructType (EmptyCtx ::> MirReferenceType ::> AnyType)
 
 dynRefDataIndex :: Index (EmptyCtx ::> MirReferenceType ::> AnyType) MirReferenceType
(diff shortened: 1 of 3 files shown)


crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1217,6 +1217,38 @@ dynRefDataIndex = skipIndex baseIndex
...
+Using the Any type is one way to break the recursion. Rather than translating
+all of a vtable's field types upfront, we instead just represent the entire
+vtable type as Any. Later, when making a virtual call, we check dynamically (at
+simulation time) that the Any's underlying type actually matches the expected
+vtable type, which allows us to unfold the vtable type definition once (and
+only once, to avoid infinitely recursing).
+
+An alternative approach would be to use Crucible's RecursiveType instead of
+Any. It is unclear if this is worth the effort, however, as we would have to
+invent a new symbol for each vtable type. Moreover, the amount of machinery
```
spernsteiner, line 1248: >we would have to invent a new symbol for each vtable type

This is probably not difficult—there's a table in the `Mir.Collection` that lists every trait that's used in a `dyn Trait` type (fully specialized, so `dyn Iterator<Item = u32>` has its own entry distinct from `dyn Iterator<Item = bool>`)

PR #1751, opened 2026-02-16
crux-mir: add support for arbitrary_self_types / DispatchFromDyn

### LINKED ISSUE #1718: `crucible-mir`: Support vtable receiver types other than references and raw pointers
Currently, `crucible-mir` only supports simulating vtable dispatches where the receiver type is a reference or a raw pointer (as well as a special case for `FnOnce::call_once`), as seen in [this code](https://github.com/GaloisInc/crucible/blob/6d4c184e18c537830bcd0c1cd5e0f7bec0d01a6e/crucible-mir/src/Mir/Trans.hs#L2581-L2589). These are not the only types that vtable receiver types can have, however. [This section of _The Rust Reference_](https://doc.rust-lang.org/reference/items/traits.html#r-items.traits.dyn-compatible.associated-functions) says the following about dispatchable functions:

> * Dispatchable functions must:
>   * ...
>   * Have a receiver with one of the following types:
>     * `&Self` (i.e. `&self`)
>     * `&mut Self` (i.e `&mut self`)
>     * [`Box<Self>`](https://doc.rust-lang.org/reference/special-types-and-traits.html#boxt)
>     * [`Rc<Self>`](https://doc.rust-lang.org/reference/special-types-and-traits.html#rct)
>     * [`Arc<Self>`](https://doc.rust-lang.org/reference/special-types-and-traits.html#arct)
>     * [`Pin<P>`](https://doc.rust-lang.org/reference/special-types-and-traits.html#pinp) where `P` is one of the types above

Using this, we can construct examples of code that `crucible-mir` is unable to translate. Here is an example involving `Box`:

```rs
trait TBox {
    fn m_box(self: Box<Self>) -> u32;
}

impl TBox for u32 {
    fn m_box(self: Box<Self>) -> u32 {
        *self
    }
}

#[crux::test]
fn ex_box() -> u32 {
    let d_box1: Box<u32> = Box::new(27);
    let d_box2: Box<dyn TBox> = d_box1;
    d_box2.m_box()
}
```

<details>

```
failed to generate virtual-call shim for method 1 of trait test/8bc1f2f8::TBox[0]::_trait3d63c23c8dfa8f88[0] method receiver is not `&dyn`/`&mut dyn`
  error, called at src/Mir/Trans.hs:2880:18 in crucible-mir-0.5.0.0.99-inplace:Mir.Trans
  mkVirtCall, called at src/Mir/Trans.hs:2899:23 in crucible-mir-0.5.0.0.99-inplace:Mir.Trans
  doVirtTailCall, called at src/Mir/Trans.hs:2961:17 in crucible-mir-0.5.0.0.99-inplace:Mir.Trans
  transVirtCall, called at src/Mir/Trans.hs:2412:21 in crucible-mir-0.5.0.0.99-inplace:Mir.Trans
  transInstance, called at src/Mir/Trans.hs:3076:22 in crucible-mir-0.5.0.0.99-inplace:Mir.Trans
  transCollection, called at src/Mir/ParseTranslate.hs:155:6 in crucible-mir-0.5.0.0.99-inplace:Mir.ParseTranslate
  translateMIR, called at src/Mir/Language.hs:257:16 in crux-mir-0.11.0.0.99-inplace:Mir.Language
```
</details>

And here is an example involving `Pin`:

```rs
use std::pin::Pin;

trait TPin {
    fn m_pin(self: Pin<&Self>) -> u32;
}

impl TPin for u32 {
    fn m_pin(self: Pin<&Self>) -> u32 {
        *self
    }
}

#[crux::test]
fn ex_pin() -> u32 {
    let d_pin1: Pin<&u32> = Pin::new(&27);
    let d_pin2: Pin<&dyn TPin> = d_pin1;
    d_pin2.m_pin()
}
```

<details>

```
test test/6dba90c6::ex_pin[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/6dba90c6::ex_pin[0] counterexamples ----
[Crux]   internal: error: in test/6dba90c6::TPin[0]::_vtbl5c94ca3d33cf5b83[0]$shim$test/6dba90c6::{impl#0}[0]::m_pin[0]
[Crux]   Translation error in an auto-generated shim: failed to generate vtable shim for test/6dba90c6::TPin[0]::_vtbl5c94ca3d33cf5b83[0] entry test/6dba90c6::{impl#0}[0]::m_pin[0] (instance test/6dba90c6::{impl#0}[0]::m_pin[0] ): unsupported MIR receiver type TyAdt $lang/0::Pin[0]::_adt565d5f2bbf11d9ca[0] $lang/0::Pin[0] (Substs [TyRef (TyUint B32) Immut])

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```
</details>

(I actually ran into this error when attempting to simulate code in the `ractor` crate.)

We should augment `crucible-mir` to be able to simulate these other receiver types:

* [ ] `Box<Self>`
* [ ] `Rc<Self>`
* [ ] `Arc<Self>`
* [ ] `Pin<P>`

Of these four, `Pin` would likely be the easiest to support. It is simply a `repr(transparent)` struct around another receiver type, so we would just have to unwrap the outer `Pin` constructor. The other three (`Box`, `Rc`, and `Arc`) would likely be more involved, as we'd have to pull the pointers out of the corresponding struct definitions.

spernsteiner:
Previously, calling a `dyn Trait` method only worked for methods whose receiver type was `&T`, `&mut T`, `*const T`, and `*mut T` (plus a special case for `dyn FnOnce`).  This extends crux-mir to support `#[feature(arbitrary_self_types)]`, which allows any receiver type that implements [`DispatchFromDyn`][], including `Box<Self>`, `Rc<Self>`, `Pin<&mut Self>`, and so on.  It works by implementing the compiler-internal `DispatchFromDyn` logic for walking over a struct type and removing the vtable from a pointer inside (like the opposite of [`CoerceUnsized`][], which crux-mir already supports).

Most of the changes are in the virtual-call shim, which now works by extracting the vtable pointer from the receiver value, wrapping it in `AnyRepr`, and passing it to the vtable shim.  The vtable shim already worked by unwrapping the `AnyRepr` to get the concrete receiver type and passing it to the implementation method, so it required almost no changes, aside from relaxing a requirement that the receiver type be `TyRef`/`TyRawPtr`.

Fixes #1718

[`DispatchFromDyn`]: https://doc.rust-lang.org/std/ops/trait.DispatchFromDyn.html
[`CoerceUnsized`]: https://doc.rust-lang.org/std/ops/trait.CoerceUnsized.html

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -13,6 +13,8 @@ This release supports [version
 * Extend the override for the `atomic_xchg` intrinsic to support storing
   pointer values in addition to integer values.
 * Support translating constant trait object values.
+* Support calling `dyn Trait` methods that have `Box`, `Rc`, `Arc`, or `Pin`
+  receivers, plus most uses of `#[feature(arbitrary_self_types)]`.
 
 # 0.6 -- 2026-01-29
 


crucible-mir/src/Mir/Trans.hs
```diff
@@ -2829,6 +2836,82 @@ mkShimHandleMap col halloc = mconcat <$> mapM mkHandle (Map.toList $ col ^. M.in
       where
 
 
+-- | Given a @dyn Trait@ receiver value, find the inner @&dyn Trait@ or similar
+-- pointer and strip off the vtable part.  Returns the receiver value with the
+-- vtable removed, along with the vtable itself.  This implements the logic of
+-- Rust's `DispatchFromDyn` trait; it's roughly the inverse of `corceUnsized`.
```
RyanGlScott:
```suggestion
-- Rust's `DispatchFromDyn` trait; it's roughly the inverse of `coerceUnsized`.
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -2839,81 +2922,78 @@ mkVirtCall
   => M.Collection
   -> M.TraitName
   -> Integer -- ^ The method index
-  -> C.TypeRepr recvTy -- ^ The type of the method receiver (should be @&dyn Trait@)
+  -> M.Ty -- ^ The MIR type of the method receiver
+  -> C.TypeRepr recvTy -- ^ The Crucible type of the method receiver (should be @&dyn Trait@)
```
RyanGlScott: Is this comment still accurate? As I understand it, the method receiver doesn't have to be `&dyn Trait` anymore, as it can also be `Arc`, `Rc`, `Pin`, etc.

Similar comments apply to the other "should be `&dyn Trait`" Haddocks elsewhere in this module.

RyanGlScott (approved):
Worth documenting this new feature in the `crucible-mir` changelog, I think.

PR #1756, opened 2026-02-26
Update mir-json dependency

glguy:
This change accounts for allocate, allocate_zeroed, and reallocate moving to core::crucible::alloc. While core typically should not have allocation capability we need it internally in core in order to implement certain override behavior (initially used here in core::slice::as_chunks)

Propagates https://github.com/GaloisInc/mir-json/pull/246


RyanGlScott:
Can you add a `crux-mir` test case demonstrating that it can simulate `as_chunks` without issues?

PR #1757, opened 2026-03-02
Register define overrides separately from declare overrides.

kquick:
Prior to this patch, register_llvm_overrides first processed "define" overrides and then "declare" overrides, however, the former was registering for both define AND declare overrides (via the use of allModuleDeclares) which meant that the first pass was registering *all* targets in the module, but without the declare overrides.  This was probably OK in one sense since the override registration seems to check for duplicates (and ensure they are exact duplicates), but perhaps not because the declares were getting registered with define overrides and not declare overrides.

This patch changes the "define" override registration to only process defines, leaving the declares to be processed separately.

There are no external API changes for this part, just functionality changes. However, while I was in this code, I also added an entrypoint that takes in specific lists of declares and defines instead of extracting them from an LLVM.Module; I will be using this new entrypoint in some of my upcoming changes. This adds the `register_llvm_overrides'` entrypoint.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
@@ -24,6 +24,7 @@ module Lang.Crucible.LLVM.Intrinsics
 , LLVMOverride(..)
 
 , register_llvm_overrides
+, register_llvm_overrides'
 , register_llvm_overrides_
 , llvmDeclToFunHandleRepr
 , declare_overrides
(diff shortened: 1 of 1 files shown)


langston-barrett:
I believe that my original thinking was that "`declare` overrides" should only match `declare`d functions in the LLVM module, being essentially overridden by `define`s if they are present, whereas "`define` overrides" should match *whether or not* there is a `define` or just a `declare`. So I think the code was fine as-is.

Unfortunately, there isn't adequate testing here to explain the behavior because `define_overrides` are just for C++, and C++ bitcode files get quite large.

Were you encountering some kind of strange or buggy behavior that motivated this change?

> This was probably OK in one sense since the override registration seems to check for duplicates

I don't think there would be duplicates? `define_overrides` and `declare_overrides` are non-overlapping.

kquick:
As noted, the original concerns were misplaced.  Closed in favor of #1760

PR #1758, opened 2026-03-03
Equality of `SymSequence`s

langston-barrett:
Trying again at #1569.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/SymSequence.hs
+++ b/crucible/src/Lang/Crucible/Simulator/SymSequence.hs
@@ -22,6 +22,7 @@ module Lang.Crucible.Simulator.SymSequence
 , headSymSequence
 , tailSymSequence
 , unconsSymSequence
+, eqSymSequence
 , traverseSymSequence
 , concreteizeSymSequence
 , concretizeSymSequence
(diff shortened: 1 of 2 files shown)


crucible/src/Lang/Crucible/Simulator/SymSequence.hs
```diff
@@ -319,6 +320,164 @@ unconsSymSequence sym mux = \s -> getSeqUncons <$> evalWithFreshCache f s
...
+  | RightLonger (Pred sym) (SymSequence sym a)
+  | Equal (Pred sym)
+  | Ite (Pred sym) (AsPrefixResult sym a) (AsPrefixResult sym a)
+  | NotPrefix
+
+reverseAsPrefixResult ::
+  IsExprBuilder sym =>
+  sym ->
+  AsPrefixResult sym a ->
+  IO (AsPrefixResult sym a)
```
RyanGlScott, line 357: Does this function need to be monadic? As far as I can tell, this function could be pure.

crucible/src/Lang/Crucible/Simulator/SymSequence.hs
```diff
@@ -319,6 +320,164 @@ unconsSymSequence sym mux = \s -> getSeqUncons <$> evalWithFreshCache f s
...
+          r' <- go r
+          i <- itePred sym p l' r'
+          pure i
+
+data AsPrefixResult sym a
+  = LeftLonger (Pred sym) (SymSequence sym a)
+  | RightLonger (Pred sym) (SymSequence sym a)
+  | Equal (Pred sym)
+  | Ite (Pred sym) (AsPrefixResult sym a) (AsPrefixResult sym a)
+  | NotPrefix
```
RyanGlScott, line 351: I think this data type could use some Haddocks. Some things that would be nice to point out:

* What are the `SymSequence` fields in `{Left,RightLonger}`? Do they represent the entirety of the left/right sequence? Just the suffix of the left/right sequence that makes it longer? Something else?
* The name "`Equal`", as I understand it, does not imply that the two sequences are necessarily equal, just that that are equal only if the `Pred` field is satisfiable. (Similarly for other `{Left,Right}Longer`.)

PR #1760, opened 2026-03-03
Allow user-specified list of defines and declares to apply overrides to

kquick:
Provides an alternate entrypoint for declaring overrides with specifications from the caller instead of extracted from the LLVM Module.

Also extends haddocks clarifying the functionality of this module.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
@@ -74,11 +74,11 @@ llvmIntrinsicTypes =
 -- as well as any additional overrides supplied by the user (internal overrides
 -- will supercede user overrides).
 --
--- The "define" overrides are applied to *both* the define and declares found
--- within a module.
+-- The "define" overrides are applied to *both* the @Define@s and @Declare@s
+-- elements found within a module.
 --
--- The "declare" overrides are applied only to the declares found within a
--- module.  The intent is that these overrides should only apply to declares,
+-- The "declare" overrides are applied only to the @Declare@s found within a
+-- module.  The intent is that these overrides should only apply to @Declare@s,
 -- whereas the "define" overrides should apply to any matching symbol in the LLVM
 -- @Module@.
 --


crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
```diff
@@ -67,9 +68,22 @@ llvmIntrinsicTypes =
    MapF.insert (knownSymbol :: SymbolRepr "LLVM_pointer") IntrinsicMuxFn $
    MapF.empty
 
--- | Match two sets of 'OverrideTemplate's against the @declare@s and @define@s
+-- | Match two sets of 'OverrideTemplate's against the @Declare@s and @Define@s
```
langston-barrett, line 71: I think these were capitalized like they appear in LLVM modules

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
```diff
@@ -67,9 +68,22 @@ llvmIntrinsicTypes =
...
 -- in a 'L.Module', registering all the overrides that apply and returning them
--- as a list.
+-- as a list.  There are internal pre-determined overrides that will be applied,
+-- as well as any additional overrides supplied by the user (internal overrides
+-- will supercede user overrides).
+--
+-- The "define" overrides are applied to *both* the define and declares found
+-- within a module.
```
langston-barrett:
```suggestion
-- The "define" overrides are applied to *both* the @define@s and @declare@s found
-- within a module.
```

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
```diff
@@ -67,9 +68,22 @@ llvmIntrinsicTypes =
...
+--
+-- The "declare" overrides are applied only to the declares found within a
```
langston-barrett:
```suggestion
-- The "declare" overrides are applied only to the @declare@s found within a
```

RyanGlScott (commented):
Without having looked too deeply at the implementations just yet, I'm not crazy about the fact that we have three separate API entrypoints named `register_llvm_overrides`, `register_llvm_overrides'`, and `register_llvm_overrides_`. I'd want to think if there is a more specific naming convention we can employ here to make these more visually distinct.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
```diff
@@ -67,9 +68,22 @@ llvmIntrinsicTypes =
...
+-- module.  The intent is that these overrides should only apply to declares,
```
langston-barrett:
```suggestion
-- module.  The intent is that these overrides should only apply to @declare@s,
```

kquick:
I didn't want to change the names of the existing functions because of downstream effects, so I modified the name of the new function to be more distinct.

crucible-llvm/src/Lang/Crucible/LLVM/Intrinsics.hs
kquick: I was attempting to invoke https://github.com/GaloisInc/llvm-pretty/blob/master/src/Text/LLVM/AST.hs#L816 and https://github.com/GaloisInc/llvm-pretty/blob/master/src/Text/LLVM/AST.hs#L834 ... what appearance are you referencing?

langston-barrett: The appearance in the textual LLVM IR format. No big deal, I'm happy either way!

kquick: Ah, I see. :-)  Since these are Haddock's, it probably makes more sense to reference Haskell objects than the IR text forms.

PR #1763, opened 2026-03-04
syntax: Redex spec

langston-barrett:
Add a declarative specification for crucible-syntax grammar and typing. This spec is self-documenting and serves as a reference for the complete grammar and typing rules. It can even generate pretty diagrams!

grammar

The spec is tested against the existing crucible-syntax test-suite to ensure that it accepts and rejects the same things as the implementation.

I used a coding agent extensively to build the spec.


RyanGlScott:
Adding myself as a reviewer as a reminder to look at this later.

langston-barrett:
I think this is fine as-is, but I might also attempt to make the Redex type-checker bidirectional. That would match the implementation more closely and allow it to handle some currently-disabled tests.

crucible-syntax/redex/fuzz.rkt
```diff
@@ -0,0 +1,308 @@
...
+        (with-handlers ([exn:fail? (lambda (e)
+                          (printf "  PARSE FAIL: ~a (exception: ~a)\n"
+                                  path (exn-message e))
+                          (set! parse-fail (add1 parse-fail))
+                          #f)])
+          (with-output-to-string
+            (lambda ()
+              (system (format "racket '~a' '~a' 2>&1"
+                             (path->string check-path)
+                             (path->string path)))))))
```
langston-barrett: Whoops, this is completely wrong - This should be running the `crucible-syntax` parser, not just running the Racket parser again. Will fix.

langston-barrett: Fixed.

langston-barrett:
Fuzzing in CI is timing out, will try with a lower number of tests. Not sure why it doesn't yield any output, though.

crucible-syntax/app/Main.hs
```diff
@@ -0,0 +1,29 @@
...
+      hPutStrLn stderr "Parse crucible-syntax files and exit 0 on success or 1 on failure"
+      exitFailure
+    files -> do
+      let ?parserHooks = defaultParserHooks
+      mapM_ parseFile files
+
+parseFile :: (IsSyntaxExtension ext, ?parserHooks :: ParserHooks ext) => FilePath -> IO ()
+parseFile path = do
+  contents <- T.readFile path
+  doParseCheck path contents False stderr
```
RyanGlScott, line 29: I'm not sure I understand why this `crucible-syntax` executable needs to exist. What does this do that `crucible-cli`'s `check` subcommand can't do?

.github/workflows/redex-fuzzer.yml
```diff
@@ -0,0 +1,115 @@
...
+          echo "well_typed=$WELL_TYPED" >> $GITHUB_OUTPUT
+          echo "non_well_typed=$NON_WELL_TYPED" >> $GITHUB_OUTPUT
+          echo "parse_ok=$PARSE_OK" >> $GITHUB_OUTPUT
+          echo "parse_fail=$PARSE_FAIL" >> $GITHUB_OUTPUT
+
+          # Fail if any parser failures
+          if [ "$PARSE_FAIL" -gt 0 ]; then
+            echo "ERROR: $PARSE_FAIL well-typed programs failed to parse!"
+            exit 1
+          fi
```
RyanGlScott, line 82: I am leery about checking in long series of shell commands like this into CI YAML files. I would prefer that these be factored out into scripts that the CI then invokes. Aside from making the YAML files more readable, this would also make it easier to reproduce the results of what CI is doing locally.

crucible-syntax/redex/typing.rkt
```diff
@@ -0,0 +1,775 @@
...
+;;                 ((reg ty) ...)
+;;                 Maps each mutable register name (prefixed with $) to its type.
+;;
+;;   V           — local variable environment
+;;                 ((x ty) ...)
+;;                 Maps each let-bound or argument name to its type.
+;;                 Extended by (let x e) statements; otherwise threaded unchanged.
+;;
+;;   ty           — a type (see grammar.rkt for the full ty non-terminal)
+;;   ty_ret       — return type of the enclosing function
```
RyanGlScott, line 40: It would help to add some comments to make it more visually obvious which are the inputs and outputs to the typing relation. I _think_ that `ty` and `ty_ret` are the outputs, but better to make this a bit more explicit.

langston-barrett:
Closing until I have a moment to revisit.

PR #1764, opened 2026-03-05
crucible-mir: Add missing ordering operations for booleans

glguy:


RyanGlScott:
Can you add a test case which exercises the semantics of `bool` operations? I'm imagining something like [`test/symb_eval/bitvector/arith.rs`](https://github.com/GaloisInc/crucible/blob/3bb5a0128012df00449dc0ef8cb2f92863a0f5d8/crux-mir/test/symb_eval/bitvector/arith.rs), but for `bool` instead of bitvectors.

PR #1770, opened 2026-03-06
add AggregateAsChunks_RefPath

glguy:
This adds a RefPath representation of an array reinterpreted as chunk of arrays in order to support various functions that reinterpret slices as arrays in std::slice

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -1151,6 +1151,13 @@ resizeMirAggregate (MirAggregate totalSize m) newSize
       (\off (MirAggregateEntry sz _ _) -> fromIntegral off + sz <= newSize)
       m
 
+-- | Split a `MirAggregate` in two at a given offset.  The first output
+-- contains all the entries below the split point, and the second contains all
+-- the entries above.  Offsets in the second output are shifted downward, so an
+-- entry at @off@ in the input will be at offset zero in the second output.
+--
+-- Returns `Left` if any entry spans the split point, or if the split point is
+-- out of range.
 mirAggregate_split ::
   Word ->
   MirAggregate sym ->
@@ -1181,6 +1188,13 @@ mirAggregate_split off (MirAggregate totalSize m) = do
             Nothing -> after
       in (before, after')
 
+-- | Split a `MirAggregate` into three parts: the part below @off1@, the part
+-- between @off1@ and @off2@, and the part above @off2@.  Offsets in the upper
+-- parts are shifted down, so entries at @0@/@off1@/@off2@ in the input will
+-- each be at offset zero in their respective outputs.
+--
+-- Returns `Left` if either offset is out of bounds, if @off1 > off2@, or if an
+-- entry in the input aggregate spans across either of the two split points.
 mirAggregate_split3 ::
   Word ->
   Word ->
@@ -1276,6 +1290,21 @@ mirAggregate_fromChunks ::
   IO (Either String (MirAggregate sym))
 mirAggregate_fromChunks sym offsetMap chunkedAg@(MirAggregate chunkedTotalSize _) = runExceptT $ do
   let chunkedEntries = mirAggregate_entries sym chunkedAg
+  -- For each initialized chunk, we collect:
+  -- 1. The offset of the chunk within the outer aggregate
+  -- 2. The "outer size" of the chunk, meaning the size of its entry within the
+  --    outer aggregate
+  -- 3. The predicate under which the chunk is initialized (each chunk, like
+  --    any aggregate entry, may be conditionally uninitialized)
+  -- 4. The "inner size" of the chunk, which is the total size of the aggregate
+  --    representing the chunk itself
+  -- 5. The map of offsets and entries within the aggrgegate.
+  --
+  -- TODO: hardcoded size=1: Currently the outer size is always 1, and the
+  -- inner size is the number of entries within the chunk.  Once we track
+  -- proper sizes/layouts for arrays, the inner and outer sizes will be
+  -- identical (both measuring the size of the chunk in bytes), and we can
+  -- remove one of the two from this list.
   chunkParts <- do
     let f :: MonadError String m => (Word, MirAggregateEntry sym) ->
           m (Maybe (Word, Word, Pred sym, Word, IntMap (MirAggregateEntry sym)))
@@ -1570,12 +1599,19 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
   -- by viewing the next @chunkSize * numChunks@ elements as an array of
   -- arrays.
   MirRef_AggregateAsChunks ::
-     -- | Size in bytes of each chunk
+     -- | Size in bytes of each chunk (must be concrete)
      !(f UsizeType) ->
-     -- | Number of chunks to produce
+     -- | Number of chunks to produce (must be concrete)
      !(f UsizeType) ->
      !(f MirReferenceType) ->
      MirStmt f MirReferenceType
+  -- | Print the internal representation of a `MirReference` for debugging.
+  -- This is similar to the behavior of @crucible::dump_rv@, but it's easier to
+  -- call an intrinsic from inside `Mir.Trans` / `Mir.TransCustom` cases than
+  -- it is to call a Rust function.
+  --
+  -- This could likely be expanded to accept all `RegValue`s (not just
+  -- `MirReferenceType`) in the future if needed.
   DebugPrintMirRef ::
      !(f (StringType Unicode)) ->
      !(f MirReferenceType) ->


crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2211,13 +2457,70 @@ refPathOverlaps sym path1 path2 = do
...
+      -- Check `off1 < end2 && off2 < end1`
+      offBv1 <- bvSizeLit off1
+      endBv1 <- bvSizeLit end1
+      overlapsPart1 <- liftIO $ bvUlt sym offBv1 end2
+      overlapsPart2 <- liftIO $ bvUlt sym off2 endBv1
+      overlaps <- liftIO $ andPred sym overlapsPart1 overlapsPart2
+      -- If the two regions overlap, conservatively assume that the rest of the
+      -- path may also overlap, and return true without considering `rrp1` and
+      -- `rrp2`.
+      return overlaps
```
RyanGlScott: What are `rrp1` and `rrp2` referring to in this comment?

Also, this can be simplified a bit:

```suggestion
      -- If the two regions overlap, conservatively assume that the rest of the
      -- path may also overlap, and return true without considering `rrp1` and
      -- `rrp2`.
      liftIO $ andPred sym overlapsPart1 overlapsPart2
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2211,13 +2457,70 @@ refPathOverlaps sym path1 path2 = do
...
+      -- Check `off1 < end2 && off2 < end1`
+      offBv2 <- bvSizeLit off2
+      endBv2 <- bvSizeLit end2
+      overlapsPart1 <- liftIO $ bvUlt sym off1 endBv2
+      overlapsPart2 <- liftIO $ bvUlt sym offBv2 end1
+      overlaps <- liftIO $ andPred sym overlapsPart1 overlapsPart2
+      -- If the two regions overlap, conservatively assume that the rest of the
+      -- path may also overlap, and return true without considering `rrp1` and
+      -- `rrp2`.
+      return overlaps
```
RyanGlScott: What are `rrp1` and `rrp2` referring to in this comment?

Also, this can be simplified a bit:


```suggestion
      -- If the two regions overlap, conservatively assume that the rest of the
      -- path may also overlap, and return true without considering `rrp1` and
      -- `rrp2`.
      liftIO $ andPred sym overlapsPart1 overlapsPart2
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1343,6 +1566,20 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
...
+     -- | Size in bytes of each chunk
+     !(f UsizeType) ->
+     -- | Number of chunks to produce
+     !(f UsizeType) ->
+     !(f MirReferenceType) ->
+     MirStmt f MirReferenceType
+  DebugPrintMirRef ::
+     !(f (StringType Unicode)) ->
+     !(f MirReferenceType) ->
+     MirStmt f UnitType
```
RyanGlScott, line 1618: Can you leave a comment here saying what the intended purpose of this intrinsic is? I ask since we already have a variety of debugging-related functions (e.g., `crucible::print_str`), so it's not entirely clear to me why you'd reach for `DebugPrintMirRef` instead.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2211,13 +2457,70 @@ refPathOverlaps sym path1 path2 = do
...
     go (Field_RefPath {} `RrpCons` _) _ = return $ falsePred sym
     go (Variant_RefPath {} `RrpCons` _) _ = return $ falsePred sym
     go (Just_RefPath {} `RrpCons` _) _ = return $ falsePred sym
     go (VectorIndex_RefPath {} `RrpCons` _) _ = return $ falsePred sym
     go (ArrayIndex_RefPath {} `RrpCons` _) _ = return $ falsePred sym
     go (AgElem_RefPath {} `RrpCons` _) _ = return $ falsePred sym
 
+    sizeWidth = knownNat @SizeBits
+    bvSizeLit :: Word -> MuxLeafT sym IO (SymBV sym SizeBits)
+    bvSizeLit = liftIO . bvLit sym sizeWidth . BV.mkBV sizeWidth . toInteger
```
RyanGlScott: There's a striking resemblance between this function and [`wordLit`](https://github.com/GaloisInc/crucible/blob/91a3d1ab8426aeb21fe80ce992d2867a61d154f3/crucible-mir/src/Mir/Intrinsics.hs#L755-L756), defined elsewhere in the same module. I'd suggest de-duplicating them.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1343,6 +1566,20 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
...
      !(f MirReferenceType) ->
      MirStmt f (StructType (EmptyCtx ::> MirReferenceType ::> UsizeType))
+  -- | Given a pointer to an element, return a pointer to an array constructed
+  -- by viewing the next @chunkSize * numChunks@ elements as an array of
+  -- arrays.
+  MirRef_AggregateAsChunks ::
+     -- | Size in bytes of each chunk
+     !(f UsizeType) ->
+     -- | Number of chunks to produce
+     !(f UsizeType) ->
```
RyanGlScott: It would be worth noting in the Haddocks here that the size and number of chunks must ultimately be concrete at simulation time.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -2902,6 +3248,22 @@ adjustRefPath bak iTypes v path0 adj = case path0 of
...
+          case mirAggregate_split3 off (off + chunkSize * numChunks) ag of
+            Left err -> leafAbort $ Unsupported callStack $ "mirAggregate_split3: " ++ err
+            Right x -> return x
+        chunkedAg <- case mirAggregate_toChunks sym 0 chunkSize numChunks midAg of
+          Left err -> leafAbort $ Unsupported callStack $ "mirAggregate_toChunks: " ++ err
+          Right x -> return x
+        chunkedAg' <- adj chunkedAg
+        midAg' <- liftIO (mirAggregate_fromChunks sym (* chunkSize) chunkedAg') >>= \case
+          Left err -> leafAbort $ Unsupported callStack $ "mirAggregate_fromChunks: " ++ err
+          Right x -> return x
```
RyanGlScott, line 3294: Can you make the error messages here more specific to the caller function (`adjustRefPath`)? In particular, the `mirAggregate_toChunks:` error message is used both here and in `readRefPath` below.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1122,6 +1151,200 @@ resizeMirAggregate (MirAggregate totalSize m) newSize
       (\off (MirAggregateEntry sz _ _) -> fromIntegral off + sz <= newSize)
       m
 
+mirAggregate_split ::
+  Word ->
+  MirAggregate sym ->
+  Either String (MirAggregate sym, MirAggregate sym)
```
RyanGlScott, line 1164: This function (as well as `mirAggregate_split3` below) deserve some Haddocks. In particular, I think it would be worth calling out that this function adjusts offsets downwards in the second `MirAggregate` that it returns, as this was not obvious to me from just glancing at the type signature.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1122,6 +1151,200 @@ resizeMirAggregate (MirAggregate totalSize m) newSize
...
+    splitLE k m' =
+      let (before, exact, after) = IntMap.splitLookup k m'
+          after' = case exact of
+            Just v -> IntMap.insert k v after
+            Nothing -> after
+      in (before, after')
+
+mirAggregate_split3 ::
+  Word ->
+  Word ->
```
RyanGlScott, line 1200: What are the differences between these two offsets? Is one required to be greater than or equal to the other?

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1122,6 +1151,200 @@ resizeMirAggregate (MirAggregate totalSize m) newSize
...
+  -- | Input aggregate
+  MirAggregate sym ->
+  Either String (MirAggregate sym)
+mirAggregate_toChunks sym off chunkSize numChunks ag = do
+  -- TODO: hardcoded size=1 (should use chunkSize instead)
+  let outerSize = 1
+  entries <- forM (init [0 .. numChunks]) $ \i -> do
+    chunk <- mirAggregate_chunk (off + i * chunkSize) chunkSize ag
+    return $ (fromIntegral $ i * outerSize,
+      MirAggregateEntry outerSize MirAggregateRepr (justPartExpr sym chunk))
```
RyanGlScott: Redundant `$`:


```suggestion
    return (fromIntegral $ i * outerSize,
      MirAggregateEntry outerSize MirAggregateRepr (justPartExpr sym chunk))
```

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -1122,6 +1151,200 @@ resizeMirAggregate (MirAggregate totalSize m) newSize
...
+  -- | Function for mapping offsets in the input to offsets in the output
+  (Word -> Word) ->
+  -- | Input aggregate
+  MirAggregate sym ->
+  IO (Either String (MirAggregate sym))
+mirAggregate_fromChunks sym offsetMap chunkedAg@(MirAggregate chunkedTotalSize _) = runExceptT $ do
+  let chunkedEntries = mirAggregate_entries sym chunkedAg
+  chunkParts <- do
+    let f :: MonadError String m => (Word, MirAggregateEntry sym) ->
+          m (Maybe (Word, Word, Pred sym, Word, IntMap (MirAggregateEntry sym)))
```
RyanGlScott, line 1310: There's a lot going on in this return type. Please add a comment explaining what all is being returned.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: > What are `rrp1` and `rrp2` referring to in this comment?

The other cases use the names `rrp1` and `rrp2` for the tails of the two `ReversedRefPath`s.  I changed the patterns on this case from `_` to `_rrp1`/`_rrp2` so the comment makes sense locally.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Added a comment in e81e88940b2a0070c5570452e80467f9016593c0:

> Print the internal representation of a `MirReference` for debugging.  This is similar to the behavior of `crucible::dump_rv`, but it's easier to call an intrinsic from inside `Mir.Trans` / `Mir.TransCustom` cases than it is to call a Rust function.
>
> This could likely be expanded to accept all `RegValue`s (not just `MirReferenceType`) in the future if needed.

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Deduplicated in 8d3a2b9f4

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Added a note in e81e88940b2a0070c5570452e80467f9016593c0

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Added the caller name in d7ce878a0

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Added haddocks for `split` and `split3` in e81e88940b2a0070c5570452e80467f9016593c0

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: Explained the two offsets in the new haddock in e81e88940b2a0070c5570452e80467f9016593c0

crucible-mir/src/Mir/Intrinsics.hs
spernsteiner: This signature is not for human consumption—it's only written out like this because I couldn't get type inference to succeed any other way.  I added a comment about `chunkParts` in e81e88940b2a0070c5570452e80467f9016593c0

PR #1772, opened 2026-03-09
syntax: Eliminate backtracking

langston-barrett:
We were trying ~60 alternatives via `<|>` for every expression, even though S-expressions are keyword-driven. This resulted in a ton of extra allocations and time. Instead, just dispatch on the head of the S-expression.

This brings the time of the test-suite down from 0.7s to 0.4s on my machine.

This has the nice side-effect of making error messages more specific. It also caused a weird regression in a few errors that I'm currently debugging.

### DIFF
--- a/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
+++ b/crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
@@ -563,22 +563,22 @@ synthExpr typeHint = do
         BVSBorrow_ -> compareBV BVSBorrow_ BVSBorrow
         Slt -> compareBV Slt BVSlt
         Sle -> compareBV Sle BVSle
-        BV -> bvExpr
-        BVConcat_ -> bvExpr
-        BVSelect_ -> bvExpr
-        BVTrunc_ -> bvExpr
-        BVZext_ -> bvExpr
-        BVSext_ -> bvExpr
-        BoolToBV_ -> bvExpr
-        BVAnd_ -> bvExpr
-        BVOr_ -> bvExpr
-        BVXor_ -> bvExpr
-        Sdiv -> bvExpr
-        Smod -> bvExpr
-        BVShl_ -> bvExpr
-        BVLshr_ -> bvExpr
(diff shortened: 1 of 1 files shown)


crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -1110,13 +1211,29 @@ synthBV :: forall m s ext.
...
+      BVOr_     -> naryBV BVOr_ BVOr 0
+      BVXor_    -> naryBV BVXor_ BVXor 0
+      Sdiv      -> binaryBV Sdiv BVSdiv
+      Smod      -> binaryBV Smod BVSrem
+      BVShl_    -> binaryBV BVShl_ BVShl
+      BVLshr_   -> binaryBV BVLshr_ BVLshr
+      BVAshr_   -> binaryBV BVAshr_ BVAshr
+      Negate    -> unaryBV Negate BVNeg
+      BVNot_    -> unaryBV BVNot_ BVNot
+      _         -> empty
```
RyanGlScott: One of the questions I usually have when adding a new construct to a language is "what parts of the code do I need to update"? In the case of `crucible-syntax`'s parser, there generally appears to be exactly one function that matches on the corresponding keyword in the S-expression language. (For instance, the only place that matches on `EmptyRef` is `atomSetter`.) This is a nice property, as you can be sure that once you've added the one case for your new language construct, then it's clear that parsing it will work as expected.

The exception to this rule is bitvector-related expressions, however. Their keywords are matched both here in `synthBV` and in `synthExpr`, so you have to remember to update both functions. Looking at the surrounding code more closely, I'm not sure if there is a particular reason why it needs to be this way, as the only place where `synthBV` is ever called is in `synthExpr` itself. I think the code would look clearer if `synthBV` were inlined into `synthExpr`.

langston-barrett: Yeah, I was thinking about the same thing but thought perhaps it was too big of a change. Will do!

PR #1773, opened 2026-03-09
crucible-jvm: Optimize `ClassRefs` instance for list type.

brianhuffman:
Benchmarking shows this change to cause a significant reduction in runtime and memory allocation for the `classRefs` function.

See galoisInc/saw-script#2975.


brianhuffman:
I added such a comment in 7850c664333a587c7d8a15c952e95fc3bc6f3e87.

PR #1774, opened 2026-03-11
`crucible-mir`: Introduce the `MonadAssert` typeclass and generalize some `Mir.Intrinsics` operations to work over it

qsctr:
This is mainly for SAW; `crucible-mir` only uses the `IO` instance internally, which should behave the same as before.

The corresponding SAW PR is GaloisInc/saw-script#3098.

### DIFF
--- a/crucible-mir/src/Mir/FancyMuxTree.hs
+++ b/crucible-mir/src/Mir/FancyMuxTree.hs
@@ -299,14 +299,14 @@ foldZipFancyMuxTree bak f z tx ty =
 
 -- | Map `f` over the potential values of the `FancyMuxTree`.  If `f` aborts,
 -- then the input value will not have a corresponding entry in the output.
-mapFancyMuxTree :: (IsSymBackend sym bak, OrdSkel b) =>
+mapFancyMuxTree :: (MonadAssert sym bak m, OrdSkel b) =>
     bak -> (Pred sym -> b -> b -> IO b) ->
-    (a -> MuxLeafT sym IO b) -> FancyMuxTree sym a -> IO (FancyMuxTree sym b)
+    (a -> MuxLeafT sym m b) -> FancyMuxTree sym a -> m (FancyMuxTree sym b)
 mapFancyMuxTree bak mux f t = do
     let sym = backendGetSym bak
     ys <- Maybe.catMaybes <$>
         mapM (\(x,p) -> runMuxLeafMA bak (f x >>= \y -> return (y, p)) p) (viewFancyMuxTree t)
-    buildFancyMuxTree sym mux ys
+    liftIO $ buildFancyMuxTree sym mux ys
 
 collapseFancyMuxTree :: MonadAssert sym bak m =>
     bak -> (Pred sym -> a -> a -> IO a) -> FancyMuxTree sym a -> m (Maybe a)
--- a/crucible-mir/src/Mir/Intrinsics.hs
+++ b/crucible-mir/src/Mir/Intrinsics.hs
@@ -2036,7 +2036,7 @@ subfieldMirRefIO ::
     Index ctx tp ->
     IO (MirReferenceMux sym)
 subfieldMirRefIO bak iTypes ctx ref idx =
-    modifyRefMuxIO bak iTypes (\ref' -> subfieldMirRefLeaf ctx ref' idx) ref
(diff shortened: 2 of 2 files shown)


crucible-mir/src/Mir/FancyMuxTree.hs
RyanGlScott: These comments still mention `runMuxLeafIO`, which no longer exists.

For that matter, should `runMuxLeafIO` still exist? The changelog mentions `runMuxLeafIO'` (with a trailing apostrophe) being removed, but not `runMuxLeafIO`.

crucible-mir/src/Mir/FancyMuxTree.hs
```diff
@@ -161,30 +196,22 @@ instance Monad m => MonadFail (MuxLeafT sym m) where
...
-runMuxLeafIO' bak f =
-  let sym = backendGetSym bak in
-  runMuxLeafT f (truePred sym) >>= \mx -> case mx of
-    Left msg -> liftIO $ addFailedAssertion bak msg
-    Right x -> return x
-
 -- | Run a MuxLeafT sub-computation under a more restrictive predicate.  If `f`
 -- aborts, this function will `leafAssert` the negation of `p`.
-subMuxLeafIO :: (IsSymBackend sym bak, MonadIO m) =>
+subMuxLeafIO :: MonadAssert sym bak m =>
```
RyanGlScott: We should think carefully about the naming conventions used in this module. This function is called `subMuxLeafIO`, but it's using a `MonadAssert` constraint. Other functions with a `MonadAssert` constraint (e.g., `runMuxLeafMA`) use an `*MA` suffix instead of `*IO`.

I would suggest leaving off the `*IO` suffix in this function. (We could always add a variant of this function named `subMuxLeafIO` that is specialized to `IO` if other code needs it.)

crucible-mir/src/Mir/FancyMuxTree.hs
qsctr: Whoops, I had renamed `runMuxLeafIO` to `runMuxLeafMA`, which is still used in several places, but I forgot to update the comment.

crucible-mir/src/Mir/FancyMuxTree.hs
qsctr: I agree, I had just forgotten to rename it to `subMuxLeafMA`.

qsctr:
Rebased over #1704

qsctr:
force-push: edit commit message

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -3142,17 +3142,17 @@ readRefMuxSim tpr' f ref =
   ovrWithBackend $ \bak -> do
     ctx <- getContext
     let iTypes = ctxIntrinsicTypes ctx
-    liftIO $ readRefMuxIO bak iTypes tpr' f ref
+    liftIO $ readRefMuxMA bak iTypes tpr' f ref
```
RyanGlScott, line 3145: Do you think it would be worth adding a `MonadAssert` instance for `OverrideSim`? That way, we could just call `readRefMuxMA` directly here instead of lifting the `IO` instance. (Similarly in `modifyRefMuxSim` and `readMirRefSim` below.)

qsctr: Hmm the thing is `readRefMuxSim` and `modifyRefMuxSim` take a callback function `f` which runs in the `MuxLeafT sym IO` monad, but if `readRefMuxMA` is run in the `OverrideSim` monad, we would need to lift the result of `f` into `MuxLeafT sym (OverrideSim ...)` in order to pass it to `readRefMuxMA`, which I'm not sure how to do easily.

Alternatively I could change the type of `f` to just be in `MuxLeafT sym (OverrideSim ...)` in the first place, which I think would work for the uses of `modifyRefMuxSim` in this module (`readRefMuxSim` doesn't seem to be used), but these functions are technically exported, and while there don't seem to be any downstream consumers from a cursory search, I'm hesitant to change the interface of these functions without a legitimate reason.

There is no issue with calling `readMirRefMA` directly in `readMirRefSim`, but I'm not sure it's worth making an instance just to avoid a single use of `liftIO` either.

RyanGlScott: Ah, I clearly didn't think through the implications of my suggestion that carefully. I retract my idea.

PR #1779, opened 2026-03-17
Execution feature for eager proving of obligations

### LINKED ISSUE #369: Interleave proof with simulation
It would be nice to have an option to interleave the attempted proof of side-conditions with symbolic simulation.  This would allow us to report safety condition violations earlier, especially in cases where symbolic simulation would otherwise not terminate, or terminate only with timeout or iteration limits.

In addition, perhaps, it would be good to have an option to spawn off separate concurrent threads to attempt proofs while symbolic simulation proceeds.  This might allow us to make better use of multiple cores (although the memory additional memory pressure might cause this to be problematic).

langston-barrett:
Fixes #369, though we should make a new issue for the async side.

I'd like to use this in anger before merging, I'm not yet confident it's correct. I wish we had a testing setup that made it easier to test execution features like this in-repo...

### DIFF
--- a/crucible/crucible.cabal
+++ b/crucible/crucible.cabal
@@ -215,6 +215,7 @@ library
     Lang.Crucible.Simulator.BoundedRecursion
     Lang.Crucible.Simulator.CallFrame
     Lang.Crucible.Simulator.Evaluation
+    Lang.Crucible.Simulator.EagerProving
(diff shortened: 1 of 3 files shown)


langston-barrett:
Closing until I have time to revisit.

PR #1781, opened 2026-03-17
Fix `onlineProve`

langston-barrett:
It was completely busted 🙃 My fault for adding it without adequate tests! It wasn't (1) assuming the assumptions nor (2) negating the goal.

### DIFF
--- a/crucible-cli/test/Overrides.hs
+++ b/crucible-cli/test/Overrides.hs
@@ -3,6 +3,7 @@
 {-# LANGUAGE GADTs #-}
 {-# LANGUAGE LambdaCase #-}
 {-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
 {-# LANGUAGE TypeApplications #-}
 {-# LANGUAGE TypeOperators #-}
 
@@ -26,6 +27,7 @@ import qualified Data.Parameterized.Map as MapF
 
 import qualified What4.Concretize as WC
 import What4.Expr.Builder
+import What4.FunctionName (FunctionName)
 import What4.Interface
 import What4.ProgramLoc
 import qualified What4.Protocol.Online as WPO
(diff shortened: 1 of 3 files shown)


crucible-cli/test/Overrides.hs
```diff
@@ -69,7 +70,11 @@ setupOverrides bak ha =
                      <*> pure (UseOverride (mkOverride "proveObligations" (proveObligations (Just sym))))
      f6 <- FnBinding <$> mkHandle ha "crucible-print-assumption-state"
                      <*> pure (UseOverride (mkOverride "crucible-print-assumption-state" (printAssumptionState (Just sym))))
-     return [(f1, InternalPos),(f2,InternalPos),(f3,InternalPos),(f4,InternalPos),(f5,InternalPos),(f6,InternalPos)]
+     f7 <- FnBinding <$> mkHandle ha "prove-offline"
+                     <*> pure (UseOverride (mkOverride "prove-offline" (proveOffline (Just sym))))
+     f8 <- FnBinding <$> mkHandle ha "prove-online"
+                     <*> pure (UseOverride (mkOverride "prove-online" (proveOnline bak (Just sym))))
+     return [(f1, InternalPos),(f2,InternalPos),(f3,InternalPos),(f4,InternalPos),(f5,InternalPos),(f6,InternalPos),(f7,InternalPos),(f8,InternalPos)]
```
RyanGlScott: This list has grown long enough that I think it would behoove us to define a helper function that adds the `InternalPos` part. That way, we don't have to repeat it eight times in this list.

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -111,19 +111,21 @@ consumeGoals onAssumption onGoal onConj = go
 consumeGoalsWithAssumptions ::
   forall asmp goal a.
   Monoid asmp =>
-  -- | Consume a 'Prove'
+  -- | What to do at 'Assuming' nodes (e.g., call proverAssume)
```
RyanGlScott: Perhaps:


```suggestion
  -- | What to do at 'Assuming' nodes (e.g., call 'proverAssume')
```

PR #1782, opened 2026-03-18
Avoid calling SMT solver for trivial goals

langston-barrett:


RyanGlScott:
I realize this is still a draft, but the title caught my eye. To my knowledge, Crucible already makes an effort to avoid using SMT solvers when dealing with trivial proof goals. See the definition of `addProofObligation` here:

https://github.com/GaloisInc/crucible/blob/53190210bd4760e7adb1f46d7625667bffc08817/crucible/src/Lang/Crucible/Backend.hs#L277-L287

Are there situations for which this wouldn't suffice? The only thing that comes to mind are _durable_ proof obligations (i.e., those added by [`addDurableProofObligation`](https://github.com/GaloisInc/crucible/blob/53190210bd4760e7adb1f46d7625667bffc08817/crucible/src/Lang/Crucible/Backend.hs#L289-L294)), but we intentionally do not optimize those away.

langston-barrett:
Yeah, I think there are two, and you just pointed out one of them! Even if we're not dropping durable proof obligations, that doesn't mean we should spin up an SMT solver to check them if they're trivially true. Additionally, the check you highlighted doesn't look at assumptions at all, whereas the check I'm adding here does. If assumptions are `false`, then the goal is always provable.

RyanGlScott:
> Additionally, the check you highlighted doesn't look at assumptions at all, whereas the check I'm adding here does.

True. For what it's worth, there are already some existing checks for trivial assumptions:

https://github.com/GaloisInc/crucible/blob/53190210bd4760e7adb1f46d7625667bffc08817/crux/src/Crux/Goal.hs#L437
https://github.com/GaloisInc/crucible/blob/53190210bd4760e7adb1f46d7625667bffc08817/crucible/src/Lang/Crucible/Backend/Assumptions.hs#L255

I hope we're not inadvertently duplicating work.

langston-barrett:
A good point. I still think this won't be redundant, though I'll be sure to do a review and check (and some debug printing just to be sure). Consider the following:
```
(let p (fresh Bool))
(let q (fresh Bool)
(assume! p)
(assume! (not p))
(assert! q)
```
The assumptions and goal are individually nontrivial - it's only their *conjunction* that What4 would collapse to `true`. And AFAIK, we only traverse the assumptions + goal here, when proving.

langston-barrett:
I asked an agent to investigate and it confirmed my hypothesis:

<details>
<summary>Agent output</summary>

```
Summary

  I've added printf-debugging to the checkTrivialGoal function in
  /work/crucible/src/Lang/Crucible/Backend/Prove.hs and verified it works with
  three test cases:

  ✅ Test 1: Contradictory Assumptions (Your Example)

  File: /work/test_trivial_contradict.cbl
  (assume! p)
  (assume! (not p))
  (assert! q)
  Result: *** TRIVIAL GOAL OPTIMIZATION FIRED ***
  - Assumptions simplify to false (contradictory)
  - asmpsAndNotGoal = false → Goal is trivially provable
  - SMT solver NOT called

  ✅ Test 2: Goal in Assumptions

  File: /work/test_trivial_implies.cbl
  (assume! p)
  (assert! p)
  Result: *** TRIVIAL GOAL OPTIMIZATION FIRED ***
  - asmpsAndNotGoal = false (because p AND not p = false)
  - SMT solver NOT called

  ❌ Test 3: Non-trivial Goal (Control)

  File: /work/test_nontrivial.cbl
  (assume! p)
  (assert! q)
  Result: Trivial goal check: Not trivially true, calling solver...
  - asmpsAndNotGoal is a real symbolic expression: and p (not q)
  - Optimization correctly does NOT fire
  - SMT solver called → COUNTEREXAMPLE found

  The debug output confirms that the HEAD commit's optimization correctly
  identifies and skips the SMT solver for trivial goals arising from
  contradictory assumptions!
```
  
</details>

crucible/src/Lang/Crucible/Backend/Prove.hs
```diff
@@ -257,6 +257,43 @@ data Prover sym m t r
...
+--
+-- We don't optimize trivially false goals because we need the solver to provide
+-- concrete counterexample values ('WE.GroundEvalFn') for debugging and error
+-- reporting.
+--
+-- You might wonder if this is redundant with e.g., the check in @Backend@'s
+-- @addProofObligation@ or with 'Crucible.Backend.Assumption.trivialAssumption'.
+-- It is not. Consider a situation with fresh booleans @p@ and @q@ and
+-- assumptions @p@ and @not p@ and goal @q@. The conjunction of the assumptions
+-- and the negation of the goal is unsatisfiable, even though each assumption
```
RyanGlScott: To clarify:


```suggestion
-- and the negation of the goal is trivially unsatisfiable, even though each assumption
```

You might also mention that this is the case because of What4's ability to simplify expressions as they're constructed.

langston-barrett:
> which may impact things like SAW

FWIW, I don't think SAW or Crux use `Crucible.Backend.Prove`: https://github.com/search?q=owner%3AGaloisInc+%2FCrucible%5C.Backend%5C.Prove%2F&type=code

Will add to changelog and update that comment for sure though!

PR #1783, opened 2026-03-19
`reverse` for `SymSequence`

langston-barrett:
... to enable more general `RecordAndReplay`

### DIFF
--- a/crucible-cli/test-data/simulate/seq-test1.cbl
+++ b/crucible-cli/test-data/simulate/seq-test1.cbl
@@ -28,5 +28,13 @@
     (let v3 (from-just (seq-head (from-just (seq-tail s2) "cdr s2")) "cadr s2"))
     (assert! (equal? 42 v3) "cadr s2 test")
 
+    (let r (seq-reverse s2))
+    (assert! (equal? 2 (seq-length r)) "reverse length")
+    (let rv1 (from-just (seq-head r) "head r"))
+    (assert! (equal? 42 rv1) "reversed first element")
+    (let r_tail (from-just (seq-tail r) "cdr r"))
+    (let rv2 (from-just (seq-head r_tail) "cadr r"))
+    (assert! (equal? 5 rv2) "reversed second element")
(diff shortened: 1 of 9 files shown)


langston-barrett:
(AFAIK the only consumer of `RecordAndReplay` is GREASE, and I'm going to update that myself)

crucible/src/Lang/Crucible/Simulator/SymSequence.hs
```diff
@@ -195,6 +197,26 @@ appendSymSequence _sym xs ys =
...
+reverseSymSequence :: forall sym a. sym -> SymSequence sym a -> IO (SymSequence sym a)
+reverseSymSequence sym = \s -> coerce (evalWithFreshCache f s)
+  where
+    f :: (SymSequence sym a -> IO (Const (SymSequence sym a) a))
+      -> SymSequence sym a -> IO (Const (SymSequence sym a) a)
+    f loop = \case
+      SymSequenceNil -> pure (coerce SymSequenceNil)
+      SymSequenceCons _ v tl ->
+        do tl' <- coerce (loop tl)
+           coerce (appendSymSequence sym tl' =<< consSymSequence sym v SymSequenceNil)
```
RyanGlScott, line 210: Do you have a sense for how efficient this is? My reading of this is that we are performing the symbolic equivalent of `tl ++ [v]` here, which could potentially be quadratic if the `SymSequence` is built up entirely with `SymSequenceCons`es. That being said, maybe this isn't actually as severe in practice, since `SymSequence`s might actually consist of more `SymSequenceAppend`s (for which reversing them is cheap).

langston-barrett: `appendSymSequence` is `O(1)`, so I don't think this will be `O(n^2)`.

langston-barrett:
> do you think it is worth adding a seq-reverse operation to crucible-syntax as well?

Sure, will do!

PR #1784, opened 2026-03-19
`crucible-mir`: break down `Mir.Intrinsics` into modules

samcowger:
`Mir.Intrinsics` currently weighs in at 3,579 lines, and I find that its size and scope makes working in it effectively to be a bit challenging. I've tried to break it down into a few separate modules along reasonable lines.

This PR should contain no functional changes, nor break compatibility with any consumers of `Mir.Intrinsics` - it re-exports everything defined in its new subsidiary modules.

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics/Dyn.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Dyn.hs
@@ -46,7 +46,7 @@ pattern DynRefCtx = Empty :> MirReferenceRepr :> AnyRepr
 -- | The representation for a @&dyn Tr@/@&mut dyn Tr@. Both use the same
 -- representation: a pair of a data value (which is either @&Ty@ or @&mut Ty@)
 -- and a vtable. The vtable is type-erased (`AnyRepr`); see @Note [Erase vtable
--- types]@ in "Mir.Intrinsics". See also `DynRefCtx`.
+-- types]@. See also `DynRefCtx`.
 pattern DynRefRepr :: () => (tp ~ DynRefType) => TypeRepr tp
 pattern DynRefRepr = StructRepr DynRefCtx
 


crucible-mir/src/Mir/Intrinsics/Syntax.hs
```diff
@@ -0,0 +1,10 @@
+{-# LANGUAGE TypeFamilies #-}
+
+module Mir.Intrinsics.Syntax where
+
+import Lang.Crucible.CFG.Extension (EmptyExprExtension, ExprExtension)
+
+-- | Sigil type indicating the MIR syntax extension
+data MIR
+
+type instance ExprExtension MIR = EmptyExprExtension
```
spernsteiner, line 149: IMO either these should be moved back to `Intrinsics.hs` alongside the `StmtExtension`, or the `StmtExtension` and related stuff should be moved here, so that the entire syntax extension is defined in one place

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -0,0 +1,929 @@
...
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TypeApplications #-}
+{-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
+{-# LANGUAGE TypeSynonymInstances #-}
+{-# LANGUAGE ViewPatterns #-}
+
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+module Mir.Intrinsics.Aggregate where
```
RyanGlScott: A request: since you are going through the effort of creating new modules, I'd greatly appreciate it if you could add explicit export lists for each new module. At the moment, it is very difficult to tell what is truly meant to be part of the public API versus what are internals that are accidentally exported. (For now, it would suffice to just export everything—we can always refine this later.)

crucible-mir/src/Mir/Intrinsics/Syntax.hs
RyanGlScott: Also, this would avoid needing to define `type instance StmtExtension MIR = MirStmt` as an orphan instance.

crucible-mir/src/Mir/Intrinsics.hs
```diff
@@ -29,1429 +29,57 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
```
RyanGlScott: I think this pragma can now be removed.

crucible-mir/src/Mir/Intrinsics/Syntax.hs
samcowger: It turns out that, if `MIR`'s module contains or depends on either `MirStmt` or `execMirStmt`, we can't define `OverrideSim` functions (e.g. `Mir.Intrinsics.Reference.newMirRefSim`) in the same module as their `IO`/`MuxLeafT` equivalents (e.g. `Mir.Intrinsics.Reference.newMirRefIO`). `newMirRefSim`'s type mentions `MIR`, so its module must depend on `MIR`'s module, but if `MIR`'s module has/needs `MirStmt` (which mentions `Mir.Intrinsics.Reference.MirReferenceType`) or `execMirStmt` (a caller of `newMirRefIO`), then we get a dependency cycle.

Because it seemed desirable to be able to define such function pairs together, I opted to carve `MIR` off from `MirStmt`/`execMirStmt`. Knowing that, would you still prefer to rejoin `MIR` with its instances, and define a separate module (or modules) for `OverrideSim` functions?

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Great idea. I'll do so, and will open an issue for refining them.

crucible-mir/src/Mir/Intrinsics.hs
samcowger: Not quite yet - this still defines a couple of orphan instances for `MIR`, pending disposition of the thread above about unorphaning them.

crucible-mir/src/Mir/Intrinsics/Syntax.hs
RyanGlScott: Ah, I see. In that case, perhaps we should split of the `OverrideSim` functions into a separate module. (For what it's worth, I don't think anyone except SAW is actually making use of these functions.)

samcowger: I am reminded that we also define helper functions for slices (`mkSlice`, `getSlicePtr`, `getSliceLen`) in a way that depends on `MIR`, so the same cycle problem arises with them if we reunite `MIR` and `MirStmt`. It would feel awfully silly to move these into a separate module - do you think we should make them extension-polymorphic rather than `MIR`-specific, as we seem to have done for enum helpers?

I suppose we could even do the same for `OverrideSim` functions, and sidestep this whole issue...

RyanGlScott: If making them extension-polymorphic sidesteps the issue, then I would be in favor of that.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Done, and see #1786.

crucible-mir/src/Mir/Intrinsics/Dyn.hs
```diff
@@ -0,0 +1,84 @@
...
+dynRefVtableIndex :: Index (EmptyCtx ::> MirReferenceType ::> AnyType) AnyType
+dynRefVtableIndex = lastIndex (incSize $ incSize zeroSize)
+
+pattern DynRefCtx :: () => (ctx ~ EmptyCtx ::> MirReferenceType ::> AnyType) => Assignment TypeRepr ctx
+pattern DynRefCtx = Empty :> MirReferenceRepr :> AnyRepr
+
+-- | The representation for a @&dyn Tr@/@&mut dyn Tr@. Both use the same
+-- representation: a pair of a data value (which is either @&Ty@ or @&mut Ty@)
+-- and a vtable. The vtable is type-erased (`AnyRepr`); see @Note [Erase vtable
+-- types]@ in "Mir.Intrinsics". See also `DynRefCtx`.
```
RyanGlScott: Now that the Note is in the same module as this comment, we can remove this bit (which is now stale):


```suggestion
-- and a vtable. The vtable is type-erased (`AnyRepr`); see @Note [Erase vtable
-- types]@. See also `DynRefCtx`.
```

samcowger: Good catch, thanks!

PR #1787, opened 2026-03-23
`crucible-mir`: Enable `UndecidableInstances` in more places for GHC 9.4 support

RyanGlScott:
GHC 9.6 or later do not require `UndecidableInstances` to build `crucible-mir` after https://github.com/GaloisInc/crucible/pull/1784, but GHC 9.4 does require it. This wasn't caught in `crucible`'s CI (which only tests GHC 9.6 through 9.10), but since SAW's CI still requires supporting GHC 9.4, this was caught when attempting to bump the `crucible` submodule in SAW. (See https://github.com/GaloisInc/saw-script/pull/3119#issuecomment-4114120763).

For backwards compatibility with GHC 9.4, this enables `UndecidableInstances` in the places that need them.

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
@@ -14,6 +14,9 @@
 {-# LANGUAGE TypeSynonymInstances #-}
 {-# LANGUAGE ViewPatterns #-}
 
+-- See: https://ghc.haskell.org/trac/ghc/ticket/11581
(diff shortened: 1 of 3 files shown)


langston-barrett:
Should we perhaps add comments that say that these are only necessary for GHC <= 9.4, so we can remove them when that GHC version is really old and no longer used by any Crucible consumers?

RyanGlScott:
Good call. I've changed each module to add a `See: https://ghc.haskell.org/trac/ghc/ticket/11581` comment next to each additional use of `UndecidableInstances`.

PR #1790, opened 2026-03-31
Label the Crux Docker images to appear in the Crux repository.

aschwerdfeger-galois:
This should list future versions of the `crux-llvm` and `crux-mir` Docker images in the new Crux repository instead of this one.

### DIFF
--- a/.github/workflows/crux-llvm-build.yml
+++ b/.github/workflows/crux-llvm-build.yml
@@ -268,6 +268,7 @@ jobs:
         include:
           - file: .github/Dockerfile-crux-llvm
             image: ghcr.io/galoisinc/crux-llvm
+            image-repo: GaloisInc/crux
             cache: ghcr.io/galoisinc/cache-crux-llvm
     steps:
       - uses: actions/checkout@v6
@@ -292,6 +293,8 @@ jobs:
         id: labels
         with:
           images: ${{ matrix.image }}
+          label-custom: |
+             org.opencontainers.image.source=https://github.com/${{ matrix.image-repo }}
 
       - uses: docker/login-action@v4
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/crux-llvm-build.yml
```diff
@@ -292,6 +293,8 @@ jobs:
         id: labels
         with:
           images: ${{ matrix.image }}
+          label-custom: |
+             org.opencontainers.image.source=https://github.com/${{ matrix.image-repo }}
```
RyanGlScott, line 297: As I understand it, this is a temporary measure until the `crux` CI is the one responsible for building the Docker images, correct? If so, it would be worth citing the relevant issue in the `crux` repo that tracks moving this part over to `crux`'s CI.

(Similarly for the related line in `crux-mir.build.yml`.)

aschwerdfeger-galois: My understanding was that the Crux code in this repository was staying here and the CI harness along with it. Is the ultimate goal to shift the code into the new repository?

langston-barrett: I'm also a bit unclear on the new repo shift. Perhaps we should have some GitHub issues tracking the goal and what needs to be done along the way?

sauclovian-g:
Don't forget to do this for the crux-mir-comp image in the SAW repo :-)

.github/workflows/crux-llvm-build.yml
RyanGlScott: Ah, perhaps I misremembered. I found https://github.com/GaloisInc/crux/issues/4, but that proposes to keep the Docker images as they are. (Eventually, I suppose we'll make a separate `crux` Docker image that includes everything from both `crux-llvm` and `crux-mir`, but that will be a different matter entirely.)

PR #1794, opened 2026-04-01
`crux-mir`: strengthen check for override type mismatch

### LINKED ISSUE #1793: `crux-mir`: catch more mistyped overrides
When `crux-mir` sees `crucible::override_(f, g)`, it checks whether `f` and `g` have the same type by examining the `FnHandle`s associated with them, which means checking equivalence of `CtxRepr`s for argument types and `TypeRepr`s for return types: https://github.com/GaloisInc/crucible/blob/8f8156a795253ef84c83bac4857120cd9682bd3e/crux-mir/src/Mir/Overrides.hs#L403-L410

This check is too lenient, as it's quite possible to accidentally override a function of one Rust type with another function of a different Rust type, and the downstream errors that result from doing so can be difficult to diagnose. For instance, one can "successfully" override `&A -> T`  with `&B -> T`, as `&A` and `&B` both use equivalent `MirReferenceRepr`s, or override `(A, B) -> T` with `(B, A) -> T`, as `(A, B)` and `(B, A`) both use equivalent `MirAggregateRepr`s.

It would be appropriate to strengthen the check to operate on the `Mir.Ty`s associated with functions' arguments and returns.

samcowger:
Fixes #1793.

### DIFF
--- a/crux-mir/src/Mir/Overrides.hs
+++ b/crux-mir/src/Mir/Overrides.hs
@@ -422,13 +422,18 @@ overrideRust cs name = do
   -- above, but are needed for type inference to succeed.
   Refl <- case testEquality (handleArgTypes fhF) (handleArgTypes fhG) of
     Just x -> return x
-    Nothing -> fail $ "type mismatch: original and override argument lists don't match: " ++
-      show (handleArgTypes fhF, handleArgTypes fhG)
+    Nothing -> panic "overrideRust"
+      [ "type mismatch: original and override argument lists don't match",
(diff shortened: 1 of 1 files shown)


crux-mir/src/Mir/Overrides.hs
RyanGlScott: I think these `fails` should be converted to `panic`s, as it should be impossible for these `testEquality` calls to fail now.

samcowger: Good idea, done.

PR #1796, opened 2026-04-01
Optimize performance of `SymSequence` `reverse`

langston-barrett:
Keep long `Cons`-sequences as `Cons`-sequences.

Add performance tests.

### DIFF
--- a/crucible/test/helpers/SymSequence.hs
+++ b/crucible/test/helpers/SymSequence.hs
@@ -5,6 +5,11 @@ module SymSequence (tests) where
 import Test.Tasty qualified as TT
 
 import SymSequence.Properties qualified as Properties
+import SymSequence.Reverse qualified as Reverse
 
 tests :: TT.TestTree
-tests = Properties.tests
+tests =
+  TT.testGroup "SymSequence"
+  [ Properties.tests
+  , Reverse.tests
+  ]
--- a/crucible/test/helpers/SymSequence/Reverse.hs
+++ b/crucible/test/helpers/SymSequence/Reverse.hs
@@ -1,0 +1,174 @@
+{-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE ImportQualifiedPost #-}
+{-# LANGUAGE RankNTypes #-}
+
+module SymSequence.Reverse (tests) where
+
+import Data.Parameterized.Nonce qualified as Nonce
+import Data.Parameterized.Some (Some(Some))
+import Lang.Crucible.Backend (SomeBackend(SomeBackend), backendGetSym)
+import Lang.Crucible.Backend.Simple (newSimpleBackend)
+import Lang.Crucible.Simulator.SymSequence (SymSequence)
+import Lang.Crucible.Simulator.SymSequence qualified as S
+import System.Timeout (timeout)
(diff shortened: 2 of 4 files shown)


crucible/test/helpers/SymSequence.hs
```diff
@@ -1,308 +1,15 @@
...
-asConstPred ::
-  WI.IsExprBuilder sym =>
-  proxy sym ->
-  WI.Pred sym ->
-  Bool
-asConstPred _proxy p =
-  case WI.asConstantPred p of
-    Just True -> True
-    Just False -> False
-    Nothing -> panic "asConstPred" ["non-constant predicate"]
```
RyanGlScott, line 81: A lot of this commit appears to be just moving definitions from `SymSequence` to `SymSequence.Properties`. Could you split this out into a separate commit to make this easier to review?

crucible/test/helpers/SymSequence/Reverse.hs
```diff
@@ -0,0 +1,174 @@
...
+---------------------------------------------------------------------
+-- Helpers
+
+mkBackend :: IO (Some SomeBackend)
+mkBackend = do
+  sym <- newExprBuilder FloatIEEERepr EmptyExprBuilderState Nonce.globalNonceGenerator
+  Some . SomeBackend <$> newSimpleBackend sym
+
+-- | Build a sequence, reverse it, and check that length is preserved.
+--   Must complete within 5 seconds.
```
RyanGlScott, line 123: Is it reasonable to assume that these tests will complete well within the five-second threshold on most machines? (I haven't checked.) I do worry about the potential for these tests to flake out on particularly slow setups, but perhaps these tests are now so fast that that isn't a real concern.

langston-barrett: Hard to say for sure. Here's the timings on my machine:
```
    reverse performance
      reverse large cons-list:                            OK (0.77s)
      reverse cons-list with periodic muxes:              OK (0.59s)
      reverse left-nested appends:                        OK (0.79s)
      reverse right-nested appends:                       OK (0.11s)
      reverse left-deep mux chain:                        OK (1.12s)
      reverse right-deep mux chain:                       OK (0.68s)
      reverse balanced appends:                           OK (0.75s)
      reverse balanced muxes:                             OK (1.13s)
```
I can post those of CI when that completes. We could definitely consider bumping to e.g., 10s or just using them to track regressions.

RyanGlScott: OK. Five seconds is likely enough wiggle room, but I agree that we should reserve the right to bump this if the tests prove flaky on certain machines.

langston-barrett: CI timings:
```
    reverse performance
      reverse large cons-list:                            OK (1.67s)
      reverse cons-list with periodic muxes:              OK (1.33s)
      reverse left-nested appends:                        OK (1.69s)
      reverse right-nested appends:                       OK (0.24s)
      reverse left-deep mux chain:                        OK (2.49s)
      reverse right-deep mux chain:                       OK (1.42s)
      reverse balanced appends:                           OK (1.70s)
      reverse balanced muxes:                             OK (2.55s)
```

PR #1802, opened 2026-04-07
Update for new llvm-pretty UnnamedMdIdx newtype wrapper.

kquick:


crucible-llvm/src/Lang/Crucible/LLVM/TypeContext.hs
```diff
@@ -193,8 +193,8 @@ lookupAlias i =
     Just stp -> return stp
     Nothing  -> throwError $ unwords ["Unknown type alias", show i]
 
-lookupMetadata :: (?lc :: TypeContext) => Int -> Maybe L.ValMd
-lookupMetadata x = view (at x) (llvmMetadataMap ?lc)
+lookupMetadata :: (?lc :: TypeContext) => L.UnnamedMdIdx -> Maybe L.ValMd
+lookupMetadata x = view (at (L.unnamedMdIdx x)) (llvmMetadataMap ?lc)
```
RyanGlScott, line 197: Since this is part of the public API, it would be worth noting that `lookupMetadata`'s type changed in `crucible-llvm`'s changelog.

RyanGlScott (approved):

PR #1806, opened 2026-04-11
Fix `execResultGlobals` on `PartialRes`

langston-barrett:
It previously ignored the possible branch inside `FinishedResult`, always returning the globals from the non-aborted side.

### DIFF
--- a/crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
+++ b/crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
@@ -480,7 +480,7 @@ execStateSimState = \case
 
 abortedGlobals ::
   Monad f =>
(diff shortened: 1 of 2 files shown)


crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
RyanGlScott: Perhaps the Haddocks here (and elsewhere?) should be generalized a bit, as this is now used for more than just `AbortedBranch`.

RyanGlScott (approved):

crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
langston-barrett: Good point, done!

PR #1809, opened 2026-04-13
Fix `uniquelyConcRegMap` blocking clause: use OR not AND

langston-barrett:
The blocking clause in `uniquelyConcRegMap` used conjunction (AND) to combine per-entry "not equal" predicates. This meant if any single RegMap entry was uniquely determined, the conjunction was UNSAT, falsely reporting the entire map as unique. Use disjunction (OR) instead so that uniqueness is only reported when all entries are uniquely determined.

Adds a `FloatModeRepr` arg to the overrides hook for this.

### DIFF
--- a/crucible-cli/test/Overrides.hs
+++ b/crucible-cli/test/Overrides.hs
@@ -64,15 +64,16 @@ mkOverrideBinding ha name override = do
 -- | Set up all test overrides
 setupOverrides ::
   ( IsSymBackend sym bak
-  , sym ~ ExprBuilder scope st fs
+  , sym ~ ExprBuilder scope st (Flags fm)
   , SymExpr sym ~ Expr scope
-  , bak ~ CBO.OnlineBackend solver scope st fs
+  , bak ~ CBO.OnlineBackend solver scope st (Flags fm)
   , WPO.OnlineSolver solver
   ) =>
   bak ->
+  FloatModeRepr fm ->
   HandleAllocator ->
   IO [(FnBinding p sym ext, Position)]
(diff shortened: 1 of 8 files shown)


crucible-cli/test/Overrides.hs
```diff
@@ -292,3 +294,31 @@ proveOnline bak _proxy =
       let prover = CB.onlineProver sym sp
       let strat = CB.ProofStrategy prover CB.keepGoing
       CB.proveCurrentObligations bak strat proveChecker
+
+-- | Test unique concretization of two integers via 'uniquelyConcRegMap'.
```
RyanGlScott, line 298: It might be worth saying that this tests whether concretization is unique or not by printing out a specific message, so this must be observed by recording the program's output.

crucible/src/Lang/Crucible/Concretize.hs
```diff
@@ -669,8 +669,8 @@ uniquelyConcRegMap bak fm iFns sFns m = do
           preds <- Ctx.zipWithM notEq concM mAssign
           p <-
             foldlMFC
-              (\p (Const p') -> W4I.andPred sym p p')
-              (W4I.truePred sym)
+              (\p (Const p') -> W4I.orPred sym p p')
+              (W4I.falsePred sym)
               preds
```
RyanGlScott, line 675: It might be worth a comment here saying what `p` is, i.e., `not (modelA_1 == modelB_1) \/ ... \/ not (modelA_n == modelB_n)`.

RyanGlScott (approved):

PR #1815, opened 2026-04-22
Add function handle symbol name to error message.

kquick:
Shows the symbol name for the function that could not be found.

Addresses https://github.com/GaloisInc/saw-script/issues/2198

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -481,7 +481,7 @@ evalStmt bak = eval
              do p <- Partial.annotateME sym mop (BadFunctionPointer lookupErr) (falsePred sym)
                 loc <- getCurrentProgramLoc sym
                 let estr = "Failed to load function handle: "
-                           <> maybe  "<unidentified>" id gsym
+                           <> fromMaybe "<unidentified>" gsym
                 let err = SimError loc (AssertFailureSimError estr
                                         (show (ME.ppFuncLookupError lookupErr)))
                 addProofObligation bak (LabeledPred p err)
--- a/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
@@ -480,7 +480,10 @@ evalStmt bak = eval
            Left lookupErr -> lift $
              do p <- Partial.annotateME sym mop (BadFunctionPointer lookupErr) (falsePred sym)
                 loc <- getCurrentProgramLoc sym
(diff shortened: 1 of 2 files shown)


RyanGlScott:
Interestingly, this error message used to display the function name here at one point in time, but the function name was removed in https://github.com/GaloisInc/crucible/pull/915. The rationale is given in https://github.com/GaloisInc/crucible/pull/915#discussion_r744785298:

> Note that the function name that is removed from this message appears here at line 9. It was just duplicated between the `MemoryOp` and the `MemoryErrorReason`.

And indeed, we can see that in the new `invoke-test.z3.good` output, the function name appears twice:

https://github.com/GaloisInc/crucible/blob/4371eb80a7560f56576bdb55708e512fcea6a3a3/crux-llvm/test-data/golden/golden/invoke-test.z3.good#L3-L9

That being said, I think we should revisit this choice. As noted in https://github.com/GaloisInc/saw-script/issues/2198, SAW doesn't display anything after the `No implementation or override found for pointer` part of the error message for performance reasons (see https://github.com/GaloisInc/saw-script/issues/2198#issuecomment-2622491308), which means that the latter occurrence of the function name simply doesn't appear in SAW's error messages. Printing the function name earlier in error messages (as is done in this PR), on the other hand, is a relatively cheap way to ensure that SAW always prints the function name. This does mean that `crux-llvm`'s error messages will have duplicate information, but I think this is a small price to pay for making the debugging experience with SAW easier.

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
```diff
@@ -480,7 +480,10 @@ evalStmt bak = eval
            Left lookupErr -> lift $
              do p <- Partial.annotateME sym mop (BadFunctionPointer lookupErr) (falsePred sym)
                 loc <- getCurrentProgramLoc sym
-                let err = SimError loc (AssertFailureSimError "Failed to load function handle" (show (ME.ppFuncLookupError lookupErr)))
+                let estr = "Failed to load function handle: "
+                           <> maybe  "<unidentified>" id gsym
```
RyanGlScott: A little shorter:


```suggestion
                           <> fromMaybe "<unidentified>" gsym
```

RyanGlScott (approved):

crucible-llvm/src/Lang/Crucible/LLVM/MemModel.hs
kquick: I will usually use `maybe .. id` to minimize the change and avoid adding an import, but apparently mis-checked this one because indeed, `Data.Maybe` is already imported.  Updated.

langston-barrett (approved):

PR #1817, opened 2026-04-27
crux-mir: store size/align in vtable

spernsteiner:
Currently, vtables in crux-mir store only the method `FunctionHandle`s.  In rustc the vtable contains a bit more information, notably including the size and alignment of the erased type.  This is used for two major purposes:

1. Size and align are used to implement `size_of_val` and `align_of_val` on `&dyn Trait`.
2. Align is used to compute the offset of the last field of custom DSTs with `dyn Trait` tails.  `CustomDST<dyn Trait>` must have the same layout in memory as `CustomDST<ErasedType>`, so the offset / padding of the final `dyn Trait` field depends on the alignment of `ErasedType`.

This branch adds the size and align fields to the vtable and uses them to implement the (non-custom DST) `dyn Trait` case of `size_of_val` and `align_of_val`.  The offset computation for custom DSTs will be part of the future migration of structs to use `MirAggregateRepr`.

Fixes the non-custom case of #1614

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1306,16 +1306,16 @@ mkTraitObject traitName' vtableName e = do
         Just x -> return x
         Nothing -> error $ "missing vtable definition for " ++ show vtableName
 
-    let info =
-            [ MirExp UsizeRepr $ R.App $ usizeLit $ fromIntegral $ vtable ^. vtSize
-            , MirExp UsizeRepr $ R.App $ usizeLit $ fromIntegral $ vtable ^. vtAlign
-            ]
-
     let mkEntry :: MirHandle -> MirExp s
         mkEntry (MirHandle _ _ fh) =
             MirExp (C.FunctionHandleRepr (FH.handleArgTypes fh) (FH.handleReturnType fh))
                 (R.App $ E.HandleLit fh)
-    vtableExp@(MirExp vtableTy _) <- return $ mkStructExp $ info ++ map mkEntry handles
+    let vtParts = VtableParts
+            { vtpMethods = map mkEntry handles
+            , vtpSize = R.App $ usizeLit $ fromIntegral $ vtable ^. vtSize
+            , vtpAlign = R.App $ usizeLit $ fromIntegral $ vtable ^. vtAlign
+            }
+    vtableExp@(MirExp vtableTy _) <- return $ mkStructExp $ vtablePartsToEntries vtParts
 
     -- Check that the vtable we constructed has the appropriate type for the
     -- trait.  A mismatch would cause runtime errors at calls to trait methods.
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -2731,11 +2731,6 @@ splitMethodArgs args argsSize =
     (R.AtomExpr arg0, fmapFC R.AtomExpr args')
 
 
-type (x :: k) :<: (xs :: Ctx.Ctx k) = Ctx.SingleCtx x Ctx.<+> xs
-
-(<:) :: forall f tp ctx. f tp -> Ctx.Assignment f ctx -> Ctx.Assignment f (tp :<: ctx)
-x <: xs = Ctx.singleton x Ctx.<++> xs
-
 data AssignUncons (f :: k -> Type) :: Ctx.Ctx k -> Type where
   AssignUncons :: f x -> Ctx.Assignment f y -> AssignUncons f (Ctx.SingleCtx x Ctx.<+> y)
 
@@ -2752,16 +2747,6 @@ assignUncons (xs Ctx.:> x) =
   case assignUncons xs of
     Right (AssignUncons y ys) -> Right $ AssignUncons y (ys Ctx.:> x)
     Left Refl -> Right $ AssignUncons x Ctx.Empty
-
-data AsFunctionHandleRepr :: C.CrucibleType -> Type where
-  AsFunctionHandleRepr ::
-    C.CtxRepr args ->
-    C.TypeRepr ret ->
-    AsFunctionHandleRepr (C.FunctionHandleType args ret)
-
-asFunctionHandleRepr :: C.TypeRepr tp -> Maybe (AsFunctionHandleRepr tp)
-asFunctionHandleRepr (C.FunctionHandleRepr args ret) = Just $ AsFunctionHandleRepr args ret
-asFunctionHandleRepr _ = Nothing
 
 unappendAssignment :: forall k f (xs :: Ctx.Ctx k) (ys :: Ctx.Ctx k).
     Ctx.Size ys ->
@@ -2960,8 +2945,7 @@ dispatchFromDyn dynTraitName recvTy recvExp die = do
 -- function.
 mkVirtCall
   :: HasCallStack
-  => M.Collection
-  -> M.TraitName
+  => M.TraitName
   -> Integer -- ^ The method index
   -> M.Ty -- ^ The MIR type of the method receiver
   -> C.TypeRepr recvTy -- ^ The Crucible type of the method receiver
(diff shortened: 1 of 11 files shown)


crux-mir/README.md
```diff
@@ -49,7 +49,7 @@ Next, navigate to the `crucible/dependencies/mir-json` directory. Install
 [the `mir-json` README][mir-json-readme].
 
 Currently, `crux-mir` supports [version
-10](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#10) of
+11](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#11) of
```
RyanGlScott, line 52: Make sure to update the relevant line in the changelog as well.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -2533,20 +2562,22 @@ callOnceVirtShimDef methodIdx = CustomMirOp $ \ops ->
       Some vtableCtx <- case vtableStructTpr of
         C.StructRepr ctx -> return $ Some ctx
         _ -> mirFail $ "callOnceVirtShimDef: vtable type is not a struct"
-      Some vtableMethodIdx <- case Ctx.intIndex (fromInteger methodIdx) (Ctx.size vtableCtx) of
+      let methodSlotIdx = numVtableInfoSlots + fromInteger methodIdx
+      Some vtableMethodSlotIdx <- case Ctx.intIndex methodSlotIdx (Ctx.size vtableCtx) of
```
RyanGlScott: Perhaps we should factor out the logic for creating a proper vtable `Index` into its own function. Manually adding an `Int` offset seems a bit fragile.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1899,6 +1899,18 @@ getLayoutFieldAsMirExp opName layoutFieldLens ty =
 
 -- Vtable handling
 
+-- | Non-method fields present at the start of every vtable.
```
RyanGlScott, line 1916: Ignorant question: does `crucible-mir`'s representation of vtables include destructors? I am presuming that the answer is "no", given that these usually appear at the start of vtables in most material I've seen on the subject (e.g., [this article](https://geo-ant.github.io/blog/2023/rust-dyn-trait-objects-fat-pointers/)), but I wanted to check.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1325,13 +1361,6 @@ align_of_val = (["core", "intrinsics", "align_of_val"], \substs -> case substs o
     _ -> Nothing
     )
   where
-    unsupportedTraitObject :: Ty -> MirGenerator h s ret a
-    unsupportedTraitObject ty =
-        mirFail $ unlines
-            [ "align_of_val does not currently support trait objects"
-            , "In the type " ++ show (PP.pretty ty)
-            ]
```
RyanGlScott, line 1333: I take it that this PR addresses part of https://github.com/GaloisInc/crucible/issues/1614? If so, it would be worth stating this in the commit message.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: It does include destructors.  In the `--print-mir` output, the first method of every vtable is `drop_in_place`.  mir-json always includes it at the start of the method list, and crucible-mir expects to find it there: https://github.com/GaloisInc/crucible/blob/2f4dfac61d1ac51f56a5b696bf3a48d72df510b6/crucible-mir/src/Mir/TransCustom.hs#L704

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Added a note to the PR description

crucible-mir/src/Mir/TransTy.hs
RyanGlScott: Ah, I suppose my confusion was about "method fields" vs. "non-method fields", but the Haddocks here do clarify that `vtableInfoTys` is specifically about non-method fields. Thanks for clarifying that!

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: I did this along with a more extensive refactor to deduplicate the logic for accessing a vtable slot, which with this PR had been duplicated in three places (`mkVirtCall`, `callOnceVirtShimDef`, and now `size_of_val`/`align_of_val`).

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1910,14 +1926,98 @@ traitVtableType col trait = vtableTy
...
+
+vtableInfo :: M.TraitName -> Int -> C.TypeRepr tp -> MirGenerator h s ret (VtableInfo tp)
+vtableInfo dynTraitName slotIdx slotTpr = do
+  Some vt@(VtableInfo ctx idx) <- vtableInfo' dynTraitName slotIdx
+  let tpr = ctx Ctx.! idx
+  Refl <- testEqualityOrFail tpr slotTpr $
+    dieMsg $ "expected slot to contain " ++ show slotTpr ++ ", but got " ++ show tpr
+  return vt
+  where
+    dieMsg s = "vtableInfo: trait" ++ show dynTraitName ++ ", slot " ++ show slotIdx ++ ": " ++ s
```
samcowger, line 1976:
```suggestion
    dieMsg s = "vtableInfo: trait " ++ show dynTraitName ++ ", slot " ++ show slotIdx ++ ": " ++ s
```

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1910,14 +1926,98 @@ traitVtableType col trait = vtableTy
...
+    Just x -> return x
+    Nothing -> die "slot index out of range for trait"
+
+  return $ Some (VtableInfo vtableTprs slotIdx')
+
+  where
+    die :: String -> MirGenerator h s ret a
+    die s = mirFail $ dieMsg s
+
+    dieMsg s = "vtableInfo': trait" ++ show dynTraitName ++ ", slot " ++ show slotIdx ++ ": " ++ s
```
samcowger, line 2002:
```suggestion
    dieMsg s = "vtableInfo': trait " ++ show dynTraitName ++ ", slot " ++ show slotIdx ++ ": " ++ s
```

crux-mir/README.md
RyanGlScott: Don't forget this!

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1309,7 +1308,7 @@ align_of_val = (["core", "intrinsics", "align_of_val"], \substs -> case substs o
             -- do here is make an effort to give a descriptive error message.
             -- TODO(#1614): Support trait objects and custom DSTs here.
             DynRefRepr -> case ty of
-                TyDynamic {} -> unsupportedTraitObject ty
+                TyDynamic dynTraitName -> getVtableUsize dynTraitName 0 e
```
samcowger:
```suggestion
                TyDynamic dynTraitName -> getVtableUsize dynTraitName 1 e
```
This probably accounts for the test failure we're seeing.

In this override and in `size_of_val`, you might consider giving this constant index a name, and/or commenting on its relationship to `TransTy.vtableInfoTys`

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1309,7 +1308,7 @@ align_of_val = (["core", "intrinsics", "align_of_val"], \substs -> case substs o
             -- do here is make an effort to give a descriptive error message.
             -- TODO(#1614): Support trait objects and custom DSTs here.
```
RyanGlScott: I think the documentation here needs to be updated, since `crucible-mir` does support trait objects now.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1301,17 +1301,26 @@ mkTraitObject traitName' vtableName e = do
...
 
+    col <- use $ cs . collection
+    vtable <- case col ^. vtables . at vtableName of
+        Just x -> return x
+        Nothing -> error $ "missing vtable definition for " ++ show vtableName
+
+    let info =
+            [ MirExp UsizeRepr $ R.App $ usizeLit $ fromIntegral $ vtable ^. vtSize
+            , MirExp UsizeRepr $ R.App $ usizeLit $ fromIntegral $ vtable ^. vtAlign
+            ]
```
samcowger: Consider documenting the relationship between this and `TransTy.vtableInfoTys`

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1234,7 +1240,7 @@ size_of_val = (["core", "intrinsics", "size_of_val"], \substs -> case substs of
...
```
RyanGlScott: I think the documentation here needs to be updated, since `crucible-mir` does support trait objects now.

RyanGlScott (approved):

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -2523,30 +2515,17 @@ callOnceVirtShimDef methodIdx = CustomMirOp $ \ops ->
...
       Some retTpr <- case vtableMethodTpr of
         C.FunctionHandleRepr _ retTpr -> return $ Some retTpr
         tpr -> mirFail $ "callOnceVirtShimDef: expected method " ++ show methodIdx
           ++ " of " ++ show dynTraitName ++ " to have FunctionHandleRepr, but got " ++ show tpr
 
       MirExp retTpr <$> doVirtCall
-        col
         dynTraitName
-        methodIdx
+        methodIdx   -- method index, not vtable slot index
```
samcowger, line 2526: If you wanted to be extra careful, you could consider `Int`-wrapping newtypes like `MethodIdx` and `SlotIdx` to avoid accidental conflation, though I think we'd probably be fine without them

samcowger (approved):

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Updated the align_of_val comment in c73adc71c604011d548b436a3d8ffbc37093e86e

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Updated the size_of_val comment in c73adc71c604011d548b436a3d8ffbc37093e86e

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Refactored in e3471eb92f0ef684c9c31f41dfb97595ad55cd70 to use named constants for the size/align slot indices

crucible-mir/src/Mir/Trans.hs
spernsteiner: Refactored in e3471eb92f0ef684c9c31f41dfb97595ad55cd70 to use a helper type with named fields instead of a specific list ordering

crux-mir/README.md
spernsteiner: Updated the changelog in 0f5937a17

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Yeah, seems okay to leave it as is for now.  Should be pretty obvious in the tests if we manage to mix it up.

PR #1818, opened 2026-04-28
Deprecate `BitSet` module

langston-barrett:
This module does not appear to be consumed by any downstream code, except the ancient `mss`:

https://github.com/search?q=owner%3AGaloisInc+%2FUtils%5C.BitSet%2F&type=code

It's also not really clear why it should be part of Crucible; the data-structure it provides is not fundamentally related to symbolic execution. We can save a bit of compile time and artifact sizes by removing it someday.


RyanGlScott (approved):
I can only fathom how this wound up in `crucible`. The commit which introduced it (https://github.com/GaloisInc/crucible/commit/ae8ec2a866cda7d564116b0e15eddafe0cbb3a05) doesn't even use `BitSet` anywhere.

PR #1822, opened 2026-04-29
Drop `generic-lens` dependency

### LINKED ISSUE #1821: Drop `generic-lens` dependency
It is only used in a few places for `field`/`setField`, but has ~two dozen modules between `generic-lens` and `generic-lens-core`. Furthermore it depends on `profunctors`, which is itself large and has many dependencies and might not otherwise be a transitive dep of our packages after #1820.

https://github.com/GaloisInc/crucible/blob/c032e7f298ff5ccb471b2837c65206ca0f9c4c79/crux/crux.cabal#L45

https://github.com/GaloisInc/crucible/blob/c032e7f298ff5ccb471b2837c65206ca0f9c4c79/crucible-concurrency/crucible-concurrency.cabal#L131

langston-barrett:
It was only used in a few places for `field`/`setField`, but has around two dozen modules between `generic-lens` and `generic-lens-core`. Furthermore it depends on `profunctors`, which is itself large and has many dependencies and might not otherwise be a transitive dep of our packages after we drop `lens` for `microlens`.

Fixes #1821.


crux/CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # next
 
+- **Breaking:** Drop the `Generic` instances from `ColorOptions`, `CruxOptions`,
+  `EarlyConfig`, and `OutputOptions`.
```
RyanGlScott, line 4: Is there a reason to drop the `Generic` instances outright? I can understand wanting to remove the generic lenses, but there could be use cases for these `Generic` instances besides defining lenses.

In any case, it would also be worth mentioning the newly defined top-level lens functions in the changelog.

crucible-concurrency/src/Cruces/ExploreCrux.hs
```diff
@@ -191,7 +190,7 @@ exploreOvr bak symOnline cruxOpts mainAct =
       Just (Crux.SomeOnlineSolver _) ->
         do ctx <- use stateContext
            todo <- liftIO $ getProofObligations bak
-           let cruxOpts' = over (field @"outputOptions") (setField @"quietMode" True . setField @"simVerbose" 0) cruxOpts
+           let cruxOpts' = set (outputOptionsL . quietModeL) True $ set (outputOptionsL . simVerboseL) 0 cruxOpts
```
RyanGlScott: Hm, I think I liked the previous way of writing this code better, as you only have to type `outputOptionsL`. (You have to replace `setField` with `set`, but that's easily done.)

RyanGlScott (approved):

crux/CHANGELOG.md
langston-barrett: There certainly could be in principle, but in practice I doubt that there really are. These are configurations for Crux itself - what would one be doing that one would need to program generically over them? I suspect they were introduced specifically for the sake of these lenses. They were certainly introduced in the same commit: https://github.com/GaloisInc/crucible/commit/04d7c0be9274a8ac904590c2a362985fc33edf87.

As for the motivation to remove them, they're just extra code that GHC has to compile. If there's no use case, then we can save on compile-times and artifact sizes by dropping them.

> In any case, it would also be worth mentioning the newly defined top-level lens functions in the changelog.

Will do.

RyanGlScott: Fair enough. I don't have a particular use case for these instances.

PR #1823, opened 2026-04-30
Drop `lens` in favor of `microlens`

### LINKED ISSUE #1820: Drop `lens` in favor of `microlens`
`lens` has many (transitive) dependencies. This leads to:

- Slow build times (the `microlens` README claims that it compiles in 3.5s whereas `lens` takes  4m10s)
- Longer GHC upgrade latency (we have to wait until all of `lens`'s dependencies support a new GHC version)
- Smaller GHC support windows (our GHC support window is limited by that of all of our transitive dependencies)
- Supply chain risks such as lack of maintenance and even supply chain attacks (see recent NPM attacks)

It's worth noting that we recently hit the GHC upgrade latency problem in elf-edit for 9.14 due to a dependency of `lens`; that concern is not theoretical.

Happily, `microlens` is often a drop-in replacement. I suggest that we drop `lens` and replace it with `microlens-*`.

langston-barrett:
Fixes #1820.


langston-barrett:
Disappointingly, this change only drops `lens` itself and `reflection` from our build plan. I had hoped for a more dramatic reduction involving transitive dependencies. Part of the reason we don't see this is a test-suite dependency on a beefy package in llvm-pretty-bc-parser (see https://github.com/GaloisInc/llvm-pretty-bc-parser/issues/363). I'll continue to investigate to see what else we might be able to drop entirely.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -2104,11 +2107,12 @@ initialValue (M.TyAdt nm _ _) = do
...
             Just <$> buildStruct' adt fldExps
         M.Enum _ -> do
-            case ifind (\_ vars -> vars ^. M.vinhabited) (adt ^. M.adtvariants) of
+            case findIndex (\vars -> vars ^. M.vinhabited) (adt ^. M.adtvariants) of
                 -- Uninhabited enums can't be initialized.
                 Nothing -> return Nothing
                 -- Inhabited enums get initialized to their first inhabited variant.
-                Just (discr, var) -> do
+                Just discr -> do
+                    let var = (adt ^. M.adtvariants) !! discr
```
langston-barrett: This is suboptimal. I can make a helper to contain the partiality, or we can leave crucible-mir with `lens`. I'd be fine with either, or just accepting this as-is.

crucible-concurrency/src/Crucibles/Execution.hs
```diff
@@ -150,7 +153,7 @@ spawnEvent tid exe = exe ^. birthdays.at tid
 
 -- | Record a thread's "birthday"
 addSpawnEvent :: ThreadID -> EventID -> Executions e -> Executions e
-addSpawnEvent tid eventid exe = exe & birthdays.at tid ?~ eventid
+addSpawnEvent tid eventid exe = exe & birthdays.at tid .~ Just eventid
```
RyanGlScott: Any reason not to use the old code? `Lens.Micro` offers `(?~)` so the old code should continue to work if you import `(?~)`.

crucible-concurrency/src/Crucibles/Execution.hs
```diff
@@ -122,22 +125,22 @@ maximalExecution e = IntSet.member (e ^. currentEventID) (e ^. maximalEvents)
 
 -- | Lens for extracting an event given its eventID. Will fail with an error if
 -- the eventID is not mapped.
-event :: HasCallStack => EventID -> Getter (Executions e) e
+event :: HasCallStack => EventID -> Getting r (Executions e) e
```
RyanGlScott: This is somewhat of a personal preference, but I think I would prefer writing `SimpleGetter (Executions e) e` instead of `Getting r (Executions e) e`. The former is equivalent to the latter, but hides the `r` type parameter (which is arguably something of an implementation detail).

Alternatively, we could add a dependency on `microlens-contra` and continue to use `Getting` as-is. That's one more dependency, but not a huge one (`microlens-contra` only depends on `base` and `microlens`).

(Similar comments apply to all other uses of `Getter`/`Getting` throughout this PR.)

crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
```diff
@@ -142,14 +143,10 @@ updateAlign :: Natural
...
-
-instance Ixed AlignInfo where
-  ix k = at k . traverse
-
-instance At AlignInfo where
-  at k f m = updateAlign k m <$> indexed f k (findExact k m)
+-- | Lens for accessing an alignment at a specific key in 'AlignInfo'.
+atAlignInfo :: Natural -> Lens' AlignInfo (Maybe Alignment)
+atAlignInfo k f m = updateAlign k m <$> f (findExact k m)
+{-# INLINE atAlignInfo #-}
```
RyanGlScott, line 149: I can see why you've removed these instances (in favor of a top-level `atAlignInfo` function), but this is nevertheless a breaking API change. Please make a note of it in the `crucible-llvm` changelog.

crucible-mir/src/Mir/Mir.hs
```diff
@@ -716,7 +719,6 @@ makeLenses ''MirBody
 makeLenses ''BasicBlock
 makeLenses ''BasicBlockData
 makeLenses ''Adt
-makePrisms ''AdtKind
```
RyanGlScott, line 719: Similarly, please make a note that the prisms/`Wrapped` instances have been removed in the `crucible-mir` changelog.

crucible-mir/src/Mir/TransTy.hs
RyanGlScott: How about just using [`ifind`](https://hackage.haskell.org/package/indexed-traversable-0.1.4/docs/Data-Foldable-WithIndex.html#v:ifind) from `indexed-traverable`? That's a package with a much smaller dependency footprint than `lens`, and we already have `indexed-traversable` in `crucible`'s dependency graph.

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
```diff
@@ -2098,8 +2101,7 @@ prog defuns =
                    return $ AnyCFG (CFG handle entry (e' : rest))
      gs <- use stxGlobals
      externs <- use stxExterns
-     fds <- uses stxForwardDecs $ fmap $
-              \(FunctionHeader _ _ _ handle _) -> SomeHandle handle
+     fds <- fmap (fmap (\(FunctionHeader _ _ _ handle _) -> SomeHandle handle)) (use stxForwardDecs)
```
RyanGlScott: Not your fault, but I'm a bit surprised that there isn't already a lens for retrieving the `_headerHandle` field of a `FunctionHeader`. That would allow simplifying this somewhat lengthy line a bit.

crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
```diff
@@ -293,7 +294,7 @@ filterCrucibleFrames (SomeFrame (MF f)) = Just (frameProgramLoc f)
 filterCrucibleFrames _ = Nothing
 
 -- | Iterate over frames in the result.
-arFrames :: Simple Traversal (AbortedResult sym ext) (SomeFrame (SimFrame sym ext))
+arFrames :: Traversal (AbortedResult sym ext) (AbortedResult sym ext) (SomeFrame (SimFrame sym ext)) (SomeFrame (SimFrame sym ext))
```
RyanGlScott: How about using `Traversal' (AbortedResult sym ext) (SomeFrame (SimFrame sym ext))` here instead of the lengthier `Traversal (AbortedResult sym ext) (AbortedResult sym ext) (SomeFrame (SimFrame sym ext)) (SomeFrame (SimFrame sym ext))`?

crucible-llvm/src/Lang/Crucible/LLVM/DataLayout.hs
langston-barrett: I don't think it is, because `AlignInfo` is not exported.

crucible-syntax/src/Lang/Crucible/Syntax/Concrete.hs
langston-barrett: I think it's because of the existential type. Added a getter for a `SomeHandle`.

crucible/src/Lang/Crucible/Simulator/ExecutionTree.hs
langston-barrett: Good catch, thanks!

crucible-mir/src/Mir/Mir.hs
langston-barrett: Done!

crucible-mir/src/Mir/TransTy.hs
langston-barrett: Perfect, thanks! `indexed-traversable` is a dep of p-u so that's definitely staying.

crucible-concurrency/src/Crucibles/Execution.hs
langston-barrett: Done!

RyanGlScott (approved):

Issue #1497, opened 2025-08-14
`crucible-mir`: Don't actually dereference pointers to ZSTs


qsctr:
Since ZSTs (zero-sized types) don't take up any space, the Rust standard library likes to skip allocations for ZSTs and instead just use an invalid pointer (something like `align_of::<T>() as *mut T`) to represent pointers to ZSTs. But then it will read from and write to the invalid pointer as usual. In Rust, this behavior is allowed: in the read case, it will return "the" ZST value, and in the write case, it will do nothing. So the address is never actually read from or written to. But in `crucible-mir` we still attempt to dereference the pointer even if it is a pointer to a ZST, so it will fail if the pointer is invalid. We should check if the pointee type is a ZST, and if it is then just return `initialValue` on reads and do nothing on writes.

RyanGlScott:
Is this related to https://github.com/GaloisInc/crucible/issues/1502, I wonder? I discovered https://github.com/GaloisInc/crucible/issues/1502 independently of this issue, but I wonder if the two issues have a similar underlying cause.

RyanGlScott:
Also, it is worth noting that this issue would be easier to fix after implementing #1499. This is because struct/tuple/array ZSTs would be zero-sized `MirAggregate`s, and reading/writing these through a pointer doesn't need to read/write any elements of the pointee, so the operation would succeed even if the pointer is invalid (as it does in concrete Rust).

spernsteiner:
As far as I know, the current state of this is:
* Deref of non-generic pointers to ZST (e.g. `*p = q` where `p: &mut ()`): rustc skips generating `Assign` statements for such accesses, so we don't need to do anything special here.
* Deref of generic pointers to ZST (e.g. `*p = q` where `p: &mut T` and the function is instantiated with `T = ()`): rustc generates the `Assign` and doesn't eliminate it after substitution.  crucible-mir has no special handling for this and will try to do the load/store as normal, which may break if the pointer points into a `MirAggregate`.  (We omit zero-sized fields from aggregates to ensure that every entry in the aggregate has a unique address.)
* `ptr` functions on pointers to ZST: #1824 modified the overrides for `ptr::read`/`ptr::write`/`ptr::copy` to be no-ops when the pointee type is a ZST.

So I think the one remaining bit of work required here is to modify crucible-mir to skip `Assign` statements with ZST types, in order to handle the generic pointer case.

RyanGlScott:
[...]

Issue #1519, opened 2025-09-03
`crux-mir`: Support `crux::test`s that return references


RyanGlScott:
It would be convenient to simulate `crux::test`s that return reference types such as `&u32`, `&[u32]`, or `&str`. All of these have `Display` impls, and it would be nice if `crux-mir` emulated the behavior of their `Display` impls when rendering their values post-simulation.

At the moment, attempting to return any sort of reference value in a `crux::test` will cause `crux-mir` to simply give up, e.g.,

```rs
// test.rs
#[crux::test]
pub fn crux_test() -> &'static str {
    "hello world"
}
```
```
$ cabal -v0 run exe:crux-mir -- test.rs
test test/166ef950::crux_test[0]: returned I don't know how to print result of type str, ok

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] All goals discharged through internal simplification.
[Crux] Overall status: Valid.
```

There is a case in the `showRegEntry` function [here](https://github.com/GaloisInc/crucible/blob/31d05de0e3f742922d353af5ba6844e7b8a0b86c/crux-mir/src/Mir/Language.hs#L721) that purports to handle references, but it doesn't go far enough. It matches on `TyRef`, but it doesn't go as far as matching on the underlying `MirReference` (or, in the case of slices, the `MirSlice`).

RyanGlScott:
Design question: what should happen if a user tries to display a null reference or pointer, such as in this example?

```rs
// test.rs
use std::ptr;

#[crux::test]
pub fn crux_test() -> *const u32 {
    ptr::null()
}
```

Two reasonable-sounding options are:

1. Make this a simulation error
2. Don't make this a simulation error, but instead display a default value (e.g., `<null>`)

I am somewhat leaning towards option (1), since it would be easier to implement, but I could be talked into option (2) as well.

RyanGlScott:
Upon further thought, I think we should go with option (2). There is already an established `crux-mir` convention of printing `<uninit>` when displaying uninitialized tuple entries (see [this code](https://github.com/GaloisInc/crucible/blob/d3e5dbe4a3550d80bbffd60fd38a5e54777f2d6e/crux-mir/src/Mir/Language.hs#L567)), so printing a default value would be more in line with existing conventions.

PR #1824, opened 2026-05-01
crux-mir: use MirAggregate for structs

spernsteiner:
Migrates structs from `StructRepr` to `MirAggregateRepr`.  This includes handling for all of the unsized / custom DST cases.

This includes a new `mirRef_agElem_unsized` intrinsic, which is like `mirRef_agElem` but dynamically detects the size and `TypeRepr` to use for the `AgElem_RefPath`.  This is used for accessing the last field of a custom DST, similar to the old `subfieldRef_untyped` intrinsic.  We need this because the normal `mirRef_agElem` intrinsic requires the field size to be known at translation time, which is not possible when working with custom DSTs.  Once `mirRef_agOffset` is implemented (as part of aggregate flattening), we can modify the custom DST cases to use that instead.

This requires corresponding saw-script changes, which I haven't started on yet, before it can be merged.

### DIFF
--- a/crucible-mir/src/Mir/Intrinsics/Reference.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Reference.hs
@@ -56,9 +56,7 @@ module Mir.Intrinsics.Reference
     dropMirRefLeaf,
     dropMirRefIO,
     subfieldMirRefLeaf,
-    subfieldMirRef_UntypedLeaf,
     subfieldMirRefIO,
-    subfieldMirRef_UntypedIO,
     subvariantMirRefLeaf,
     subvariantMirRefIO,
     subindexMirRefSim,
@@ -128,8 +126,6 @@ import Data.Parameterized.Context
     Index,
     adjustM,
     field,
-    intIndex,
-    size,
     (!),
     (::>),
     pattern Empty,
@@ -934,48 +930,6 @@ subfieldMirRefLeaf ctx ref idx =
     let tpr = ctx ! idx
     return $ MirReference tpr root (Field_RefPath ctx path idx)
 
--- | Mimic `subfieldMirRefLeaf`, but infer the appropriate `CtxRepr` and `Index`
--- at simulation time. If @expectedTy@ is provided, this will assert that it
--- matches the actual type of the field during simulation.
-subfieldMirRef_UntypedLeaf ::
-    MirReference sym ->
-    Int ->
-    Maybe (Some TypeRepr) ->
-    MuxLeafT sym IO (MirReference sym)
-subfieldMirRef_UntypedLeaf ref fieldNum expectedTy =
-  case ref of
-    MirReference_Integer _ ->
-      bail $ "attempted untyped subfield on the result of an integer-to-pointer cast"
-    MirReference structReprHopefully refRoot refPath ->
-      case structReprHopefully of
-        StructRepr structCtx ->
-          do
-            Some fieldIdx <-
-              case intIndex fieldNum (size structCtx) of
-                Just someIdx -> pure someIdx
-                Nothing ->
-                  bail $ unwords $
-                    [ "out-of-bounds field access:"
-                    , "field", show fieldNum, "of struct", show structCtx ]
-            let fieldRepr = structCtx ! fieldIdx
-            () <- case expectedTy of
-              Nothing -> pure ()
-              Just (Some expected) ->
-                case testEquality expected fieldRepr of
-                  Just Refl -> pure ()
-                  Nothing ->
-                    bail $ unwords $
-                      [ "expected field type", show expected
-                      , "did not match actual field type", show fieldRepr ]
-            let fieldPath = Field_RefPath structCtx refPath fieldIdx
-            pure $ MirReference fieldRepr refRoot fieldPath
-        notAStruct ->
-          bail $ unwords $
-            [ "untyped subfield requires a reference to a struct, but got a reference to"
-            , show notAStruct ]
-  where
-    bail msg = leafAbort $ GenericSimError $ msg
-
 subfieldMirRefIO ::
     IsSymBackend sym bak =>
     bak ->
@@ -986,17 +940,6 @@ subfieldMirRefIO ::
     IO (MirReferenceMux sym)
 subfieldMirRefIO bak iTypes ctx ref idx =
     modifyRefMuxMA bak iTypes (\ref' -> subfieldMirRefLeaf ctx ref' idx) ref
-
-subfieldMirRef_UntypedIO ::
-    IsSymBackend sym bak =>
-    bak ->
-    IntrinsicTypes sym ->
-    MirReferenceMux sym ->
-    Int ->
-    Maybe (Some TypeRepr) ->
-    IO (MirReferenceMux sym)
-subfieldMirRef_UntypedIO bak iTypes ref fieldNum expectedTy =
-    modifyRefMuxMA bak iTypes (\ref' -> subfieldMirRef_UntypedLeaf ref' fieldNum expectedTy) ref
 
 
 subvariantMirRefLeaf ::
--- a/crucible-mir/src/Mir/Intrinsics/Reference.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Reference.hs
@@ -1120,10 +1120,9 @@ mirRef_agElem_unsizedLeaf ::
     SymGlobalState sym ->
     IntrinsicTypes sym ->
     RegValue sym UsizeType ->
-    TypeRepr tp ->
     MirReference sym ->
     MuxLeafT sym IO (MirReference sym)
-mirRef_agElem_unsizedLeaf bak gs iTypes off tpr ref =
+mirRef_agElem_unsizedLeaf bak gs iTypes off ref =
   typedLeafOp "MirAggregate unsized element projection" MirAggregateRepr ref $ \root path -> do
     let ref' = MirReference MirAggregateRepr root path
     offConcrete <- case asBV off of
@@ -1131,11 +1130,10 @@ mirRef_agElem_unsizedLeaf bak gs iTypes off tpr ref =
       Nothing -> leafAbort $ GenericSimError $
         "mirRef_agElem_unsized: offset must be concrete, but got " ++ show (printSymExpr off)
     MirAggregate _ m <- readMirRefLeaf bak gs iTypes MirAggregateRepr ref'
-    MirAggregateEntry sz _ _ <- case IntMap.lookup offConcrete m of
-      Just entry -> return entry
-      Nothing -> leafAbort $ GenericSimError $
-        "mirRef_agElem_unsized: no entry at offset " ++ show offConcrete
-    return $ MirReference tpr root (AgElem_RefPath off sz tpr path)
+    (sz, Some entryTpr) <- case IntMap.lookup offConcrete m of
+      Just (MirAggregateEntry sz entryTpr _) -> return (sz, Some entryTpr)
+      Nothing -> return (0, Some MirAggregateRepr)
+    return $ MirReference entryTpr root (AgElem_RefPath off sz entryTpr path)
 
 mirRef_agElem_unsizedIO ::
     IsSymBackend sym bak =>
@@ -1143,11 +1141,10 @@ mirRef_agElem_unsizedIO ::
     SymGlobalState sym ->
     IntrinsicTypes sym ->
     RegValue sym UsizeType ->
-    TypeRepr tp ->
     MirReferenceMux sym ->
     IO (MirReferenceMux sym)
-mirRef_agElem_unsizedIO bak gs iTypes off tpr ref =
-    modifyRefMuxMA bak iTypes (mirRef_agElem_unsizedLeaf bak gs iTypes off tpr) ref
+mirRef_agElem_unsizedIO bak gs iTypes off ref =
+    modifyRefMuxMA bak iTypes (mirRef_agElem_unsizedLeaf bak gs iTypes off) ref
 
 
 refRootEq ::
--- a/crucible-mir/src/Mir/Intrinsics/Reference.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Reference.hs
@@ -1069,12 +1069,11 @@ mirRef_agElem_unsizedLeaf ::
     MuxLeafT sym IO (MirReference sym)
 mirRef_agElem_unsizedLeaf bak gs iTypes off ref =
   typedLeafOp "MirAggregate unsized element projection" MirAggregateRepr ref $ \root path -> do
-    let ref' = MirReference MirAggregateRepr root path
     offConcrete <- case asBV off of
       Just bv -> return $ fromIntegral $ BV.asUnsigned bv
       Nothing -> leafAbort $ GenericSimError $
         "mirRef_agElem_unsized: offset must be concrete, but got " ++ show (printSymExpr off)
-    MirAggregate _ m <- readMirRefLeaf bak gs iTypes MirAggregateRepr ref'
+    MirAggregate _ m <- readMirRefLeaf bak gs iTypes MirAggregateRepr ref
     (sz, Some entryTpr) <- case IntMap.lookup offConcrete m of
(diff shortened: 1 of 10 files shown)


crucible-mir/src/Mir/TransTy.hs
```diff
@@ -899,19 +884,14 @@ buildTupleM tupleTy xs = buildTupleMaybeM tupleTy (map Just xs)
...
-            -- Should be impossible; all struct-like types should have field info.
-            Nothing -> panic "buildTupleMaybeM"
-                ["missing field_offsets in layout for", show tupleTy]
-    when (length fieldTys /= length fieldOffsets || length fieldTys /= length xs) $
-        mirFail $ "buildTupleMaybeM: length mismatch:\n  fieldTys = " ++ show fieldTys
-            ++ "\n  fieldOffsets = " ++ show fieldOffsets ++ "\n  xs = " ++ show xs
+    offsetsAndTys <- tyFieldsM tupleTy
+    when (length offsetsAndTys /= length xs) $
+        mirFail $ "buildTupleMaybeM: length mismatch:\n  offsetsAndTys = " ++ show offsetsAndTys
+            ++ "\n  xs = " ++ show xs
```
RyanGlScott: For the sake of readability, it would be nicer to avoid manually interspersing newlines in this string. I think something like this would be easier to read:

```hs
        mirFail $ intercalate "\n"
          [ "buildTupleMaybeM: length mismatch:"
          , "  offsetsAndTys = " ++ show offsetsAndTys"
          , "  xs = " ++ show xs
          ]
```

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -636,17 +636,22 @@ ptr_read_impl what substs =
   case substs of
     Substs [ty] -> Just $ CustomOp $ \_ ops -> case ops of
         [MirExp MirReferenceRepr ptr] -> do
+            sz <- tySizeM ty
             Some tpr <- tyToReprM ty
-            MirExp tpr <$> readMirRef tpr ptr
+            case (sz, tpr) of
+                (0, MirAggregateRepr) -> MirExp tpr <$> mirAggregate_zst
+                _ -> MirExp tpr <$> readMirRef tpr ptr
```
RyanGlScott, line 645: Is there any relation between this change and https://github.com/GaloisInc/crucible/issues/1497?

crucible-mir/src/Mir/Intrinsics/Syntax.hs
```diff
@@ -213,6 +214,11 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
      !(TypeRepr tp) ->
      !(f MirReferenceType) ->
      MirStmt f MirReferenceType
+  MirRef_AgElem_Unsized ::
+     !(f UsizeType) ->
+     !(TypeRepr tp) ->
+     !(f MirReferenceType) ->
+     MirStmt f MirReferenceType
```
RyanGlScott, line 213: It might be worth some brief documentation saying how this differs from `MirRef_AgElem`.

crux-mir/src/Mir/Language.hs
```diff
@@ -674,6 +667,36 @@ showRegEntry col mty entry@(C.RegEntry tp rv) =
...
+        Just x -> return x
+        Nothing -> fail $ "missing fields for " ++ tyDesc
+      strs <- forM fields $ \(off, ty) -> do
+        case col ^? layouts . ix ty . _Just . laySize of
+          Just 0 -> showZSTValue ty
+          Just _ -> do
+            case IntMap.lookup (fromIntegral off) m of
+              Just (MirAggregateEntry _ elemTpr elemRvPart) ->
+                goMaybe ty elemTpr elemRvPart
+              Nothing -> return "<uninit>"
```
RyanGlScott, line 696: Is it an established convention that `crux-mir` should succeed if it tries to display an uninitialized location in the returned value? I wonder if it should instead fail here. (See also https://github.com/GaloisInc/crucible/issues/1519#issuecomment-3250264917.)

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1676,41 +1603,74 @@ fieldDataRef ::
...
+          Some elemTpr <- tyToReprM elemTy
+          elemSz <- tySizeM elemTy
+          ref'' <- mirRef_agElem_constOffset 0 elemSz elemTpr ref'
+          return $ MirPlace elemTpr ref'' meta
+        Nothing -> do
+          -- Output is a reference to a nested custom DST.  No additional
+          -- projection is needed.
+          return $ MirPlace MirAggregateRepr ref' meta
+
+    _ -> do
```
RyanGlScott, line 1672: Style nit pick: I'd prefer to explicitly match on `NoMeta` here instead of a wildcard pattern such that GHC will warn us to update this code if we ever add more cases to `PtrMetadata` in the future.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -885,6 +873,30 @@ construct MIR must be very careful to ensure that the type of that MIR is
 present.
 -}
 
+buildAggregateMaybeM :: M.Ty -> [Maybe (MirExp s)] -> MirGenerator h s ret (MirExp s)
```
samcowger, line 884: Perhaps document that these expressions' order must match the order in which their types appear in the given `agTy`?

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1203,7 +1183,13 @@ tyFields col ty = do
             M.TyTuple x -> x
             M.TyClosure x -> x
             M.TyCoroutineClosure x -> x
-            M.TyAdt {} -> panic "tyFields" ["TODO: tyFields struct case"]  -- For now, we only call this on tuples
+            M.TyAdt adtName _ _ ->
+                case col ^. M.adts . at adtName of
+                    Just adt ->
+                        case adt ^. M.adtkind of
+                            M.Struct -> map (\f -> f ^. M.fty) $ M.onlyVariant adt ^. M.vfields
+                            ak -> panic "tyFields" ["bad adt kind", show ak, "for type", show ty]
```
samcowger: Is it theoretically appropriate to call this on single-variant enums or single-field unions? If so, it would be nice to make a note of that here for future reference.

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1220,126 +1206,24 @@ tyFieldsM ty = do
...
+getStructField :: M.Ty -> Int -> MirExp s -> MirGenerator h s ret (MirExp s)
+getStructField structTy i (MirExp structTpr e0) = do
+  let tpr' = MirAggregateRepr
+  Refl <- testEqualityOrFail structTpr tpr' $ "bad representation " ++ show structTpr ++
+    " for struct type " ++ show structTy ++ ": expected " ++ show tpr'
+  fields <- tyFieldsM structTy
+  (off, ty) <- case fields ^? ix i of
+    Just x -> return x
+    Nothing -> mirFail $ "tupleFieldRef: field index " ++ show i ++
+      " is out of range for tuple " ++ show structTy
```
samcowger: This should refer to structs/`getStructField`

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1220,126 +1206,24 @@ tyFieldsM ty = do
...
+  -- Assume `NoMeta`.  Calling `getStructField` requires holding a struct by
+  -- value, which is only possible if the struct is sized.
+  Some valTpr <- tyToReprM ty
```
samcowger: I'm not sure of the purpose of this comment - is it meant to communicate that we know that `tyToReprM` won't fail on account of seeing e.g. `TyDynamic`?

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1754,41 +1586,74 @@ fieldDataRef ::
...
-      DynMeta _vtable ->
-        mirFail "expected slice metadata for slice field access, but found vtable"
-      SliceMeta _len ->
-        return $ MirPlace innerRepr elemRef meta
+structFieldRef structTy i ref meta = do
+  fields <- tyFieldsM structTy
+  (off, ty) <- case fields ^? ix i of
+    Just x -> return x
+    Nothing -> mirFail $ "tupleFieldRef: field index " ++ show i ++
+      " is out of range for tuple " ++ show structTy
```
samcowger: This should refer to structs/`structFieldRef`

crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1754,41 +1586,74 @@ fieldDataRef ::
...
+            M.TyAdt name _ _ -> name
+            _ -> panic "structFieldRef" ["expected struct type, but got", show structTy]
+      -- This is `Just elemTy` if the last field of `structTy` is a slice or
+      -- str, or `Nothing` if it's another struct DST.
+      optElemTy <- findLastFieldM adtName >>= \case
+        Just f | M.TySlice elemTy <- f ^. M.fty -> return $ Just elemTy
+        Just f | M.TyStr <- f ^. M.fty -> return $ Just $ M.TyUint M.B8
+        _ -> return Nothing
+
+      let offExp = R.App $ usizeLit $ fromIntegral off
```
samcowger, line 1650: Do we need to pad this offset out to the alignment of the slice element type? Or does `off` already include any necessary padding?

crucible-mir/src/Mir/Intrinsics/Reference.hs
```diff
@@ -1109,6 +1056,38 @@ mirRef_agElemIO ::
...
+    IsSymBackend sym bak =>
+    bak ->
+    SymGlobalState sym ->
+    IntrinsicTypes sym ->
+    RegValue sym UsizeType ->
+    MirReference sym ->
+    MuxLeafT sym IO (MirReference sym)
+mirRef_agElem_unsizedLeaf bak gs iTypes off ref =
+  typedLeafOp "MirAggregate unsized element projection" MirAggregateRepr ref $ \root path -> do
+    let ref' = MirReference MirAggregateRepr root path
```
samcowger: Isn't this just a redefinition of `ref`?

crucible-mir/src/Mir/Intrinsics/Reference.hs
```diff
@@ -1109,6 +1056,38 @@ mirRef_agElemIO ::
...
+    offConcrete <- case asBV off of
+      Just bv -> return $ fromIntegral $ BV.asUnsigned bv
+      Nothing -> leafAbort $ GenericSimError $
+        "mirRef_agElem_unsized: offset must be concrete, but got " ++ show (printSymExpr off)
+    MirAggregate _ m <- readMirRefLeaf bak gs iTypes MirAggregateRepr ref'
```
samcowger: IIRC, we were hesitant to add a reference path projection that actually read from its operand, since we prefer for them to be relatively simple/infallible operations. Have we determined that this is inevitable in general, or temporarily inevitable in order to implement `agElem_unsized` until it's replaced with an `agOffset`-style intrinsic?

Also, do I understand correctly that we should be able to replace all uses of `agElem_unsized` with `agOffset`, once it and proper flattening are implemented?

crux-mir/src/Mir/Language.hs
spernsteiner: Currently we print `<uninit>` for uninitialized tuple entries as well, and we previously did something similar for tuples (but not for struct) prior to the addition of `MirAggregate`.  This output is mainly for testing / debugging, and IMO it's a bit user-friendlier to handle uninitialized fields this way.

RyanGlScott: I missed that this was already an established convention---I take back my suggestion.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: The last case covers not only `NoMeta`, but also the case where the struct pointer has `SliceMeta`/`DynMeta` but we are accessing a non-last field.

crucible-mir/src/Mir/Intrinsics/Reference.hs
spernsteiner: Right—this is a temporary measure, used only in one specific place (last-field projections on custom DSTs), and we can get rid of it once `AgOffset_RefPath` is implemented.

We previously used `subfieldRef_untyped` for certain last-field projections, but with the move to `MirAggregateRepr` for structs, there's no longer enough type information in the `MirReference` to use that strategy, so this new hack has to read from the reference to get the information it needs.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Switched to use `intercalate` in 17be57d91

crucible-mir/src/Mir/Intrinsics/Syntax.hs
spernsteiner: Added a doc comment in 17be57d91

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Added a comment in 17be57d91 that mentions the order

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Added a comment about the enum/union cases in 17be57d91

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Fixed the message in 17be57d91

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Clarified the comment in 17be57d91.  This code is very similar to the `NoMeta` case of `structFieldRef`; this comment was meant to explain why we don't need to cover the other `structFieldRef` cases too.

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Fixed the message in 17be57d91

crucible-mir/src/Mir/TransTy.hs
spernsteiner: Added a comment in 17be57d91.  `off` indeed includes the padding already in the slice case.

crucible-mir/src/Mir/Intrinsics/Reference.hs
spernsteiner: Removed in 17be57d91.  For some reason I was thinking `ref'` would have a more refined type than `ref`, but `MirReference` actually has no type arguments other than `sym`.

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Yes, this makes derefs through `ptr::read`/`ptr::write` be no-ops on ZSTs, just as (IIRC) normal `*p` dereferences already are.  I was seeing errors on some of the tests due to `MirAggregate` containing zero-sized entries (which conflict with non-zero-sized entries at the same offset), and it turned out to be `ptr::write` storing ZST marker types into the aggregate.

RyanGlScott: Great! If this fixes #1497, that would be great to note in a commit message.

spernsteiner:
Rebased and added one new commit: c6afa9e38f3e4e218db68792b5bae4b57e8c16b1.  This fixes a strange test failure on the corresponding saw-script branch (https://github.com/GaloisInc/saw-script/pull/3228): the `crux_spec_for_array` started failing, even though it doesn't use structs and should have been unaffected by these changes.  The immediate cause turned out to be an error about muxing together `MethodSpecBuilder`s while running `f_equiv_spec`.  This new commit (and its counterpart in saw-script) allows muxing to succeed as long as the two input `MethodSpecBuilder`s are the same, as indicated by them having the same nonce.  This is the same strategy we use to allow muxing together `MethodSpec`s.

RyanGlScott (approved):

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: I think there's still one case remaining for #1497—I commented there with details.

PR #1838, opened 2026-05-26
crucible-llvm: GEP overflow check respects laxArith option

AmeliaRose802:
When the user enables lax loads/stores (e.g. via SAW's `enable_lax_loads_and_stores`), arithmetic overflow checks during GEP index computation should be suppressed for consistency with the rest of the lax-arithmetic path. Currently the GEP path always emits an overflow assertion regardless of the option.

This patch threads the `laxArith` setting through `Translation/Instruction.hs` so the GEP overflow assertion is omitted when lax arithmetic is enabled.

1 file changed, +7/-5. Independent single-commit PR; no dependency on other patches in the series.

### DIFF
--- a/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
+++ b/crucible-llvm/src/Lang/Crucible/LLVM/Translation/Instruction.hs
@@ -443,7 +443,7 @@ insertValue _ _ _ = fail "invalid insertValue instruction"
 
 
 evalGEP :: forall s arch ret wptr.
-  wptr ~ ArchWidth arch =>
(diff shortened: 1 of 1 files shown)


RyanGlScott:
Thanks for the PR, @AmeliaRose802! This looks like the right fix, but please make sure to include a test case in the `crux-llvm` test suite that ensures that the issue remains fixed going forward.

It is possible that there are other checks that `crucible-llvm`'s semantics for `getelementptr` performs which would need to be suppressed if `enable_lax_loads_and_stores` is enabled, but I haven't thoroughly audited the code to see if this is the case. What you have in this PR seems like a good start.

PR #1841, opened 2026-05-29
crucible-mir: remove allocation-agnostic indexing

samcowger:
Remove the `MirSubindexRef` MIR intrinsic, and replace its uses with both new (`MirRef_ArrayIndex`, `MirRef_VecIndex`) and existing (`MirRef_AgElem`) allocation-specific indexing forms. This substitution ought to preserve existing indexing functionality.

This is possible because all current indexing occurs in places where we know whether the backing allocation is a `MirAggregate`, an `Array`, or a `Vector`, and it's desirable as a precursor to aggregate-flattening (#1499), because it avoids a potential issue with `Vector`s and/or `Array`s appearing in flattened aggregates. (In short, in a world where all aggregates are flattened and we have a reference to e.g. a `Vector` field within an aggregate, and want to access an element of the `Vector`, it turns out to be difficult to know that we need to extend that reference with the special `VectorIndex_RefPath` variant, rather than simply offsetting further forward in the aggregate on the assumption that the `Vector`'s elements have been flattened into the aggregate.)

This will require some small changes to SAW, which uses the indexing intrinsic this removes in a few places, but all such uses can be replaced with allocation-specific alternatives. I'll open a corresponding PR for that shortly.

### DIFF
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -29,6 +29,10 @@ This release supports [version
   `Mir.Intrinsics`. These are only used by SAW at the moment.
 * Remove `Mir.FancyMuxTree.runMuxLeafIO'`, which does not seem to be used
   anywhere.
+* Remove allocation-agnostic indexing intrinsic (`Mir.Generator.subindexRef`,
+  `Mir.Intrinsics.Reference.subindexMirRef{IO,Leaf,Sim}`), and migrate existing
+  uses to allocation-specific intrinsics for aggregates, sybmolic arrays, and
+  vectors (`Mir.Generator.{mirRef_agElem,mirRef_arrayIndex,mirRef_vecIndex}`).
 
 # 0.6 -- 2026-01-29
 
--- a/crucible-mir/src/Mir/Generator.hs
+++ b/crucible-mir/src/Mir/Generator.hs
@@ -629,15 +629,6 @@ subvariantRef ::
   MirGenerator h s ret (R.Expr MIR s MirReferenceType)
 subvariantRef tp ctx ref idx = G.extensionStmt (MirSubvariantRef tp ctx ref idx)
 
-subindexRef ::
-  C.TypeRepr tp ->
-  R.Expr MIR s MirReferenceType ->
-  R.Expr MIR s UsizeType ->
-  -- | Size of the element, in bytes
-  Word ->
-  MirGenerator h s ret (R.Expr MIR s MirReferenceType)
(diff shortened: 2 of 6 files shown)


spernsteiner:
> all current indexing occurs in places where we know whether the backing allocation is a `MirAggregate`, an `Array`, or a `Vector`

Worth noting that "indexing" here refers to Crucible-level indexing, not Rust-level index expressions.  In particular, for slices, slice construction like `&my_array as &[_]` does a Crucible indexing operation (to project into the first element of the array), and slice indexing like `my_slice[3]` does a Crucible *offset* operation (not indexing).  And offset operations are still allocation agnostic AIUI.

spernsteiner (approved):

RyanGlScott (approved):
One thing that dawned on me while reading this PR: we have pretty inconsistent naming conventions for operations that perform MIR reference offsets. We have things like `mirRef_offset*`, `mirRef_arrayIndex*`, and `mirRef_vecIndex*`, but at the same time, we also have `subfieldMirRef*`, `subjustMirRef*`, and `subvariantMirRef*`. It's not clear to me if one naming convention is better than the other...

In any case, this is a distinction that existed well before this PR, and I don't propose that we try to disentangle this here. I'm content with using the names that you've chosen.

samcowger:
> One thing that dawned on me while reading this PR: we have pretty inconsistent naming conventions for operations that perform MIR reference offsets. We have things like `mirRef_offset*`, `mirRef_arrayIndex*`, and `mirRef_vecIndex*`, but at the same time, we also have `subfieldMirRef*`, `subjustMirRef*`, and `subvariantMirRef*`. It's not clear to me if one naming convention is better than the other...

I thoroughly agree. I put all of these operations into separate modules in https://github.com/GaloisInc/crucible/pull/1784 partially for repo hygiene, but an ulterior motive was to make it much easier to ditch these disparate naming conventions altogether in favor of simple import-qualification, as in:
```hs
import Mir.Intrinsics.Reference qualified as MirRef

foo = MirRef.arrayIndex
```

spernsteiner:
> We have things like `mirRef_offset*`, `mirRef_arrayIndex*`, and `mirRef_vecIndex*`, but at the same time, we also have `subfieldMirRef*`, `subjustMirRef*`, and `subvariantMirRef*`. It's not clear to me if one naming convention is better than the other...

`subfield`, `subjust`, etc, was the original naming convention, but eventually we started adding things like "array as `MirVector`" that don't really fit into that convention, and stuck with the newer `mirRef_foo` naming for newer operations as well.

Issue #1485, opened 2025-08-01
`crucible-mir-syntax`: Concrete syntax for remaining MIR syntax formers


RyanGlScott:
#1484 introduced support for parsing syntax that corresponds to some of the constructors of [`MirStmt`](https://github.com/GaloisInc/crucible/blob/9cb704a783676d900c8c08b06ecda68b4c6fcf30/crucible-mir/src/Mir/Intrinsics.hs#L836-L988). We should extend this parser with syntax for all of these constructors:

- [ ] `MirIntegerToRef`
- [ ] `MirGlobalRef`
- [ ] `MirConstRef`
- [ ] `MirSubfieldRef`
- [ ] `MirSubfieldRef_Untyped`
- [ ] `MirSubvariantRef`
- [ ] `MirSubindexRef`
- [ ] `MirSubjustRef`
- [ ] `MirRef_VectorAsMirVector`
- [ ] `MirRef_ArrayAsMirVector`
- [ ] `MirRef_Eq`
- [ ] `MirRef_Offset`
- [ ] `MirRef_OffsetWrap`
- [ ] `MirRef_TryOffsetFrom`
- [ ] `MirRef_PeelIndex`
- [ ] `VectorSnoc`
- [ ] `VectorHead`
- [ ] `VectorTail`
- [ ] `VectorInit`
- [ ] `VectorLast`
- [ ] `VectorConcat`
- [ ] `VectorTake`
- [ ] `VectorDrop`
- [ ] `ArrayZeroed`
- [ ] `MirVector_Uninit`
- [ ] `MirVector_FromVector`
- [ ] `MirVector_FromArray`
- [ ] `MirVector_Resize`

Issue #1384, opened 2025-04-21
`crucible-mir`: `usize`/`isize` are always hard-coded to 32 bits


RyanGlScott:
Rust's [`usize`](https://doc.rust-lang.org/nightly/std/primitive.usize.html) and [`isize`](https://doc.rust-lang.org/nightly/std/primitive.isize.html) types are intended to be pointer sized. `crucible-mir`, on the other hand, currently hard-codes `usize` and `isize` to always be 32 bits in size:

https://github.com/GaloisInc/crucible/blob/5fc34c708443c69533f2db90c540c7136c2b86cf/crucible-mir/src/Mir/Intrinsics.hs#L147-L154

This is unfortunate, as it means that there are valid programs involving `usize`/`isize` on 64-bit machines that cannot be reasoned about using `crucible-mir`. For instance, MIRI would have no problems with the following program on a 64-bit machine:

```rs
fn add1(x: usize) -> usize {
    x + 1
}

#[crux::test]
fn f() -> usize {
    let x = u32::MAX as usize;
    add1(x)
}
```

This is because `u32::MAX` would be well below the maximum value of a 64-bit `usize`, so `add1(x)` would overflow. On the other hand, `crux-mir` will fail when it simulates this program:

```
$ crux-mir test.rs
test test/150b1418::f[0]: returned 0, [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/150b1418::f[0] counterexamples ----
[Crux] Found counterexample for verification goal
[Crux]   test.rs:2:5: 2:10: error: in test/150b1418::add1[0]
[Crux]   attempt to compute `move _2 + const 1_usize`, which would overflow

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Invalid.
```

This is because even on 32-bit machines, `crucible-mir` thinks `usize` values are 32 bits in size, so `add1(x)` will trigger a checked overflow assertion failure. (Note also that `crucible-mir` claims that `f` returns `0`.)

[...]

PR #1842, opened 2026-06-02
`crucible-mir`: flattened `MirAggregate`s

samcowger:
This introduces an invariant to `MirAggregate`s that they are "flat," i.e. that they never contain immediate `MirAggregateEntry`s that are themselves `MirAggregate`s. This implements most of the remainder of #1499, though it does not wrap local variables in `MirAggregate`s, as that issue dictates.

This is in draft form right now because it's based on #1841, and because I  don't yet have a patch for `saw-script` to accommodate these changes. I've also held off on writing up these changes in `{crucible,crux}-mir`'s CHANGELOG, in case they need to be massaged when they get ported to `saw-script`. However, I believe it's otherwise ready for review.

### DIFF
--- a/crucible-mir/src/Mir/FancyMuxTree.hs
+++ b/crucible-mir/src/Mir/FancyMuxTree.hs
@@ -262,11 +262,16 @@ leafUpdatePartExpr bak mux x (PE q y) = do
             return $ mkPE pq xy
 
 -- | Like `leafUpdatePartExpr`, but instead of unconditionally writing a new
--- value, modify an already-existing value if one exists.
+-- value, modify an already-existing value, conditional on the current leaf
+-- being active.
 leafModifyPartExpr ::
     (IsSymBackend sym bak, MonadIO m) =>
     bak ->
     (Pred sym -> a -> a -> m a) ->
+    -- | If some value @v@ inhabits the leaf and its leaf predicate is true,
+    -- this will receive @Just v@. If no value inhabits the leaf, this will
+    -- receive @Nothing@. If the leaf predicate is known to be false, this will
+    -- not be called.
     (Maybe a -> m a) ->
     PartExpr (Pred sym) a ->
     MuxLeafT sym m (PartExpr (Pred sym) a)
--- a/crucible-mir/src/Mir/FancyMuxTree.hs
+++ b/crucible-mir/src/Mir/FancyMuxTree.hs
@@ -403,7 +403,7 @@ zipFancyMuxTrees' :: MonadAssert sym bak m =>
     FancyMuxTree sym a -> FancyMuxTree sym b -> m c
 zipFancyMuxTrees' bak f mux tx ty = zipFancyMuxTrees bak f mux tx ty >>= \my -> case my of
     Just y -> return y
-    Nothing -> liftIO $ addFailedAssertion bak $ GenericSimError $
+    Nothing -> maFail bak $ GenericSimError $
         "attempted to read empty mux tree"
 
 mergeFancyMuxTree :: (IsExprBuilder sym, OrdSkel a, MonadIO m) =>
--- a/crucible-mir/src/Mir/FancyMuxTree.hs
+++ b/crucible-mir/src/Mir/FancyMuxTree.hs
@@ -261,6 +261,30 @@ leafUpdatePartExpr bak mux x (PE q y) = do
             xy <- lift $ mux p x y
             return $ mkPE pq xy
 
+-- | Like `leafUpdatePartExpr`, but instead of unconditionally writing a new
+-- value, modify an already-existing value if one exists.
+leafModifyPartExpr ::
+    (IsSymBackend sym bak, MonadIO m) =>
+    bak ->
+    (Pred sym -> a -> a -> m a) ->
+    (Maybe a -> m a) ->
+    PartExpr (Pred sym) a ->
+    MuxLeafT sym m (PartExpr (Pred sym) a)
+leafModifyPartExpr _bak _mux modify Unassigned =
+    mkPE <$> leafPredicate <*> lift (modify Nothing)
+leafModifyPartExpr bak mux modify (PE oldPred oldVal) = do
+    let sym = backendGetSym bak
+    currPred <- leafPredicate
+    case asConstantPred currPred of
+        Just True -> mkPE (truePred sym) <$> lift (modify (Just oldVal))
+        Just False -> return $ PE oldPred oldVal
+        Nothing -> do
+            newPred <- liftIO $ orPred sym currPred oldPred
+            muxed <- lift $ do
+                newVal <- modify (Just oldVal)
+                mux currPred newVal oldVal
+            return $ mkPE newPred muxed
+
 -- | Set a PartExpr to Unassigned, conditional on the current leaf being
 -- active.
 leafClearPartExpr :: (IsSymBackend sym bak, MonadIO m) =>
--- a/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
@@ -68,12 +68,13 @@ import GHC.TypeLits
     TypeError,
   )
 
-import Control.Monad (foldM, forM, forM_, liftM, when)
+import Control.Monad (foldM, forM, forM_, liftM, unless, when)
 import Control.Monad.Except (MonadError (..), runExceptT, throwError)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Trans (lift)
 
 import Data.BitVector.Sized qualified as BV
+import Data.Either (fromRight)
 import Data.IntMap (IntMap)
 import Data.IntMap qualified as IntMap
 import Data.Kind (Type)
@@ -104,7 +105,7 @@ import Lang.Crucible.Types
   )
 
 import What4.Concrete (fromConcreteBV)
-import What4.Interface (IsExprBuilder (..), Pred, SymExpr, asAffineVar, asBV)
+import What4.Interface (IsExprBuilder (..), Pred, SymExpr, asAffineVar, asBV, asConstantPred)
 import What4.Partial (PartExpr, justPartExpr, mergePartial, pattern PE, pattern Unassigned)
 
 import Mir.FancyMuxTree
@@ -116,7 +117,7 @@ import Mir.FancyMuxTree
     subMuxLeafMA,
   )
 import Mir.Intrinsics.Size (UsizeType, wordLit)
-import Mir.Mir (OpSize)
+import Mir.Mir (OpSize (..))
 
 --------------------------------------------------------------
 -- A MirAggregateType is a collection of elements of any type, with each entry
@@ -293,6 +294,8 @@ readMirAggregateWithSymOffset ::
   MirAggregate sym ->
   MuxLeafT sym m (RegValue sym tp)
 readMirAggregateWithSymOffset bak iteFn off readSize tpr ag@(MirAggregate totalSize m)
+  | MirAggregateRepr <- tpr =
+      readSubaggregateWithSymOffset bak iteFn readSize off ag
   | Just (fromIntegral . BV.asUnsigned -> off') <- asBV off = do
       case IntMap.lookup off' m of
         Nothing -> leafAbort $ agNoValueAtOffsetSimError off' totalSize
@@ -333,6 +336,39 @@ readMirAggregateWithSymOffset bak iteFn off readSize tpr ag@(MirAggregate totalS
     offsetLit = wordLit sym
     iteFn' = liftIteFnMaybe sym tpr iteFn
 
+readSubaggregateWithSymOffset ::
+  forall sym bak m.
+  MonadAssert sym bak m =>
+  bak ->
+  (Pred sym -> MirAggregate sym -> MirAggregate sym -> IO (MirAggregate sym)) ->
+  OpSize ->
+  RegValue sym UsizeType ->
+  MirAggregate sym ->
+  MuxLeafT sym m (MirAggregate sym)
+readSubaggregateWithSymOffset bak iteFn readSize off ag@(MirAggregate sz _)
+  | Just (fromIntegral . BV.asUnsigned -> off') <- asBV off = do
+      let err e = leafAbort $ GenericSimError $ "readSubaggregateWithSymOffset: " <> e
+      either err pure (readConcrete off' ag)
+  | otherwise = do
+      liftIO $ foldM
+        (\ag' candidateOff -> do
+          isTheOffset <- bvEq sym off =<< offsetLit candidateOff
+          iteFn isTheOffset (fromRight ag' $ readConcrete candidateOff ag') ag')
+        ag
+        (init [0 .. sz])
+  where
+    snd3 (_, x, _) = x
+    sym = backendGetSym bak
+    offsetLit = wordLit sym
+
+    readConcrete off' ag' = do
+      case readSize of
+        All ->
+          snd <$> mirAggregate_split off' ag'
+        Width width ->
+          snd3 <$> mirAggregate_split3 off' (off' + width) ag'
+
+
 adjustMirAggregateWithSymOffset ::
   forall sym bak tp.
   (IsSymBackend sym bak) =>
@@ -344,6 +380,8 @@ adjustMirAggregateWithSymOffset ::
   MirAggregate sym ->
   MuxLeafT sym IO (MirAggregate sym)
 adjustMirAggregateWithSymOffset bak iteFn off tpr f ag@(MirAggregate totalSize m)
+  | MirAggregateRepr <- tpr =
+      adjustSubaggregateWithSymOffset bak iteFn off f ag
   | Just (fromIntegral . BV.asUnsigned -> off') <- asBV off = do
       MirAggregateEntry sz tpr' rvPart <- case IntMap.lookup off' m of
         Just x -> return x
@@ -398,6 +436,46 @@ adjustMirAggregateWithSymOffset bak iteFn off tpr f ag@(MirAggregate totalSize m
     offsetLit = wordLit sym
     iteFn' = liftIteFnMaybe sym tpr iteFn
 
+adjustSubaggregateWithSymOffset ::
+  forall sym bak.
+  (IsSymBackend sym bak) =>
+  bak ->
+  (Pred sym -> MirAggregate sym -> MirAggregate sym -> IO (MirAggregate sym)) ->
+  RegValue sym UsizeType ->
+  (MirAggregate sym -> MuxLeafT sym IO (MirAggregate sym)) ->
+  MirAggregate sym ->
+  MuxLeafT sym IO (MirAggregate sym)
+adjustSubaggregateWithSymOffset bak iteFn off f ag@(MirAggregate sz _)
+  | Just (fromIntegral . BV.asUnsigned -> off') <- asBV off =
+      adjustConcrete off'
+  | otherwise = do
+      foldM
+        (\ag' candidateOff -> do
+          isTheOffset <- liftIO $ bvEq sym off =<< offsetLit candidateOff
+          subAg <- adjustConcrete candidateOff
+          liftIO $ iteFn isTheOffset subAg ag')
+        ag
+        (init [0 .. sz])
+  where
+    adjustConcrete off' = do
+      case mirAggregate_split off' ag of
+        Right (MirAggregate leftSz leftAg, r@(MirAggregate rightSz _rightAg)) -> do
+          MirAggregate rightSz' rightAg' <- f r
+          let rightAg'' = IntMap.mapKeysMonotonic (\o -> o + fromIntegral off') rightAg'
+          when (rightSz /= rightSz') $
+            panic "adjustSubaggregateWithSymOffset"
+              [ "adjusting function changed aggregate size"
+              , "was " <> show rightSz <> ", became " <> show rightSz'
+              ]
+          unless (IntMap.disjoint leftAg rightAg'') $
+            panic "adjustSubaggregateWithSymOffset"
+              [ "aggregates weren't disjoint" ]
+          pure $ MirAggregate (leftSz + rightSz) (rightAg'' <> leftAg)
+        Left err -> leafAbort $ GenericSimError $
+          "adjustSubaggregateWithSymOffset: " <> err
+
+    sym = backendGetSym bak
+    offsetLit = wordLit sym
 
 -- | Given a list of valid entry spans @(fromOffset, toOffset)@ in this
 -- aggregate, create a predicate that the provided offset appears among their
@@ -565,11 +643,39 @@ mirAggregate_entries :: sym -> MirAggregate sym -> [(Word, MirAggregateEntry sym
 mirAggregate_entries _sym (MirAggregate _totalSize m) =
   [(fromIntegral off, entry) | (off, entry) <- IntMap.toList m]
 
+foldMWithKey :: Monad m => (b -> Int -> a -> m b) -> b -> IntMap a -> m b
+foldMWithKey f z m = IntMap.foldlWithKey' f' (pure z) m
+  where
+    f' accM key val = do
+      acc <- accM
+      f acc key val
+
 mirAggregate_insert ::
+  (IsSymInterface sym) =>
   Word ->
   MirAggregateEntry sym ->
   MirAggregate sym ->
   Either String (MirAggregate sym)
+mirAggregate_insert outerAgOff (MirAggregateEntry entrySz MirAggregateRepr entryP) outerAg = case entryP of
+  Unassigned -> pure outerAg
+  PE p (MirAggregate innerAgSz innerAgEntries) -> case asConstantPred p of
+    _ | innerAgSz /= entrySz -> Left $
+      "mirAggregate_insert: when adding a `MirAggregate` as an entry, " <>
+      "the `MirAggregateEntry`-derived size was " <> show entrySz <> ", " <>
+      "but the `MirAggregate` value had size " <> show innerAgSz
+    Just False ->
+      pure outerAg
+    Just True -> do
+      let addEntry ag (fromIntegral -> entryOff) entry =
+            mirAggregate_insert (outerAgOff + entryOff) entry ag
+      foldMWithKey addEntry outerAg innerAgEntries
+    -- Note: this case doesn't seem to show up in practice, but if it does,
+    -- consider lifting this function to a monadic context in which it can
+    -- manipulate predicates, so that it can add entries the conjunction of
+    -- their own individual predicates and the overarching entry's predicate,
+    -- `entryP`.
+    Nothing -> Left $
+      "mirAggregate_insert: unsupported: `MirAggregate` as entry with symbolic predicate"
 mirAggregate_insert off entry@(MirAggregateEntry sz tpr _) (MirAggregate totalSize m) = do
   -- For now, we require that either there are no entries overlapping the byte
   -- range `off .. off + sz`, or there is an entry covering exactly that range
@@ -943,6 +1049,10 @@ mirAggregate_getIO ::
     TypeRepr tp ->
     MirAggregate sym ->
     IO (RegValue sym tp)
+mirAggregate_getIO bak off sz MirAggregateRepr ag =
+  case mirAggregate_split3 off (off + sz) ag of
+    Right (_toLeft, mid, _toRight) -> pure mid
+    Left err -> addFailedAssertion bak $ GenericSimError $ "mirAggregate_getIO: " <> err
 mirAggregate_getIO bak off sz tpr (MirAggregate _ m) = do
   -- TODO: deduplicate logic between this and readMirAg* concrete case?
   MirAggregateEntry sz' tpr' rvPart <- case IntMap.lookup (fromIntegral off) m of
--- a/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
@@ -116,6 +116,7 @@ import Mir.FancyMuxTree
     subMuxLeafMA,
   )
 import Mir.Intrinsics.Size (UsizeType, wordLit)
+import Mir.Mir (OpSize)
 
 --------------------------------------------------------------
 -- A MirAggregateType is a collection of elements of any type, with each entry
@@ -287,10 +288,11 @@ readMirAggregateWithSymOffset ::
   bak ->
   (Pred sym -> RegValue sym tp -> RegValue sym tp -> IO (RegValue sym tp)) ->
   RegValue sym UsizeType ->
+  OpSize ->
   TypeRepr tp ->
   MirAggregate sym ->
   MuxLeafT sym m (RegValue sym tp)
-readMirAggregateWithSymOffset bak iteFn off tpr ag@(MirAggregate totalSize m)
+readMirAggregateWithSymOffset bak iteFn off readSize tpr ag@(MirAggregate totalSize m)
   | Just (fromIntegral . BV.asUnsigned -> off') <- asBV off = do
       case IntMap.lookup off' m of
         Nothing -> leafAbort $ agNoValueAtOffsetSimError off' totalSize
--- a/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
+++ b/crucible-mir/src/Mir/Intrinsics/Aggregate.hs
@@ -48,9 +48,7 @@ module Mir.Intrinsics.Aggregate
     mirAggregate_split,
     mirAggregate_split3,
     mirAggregate_chunk,
-    mirAggregate_toChunks,
     mirAggregate_concat,
-    mirAggregate_fromChunks,
     concreteAllocSize,
     mirAggregate_uninitIO,
     mirAggregate_zstSim,
@@ -68,8 +66,7 @@ import GHC.TypeLits
     TypeError,
   )
 
-import Control.Monad (foldM, forM, forM_, liftM, unless, when)
-import Control.Monad.Except (MonadError (..), runExceptT, throwError)
+import Control.Monad (foldM, forM, unless, when)
 import Control.Monad.IO.Class (MonadIO, liftIO)
 import Control.Monad.Trans (lift)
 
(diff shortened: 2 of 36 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -1531,6 +1534,10 @@ evalPlaceProj ty (MirPlace tpr ref NoMeta) M.Deref = do
         CTyBox t -> doRef t
         _ -> mirFail $ "deref not supported on " ++ show ty
   where
+    ptrOpSize, fatPtrOpSize :: OpSize
+    ptrOpSize = Width 8
+    fatPtrOpSize = Width 16
```
RyanGlScott: I think that these values will depend on the pointer width, which may be platform-specific. (See https://github.com/GaloisInc/crucible/issues/1384.) I think that these are reasonable choices for now, but we may want to mention https://github.com/GaloisInc/crucible/issues/1384 in a comment here in case we want to generalize this later.

crucible-mir/src/Mir/Mir.hs
```diff
@@ -760,7 +762,23 @@ onlyVariant (Adt name kind _ _ _ _ _) = error $
...
 
 
+--------------------------------------------------------------------------------
+-- OpSize
+
+-- | Several operations need to know how many bytes are occupied by the Rust
+-- values being operated on - use this structure to provide that information.
+data OpSize
+  = -- | A statically-known width, in bytes.
+    Width Word
```
RyanGlScott, line 767: Is there a requirement here that the number of bytes be non-zero?

crucible-mir-cli/test-data/ref.cbl
```diff
@@ -2,8 +2,8 @@
   (start start:
     (let r (ref-new (Bitvector 8)))
     (let v-expected (bv 8 42))
-    (ref-write (Bitvector 8) r v-expected)
-    (let v-actual (ref-read (Bitvector 8) r))
+    (ref-write (Bitvector 8) 8 r v-expected)
+    (let v-actual (ref-read (Bitvector 8) 8 r))
```
RyanGlScott: It's a bit unfortunate that all uses of `ref-write` and `ref-read` now require you to specify the number of bytes to read, as I suspect that most of the time, this will just be the size of the type being written to or read from. For `Bitvector`s, figuring out this number of bytes is not so bad, but for compound types like structs, I suspect that this will be somewhat annoying.

I wonder if we should instead leave the types of `ref-write` and `ref-read` alone and instead introduce variants of these statements that accept an explicit number of bytes instead?

crux-mir/test/symb_eval/debug/dump_nested.good
```diff
@@ -0,0 +1,23 @@
...
+, 8..9 -> 0x1:[8]
+, 10..12 -> 0x2:[16]
+, 12..16 -> 0x3:[32]
+, 16..17 -> 0x5:[8]
+, 24..26 -> 0x6:[16]
+, 32..36 -> 0x7:[32]
+, 40..48 -> 0x8:[64]
+, size 48}
+nested_array = {0..2 -> 0x1:[16], 2..4 -> 0x2:[16], 4..6 -> 0x3:[16], 6..8 -> 0x4:[16], size 8}
+slice = {[((MirReference IntrinsicRepr MirAggregate [] IntrinsicRepr MirAggregate [] (RefCell_RefRoot <NONCE>) (AgOffset_RefPath 0x6:[64] Empty_RefPath)),true)]
```
RyanGlScott, line 11: Wow, this output is pretty ugly. Not really your fault, as `dump_rv` was using `Show` under the hood before this PR, but it's _really_ ugly now that we're printing full nonces. We should consider ways to make this output a little nicer.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -135,6 +134,17 @@ import Mir.Intrinsics.Size (UsizeType, wordLit)
 -- to partially overlap old ones - they must either be disjoint from all
 -- existing entries, or fully overwrite an existing entry.  Read operations
 -- must also touch exactly one entry.
+--
+-- We enforce an invariant on `MirAggregate`s that they are "flattened" - that
+-- is, no `MirAggregate` will contain an _immediate_ `MirAggregateEntry` that is
+-- itself a `MirAggregate`. (A `MirAggregate` can still appear nested within
+-- another `MirAggregate.) Clients shouldn't need to care about this flattening;
```
spernsteiner:
```suggestion
-- is, no `MirAggregate` will contain an _immediate_ `MirAggregateEntry` that is
-- itself a `MirAggregate`. (A `MirAggregate` can still appear nested within
-- another `MirAggregate` indirectly, e.g. a `MirAggregate` inside a `RustEnumRepr`
-- inside another `MirAggregate`.) Clients shouldn't need to care about this flattening;
```
Clarifies why this parenthetical doesn't directly contradict the sentence before.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -331,6 +344,39 @@ readMirAggregateWithSymOffset bak iteFn off tpr ag@(MirAggregate totalSize m)
...
+  | Just (fromIntegral . BV.asUnsigned -> off') <- asBV off = do
+      let err e = leafAbort $ GenericSimError $ "readSubaggregateWithSymOffset: " <> e
+      either err pure (readConcrete off' ag)
+  | otherwise = do
+      liftIO $ foldM
+        (\ag' candidateOff -> do
+          isTheOffset <- bvEq sym off =<< offsetLit candidateOff
+          iteFn isTheOffset (fromRight ag' $ readConcrete candidateOff ag') ag')
+        ag
+        (init [0 .. sz])
```
spernsteiner: I'm skeptical of the logic here, specifically the use of `liftIO` and `fromRight`.  If I read from a symbolic offset within `[(u32, u32)]`, it should effectively assert `off % 4 == 0` as a side effect, since otherwise the read will fail due to bad alignment.

I think this should pass:
```Rust
let xs: &[(u32, u32)] = ...;
let i = usize::symbolic_where("i", |&i| i < xs.len());
xs[i]
```
And this should fail:
```
let xs: &[(u32, u32)] = ...;
let off = usize::symbolic_where("i", |&off| off <= (xs.len() - 1) * mem::size_of_val(&xs[0]));
unsafe { *(xs.as_ptr().byte_add(off) }
```

spernsteiner: Also, `ag` is probably the wrong default for the case when all `readConcrete` calls fail.  And this should probably be passing `ag` to `readConcrete` rather than `ag'`.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -396,6 +444,46 @@ adjustMirAggregateWithSymOffset bak iteFn off tpr f ag@(MirAggregate totalSize m
...
+adjustSubaggregateWithSymOffset ::
+  forall sym bak.
+  (IsSymBackend sym bak) =>
+  bak ->
+  (Pred sym -> MirAggregate sym -> MirAggregate sym -> IO (MirAggregate sym)) ->
+  RegValue sym UsizeType ->
+  (MirAggregate sym -> MuxLeafT sym IO (MirAggregate sym)) ->
+  MirAggregate sym ->
+  MuxLeafT sym IO (MirAggregate sym)
+adjustSubaggregateWithSymOffset bak iteFn off f ag@(MirAggregate sz _)
```
spernsteiner: Why doesn't this need a read/write size argument, like the `OpSize` passed to `readSubaggregateWithSymOffset`?

crucible-mir/src/Mir/Mir.hs
samcowger: No, I think the operations that accept widths will (or at least should) gracefully handle widths of zero. For writes, that means trivially succeeding without updating the aggregate, and for reads, that probably means evincing a zero-sized aggregate value regardless of the aggregate being read from.

crucible-mir/src/Mir/Trans.hs
samcowger: Good call, I'll do that.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -563,11 +651,39 @@ mirAggregate_entries :: sym -> MirAggregate sym -> [(Word, MirAggregateEntry sym
...
+    _ | innerAgSz /= entrySz -> Left $
+      "mirAggregate_insert: when adding a `MirAggregate` as an entry, " <>
+      "the `MirAggregateEntry`-derived size was " <> show entrySz <> ", " <>
+      "but the `MirAggregate` value had size " <> show innerAgSz
+    Just False ->
+      pure outerAg
+    Just True -> do
+      let addEntry ag (fromIntegral -> entryOff) entry =
+            mirAggregate_insert (outerAgOff + entryOff) entry ag
+      foldMWithKey addEntry outerAg innerAgEntries
```
spernsteiner: I think this needs to clear out all existing entries covered by the inner aggregate before writing new ones.  (Note aggregates can contain empty/uninitialized space.)  Suppose we have code like this:
```Rust
let mut xs = [MaybeUninit::new((1, 2))];
xs[0] = MaybeUninit::uninit();
let x = unsafe { xs[0].assume_init_read() };
x.0 + x.1
```
I think what will happen here is the write to `xs[0]` will have no effect, since the rvalue `MaybeUninit::uninit()` produces an empty aggregate.  So the final result will be 3, when instead it should fail due to reading an uninitialized aggregate entry on the access of `x.0`.  (According to Rust semantics this should actually detect UB on the `assume_init_read()`, but with our current design, such errors are deferred until the uninitialized value is actually used.)

crucible-mir-cli/test-data/ref.cbl
samcowger: > I wonder if we should instead leave the types of `ref-write` and `ref-read` alone

We could probably get away with that by generating `Mir{Read,Write}Ref` statements that use the `All` variant of `OpSize` - that should suffice for basically any non-aggregate reads/writes, where the size is ignored anyway.

Thinking ahead to potential sized versions of `ref-{read,write}`, if I implemented them, I would want to document them with examples of what kinds of `.cbl` programs require them - but it's not actually clear to me that this syntax is capable of expressing programs that generate `MirAggregate` values. `MirAggregateRepr` doesn't seem to be a valid parse target, and I don't think aggregates are used as a representation for any of the parseable `TypeRepr`s. This makes me think we should leave these operations unimplemented for now, and perhaps note somewhere around [here](https://github.com/GaloisInc/crucible/blob/591af6927eb469be95c744617bdbfb194870a87c/crucible-mir-syntax/src/Lang/Crucible/MIR/Syntax.hs#L58-L69) that adding `MirAggregateRepr` as a parse target should entail implementing those operations. Does that sound reasonable?

spernsteiner (commented):
Only reviewed `Aggregate.hs` so far

crux-mir/test/symb_eval/debug/dump_nested.good
samcowger: I agree; I've opened #1843.

crucible-mir-cli/test-data/ref.cbl
spernsteiner: > that should suffice for basically any non-aggregate reads/writes, where the size is ignored anyway

Non-aggregate writes should not be ignoring the size—they need to check which entries in the destination potentially overlap the one being written, as described in the non-`MirAggregateRepr` case of `mirAggregate_insert`.  For example, if you write a `u8` at offset 1 followed by a `u16` at offset 0, the latter write should error due to the overlap.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Hmm, good question. I think it should either have one, or it should allow its adjustment function to produce a smaller aggregate than it's provided, and I think the former probably more appropriate.

A test case that demonstrates the need for this:
```rs
extern crate crucible;
use crucible::Symbolic as _;

#[crux::test]
fn test() {
    let mut xs: [[u16; 2]; 2] = [[1, 2], [3, 4]];
    let i = usize::symbolic_where("i", |&i| i < xs.len());
    xs[i] = [0, 0];
}
```

crucible-mir-cli/test-data/ref.cbl
samcowger: Hmm, I'm not sure we're on the same page. I'm using "non-aggregate writes" to mean writes of non-aggregate values to _non-aggregate_ places, like `let mut x = 0u8; x = 1`, but it sounds like you're using it to mean writes of non-aggregate values to _aggregate_ places, like `let mut x = [0u8, 1]; x[0] = 1`. Writes under your interpretation do indeed use sizes to check for overlap.

RyanGlScott: > it's not actually clear to me that this syntax is capable of expressing programs that generate `MirAggregate` values.

It's not capable of expressing `MirAggregate` values. `crucible-mir-syntax` is still a basic proof-of-concept at the moment, and it has a ways to go before it is as fully features as, say, `crucible-llvm-syntax`. See https://github.com/GaloisInc/crucible/issues/1485.

> This makes me think we should leave these operations unimplemented for now, and perhaps note somewhere around [here](https://github.com/GaloisInc/crucible/blob/591af6927eb469be95c744617bdbfb194870a87c/crucible-mir-syntax/src/Lang/Crucible/MIR/Syntax.hs#L58-L69) that adding MirAggregateRepr as a parse target should entail implementing those operations. Does that sound reasonable?

That sounds reasonable to me.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Thanks for calling this out; I've been a little hazy on how this logic (and similar logic in `adjustSubaggregateWithSymOffset`) ought to work, but I think this helps clarify it for me, and I think I understand how to fix it.

crucible-mir-cli/test-data/ref.cbl
samcowger: Alright, done.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: I think I've fixed both this logic and that of `adjustSubaggregateWithSymOffset`, and have included these examples, and a couple others based on them, as tests in `crux-mir/test/symb_eval/array/mux_offset.rs`.

samcowger: Fixed, see above thread.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Fixed, and added this example as a test at `crux-mir/test/symb_eval/array/nested.rs`.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Quite right, good catch. I've fixed this and added this example as an xfail test at `crux-mir/test/conc_eval/mem/maybe_uninit_array.rs`.

RyanGlScott:
This is looking promising, and the [corresponding SAW pull request](https://github.com/GaloisInc/saw-script/pull/3354) appears to work as well. Let me know when this is ready for a final review.

samcowger:
> This is looking promising, and the [corresponding SAW pull request](https://github.com/GaloisInc/saw-script/pull/3354) appears to work as well. Let me know when this is ready for a final review.

Indeed, I was waiting for the SAW PR to pass CI, which it's now done. Given that, I believe this is ready for review.

(My apologies for the size of this PR - when I started out on it, I didn't anticipate it would grow this large! You might prefer to review this commit-by-commit, as plenty of these commits are boring/functionality-preserving.)

crucible-mir/src/Mir/Mir.hs
```diff
@@ -755,7 +757,23 @@ onlyVariant (Adt name kind _ _ _ _ _) = error $
...
+  | -- | A dynamic width, generally meant for reading entire allocations,
+    -- regardless of size.
+    All
+  deriving (Eq, Ord)
 
+instance Pretty OpSize where
+  pretty s = case s of
+    Width w -> pretty w
+    All -> "<alloc>"
```
RyanGlScott: What is the rationale for printing `<alloc>` here? Would `<all>` make more sense?

crucible-mir/src/Mir/Intrinsics/Syntax.hs
```diff
@@ -165,11 +164,16 @@ data MirStmt :: (CrucibleType -> Type) -> CrucibleType -> Type where
...
+     -- | The size of the value to read, in bytes. This must be known concretely
+     -- at symbolic execution time.
+     !OpSize ->
      !(f MirReferenceType) ->
      MirStmt f tp
   MirWriteRef ::
      !(TypeRepr tp) ->
      !(f MirReferenceType) ->
+     -- | The size of the value to write.
+     !OpSize ->
```
RyanGlScott: The Haddocks between the two uses of `OpSize` here are a bit lopsided, as the former use has a lot more detail that the latter doesn't. Personally, I'd be inclined just to say "The size of the value to read." as the Haddocks for the former, as `OpSize`s use bytes by definition, and due to the types involved, they cannot be symbolic.

crucible-mir/src/Mir/FancyMuxTree.hs
```diff
@@ -261,6 +261,30 @@ leafUpdatePartExpr bak mux x (PE q y) = do
             xy <- lift $ mux p x y
             return $ mkPE pq xy
 
+-- | Like `leafUpdatePartExpr`, but instead of unconditionally writing a new
+-- value, modify an already-existing value if one exists.
+leafModifyPartExpr ::
+    (IsSymBackend sym bak, MonadIO m) =>
+    bak ->
+    (Pred sym -> a -> a -> m a) ->
+    (Maybe a -> m a) ->
```
RyanGlScott, line 275: Perhaps worth documenting that `Nothing` = no value present, and `Just x ` = the value `x` was present.

crucible-mir/src/Mir/FancyMuxTree.hs
```diff
@@ -261,6 +261,30 @@ leafUpdatePartExpr bak mux x (PE q y) = do
             xy <- lift $ mux p x y
             return $ mkPE pq xy
 
+-- | Like `leafUpdatePartExpr`, but instead of unconditionally writing a new
+-- value, modify an already-existing value if one exists.
```
RyanGlScott: I think it's more than just modifying an already-existing value, as it only modifies the value if the leaf predicate is true. (In other words, this function doesn't call `modify` in the `Just False` case below.)

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1549,35 +1550,6 @@ slice_from_ref = slice_from Immut
...
-    where
-        hookLoc = case mut of
-            Immut -> "as_chunks_unchecked"
-            Mut -> "as_chunks_unchecked_mut"
-
-slice_as_chunks_cast_hook :: (ExplodedDefId, CustomRHS)
-slice_as_chunks_cast_hook = slice_as_chunks_cast_hook_common Immut
-
-slice_as_chunks_mut_cast_hook :: (ExplodedDefId, CustomRHS)
-slice_as_chunks_mut_cast_hook = slice_as_chunks_cast_hook_common Mut
```
RyanGlScott, line 1579: Worth filing a follow-up issue on the `mir-json` side as a reminder to remove the `crucible_cast_hook` patch now that `crucible-mir` no longer needs it.

crucible-mir/src/Mir/TransTy.hs
RyanGlScott: There are still some references to "size 1" here. Are these out of date now that Rust `union` types' `MirAggregate` representations use the same size as `_adtSize`?

crux-mir/test/symb_eval/union/same_repr.rs
```diff
@@ -0,0 +1,41 @@
+// FAIL: attempt to mux differently-initialized unions
```
RyanGlScott: I'm a bit confused by this comment. This says that it is expected to fail, but `same_repr.good` shows that it can discharge all of the proof goals successfully.

crux-mir/test/Test.hs
```diff
@@ -211,12 +211,18 @@ sanitizeGoldenTestOutput = go
       | Just (before, _matched :: BS.ByteString, after, [preDisamb])
           <- matchM disambRE out
       = before <> preDisamb <> "<DISAMB>::" <> go after
+      | Just (before, _matched :: BS.ByteString, after, [] :: [BS.ByteString])
+          <- matchM nonceRE out
+      = before <> "<NONCE>" <> go after
```
RyanGlScott, line 218: This find-and-replace is only needed to sanitize the pretty-printed output of a `MirReference`, correct? It might be worth leaving a TODO comment here reminding us to remove this hack once https://github.com/GaloisInc/crucible/issues/1843 is fixed.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -135,6 +134,18 @@ import Mir.Intrinsics.Size (UsizeType, wordLit)
 -- to partially overlap old ones - they must either be disjoint from all
 -- existing entries, or fully overwrite an existing entry.  Read operations
 -- must also touch exactly one entry.
+--
+-- We enforce an invariant on `MirAggregate`s that they are "flattened" - that
+-- is, no `MirAggregate` will contain an _immediate_ `MirAggregateEntry` that is
+-- itself a `MirAggregate`. (A `MirAggregate` can still appear _nested_ within
+-- another `MirAggregate indirectly, e.g. a `MirAggregate` inside a
+-- `RustEnumRepr` inside another `MirAggregate`.) Clients shouldn't need to care
```
RyanGlScott, line 144: What happens if you put a `MirAggregate` inside of a `repr(transparent)` enum inside of another `MirAggregate`? Does this result in a single `MirAggregate`, or does that result in a `MirAggregate` inside of a `MirAggregate`?

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -331,6 +345,66 @@ readMirAggregateWithSymOffset bak iteFn off tpr ag@(MirAggregate totalSize m)
...
+  | Just (fromIntegral . BV.asUnsigned -> off') <- asBV off =
+    readConcrete off'
+  | otherwise = do
+      -- Do some static winnowing of the offsets we need to check. Don't try to
+      -- read at an offset if:
+      -- - There isn't a large enough sub-aggregate starting at that offset
+      -- - Extracting a sub-aggregate starting at that offset would split an
+      --   aggregate entry
+      let sizedAgOffsets = case readSize of
+            All -> init [0 .. agSize] -- not sure this `init` is necessary...
```
RyanGlScott: What happens if `agSize` is `0` here? That would result in `[0 .. agSize]` evaluating to `[]`, and calling `init []` would crash.

(Similar comments for the code in `adjustSubaggregateWithSymOffset` as well.)

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -331,6 +345,66 @@ readMirAggregateWithSymOffset bak iteFn off tpr ag@(MirAggregate totalSize m)
...
+      -- left with `Unassigned` in failure cases, and can propagate that error
+      -- via `leafReadPartExpr`.
+      let sizeMsg = case readSize of All -> ""; Width w -> " of size " <> show w
+      leafReadPartExpr bak agPMux $
+        GenericSimError $
+          "readSubaggregateWithSymOffset: in aggregate of size " <> show agSize <>
+          ", no subaggregate value" <> sizeMsg <> " at symbolic offset"
+  where
+    readConcrete off' = do
+      let sz = case readSize of All -> agSize - off'; Width w -> w
```
RyanGlScott: Formatting nitpick: instead of using semicolons here, could you use whitespace to separate the cases?


```suggestion
      let sz = case readSize of
                 All -> agSize - off'
                 Width w -> w
```

And similarly for other uses of semicolons in this file.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -563,11 +704,42 @@ mirAggregate_entries :: sym -> MirAggregate sym -> [(Word, MirAggregateEntry sym
 mirAggregate_entries _sym (MirAggregate _totalSize m) =
   [(fromIntegral off, entry) | (off, entry) <- IntMap.toList m]
 
+foldMWithKey :: Monad m => (b -> Int -> a -> m b) -> b -> IntMap a -> m b
+foldMWithKey f z m = IntMap.foldlWithKey' f' (pure z) m
+  where
+    f' accM key val = do
+      acc <- accM
+      f acc key val
```
RyanGlScott: If you don't want to redefine this function yourself, you can always use [`ifoldlM`](https://hackage-content.haskell.org/package/indexed-traversable-0.1.5/docs/Data-Foldable-WithIndex.html#v:ifoldlM) from the `indexed-traversable` library, which `crucible-mir` already depends on transitively.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1551,6 +1545,12 @@ evalPlaceProj ty (MirPlace tpr ref NoMeta) M.Deref = do
         CTyBox t -> doRef t
         _ -> mirFail $ "deref not supported on " ++ show ty
   where
+    -- These values presume 64-bit architectures. In the longer term, we'd
+    -- prefer not to hardcode that presumption - see #1384.
+    ptrOpSize, fatPtrOpSize :: OpSize
+    ptrOpSize = Width 8
+    fatPtrOpSize = Width 16
```
RyanGlScott: Thinking about this more, perhaps we should define `ptrOpSize`/`fatPtrOpSize` in terms of `SizeBits`? Similarly for the call to `buildWrapperAggregate ptrSize` below, which defines `ptrSize = 8` separately. Then we could move the comment about #1384 near the definition of `SizeBits`, and then if we decide to fix `SizeBits` later, then we wouldn't need to remember to update all of these places in `Mir.Trans`.

crucible-mir/src/Mir/Mir.hs
samcowger: `<all>` would be fine, I think - I'll change it.

crucible-mir/src/Mir/Intrinsics/Syntax.hs
samcowger: Ah, good catch - I think I must have made this symbolic in an earlier iteration of this work, since this comment is quite nonsensical in this context.

crucible-mir/src/Mir/TransCustom.hs
samcowger: Good call, I've opened https://github.com/GaloisInc/mir-json/issues/293.

crucible-mir/src/Mir/TransTy.hs
samcowger: Yes, they are - I'll update them.

crux-mir/test/symb_eval/union/same_repr.rs
samcowger: > I'm a bit confused by this comment

As you should be - it makes no sense. I think this was an artifact of my defining this test by first copying its sibling `init_cond.rs`. I'll remove it.

crux-mir/test/Test.hs
samcowger: Yes, I think that's right - I'll make a note.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: If `agSize` is `0`, then we have `[0 .. 0]`, which (for better or worse) Haskell evaluates as `[0]`, so this ought to remain well-defined.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Sure, I can change that.

crucible-mir/src/Mir/Trans.hs
samcowger: Good call, I'll do that.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Good question - my instinct would be that that yields an aggregate within (a `RustEnumRepr` within) an aggregate, but it's possible that our special cases for `repr(transparent)` translation would fire and cause the resulting aggregate to end up flattened. Why do you ask?

RyanGlScott: I am curious to know if `crucible-mir` actually handles this case correctly!

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
RyanGlScott: Oops, my mental model of how `[x .. y]` works was incorrect. Sorry for the noise.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: As some evidence that it does, `crux-mir` can simulate this without issue:
```rs
#[repr(transparent)]
#[derive(Clone, Debug, Eq, PartialEq)]
enum E2 {
    A([u8; 2]),
}

#[repr(transparent)]
#[derive(Clone, Debug, Eq, PartialEq)]
enum E4 {
    A([u8; 4]),
}

#[crux::test]
fn test() {
    let mut e2s = [E2::A([1, 2]), E2::A([3, 4])];
    assert_eq!(e2s[0], E2::A([1, 2]));
    e2s[0] = E2::A([5, 6]);
    assert_eq!(e2s[0], E2::A([5, 6]));

    let mut xss: [[u8; 1]; 4] = unsafe { std::mem::transmute(e2s) };
    assert_eq!(xss, [[5], [6], [3], [4]]);

    xss[1] = [7];
    xss[2] = [8];
    let e4s: [E4; 1] = unsafe { std::mem::transmute(xss) };
    assert_eq!(e4s, [E4::A([5, 7, 8, 4])]);
}
```

Unfortunately, I've also managed to produce some evidence that it doesn't, as `crux-mir` will choke on this (with `E2` defined as above):
```rs
#[crux::test]
fn test() {
    let e2s = [E2::A([1, 2]), E2::A([3, 4])];
    let _ = e2s.clone();
}
```

It fails with `bad MirAggregate insert: entry out of range: covers 2..3, but total size is 2`, which is reminiscent of the error that was resolved in ecea9e158c70d7388e933a1b9d5482e401c52226. I'm not yet sure what's going wrong here.

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
samcowger: Good call, done.

crucible-mir/src/Mir/FancyMuxTree.hs
samcowger: Good point, I've clarified this behavior.

crucible-mir/src/Mir/FancyMuxTree.hs
samcowger: Done

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
spernsteiner: Reduced example:
```Rust
#[derive(Clone)]
#[repr(transparent)]
struct S([u8; 2]);

#[cfg_attr(crux, crux::test)]
pub fn crux_test() {
    let xs = [S([1, 2]), S([3, 4]), S([5, 6])];
    let y = [xs.iter().next().unwrap().clone()];
}
```
Note that `let y = [xs[0].clone()]` succeeds.  The difference according to `dump_rv` is that `&xs[0]` has an `AgOffset_RefPath` with offset zero, while `xs.iter().next().unwrap()` has no `AgOffset` at all.

spernsteiner:
Rebased onto current `master`

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
spernsteiner: The issue was that `readRefPath` with `Empty_RefPath` would read out the entire aggregate, even if the `readSize` was smaller.  So `*xs.as_ptr()`, whose Rust representation is `[u8; 2]`, would actually return an aggregate with 6 elements.  I pushed e1be8476fc8db8a728f201ee10a5b5025e62760b, which needs a bit of cleanup (it duplicates code from the `AgOffset_RefPath` case, and uses `readMirAggregateWithSymOffset` even though the offset is always concretely zero) but fixes the `E2` and `S` examples from this thread.

spernsteiner:
I somehow missed a bunch of recent changes when I did the rebase, so I've restored the previous commit e93d110.

RyanGlScott:
Let me know when you've pushed a resolution to https://github.com/GaloisInc/crucible/pull/1842#discussion_r3648336059, and I'm happy to take a final look (and approve).

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
spernsteiner: Pushed db977b271df208e48a19950cb0333bc4bccb92ac with a cleaner fix for this.

RyanGlScott (approved):

spernsteiner:
@RyanGlScott Can you take a look at https://github.com/GaloisInc/crucible/pull/1842/changes/db977b271df208e48a19950cb0333bc4bccb92ac..c5e5e812e980196ca3400ee5793a393544df54dd?  These are just some minor fixes based on stuff I saw while reviewing the code.

RyanGlScott:
Looks reasonable to me. Consider the PR re-approved.

spernsteiner:
Oops, I merged this not realizing that https://github.com/GaloisInc/saw-script/pull/3354 is not quite ready to go yet.  I'll try to finish that up tomorrow to avoid blocking other saw-script submodule updates.

Issue #1109, opened 2023-09-27
`crucible-mir`: Reconsider `MaybeRepr` conventions for structs


RyanGlScott:
Currently, `crucible-mir` uses the following conventions for translating structs and tuples into `RegValue`s:

* A tuple value `(v_1 : t_1, ..., v_n : t_n)` is translated to a `RegValue` of type `StructRepr [MaybeRepr t_1, ..., MaybeRepr t_n]`. The use of `MaybeRepr` reflects the fact that MIR can partially initialize the fields of a struct, so an `Unassigned` value indicates that a field has not yet been initialized.
* A struct value `S { v_1 : t_1, ..., v_n : t_n }` is translated to a `RegValue` of type `AnyRepr`, where the packed representation is of type `StructRepr [f_1, ..., f_n]`, where each `f_i` is one of the following:
  * `t_i`, if the type `t_i` has a reasonable initial value (see [`canInitialize`](https://github.com/GaloisInc/crucible/blob/602cb8f04ec5f42e76a2ba0833f7e91a95d7109c/crucible-mir/src/Mir/TransTy.hs#L252-L277) for how this is determined).
  * `MaybeRepr t_i` otherwise.

Despite the fact that tuples and structs are conceptually quite similar, they use rather different `StructRepr` conventions. I propose to simplify things by making both tuples and structs use the `StructRepr [MaybeRepr t_1, ..., MaybeRepr t_n]`—that is, we should no longer use `canInitialize`. I have two motivations for this:

1. This makes the treatment of tuples and structs much more uniform throughout the code.
2. In SAW, it is convenient to be able to have tuple and struct values with uninitialized fields. Notably, this is convenient to use for SAW overrides, as we want to invalidate mutable allocations used in an override after it has been applied to ensure that they are used in a sound manner. With the `StructRepr [MaybeRepr t_1, ..., MaybeRepr t_n]` convention, created one of these structs is as simple as `Empty :> Unassigned :> ... :> Unassigned`.

[...]

PR #1849, opened 2026-06-13
`crucible-mir`: `crucible::alloc::reallocate` doesn't reallocate in-place

samcowger:
Accommodates https://github.com/GaloisInc/mir-json/pull/276.

The existing implementation of our `reallocate` override might overwrite a smaller `MirAggregate` with a larger one. With #1842's changes, this becomes an error.

### DIFF
--- a/crux-mir/test/conc_eval/vec/reallocate.rs
+++ b/crux-mir/test/conc_eval/vec/reallocate.rs
@@ -1,0 +1,15 @@
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> (i32, i32) {
+    let orig_capacity = 1;
+    let mut v = Vec::with_capacity(orig_capacity);
+    while v.len() < v.capacity() {
+        v.push(0);
+    }
+    v.push(1);
+    assert_ne!(v.capacity(), orig_capacity);
+    (v[0], *v.last().unwrap())
+}
+
+pub fn main() {
+    println!("{:?}", crux_test());
+}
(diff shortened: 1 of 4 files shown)


RyanGlScott (commented):
This looks generally reasonable, although I do have one question: what should happen to the old pointer after `reallocate` is called on it? Namely, what should happen if someone tries to access the old pointer?

Rust's [`realloc`](https://doc.rust-lang.org/std/alloc/trait.GlobalAlloc.html#method.realloc) function says that attempting to do this is explicitly undefined behavior. I wonder if we should model something similar? (See also https://github.com/GaloisInc/crucible/issues/1109.)

spernsteiner:
I think it's fine to do nothing with the old pointer for now.  We currently don't check for use-after-`dealloc`, so treating `realloc` the same way seems fine until we fix that.

crux-mir/test/conc_eval/vec/reallocate.rs
```diff
@@ -0,0 +1,13 @@
+#[cfg_attr(crux, crux::test)]
+fn crux_test() -> (i32, i32) {
+    let mut v = Vec::with_capacity(1);
+    while v.len() < v.capacity() {
+        v.push(0);
+    }
+    v.push(1);
+    (v[0], *v.last().unwrap())
```
spernsteiner, line 10: Maybe assert here that the capacity actually increased?

RyanGlScott:
Fair enough. That being said, it might be worth adding a disclaimer to `reallocate`'s docstring that the behavior of accessing the old pointer is unspecified.

spernsteiner (approved):

RyanGlScott (approved):

samcowger:
> Fair enough. That being said, it might be worth adding a disclaimer to `reallocate`'s docstring that the behavior of accessing the old pointer is unspecified.

I've done so - and for good measure, in efa292b, I've also updated a testcase to avoid leveraging that very behavior.

spernsteiner:
Since the mir-json part of this is merged but the crucible part is not, this is blocking further changes to mir-json, such as the toolchain upgrade https://github.com/GaloisInc/mir-json/pull/277 + https://github.com/GaloisInc/crucible/pull/1851 (though that upgrade is not urgent)

PR #1850, opened 2026-06-15
Improve coverage reporting

### LINKED ISSUE #1627: Improve report-coverage metric
Add the following features to the [report-coverage](https://github.com/GaloisInc/crucible/tree/master/crux-mir/report-coverage) tool:
* optionally disable hash checking (for compiling large crates)
* add a filter for skipping certain regexpressions from the coverage count
* add an average/total coverage of functions and branches
* don't show `0/0` for visited functions without any branches

podhrmic:
Closes #1627

### DIFF
--- a/crux-mir/report-coverage/src/main.rs
+++ b/crux-mir/report-coverage/src/main.rs
@@ -841,7 +841,7 @@ impl Filter {
 }
 
 
-fn report_all(reporter: &mut Reporter, cov: &Coverage) -> (f32, i32) {
+fn report_all(reporter: &mut Reporter, cov: &Coverage) -> (f32, usize) {
 
     let mut summary = vec![];
 
@@ -916,28 +916,22 @@ fn report_all(reporter: &mut Reporter, cov: &Coverage) -> (f32, i32) {
     }
 
     let mut sum_coverage = 0;
-    let mut iter = 0;
-    for (fun, called, seen,tot) in summary.into_iter() {
-        let mut skip = false;
+    let visited_fns = summary.len();
+    'outer: for (fun, called, seen,tot) in summary.into_iter() {
         if let Some(filters) = &reporter.filters {
             // see if the function contains any of the filters
             for val in filters {
                 if fun.contains(&val.filename) {
-                    skip = true;
-                    break
+                    continue 'outer;
                 }
             }
         }
-        if skip {
-            continue;
-        }
         sum_coverage = sum_coverage + reporter.coverage_stats(fun, called, seen, tot);
-        iter = iter + 1;
     }
 
-    let average_coverage = sum_coverage as f32 / iter as f32;
+    let average_coverage = sum_coverage as f32 / visited_fns as f32;
 
-    (average_coverage, iter)
+    (average_coverage, visited_fns)
 
 }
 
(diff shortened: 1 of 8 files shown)


crux-mir/report-coverage/src/main.rs
```diff
@@ -896,11 +897,16 @@ fn report_all(reporter: &mut Reporter, cov: &Coverage) {
         summary.push((fun, fn_cov.fn_called, seen, tot));
     }
 
-
+    let mut sum_coverage = 0;
+    let mut iter = 0;
     for (fun, called, seen,tot) in summary.into_iter() {
-        reporter.coverage_stats(fun, called, seen, tot);
+        sum_coverage = sum_coverage + reporter.coverage_stats(fun, called, seen, tot);
+        iter = iter + 1;
```
RyanGlScott: Instead of computing `iter` like this, I think you can just use `summary.len()` instead.

crux-mir/report-coverage/src/main.rs
```diff
@@ -972,15 +978,21 @@ fn main() {
...
         if let Some(old_trans_hash) = trans_hash {
-            assert!(
-                cur_trans_hash == old_trans_hash,
-                "translation hashes for {:?} and {:?} do not match",
-                report_path_str, m.value_of_os("input").unwrap(),
-            );
+            if m.is_present("ignore-hash") {
+                println!("Warning! Translation hashes for {:?} and {:?} do not match",
+                    report_path_str, m.value_of_os("input").unwrap());
+            }
```
RyanGlScott: The flag name "`ignore-hash`" is a bit of a misnomer, given that the code _is_ paying attention to the hashes---it's just warning about hash mismatches now rather than erroring out.

crux-mir/report-coverage/src/main.rs
```diff
@@ -907,12 +914,26 @@ fn report_all(reporter: &mut Reporter, cov: &Coverage) {
...
+            for val in filters {
+                if fun.contains(&val.filename) {
+                    skip = true;
+                    break
+                }
+            }
+        }
+        if skip {
+            continue;
+        }
```
RyanGlScott: This is a bit inefficient: we want to continue to the next iteration of the outermost loop (the `for (fun, called, seen,tot) in summary.into_iter() { ... }` one) as soon as we find a function name mentioned in a filter, but the innermost loop (the `for val in filters { ... }` one) must search through all filters before it can call `continue`. Consider calling `continue` on the outermost loop as soon as you find a matching filter, i.e.,

```rs
    'outer: for (fun, called, seen,tot) in summary.into_iter() {
        if let Some(filters) = &reporter.filters {
            // see if the function contains any of the filters
            for val in filters {
                if fun.contains(&val.filename) {
                    continue 'outer;
                }
            }
        }
        ...
    }
```

crux-mir/report-coverage/src/main.rs
```diff
@@ -26,16 +26,29 @@ fn parse_args() -> ArgMatches<'static> {
              .long("filter")
              .takes_value(true)
              .value_name("FILE[:[LINE]-[LINE]]")
-             .help("only report uncovered branches in the indicated source region(s)")
+             .help("only report uncovered branches in the indicated source region(s).\
+                    Use filter to remove functions from the calculated coverage,\
+                    in that case the visited functions that contain the filter value\
+                    are skipped (useful for removing tests and similar from the coverage")
```
RyanGlScott: Some of the help text here feels a bit redundant. Here is my attempt at condensing it a bit:

```suggestion
             .help("Only report uncovered branches in the indicated source region(s).\
                    Any visited functions whose names contain the filter value are skipped\
                    when computing the calculated coverage.")
```

RyanGlScott (commented):
Looks generally reasonable, but do make sure to update the expected test output in the test suite.

crux-mir/report-coverage/src/main.rs
podhrmic: Good point, changed to `warn-hash`

crux-mir/report-coverage/src/main.rs
podhrmic: Thanks, I had no idea Rust supports labels like that!

RyanGlScott (approved):

PR #1851, opened 2026-06-15
update Rust to nightly 2026-03-21

spernsteiner:
Most changes here are minor.  The most notable ones are:

* Remove variant `vdiscr` field (JSON: `discr`) in favor of `discrval` (JSON: `discr_value`).  `vdiscr` held the `DefId` of the MIR representation of the constant discriminant value, and we previously had some logic to pattern-match on the MIR and extract the actual integer value.  `discrval` holds the value directly.  This change was necessary because mir-json was failing to obtain the MIR representation for `vdiscr` in some cases after the toolchain upgrade.
* Add `str` and thin-pointer cases for `AKRawPtr` (rustc: `AggregateKind::RawPtr`), which implements `ptr::from_raw_parts`.  The `dyn Trait` case is still unimplemented, since we don't have any Crucible representation for vtable pointers.
* Add many new test cases for mir-json library patches.  Previously we had about 10 patches in mir-json that changed library behavior in ways that were completely untested in crux-mir, making it hard to tell whether the patch is correct and whether it's still needed.

mir-json PR: https://github.com/GaloisInc/mir-json/pull/277

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -414,6 +414,15 @@ evalOperand (M.Move lv) = evalPlace lv >>= readPlace (M.typeOf lv)
 evalOperand (M.OpConstant (M.Constant conty constval)) = do
     Some tpr <- tyToReprM conty
     transConstVal conty (Some tpr) constval
+evalOperand (M.OpRuntimeChecks rc) = do
+    let enable = case rc of
+            -- Disable undefined behavior checks.
+            -- TODO: re-enable this later, and fix the tests that break
+            -- (see https://github.com/GaloisInc/mir-json/issues/107)
+            UbChecks -> False
+            ContractChecks -> True
+            OverflowChecks -> True
+    return $ MirExp C.BoolRepr $ R.App $ E.BoolLit enable
 evalOperand (M.Temp rv) = evalRval rv
 
 -- | Dereference a `MirExp` (which must be `MirReferenceRepr` pointing to a
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -101,7 +101,6 @@ import qualified Mir.DefId as M
 
 import Mir.Intrinsics
 import Mir.Generator
-import Mir.GenericOps
 import Mir.TransTy
 
 import Mir.PP (fmt, fmtDoc)
@@ -3079,7 +3078,7 @@ transVirtCall colState intrName' methName dynTraitName methIndex
 
 mkDiscrMap :: M.Collection -> Map M.AdtName [Integer]
 mkDiscrMap col = mconcat
-    [ Map.singleton (adt ^. M.adtname) (adtIndices adt col)
+    [ Map.singleton (adt ^. M.adtname) (map (\v -> v ^. M.discrval) $ adt ^. M.adtvariants)
     | adt <- Map.elems $ col ^. M.adts, M.isEnum (adt ^. M.adtkind) ]
 
 -- | Gather all of the 'M.DefId's in a 'M.Collection' and construct a
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -714,18 +714,6 @@ evalBinOp bop mat me1 me2 =
                 (S.app $ E.BVEq w y $ S.app $ eBVLit w ((1 `shiftL` w') - 1))
       where w' = fromIntegral $ intValue w
 
--- Nullary ops in rust are used for resource allocation, so are not interpreted
-transNullaryOp ::  M.NullOp -> M.Ty -> MirGenerator h s ret (MirExp s)
-transNullaryOp nop ty =
-  case nop of
-    M.AlignOf -> getLayoutFieldAsMirExp "AlignOf" layAlign ty
-    M.SizeOf -> getLayoutFieldAsMirExp "SizeOf" laySize ty
-    M.UbChecks -> do
-      -- Disable undefined behavior checks.
-      -- TODO: re-enable this later, and fix the tests that break
-      -- (see https://github.com/GaloisInc/mir-json/issues/107)
-      return $ MirExp C.BoolRepr $ R.App $ E.BoolLit False
-
 transUnaryOp :: M.UnOp -> M.Operand -> MirGenerator h s ret (MirExp s)
 transUnaryOp uop op = do
     mop <- evalOperand op
@@ -1373,7 +1361,6 @@ evalRval (M.Len lv) =
         ty -> mirFail $ "don't know how to take Len of " ++ show ty
 evalRval (M.Cast ck op ty) = evalCast ck op ty
 evalRval (M.BinaryOp binop op1 op2) = transBinOp binop op1 op2
-evalRval (M.NullaryOp nop nty) = transNullaryOp  nop nty
 evalRval (M.UnaryOp uop op) = transUnaryOp  uop op
 evalRval (M.Discriminant lv discrTy) = do
     let enumTy = typeOf lv
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -1034,6 +1034,10 @@ evalCast' ck ty1 e ty2  = do
       -- we decide to track that, this needs to be updated.
       (M.UnsafeFnPointer, _, _) | ty1 == ty2 -> pure e
 
+      -- Subtype is a no-op, as it is only present in the MIR to making subtyping
+      -- explicit during optimizations and codegen.
+      (M.Subtype, _, _) -> pure e
+
       _ -> mirFail $ "unimplemented cast: " ++ (show ck) ++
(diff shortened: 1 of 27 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -1410,21 +1410,30 @@ evalRval rv@(M.Aggregate ak ops) =
...
+                    return $ MirExp MirSliceRepr tup
+                (TyStr, MirReferenceRepr, UsizeRepr) -> do
+                    let tup = S.mkStruct
+                            (Ctx.Empty Ctx.:> MirReferenceRepr Ctx.:> knownRepr)
+                            (Ctx.Empty Ctx.:> ptr Ctx.:> meta)
+                    return $ MirExp MirSliceRepr tup
+                (_, MirReferenceRepr, MirAggregateRepr)
+                  | Nothing <- findUnsizedTail col ty
+                  , TyTuple [] <- typeOf opMeta -> do
+                    return $ MirExp MirReferenceRepr ptr
```
RyanGlScott, line 1443: Might be worth a comment here that this covers all thin pointers (i.e., references and raw pointers), and that the metadata should be `()`.

crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -1028,6 +1029,39 @@ unchecked_shr =
     , makeUncheckedArith "unchecked_shr" Shr
     )
 
+
+carrying_mul_add :: (ExplodedDefId, CustomRHS)
+carrying_mul_add = (["core", "intrinsics", "carrying_mul_add"],
```
RyanGlScott, line 1035: Worth adding a test case exercising this new override?

crucible-mir/src/Mir/Trans.hs
spernsteiner: Added comments for the `AKRawPtr` thin pointer / `()` metadata case in 50eb61000

crucible-mir/src/Mir/TransCustom.hs
RyanGlScott: In particular, does calling something like [`u8::carrying_mul_add`](https://doc.rust-lang.org/std/primitive.u8.html#method.carrying_mul_add) suffice to make use of this intrinsic? On Rust 1.91.0, it uses a non-intrinsic-based implementation, but I wonder if things are different with this new nightly.

RyanGlScott (approved):
LGTM, aside from one suggestion.

crucible-mir/src/Mir/TransCustom.hs
spernsteiner: Added a test in 3eb1ac9b54e9ae2edfa97054912341bc1a07e696.  I confirmed via `--print-mir` that it actually calls the `carrying_mul_add` intrinsic.



> On Rust 1.91.0, it uses a non-intrinsic-based implementation

Where are you seeing this?  According to the 1.91.0 sources [here](https://doc.rust-lang.org/1.91.0/src/core/num/uint_macros.rs.html#3066), it should be using the intrinsic.

RyanGlScott: With `rustc 1.91.0-nightly (02c7b1a7a 2025-09-13)`, if I run `crux-mir --print-mir` on the following program:

```rs
#[crux::test]
pub fn f() -> (u8, u8) {
    5u8.carrying_mul_add(2, 0, 0)
}
```

Then I get the following:

<details>

```
FNs
fn core/4cac97e0::intrinsics[0]::carrying_mul_add[0]::_instb5fbeefa9be52936[0]( _1 : u8
, _2 : u8
, _3 : u8
, _4 : u8 ) -> (u8, u8) {
   let mut _0 : (u8, u8);
   bb0: {
      call( core/4cac97e0::intrinsics[0]::fallback[0]::{impl#2}[0]::carrying_mul_add[0]( copy(_1)
      , copy(_2)
      , copy(_3)
      , copy(_4) )
      , (_0, bb1) )
   }
   bb1: {
      return;
   }
}
fn core/4cac97e0::intrinsics[0]::fallback[0]::{impl#2}[0]::carrying_mul_add[0]( _1 : u8
, _2 : u8
, _3 : u8
, _4 : u8 ) -> (u8, u8) {
   let mut _0 : (u8, u8);
   let const _5 : u16;
   let mut _6 : u16;
   let mut _7 : u16;
   let mut _8 : u16;
   let mut _9 : u16;
   let mut _10 : (u16, bool);
   let mut _11 : u16;
   let mut _12 : (u16, bool);
   let mut _13 : u16;
   let mut _14 : (u16, bool);
   let mut _15 : u8;
   let mut _16 : u8;
   let mut _17 : u16;
   let mut _18 : bool;
   bb0: {
      _8 = Cast(Misc, copy(_1), u16);
      _9 = Cast(Misc, copy(_2), u16);
      _10 = copy(_8) *? copy(_9);
      assert move(_10.1) == False -> bb1
   }
   bb1: {
      _7 = move(_10.0);
      _11 = Cast(Misc, copy(_3), u16);
      _12 = copy(_7) +? copy(_11);
      assert move(_12.1) == False -> bb2
   }
   bb2: {
      _6 = move(_12.0);
      _13 = Cast(Misc, copy(_4), u16);
      _14 = copy(_6) +? copy(_13);
      assert move(_14.1) == False -> bb3
   }
   bb3: {
      _5 = move(_14.0);
      _15 = Cast(Misc, copy(_5), u8);
      _18 = const(8) < const(16);
      assert move(_18) == True -> bb4
   }
   bb4: {
      _17 = copy(_5) >> const(8);
      _16 = Cast(Misc, move(_17), u8);
      _0 = Aggregate(tup, [copy(_15), copy(_16)]);
      return;
   }
}
fn core/4cac97e0::num[0]::{impl#6}[0]::carrying_mul_add[0]( _1 : u8
, _2 : u8
, _3 : u8
, _4 : u8 ) -> (u8, u8) {
   let mut _0 : (u8, u8);
   bb0: {
      call( core/4cac97e0::intrinsics[0]::carrying_mul_add[0]::_instb5fbeefa9be52936[0]( copy(_1)
      , copy(_2)
      , copy(_3)
      , copy(_4) )
      , (_0, bb1) )
   }
   bb1: {
      return;
   }
}
fn test/4f1d3389::f[0]() -> (u8, u8) {
   let mut _0 : (u8, u8);
   bb0: {
      call( core/4cac97e0::num[0]::{impl#6}[0]::carrying_mul_add[0]( const(5)
      , const(2)
      , const(0)
      , const(0) )
      , (_0, bb1) )
   }
   bb1: {
      return;
   }
}
ADTs
TRAITs
VTABLEs
LAYOUTs
bool
   size = 1
   align = 1
   fields = Nothing
u8
   size = 1
   align = 1
   fields = Nothing
u16
   size = 2
   align = 2
   fields = Nothing
u32
   size = 4
   align = 4
   fields = Nothing
(u8, u8)
   size = 2
   align = 1
   fields = Just [0,1]
(u16, bool)
   size = 4
   align = 2
   fields = Just [0,2]
fnDef core/4cac97e0::intrinsics[0]::carrying_mul_add[0]::_instb5fbeefa9be52936[0]
   size = 0
   align = 1
   fields = Just []
fnDef core/4cac97e0::intrinsics[0]::fallback[0]::{impl#2}[0]::carrying_mul_add[0]
   size = 0
   align = 1
   fields = Just []
fnDef core/4cac97e0::num[0]::{impl#6}[0]::carrying_mul_add[0]
   size = 0
   align = 1
   fields = Just []
INTRINSICSs
core/4cac97e0::intrinsics[0]::carrying_mul_add[0]::_instb5fbeefa9be52936[0] = IkIntrinsic core/4cac97e0::intrinsics[0]::carrying_mul_add[0]<<u8,u8>>
core/4cac97e0::intrinsics[0]::fallback[0]::{impl#2}[0]::carrying_mul_add[0] = IkItem core/4cac97e0::intrinsics[0]::fallback[0]::{impl#2}[0]::carrying_mul_add[0]<<>>
core/4cac97e0::num[0]::{impl#6}[0]::carrying_mul_add[0] = IkItem core/4cac97e0::num[0]::{impl#6}[0]::carrying_mul_add[0]<<>>
test/4f1d3389::f[0] = IkItem test/4f1d3389::f[0]<<>>
STATICS
```
</details>

Note that it doesn't call the actual intrinsic here, but rather a [fallback implementation (`std::intrinsics::fallback::carrying_mul_add`)](https://doc.rust-lang.org/1.91.0/std/intrinsics/fallback/trait.CarryingMulAdd.html).

spernsteiner: Interesting!  I didn't know intrinsics could have fallback implementations like that.  According to https://doc.rust-lang.org/1.91.1/unstable-book/language-features/intrinsics.html#intrinsics-with-fallback-logic:

> Many intrinsics can be written in pure rust, albeit inefficiently or without supporting some features that only exist on some backends. Backends can simply not implement those intrinsics without causing any code miscompilations or failures to compile.

It seems like we used to take advantage of intrinsic fallbacks but no longer do.  I'm planning to merge this branch anyway, but I've opened https://github.com/GaloisInc/mir-json/issues/291 to track this.

PR #1854, opened 2026-06-22
`.github/dependabot.yml`: Don't recommend minor/patch updates

### LINKED ISSUE #1846: ci: Disable Dependabot updates for minor/patch versions
Dependabot keeps suggesting that we update to new patch versions of our dependencies. However, it is customary in the GHA ecosystem to update major version tags like `vX` to point to the latest minor/patch release like `vX.Y.Z`. We currently pin our actions dependencies to their major version tag. This seems to strike a nice balance between reliability of CI builds and effort spent on updates. Let's configure Dependabot to not suggest these minor/patch version updates.

> I guess Dependabot is just going to keep recommending that we update minor/patch versions even though we explicitly configured it to use the latest release in a specific major version? Rather than keep closing these PRs, perhaps we should configure Dependabot to ignore minor/patch versions. Based on my read of the [GitHub Actions docs](https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference), we'd need to put something like this in `.github/dependabot.yml`:

```yaml
updates:
  - package-ecosystem: "github-actions"
    ...
    ignore:
      - dependency-name: "*"
        update-types:
          - "version-update:semver-minor"
          - "version-update:semver-patch"
```

_Originally posted by @RyanGlScott in https://github.com/GaloisInc/crucible/issues/1845#issuecomment-4649120130_

RyanGlScott:
Fixes #1846.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -8,3 +8,15 @@ updates:
   # https://docs.zizmor.sh/audits/#dependabot-cooldown
   cooldown:
     default-days: 7
+  # We pin github-actions dependencies to major version tags (e.g., vX) so that
+  # we automatically get the latest minor/patch release (e.g., vX.Y.Z) without
+  # needing to constantly update the tag number. For inexplicable reasons,
+  # however, dependabot will sometimes recommend that you switch out major
+  # version tags with explicit minor/patch version tags, which would defeat
+  # the point of using major version tags. As such, we explicitly instruct
+  # dependabot not to recommend minor/patch updates to avoid this.
+  ignore:
+    - dependency-name: "*"
+      update-types:
+        - "version-update:semver-minor"
+        - "version-update:semver-patch"


langston-barrett (approved):

kquick:
Is there a way to tell dependabot to use haskell pvp version interpretation instead of semver?

RyanGlScott:
I'm not quite sure what you're asking here. As far as I can tell, all GitHub actions use SemVer, so I don't see the value in telling dependabot to use Haskell PVP for actions.

kquick:
I was hoping that because the changes noted `...: semver-...` that dependabot would have the notion of other versioning schemes.  Alas, as you noted, it doesn't seem to (https://docs.github.com/en/code-security/reference/supply-chain-security/dependabot-options-reference#update-types-allow is the best reference I could find to what's supported here).

PR #1856, opened 2026-06-25
`Mir.TransCustom`: Add `black_box` override

### LINKED ISSUE #1855: `crucible-mir`: Support the `black_box` intrinsic
[`std::hint::black_box`](https://doc.rust-lang.org/src/core/hint.rs.html#490) is powered by the [`black_box`](https://doc.rust-lang.org/std/intrinsics/fn.black_box.html) intrinsic:

```rs
pub const fn black_box<T>(dummy: T) -> T;
```

Per the documentation for `std::hint::black_box`:

> An identity function that **_hints_** to the compiler to be maximally pessimistic about what `black_box` could do.
> 
> Unlike [`std::convert::identity`](https://doc.rust-lang.org/std/convert/fn.identity.html), a Rust compiler is encouraged to assume that `black_box` can use `dummy` in any possible valid way that Rust code is allowed to without introducing undefined behavior in the calling code. This property makes `black_box` useful for writing code in which certain optimizations are not desired, such as benchmarks.

Among other uses, `black_box` is used in the `ctutils` crate ([here](https://github.com/RustCrypto/utils/blob/2afc28d5bfc4538296d7b21394892e4c46b612a6/ctutils/src/choice.rs#L402)).

For `crucible-mir`'s needs, it would suffice to override `black_box` with the same implementation as `std::convert::identity`.

RyanGlScott:
Fixes #1855.

### DIFF
--- a/crucible-mir/src/Mir/TransCustom.hs
+++ b/crucible-mir/src/Mir/TransCustom.hs
@@ -126,6 +126,7 @@ customOpDefs = Map.fromList $ [
                          , align_of_val
                          , intrinsics_assume
                          , assert_inhabited
+                         , black_box
                          , unlikely
                          , bitreverse
                          , volatile_load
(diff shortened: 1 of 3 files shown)


crucible-mir/src/Mir/TransCustom.hs
```diff
@@ -2177,13 +2178,24 @@ atomic_funcs =
...
 unlikely :: (ExplodedDefId, CustomRHS)
 unlikely = (name, rhs)
     where
         name = ["core", "intrinsics", "unlikely"]
-        rhs _substs = Just $ CustomOp $ \_ ops -> case ops of
-          [op] -> pure op
-          _ -> mirFail $ "bad arguments to intrinsics::unlikely: " ++ show ops
+        rhs = identity_rhs "unlikely"
+
+-- An implementation for intrinsics whose overrides take a single argument and
```
samcowger:
```suggestion
-- | An implementation for intrinsics whose overrides take a single argument and
```

samcowger (approved):

PR #1863, opened 2026-07-20
`crux-mir`: Use `nightly-2026-03-21` consistently in CI

RyanGlScott:
A follow-up to https://github.com/GaloisInc/crucible/pull/1851.


spernsteiner (approved):
I guess this doesn't directly use the mir-json docker image, so there's no need to bring in https://github.com/GaloisInc/mir-json/pull/292 with a submodule bump.

PR #1866, opened 2026-07-29
crux-mir: minor fixes to support flattened aggregates in SAW

spernsteiner:
* Convert `mirRef_overlapsIO` to use `MonadAssert`.  This allows SAW override matching to handle failures, which can happen when the inputs are empty mux trees (or when they're nonempty but their intersection is empty).
* Add `mirAggregate_keysSet :: MirAggregate sym -> IntSet`.  This is convenient for checking that entries are present at all the necessary offsets in some SAW helper functions for manipulating aggregates.


crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -723,6 +720,9 @@ mirAggregate_entries :: sym -> MirAggregate sym -> [(Word, MirAggregateEntry sym
 mirAggregate_entries _sym (MirAggregate _totalSize m) =
   [(fromIntegral off, entry) | (off, entry) <- IntMap.toList m]
 
+mirAggregate_keysSet :: MirAggregate sym -> IntSet
+mirAggregate_keysSet (MirAggregate _ m) = IntMap.keysSet m
```
RyanGlScott, line 724: There's a bit of overlap in functional between `mirAggregate_entries` and `mirAggregate_keysSet`. Would it make sense to combine these into a single function of type `MirAggregateSym sym -> IntMap (MirAggregateEntry sym)`?

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
```diff
@@ -36,13 +36,9 @@ module Mir.Intrinsics.Aggregate
     readMirAggregateWithSymOffset,
     readSubaggregateWithConcreteOffset,
     adjustMirAggregateWithSymOffset,
-    offsetInSpans,
-    foldRuns,
-    addToRun,
-    mkRun,
-    Run (..),
```
RyanGlScott, line 43: This commit is doing more than just exposing `mirAggregate_keysSet`, as it's also un-exporting these definitions. This seems unlikely to affect anyone, but probably worth noting this in the commit message.

RyanGlScott (approved):

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
spernsteiner: I split the export removals into a separate commit: 817d443bc14fa922d3aa6b1aa995933ca6bac166

crucible-mir/src/Mir/Intrinsics/Aggregate.hs
spernsteiner: I like it better this way because it makes it clearer which aspects of the `MirAggregate` implementation other code is currently relying on.  This is in line with the changes @samcowger made in https://github.com/GaloisInc/crucible/pull/1842 to have external code go through accessors instead of manipulating the `MirAggregate`'s inner `IntMap` directly.

PR #1869, opened 2026-08-06
`crucible-mir`: Support mir-json schema version 12

### LINKED ISSUE #1864: Update to match `mir-json` schema version 12 (and also bugs in processing of 11)
Here are some mismatches between the output of `mir-json` (version 12) and how we process it:

  * `FloatKind`: There are more floating point types (F16 and F128 are missing)
  * `Instance`: VtableShim should be VTableShim  (this is also a problem in version 11)
  * `Variant`: field `discr` is not there anymore
  * `Mutability`: `mir-json` does not produce `MutMutable` and `MutImmutable`
  * `StatemntKind`: unhandled `FakeRead`, `Retag`, `PlaceMention`, `AscribeUserType`, `Coverage`,
`BackwardIncompatibleDropHint` (not handling some of these is probably OK, we should check);  `Deinit` is never produced by `mir-json`
  * `PlaceElem`: `Subtype` never produced; not handled `OpaqueCast` and `UnwrapUnsafeBinder`
  * `Rvalue`: unhandled: `WrapUnsafeBinder`;  not prduced: `Len`, `NullaryOp`, `ShallowInitBox`
  * `Operand`: not handled: `RuntimeChecks`
  *  `TerminatorKind`: not handled: `Yield`, `FalseEdge`, `FalseUnwind`, `CoroutineDrop`, `Unsupported` (these are probably OK?)
  * `BorrowKind`: `Unique` is never produced
  * `CastKind`: missing handler (e.g., `subtype`)
  * `Constant`: never produced `initializer`
  * `InlineTy`: missing cases `Pat`, `Error`, `Infer`, `Bound`, `Placeholder`, `Alias`

RyanGlScott:
Also ensure that we are up to date with what `mir-json` does (and does not) generate. This involves removing several `Mir.Mir` constructors that `mir-json` no longer produces and adding new constructors for recently-added `mir-json`  nodes (but not going as far as translating them into Crucible).

Fixes #1864.

### DIFF
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -224,16 +224,15 @@ instance Pretty Rvalue where
     pretty (Ref Mutable b _c) = pretty "&mut" <+> pretty b
     pretty (AddressOf Immut b) = pretty "&raw" <+> pretty b
     pretty (AddressOf Mut b) = pretty "&raw mut" <+> pretty b
-    pretty (Len a) = pretty_fn1 "len" a
     pretty (Cast a b c) = pretty_fn3 "Cast" a b c
     pretty (BinaryOp a b c) = pretty b <+> pretty a <+> pretty c
     pretty (UnaryOp a b) = pretty a <+> pretty b
     pretty (Discriminant a b) = pretty_fn2 "Discriminant" a b
     pretty (Aggregate a b) = pretty_fn2 "Aggregate" a b
     pretty (RAdtAg a) = pretty a
-    pretty (ShallowInitBox ptr ty) = pretty_fn2 "ShallowInitBox" ptr ty
     pretty (CopyForDeref lv) = pretty_fn1 "CopyForDeref" lv
     pretty (ThreadLocalRef a b) = pretty_fn2 "ThreadLocalRef" a b
+    pretty (WrapUnsafeBinder a b) = pretty_fn2 "WrapUnsafeBinder" a b
 
 instance Pretty AdtAg where
   pretty (AdtAg (Adt nm _kind _vs _ _ _ _) i ops _ optField) = case optField of
--- a/crucible-mir/src/Mir/PP.hs
+++ b/crucible-mir/src/Mir/PP.hs
@@ -391,7 +391,6 @@ instance Pretty ConstVal where
     pretty (ConstArray cs)     = list (map pretty cs)
     pretty (ConstRepeat cv i)  = brackets (pretty cv <> semi <+> pretty i)
     pretty (ConstFunction a)   = pr_id a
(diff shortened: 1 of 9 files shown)


crucible-mir/src/Mir/PP.hs
```diff
@@ -250,6 +265,10 @@ instance Pretty TerminatorKind where
       pretty "assert" <+> pretty op <+> pretty "==" <+> pretty expect
                     <+> arrow <+> pretty target1
     pretty InlineAsm = pretty "inlineasm;"
+    pretty Yield = pretty "yield;"
+    pretty FalseEdge = pretty "falseEdge;"
+    pretty FalseUnwind = pretty "falseUnwind;"
+    pretty CoroutineDrop = pretty "coroutine_drop;"
```
samcowger, line 271: Is there a reason we're mixing camel and snake case here?

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1544,6 +1545,7 @@ evalPlaceProj ty (MirPlace tpr ref NoMeta) M.Deref = do
         M.TyRawPtr t _ -> doRef t
         CTyBox t -> doRef t
         _ -> mirFail $ "deref not supported on " ++ show ty
+  | otherwise = mirFail $ "deref not supported on " ++ show meta
```
samcowger: In this commit's spirit, would you consider also converting this to a `case meta of` expression and/or listing out the other two `PtrMetadata` constructors, to trigger a compilation warning/error when/if we change those constructors?

samcowger (approved):

crucible-mir/src/Mir/PP.hs
RyanGlScott: Hah, I actually based these off of how the Rust compiler itself pretty-prints them: see https://github.com/rust-lang/rust/blob/7608eb7b07eaf93f16d7cf5bcb2098eca87503df/compiler/rustc_middle/src/mir/pretty.rs#L982-L1021. (I have no idea why Rust's pretty-printing for types mixes camelCase and snake_case, but it does.)

That being said, we don't have to rigidly follow these conventions, as `crucible-mir`'s pretty-printer is only used for display purposes. If you'd like, I could change everything to use snake_case uniformly.

samcowger: How odd. I think it's probably good for us to follow `rustc`'s lead in general, so I'm fine leaving this as-is.

crucible-mir/src/Mir/Trans.hs
RyanGlScott: Yes, good call. I've done this in a follow-up commit (fdd931de4779fa4f2ee2815f823a58834bb0deef).

Issue #1865, opened 2026-07-29
`crucible-mir` crashes when simulating infinite floating-point values


RyanGlScott:
`crux-mir` crashes on the following program:

```rs
#[crux::test]
pub fn f() -> bool {
    let f1 = f32::INFINITY;
    let f2 = f32::INFINITY;
    f1 == f2
}
```
```
$ cabal -v0 run exe:crux-mir -- test.rs
test test/6f913a18::f[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/6f913a18::f[0] counterexamples ----
[Crux] Found counterexample for verification goal
[Crux]   test.rs:5:5: 5:7: error: in test/6f913a18::f[0]
[Crux]   Translation error in test/6f913a18::f[0]: cannot parse float constant: "+Inf"
[Crux]   Context:
[Crux]     test.rs:5:5: 5:7: test/6f913a18::f[0]

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
[Crux] Overall status: Invalid.
```

RyanGlScott:
`crucible-mir`'s handling of special floating-point values in general seems rather broken. While the program below doesn't cause `crucible-mir` to crash, it does cause other oddities:

```rs
#[crux::test]
pub fn f() -> f32 {
    f32::NAN
}

#[crux::test]
pub fn g() -> bool {
    f32::NAN.is_nan()
}
```
```
$ cabal -v0 run exe:crux-mir -- test.rs
test test/2686f755::f[0]: returned 269653970229347386159395778618353710042696546841345985910145121736599013708251444699062715983611304031680170819807090036488184653221624933739271145959211186566651840137298227914453329401869141179179624428127508653257226023513694322210869665811240855745025766026879447359920868907719574457253034494436336205824 % 1, ok
test test/2686f755::g[0]: returned false, ok

[Crux-MIR] ---- FINAL RESULTS ----
[Crux] All goals discharged through internal simplification.
[Crux] Overall status: Valid.
```

The output of `g` is `false`, but it should actually be `true`. We can see why this happens in the output of `f`, which incorrectly claims that `f32::NAN` computes an enormous rational value.

RyanGlScott:
[...]

PR #1870, opened 2026-08-07
`crux-mir`: `symbolic_f{16,32,64,128}` overrides, `slice::fill` test case

RyanGlScott:
This bumps the `mir-json` submodule to bring in the changes from https://github.com/GaloisInc/mir-json/pull/303 and https://github.com/GaloisInc/mir-json/pull/300 and adds corresponding test cases. The former change requires adding `crux-mir` overrides to support the newly added `symbolic_f{16,32,64,128}` functions.

### DIFF
--- a/crux-mir/test/symb_eval/float/symbolic.rs
+++ b/crux-mir/test/symb_eval/float/symbolic.rs
@@ -1,0 +1,31 @@
+#![feature(f16)]
+#![feature(f128)]
+
+extern crate crucible;
+use crucible::*;
+
+fn symbolic_test<S: PartialEq + Symbolic>() -> S {
+    let x = S::symbolic("x");
+    crucible_assert!(x == x);
+    x
+}
+
+#[crux::test]
+fn f16_test() -> f16 {
+    symbolic_test()
+}
+
+#[crux::test]
+fn f32_test() -> f32 {
+    symbolic_test()
+}
+
+#[crux::test]
+fn f64_test() -> f64 {
+    symbolic_test()
+}
+
+#[crux::test]
+fn f128_test() -> f128 {
+    symbolic_test()
+}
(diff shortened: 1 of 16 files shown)


crux-mir/test/symb_eval/float/symbolic.rs
```diff
@@ -0,0 +1,31 @@
+#![feature(f16)]
+#![feature(f128)]
+
+extern crate crucible;
+use crucible::*;
+
+fn symbolic_test<S: PartialEq + Symbolic>() -> S {
```
podhrmic, line 7: Would it be appropriate to have another, more interesting test that actually requires some computation with the float?

E.g. `crucible_assert( (x+x) == 2x)` or something like that, properly handling the edge cases (zero, NaN, maybe something else)? 

Or should such test be in a different directory?

podhrmic (commented):
Thanks for getting this done so quickly! My only comment is about the depth of the float test (see above).

crux-mir/test/symb_eval/float/symbolic.rs
RyanGlScott: My inclination would be to wait until https://github.com/GaloisInc/crucible/issues/1865 is addressed before doing this. Currently, `crucible-mir`'s representation of floats is somewhat unsound, and a theorem like `x+x == 2*x` actually _shouldn't_ hold if `x` is NaN, but `crucible-mir` won't detect this at the moment. We'd need to switch `crucible-mir`'s representation from `RealValType` (i.e., algebraic reals) to `FloatType` (i.e., IEEE floats) in order for `crucible-mir` to be able to do this properly.

samcowger: Given that `crucible-mir`'s floating-point machinery is already unsound, I wonder whether we should avoid adding more interfaces for it before we fix it, perhaps via #1865. What do you think?

RyanGlScott: Coincidentally, I am working on a fix for #1865 as we speak. I'm happy to submit another PR (based on this one) and merge everything en masse once we're happy with it.

RyanGlScott: See https://github.com/GaloisInc/crucible/pull/1871.

PR #1871, opened 2026-08-07
`{crucible,crux}-mir`: Improve handling of floating point

### LINKED ISSUE #1865: `crucible-mir` crashes when simulating infinite floating-point values

RyanGlScott:
This changes the Crucible representation of primitive floating point types (e.g., `f32` and `f64`) from `RealValRepr` (i.e., rationals) to `FloatRepr`, the latter of which can properly handle the various nuances of IEEE-754 floats. I demonstrate the new capabilities in a slew of new test cases.

Fixes https://github.com/GaloisInc/crucible/issues/1865.

----

(Reviewer's note: this is based on top of https://github.com/GaloisInc/crucible/pull/1870. Only the last three commits in this series are unique to this PR.)

### DIFF
--- a/crux-llvm/src/Crux/LLVM/Simulate.hs
+++ b/crux-llvm/src/Crux/LLVM/Simulate.hs
@@ -195,7 +195,7 @@ simulateLLVMFile ::
   LLVMOptions ->
   Crux.SimulatorCallbacks msgs st Crux.Types.CruxSimulationResult
 simulateLLVMFile llvm_file llvmOpts =
-  Crux.SimulatorCallbacks $
+  Crux.SimulatorCallbacks $ \_fm ->
     do bbMapRef <- newIORef (Map.empty :: LLVMAnnMap sym)
        let ?recordLLVMAnnotation =
              \callStack an bb ->
--- a/crux/CHANGELOG.md
+++ b/crux/CHANGELOG.md
@@ -6,6 +6,7 @@
   `colorOptionsL`, `simVerboseL`, `printFailuresL`, `quietModeL` (from
   `Crux.Config.Common`), and `colorOptionsL`, `noColorsErrL`, `noColorsOutL`
   (from `Crux.Config.Load`).
+- Add a `FloatModeRepr` argument to `SimulatorCallbacks`.
 
 # 0.9 -- 2026-01-29
 
--- a/crux/CHANGELOG.md
+++ b/crux/CHANGELOG.md
@@ -7,6 +7,8 @@
   `Crux.Config.Common`), and `colorOptionsL`, `noColorsErrL`, `noColorsOutL`
   (from `Crux.Config.Load`).
 - Add a `FloatModeRepr` argument to `SimulatorCallbacks`.
+- Add `baseFreshFloatOverride` and `baseFreshFloatOverride'` to
+  `Crux.Overrides`.
 
 # 0.9 -- 2026-01-29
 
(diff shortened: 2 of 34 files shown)


crux/CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
   `colorOptionsL`, `simVerboseL`, `printFailuresL`, `quietModeL` (from
   `Crux.Config.Common`), and `colorOptionsL`, `noColorsErrL`, `noColorsOutL`
   (from `Crux.Config.Load`).
+- Add a `FloatModeRepr` argument to `SimulatorCallbacks`.
+- Add `baseFreshFloatOverride` and `baseFreshFloatOverride'` to
+  `Crux.Overrides`.
```
RyanGlScott, line 11: Although I remembered to update the `crux` changelog, I forgot to update the `crucible-mir` and `crux-mir` changelogs to mention the changes to floating-point representation. I'll make sure to do that before landing.

crux-llvm/src/Crux/LLVM/Simulate.hs
```diff
@@ -195,7 +195,7 @@ simulateLLVMFile ::
   LLVMOptions ->
   Crux.SimulatorCallbacks msgs st Crux.Types.CruxSimulationResult
 simulateLLVMFile llvm_file llvmOpts =
-  Crux.SimulatorCallbacks $
+  Crux.SimulatorCallbacks $ \_fm ->
```
spernsteiner, line 198: Does crux-llvm (and crucible-jvm/crucible-go) not support floats?  How have we been able to get by without this float mode argument until now?

spernsteiner (approved):

podhrmic (approved):

crux-llvm/src/Crux/LLVM/Simulate.hs
RyanGlScott: `crucible-{jvm,go,llvm}` do support floats, and moreover, they do so in a way that respects the `FloatMode`. The reason why we need the `FloatMode` here in `crux-mir` (and not in, say, `crux-llvm`) is a bit subtle. Picking `crucible-llvm` as a specific example, `crucible-llvm`'s implements all floating-point operations by dispatching through methods of the [`IsInterpretedFloatExprBuilder`](https://hackage-content.haskell.org/package/what4-1.7.3/docs/What4-InterpretedFloatingPoint.html#t:IsInterpretedFloatExprBuilder) class, which is carefully designed to have a `FloatMode`-agnostic API. The only time when the `FloatMode` is consulted is during simulation, which happens behind the scenes in `crucible`'s internals. As such, `crucible-llvm`'s translation is able to get away without needing to explicitly pass a `FloatMode` around.

Actually, most of `crucible-mir`'s translation is agnostic to the `FloatMode` as well, as it also uses the `IsInterpretedFloatExprBuilder` class for most operations. The one exception is the machinery that `crux-mir` uses to display the return values of functions marked as `crux::test`s, where it needs to concretize values if that is possible. This beyond the scope of what `IsInterpretedFloatExprBuilder` can do, and it's not entirely clear to me if it's even possible to add such a method to that class with the way the types work out. As far as I can tell, this aspect of `crux-mir` is unique---none of the other Crux backends do anything quite like it.

PR #1876, opened 2026-08-11
crucible-mir: support subslicing projections on arrays

### LINKED ISSUE #1494: `crucible-mir`: support `Subslice` projections on arrays
#1491 implements `Subslice` projections on slices, but Crucible's representation of `MirReference`s to arrays is not expressive enough to implement the same projection on arrays. A Crucible reference to an array is a reference to the entire array, and cannot have `mirRef_offset` applied to it - unlike a slice, a Crucible reference to which is represented as a reference to its first element.

A couple possible ways past this:
- Make Crucible array references actually point to the first element of the array, mimicking Crucible's approach to slices and Rust itself's approach to arrays. Then, they'd be amenable to `mirRef_offset`.
- Implement a new reference path extension, something like `MirRef_Subarray`. Such a path extension would behave somewhat like `MirRef_Index`, but instead of offering read/write access to a single element of the parent array, it would offer read/write access to a subarray of elements of the parent array.

samcowger:
Fixes #1494.

### DIFF
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -862,7 +862,7 @@ typeOfProj elm baseTy = case elm of
     Index{}              -> peelIdx baseTy
     ConstantIndex{}      -> peelIdx baseTy
     Downcast i           -> TyDowncast baseTy i   --- TODO: check this
-    Subslice{}           -> TySlice (peelIdx baseTy)
+    Subslice f t e       -> subslice baseTy f t e
     OpaqueCast t         -> t
     UnwrapUnsafeBinder t -> t
   where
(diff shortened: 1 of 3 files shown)


crucible-mir/src/Mir/Mir.hs
```diff
@@ -878,6 +878,16 @@ typeOfProj elm baseTy = case elm of
...
     peelIdx t             = t
 
+    subslice :: Ty -> Int -> Int -> Bool -> Ty
+    subslice (TyArray elemTy len) fromIndex toIndex fromEnd =
+      let firstIndex = fromIndex
+          lastIndex
+            | fromEnd = len - toIndex
+            | otherwise = toIndex
+          newLen = lastIndex - firstIndex
+      in TyArray elemTy newLen
```
RyanGlScott, line 888: I assume that we don't need a `TyStr` case here because you can't write patterns like `[_, rest @ ..]` on type `&str`?

RyanGlScott (approved):

crucible-mir/src/Mir/Mir.hs
samcowger: That's right, you can't decompose `&str`s with these patterns.

PR #1880, opened 2026-08-17
`crucible-mir`: Support translating the `Cmp` binop

### LINKED ISSUE #1765: `crucible-mir`: Support translating `Cmp` binary operations
Recent versions of MIR include a binary [`Cmp`](https://doc.rust-lang.org/1.91.0/nightly-rustc/rustc_middle/mir/enum.BinOp.html#variant.Cmp) operator, which performs a three-way comparison that returns an `Ordering`. The easiest way to produce MIR code that uses a `Cmp` binop is by calling the [`three_way_compare`](https://doc.rust-lang.org/std/intrinsics/fn.three_way_compare.html) intrinsic, as seen in the following program:

```rs
#![feature(core_intrinsics)]

pub fn f(x: u8, y: u8) -> std::cmp::Ordering {
    std::intrinsics::three_way_compare(x, y)
}

#[crux::test]
pub fn g() -> std::cmp::Ordering {
    f(0, 1)
}
```

`crucible-mir` does not currently support translating `Cmp` binops, so `crux-mir` will fail on the program above:

```
test test/08fc691d::g[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/08fc691d::g[0] counterexamples ----
[Crux]   test.rs:4:5: 4:45: error: in test/08fc691d::f[0]
[Crux]   Translation error in test/08fc691d::f[0]: No translation for binop: Cmp Just Unsigned for BVRepr 8 and BVRepr 8
[Crux]   Context:
[Crux]     test.rs:4:5: 4:45: test/08fc691d::f[0]
[Crux]     test.rs:9:5: 9:12: test/08fc691d::g[0]

[Crux] Goal status:
[Crux]   Total: 1
[Crux]   Proved: 0
[Crux]   Disproved: 1
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

We should add `Cmp` support. Perhaps we should also add an override for `three_way_compare` just to be on the safe side (it is unclear to me if `rustc` will always desugar applications of `three_way_compare` to `Cmp` binops).

Fixing this would allow removing a `mir-json` patch introduced in https://github.com/GaloisInc/mir-json/commit/16a6701ac35c0a714163946daf7192ec0a59c908.

RyanGlScott:
This is necessary to support invoking the `three_way_compare` intrinsic, which is used in the implementations of `partial_cmp` and `cmp` for primitive integer and char types. Eventually, this will allow us to remove the `mir-json` patch introduced in https://github.com/GaloisInc/mir-json/commit/16a6701ac35c0a714163946daf7192ec0a59c908.

Fixes #1765.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -663,6 +663,8 @@ evalBinOp bop mat me1 me2 =
               (M.Ge, Just M.Unsigned) -> return (MirExp (C.BoolRepr) (S.app $ E.BVUle n e2 e1), noOverflow)
               (M.Ge, Just M.Signed)   -> return (MirExp (C.BoolRepr) (S.app $ E.BVSle n e2 e1), noOverflow)
 
+              (M.Cmp, Just arithTy) -> threeWayCompare arithTy n e1 e2
+
               (M.Ne, _) -> return (MirExp (C.BoolRepr) (S.app $ E.Not $ S.app $ E.BVEq n e1 e2), noOverflow)
               (M.Beq, _) -> return (MirExp (C.BoolRepr) (S.app $ E.BVEq n e1 e2), noOverflow)
               _ -> mirFail $ "No translation for binop: " ++ show bop ++ " " ++ show mat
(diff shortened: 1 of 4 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -774,6 +776,55 @@ evalBinOp bop mat me1 me2 =
...
+            buildOrdering discr = do
+              MirExp orderingTpr ordering <- buildEnum orderingAdt discr []
+              Refl <- expectEnumOrFail orderingDiscrTpr orderingVariantsCtx orderingTpr
+              pure ordering
+        let bvLt = case arithTy of
+                     M.Unsigned -> E.BVUlt
+                     M.Signed   -> E.BVSlt
+
+        -- We only @x < y@ and @x == y@ below, and if neither are true, then
+        -- @x < y@ must hold. We are allowed to assume this because the Cmp
```
```suggestion
        -- @x > y@ must hold. We are allowed to assume this because the Cmp
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -774,6 +776,55 @@ evalBinOp bop mat me1 me2 =
...
+                (R.Expr MIR s (RustEnumType orderingDiscrTp orderingVariantsCtx))
+            buildOrdering discr = do
+              MirExp orderingTpr ordering <- buildEnum orderingAdt discr []
+              Refl <- expectEnumOrFail orderingDiscrTpr orderingVariantsCtx orderingTpr
+              pure ordering
+        let bvLt = case arithTy of
+                     M.Unsigned -> E.BVUlt
+                     M.Signed   -> E.BVSlt
+
+        -- We only @x < y@ and @x == y@ below, and if neither are true, then
```
```suggestion
        -- We only check @x < y@ and @x == y@ below, and if neither are true, then
```

samcowger (approved):

PR #1882, opened 2026-08-18
`crucible-mir`: Partial support for raw pointer comparisons

### LINKED ISSUE #311: Regression on memchr
https://github.com/GaloisInc/mir-json/commit/c363a8c85f013a87f73a204d4e7b09950ccb353a

This commit was lost in a recent update to a new rustc, but we still don't support the feature.

#277 has a discussion about this specific change. We couldn't figure out how to reproduce the problem. In my particular case it's related to a static initializer that fails once again after updating to the latest saw-suite

RyanGlScott:
Rust implements raw pointer comparisons by comparing the pointers' addresses, but `crucible-mir` references lack addresses. Previously, attempting to perform a raw pointer comparison would result in an immediate translation error in `crucible-mir`, but we now implement a partial approach that correctly handles cases where the two pointers are derived from the same allocation. (If the pointers are not derived from the same allocation, then translation fails as before.)

The primary motivation behind this improvement is being able to simulate the generic implementation of the `memchr` crate. This PR also bumps the `mir-json` submodule to bring in the changes from https://github.com/GaloisInc/mir-json/pull/312 and adds a test case confirming that `memchr` can now be simulated. As such, this PR fixes https://github.com/GaloisInc/mir-json/issues/311.

### DIFF
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -573,7 +573,64 @@ transBinOp bop op1 op2 = do
             elemSize <- tySizeM elemTy
             newRef <- mirRef_offsetWrap e1 e2 elemSize
             pure $ MirExp MirReferenceRepr newRef
+        -- Raw pointer inequality operations. Note that pointer /equality/
+        -- operations are included elsewhere (in evalBinOp).
+        Lt
+          | MirExp MirReferenceRepr e1 <- me1
+          , MirExp MirReferenceRepr e2 <- me2 -> do
+            offset <- evalPtrOffset e1 e2
+            let offsetNeg = R.App $ E.BVSlt (knownNat @SizeBits) offset zeroIsize
+            pure $ MirExp C.BoolRepr offsetNeg
+        Le
+          | MirExp MirReferenceRepr e1 <- me1
+          , MirExp MirReferenceRepr e2 <- me2 -> do
+            offset <- evalPtrOffset e1 e2
+            let offsetNonPos = R.App $ E.BVSle (knownNat @SizeBits) offset zeroIsize
+            pure $ MirExp C.BoolRepr offsetNonPos
+        Gt
+          | MirExp MirReferenceRepr e1 <- me1
+          , MirExp MirReferenceRepr e2 <- me2 -> do
+            offset <- evalPtrOffset e1 e2
+            let offsetPos = R.App $ E.BVSlt (knownNat @SizeBits) zeroIsize offset
+            pure $ MirExp C.BoolRepr offsetPos
+        Ge
+          | MirExp MirReferenceRepr e1 <- me1
+          , MirExp MirReferenceRepr e2 <- me2 -> do
+            offset <- evalPtrOffset e1 e2
+            let offsetNonNeg = R.App $ E.BVSle (knownNat @SizeBits) zeroIsize offset
+            pure $ MirExp C.BoolRepr offsetNonNeg
         _ -> fst <$> evalBinOp bop mat me1 me2
+  where
+    zeroIsize :: R.Expr MIR s IsizeType
+    zeroIsize = R.App $ isizeLit 0
+
+    -- Compute the offset (in bytes) between two raw pointers. If the two
+    -- pointers are not derived from the same allocation, raise an error.
+    --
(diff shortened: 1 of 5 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -573,7 +574,62 @@ transBinOp bop op1 op2 = do
             elemSize <- tySizeM elemTy
             newRef <- mirRef_offsetWrap e1 e2 elemSize
             pure $ MirExp MirReferenceRepr newRef
+        Lt
+          | MirExp MirReferenceRepr e1 <- me1
+          , MirExp MirReferenceRepr e2 <- me2 -> do
+            offset <- evalPtrOffset e1 e2
+            let zeroIsize = R.App $ E.BVLit (knownNat @SizeBits) $ BV.mkBV (knownNat @SizeBits) 0
```
samcowger: You might consider extracting `zeroIsize` to be `where`-bound. Also, there exists a utility function, `Mir.Intrinsics.Size.usizeLit`, which is in scope here and could replace `BV.mkBV (knownNat @SizeBits)`.

crux-mir/test/conc_eval/ptr/cmp_same_allocation.rs
```diff
@@ -0,0 +1,31 @@
+// A test exercising the subset of raw pointer comparisons that crucible-mir
+// currently supports.
```
samcowger, line 2: I think it would be worthwhile to also include an xfail test for comparisons of pointers derived from different allocations, to better delineate what we do and don't support. Perhaps:
```rs
fn crux_test() {
    let a: [u32; 3] = [1, 2, 3];
    let a_s: &[u32] = &a;
    let a_p: *const u32 = a_s.as_ptr();

    let b: [u32; 3] = [4, 5, 6];
    let b_s: &[u32] = &b;
    let b_p: *const u32 = b_s.as_ptr();

    assert!(b_p < a_p || b_p >= a_p);
}
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -573,7 +574,62 @@ transBinOp bop op1 op2 = do
...
+    evalPtrOffset e1 e2 = do
+        elemTy <- case typeOf op1 of
+            TyRawPtr ty _mut -> pure ty
+            t -> mirFail $
+              "Pointer comparison on expression with unexpected type: " <>
+              show (pretty t)
+        elemSize <- tySizeM elemTy
+        maybeOffset <- mirRef_tryOffsetFrom e1 e2 elemSize
+        let errMsg = R.App $ E.StringLit $ fromString $
+              "Pointer comparison (" ++ show bop ++ ") between pointers not in same allocation"
```
samcowger: This will print e.g. `...comparison (Lt) between...`, but I think `...comparison (<) between...` would look neater:
```suggestion
              "Pointer comparison (" ++ show (pretty bop) ++ ") between pointers not in same allocation"
```

crucible-mir/src/Mir/Trans.hs
```diff
@@ -573,7 +574,62 @@ transBinOp bop op1 op2 = do
...
+            let zeroIsize = R.App $ E.BVLit (knownNat @SizeBits) $ BV.mkBV (knownNat @SizeBits) 0
+            let offsetPos = R.App $ E.BVSlt (knownNat @SizeBits) zeroIsize offset
+            pure $ MirExp C.BoolRepr offsetPos
+        Ge
+          | MirExp MirReferenceRepr e1 <- me1
+          , MirExp MirReferenceRepr e2 <- me2 -> do
+            offset <- evalPtrOffset e1 e2
+            let zeroIsize = R.App $ E.BVLit (knownNat @SizeBits) $ BV.mkBV (knownNat @SizeBits) 0
+            let offsetNonNeg = R.App $ E.BVSle (knownNat @SizeBits) zeroIsize offset
+            pure $ MirExp C.BoolRepr offsetNonNeg
```
samcowger, line 601: Maybe comment here that equality and non-equality are checked in `evalBinOp`?

samcowger (approved):

RyanGlScott:
I've pushed a separate commit which bumps the `mir-json` submodule to include the changes from https://github.com/GaloisInc/mir-json/pull/312 and adds a `memchr`-related `crux-mir` test case.

PR #1886, opened 2026-08-19
Update two support `mir-json` evaluating static initializers.

yav:
This bumps the expected `mir-json` schema to 13.

The main change is handling the values in union constants.

### DIFF
--- a/crucible-mir/src/Mir/Mir.hs
+++ b/crucible-mir/src/Mir/Mir.hs
@@ -719,8 +719,7 @@ data Static   = Static {
     _sName          :: DefId            -- ^ name of fn that initializes this static
   , _sTy            :: Ty
   , _sMutable       :: Bool             -- ^ true for "static mut"
-  , _sConstVal      :: Maybe ConstVal   -- ^ 'Just' if this static is initialized
-                                        -- with a constant value. 'Nothing' otherwise.
+  , _sConstVal      :: ConstVal         -- ^ Initial value
   }
   deriving (Show, Eq, Ord, Generic)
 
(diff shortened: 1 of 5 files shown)


RyanGlScott (commented):
Nice! Two requests:

* Can you refer back to https://github.com/GaloisInc/mir-json/pull/304 in the commit message?
* Can you update the definition of `Static` here to use `ConstVal` instead of `Maybe ConstVal`?

  https://github.com/GaloisInc/crucible/blob/4042fa45e5fa01a9356bd7724bc5a4c0b30ac1a6/crucible-mir/src/Mir/Mir.hs#L722-L723

crucible-mir/src/Mir/Trans.hs
```diff
@@ -3432,62 +3408,34 @@ transStatics ::
...
+                              ++ ", expected " ++ show repr ++ ", got " ++ show constty'
+
+             
           Nothing -> error $ "BUG: cannot find global for " ++ fmt staticName
 
   -- TODO: make the name of the static initialization function configurable
   let initName = FN.functionNameFromText "static_initializer"
   initHandle <- FH.mkHandle' halloc initName Ctx.empty C.UnitRepr
   let allStatics :: [Static]
       allStatics = Map.elems (colState ^. collection.statics)
```
RyanGlScott: I think we could change the name "`allStatics`" to just "`statics`" now. The name "`allStatics`" was given to distinguish it from `constValStatics` and `nonConstValStatics` below, but now that the latter two are gone, there isn't a distinction to be made.

crucible-mir/src/Mir/Mir.hs
RyanGlScott: This documentation is now out of date after the changes in this PR.

RyanGlScott (approved):
Note that the commit message references #304, which GitHub considers to be different from https://github.com/GaloisInc/mir-json/pull/304.

PR #1887, opened 2026-08-21
`crucible-mir`: Split the `Misc` `CastKind`

### LINKED ISSUE #1223: `crucible-mir`: Split out the `Misc` `CastKind` into more precise constructors

RyanGlScott:
Previously, `Misc` covered many separate `CastKind`s, which made it difficult to tell which `crucible-mir` casts correspond to which MIR operations. This splits `Misc` into multiple data constructors, one for each distinct cast kind. Note that we coalesce `ArrayToPointer` and `MutToConstPointer` into `PtrToPtr`, as `rustc` itself normalizes the former to the latter.

Fixes #1223.


crucible-mir/src/Mir/Trans.hs
```diff
@@ -1148,16 +1131,22 @@ evalCast' ck ty1 e ty2  = do
...
         | Just adt1 <- findAdt' col an1
         , Just fieldTy1 <- reprTransparentFieldTy col adt1
-        -> evalCast' M.Misc (M.TyRawPtr fieldTy1 m1) e ty2
+        -> evalCast' M.PtrToPtr (M.TyRawPtr fieldTy1 m1) e ty2
+
+      -- Pointer-to-pointer casts to the same type are no-ops. Usually, rustc
+      -- will optimize away these sorts of trivial casts, but they can
+      -- sometimes arise when compiling instantiations of polymorphic functions
+      -- that perform pointer-to-pointer casts.
+      (M.PtrToPtr,a,b) | a == b -> return e
```
samcowger: Nit:
```suggestion
      (M.PtrToPtr, a, b) | a == b -> return e
```

samcowger (approved):

crucible-mir/src/Mir/Trans.hs
RyanGlScott: Done in 0650905acbb77edd6e2ee52f60fd53935f53c70e.

crucible-mir/src/Mir/Trans.hs
```diff
@@ -1148,16 +1131,22 @@ evalCast' ck ty1 e ty2  = do
...
 
       -- Arbitrary pointer-to-pointer casts are allowed as long as the source
       -- and destination *pointer* types have the same Crucible representation
       -- (i.e. both MirReferenceRepr, or both MirSliceRepr, or both DynRefRepr).
       -- This is similar to calling `transmute`.
-      (M.Misc, M.TyRawPtr _ _, M.TyRawPtr _ _)
+      (M.PtrToPtr, M.TyRawPtr _ _, M.TyRawPtr _ _)
          | ty1 == ty2 -> return e
          | tyToRepr col ty1 == tyToRepr col ty2 -> return e
```
spernsteiner, line 1150: `ty1 == ty2` case is redundant with the `a == b` case above

spernsteiner (approved):
I see the cases for casting C-style enums to ints have been removed.  Looks like rustc now handles these by emitting a `Discriminant` operation followed by a cast, rather than a direct cast from the enum to int.

RyanGlScott:
> I see the cases for casting C-style enums to ints have been removed. Looks like rustc now handles these by emitting a `Discriminant` operation followed by a cast, rather than a direct cast from the enum to int.

Quite so! I made sure to mention this in the commit message for https://github.com/GaloisInc/crucible/pull/1887/changes/89a0d1279b354c3028428cd4b0dae5adfb69ebcc.

Also, it's worth noting that this behavior is tested here in the `test/conc_eval/enum/cmp.rs` test case:

https://github.com/GaloisInc/crucible/blob/4042fa45e5fa01a9356bd7724bc5a4c0b30ac1a6/crux-mir/test/conc_eval/enum/cmp.rs#L28-L36

crucible-mir/src/Mir/Trans.hs
RyanGlScott: I've removed this case in 1afcf42ad160cd97b78ddfeafb46c64f1b5e4baf.

PR #1890, opened 2026-08-25
Leave unsupported statics uninitialized

### LINKED ISSUE #1888: Do something with Unsupported in ConstVal
Currently, this is a parse error when we import the JSON, but we should probably do something a little lazier (i.e., parse it as `Unsupported`, and only cause an error if someone uses it).

yav:
Fixes #1888

### DIFF
--- a/crucible-mir/src/Mir/JSON.hs
+++ b/crucible-mir/src/Mir/JSON.hs
@@ -627,6 +627,8 @@ instance FromJSON ConstVal where
               <*> v .: "trait_id"
               <*> v .: "vtable"
 
+        Just (String "unsupported") -> pure ConstUnsupported
+
         o -> do
             fail $ "parseJSON - bad rendered constant kind: " ++ show o
 
--- a/crucible-mir/src/Mir/Trans.hs
+++ b/crucible-mir/src/Mir/Trans.hs
@@ -3405,13 +3405,17 @@ transStatics colState halloc = do
                 constval = static ^. sConstVal
                 constty = static ^. sTy
             Some tpr <- tyToReprM constty
-            MirExp constty' constval' <- transConstVal constty (Some tpr) constval
-            case testEquality repr constty' of
-              Just Refl -> G.writeGlobal g constval'
-              Nothing -> error $ "BUG: invalid type for constant initializer " ++ fmt staticName
-                              ++ ", expected " ++ show repr ++ ", got " ++ show constty'
+            case constval of
+              -- If the initializer is unsupported, leave it uninitialized
+              ConstUnsupported -> pure ()
+              _ ->
+                do
+                  MirExp constty' constval' <- transConstVal constty (Some tpr) constval
+                  case testEquality repr constty' of
+                    Just Refl -> G.writeGlobal g constval'
+                    Nothing -> error $ "BUG: invalid type for constant initializer " ++ fmt staticName
+                                    ++ ", expected " ++ show repr ++ ", got " ++ show constty'
 
-             
           Nothing -> error $ "BUG: cannot find global for " ++ fmt staticName
 
   -- TODO: make the name of the static initialization function configurable
--- a/crucible-mir/CHANGELOG.md
+++ b/crucible-mir/CHANGELOG.md
@@ -1,7 +1,7 @@
 # next
 
 This release supports [version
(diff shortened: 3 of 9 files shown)


crucible-mir/src/Mir/Trans.hs
```diff
@@ -3405,13 +3405,17 @@ transStatics colState halloc = do
...
                 constty = static ^. sTy
             Some tpr <- tyToReprM constty
-            MirExp constty' constval' <- transConstVal constty (Some tpr) constval
-            case testEquality repr constty' of
-              Just Refl -> G.writeGlobal g constval'
-              Nothing -> error $ "BUG: invalid type for constant initializer " ++ fmt staticName
-                              ++ ", expected " ++ show repr ++ ", got " ++ show constty'
+            case constval of
+              -- If the initializer is unsupported, leave it uninitialized
+              ConstUnsupported -> pure ()
```
RyanGlScott, line 3410: Please include a `crux-mir` test case that exercises this new code path. Something like this should suffice, I think:

```rs
pub union Mix {
  f1: (bool, u8),
  f2: (u8, bool),
}

// Z0 has a value which is half from one of the union and half from the other
pub static Z0: Mix = {
    let mut u = Mix { f1: (true, 99) };
    u.f2.0 = 99; // overwrite through *the other* union variant, but only half the value
    u
};

#[cfg_attr(crux, crux::test)]
pub fn crux_test() -> (u8, bool) {
    unsafe { Z0.f2 }
}

pub fn main() {
}
```

-----

On a related note, I think the error message that you get from this program is a bit confusing:

```
test test/51c6de32::crux_test[0]: [Crux] Attempting to prove verification conditions.
FAILED

failures:

---- test/51c6de32::crux_test[0] counterexamples ----
[Crux]   test.rs:15:14: 15:19: error: in test/51c6de32::crux_test[0]
[Crux]   Attempted to read uninitialized reference cell
[Crux]   Context:
[Crux]     test.rs:15:14: 15:19: test/51c6de32::crux_test[0]
[Crux]   test.rs:15:14: 15:19: error: in test/51c6de32::crux_test[0]
[Crux]   Context:
[Crux]     test.rs:15:14: 15:19: test/51c6de32::crux_test[0]

[Crux] Goal status:
[Crux]   Total: 2
[Crux]   Proved: 0
[Crux]   Disproved: 2
[Crux]   Incomplete: 0
[Crux]   Unknown: 0
```

It's not very clear that `Attempted to read uninitialized reference cell` arises because one is trying to simulate an unsupported constant. In situations like these, I think it would help if `crucible-mir` were to emit some kind of warning beforehand. Could you open an issue about implementing this warning? (This will also be relevant for https://github.com/GaloisInc/crucible/issues/1534, which describes another possible way in which statics may fail to translate.)

crucible-mir/src/Mir/JSON.hs
```diff
@@ -627,6 +627,8 @@ instance FromJSON ConstVal where
               <*> v .: "trait_id"
               <*> v .: "vtable"
 
+        Just (String "unsupported") -> pure ConstUnsupported
```
RyanGlScott, line 630: A bit belated, but please make sure to update the `mir-json` schema version numbers in the following places:

* https://github.com/GaloisInc/crucible/blob/47d8d8519e925c3c6816f1f9030dab1d42164132/crux-mir/README.md#L51-L53
* https://github.com/GaloisInc/crucible/blob/47d8d8519e925c3c6816f1f9030dab1d42164132/crux-mir/CHANGELOG.md#L3-L5
* https://github.com/GaloisInc/crucible/blob/47d8d8519e925c3c6816f1f9030dab1d42164132/crucible-mir/CHANGELOG.md#L3-L5

yav: Ah, yes, I'll update these.  I wonder if there is some clever way to centralize this for future edits, seems easy to forget all the places.

RyanGlScott (approved):

crucible-mir/src/Mir/Trans.hs
RyanGlScott: > Could you open an issue about implementing this warning?

JFTR, this is now tracked in https://github.com/GaloisInc/crucible/issues/1891.

RyanGlScott:
I think you may have accidentally changed the `mir-json` submodule to the wrong commit, but otherwise LGTM.

PR #1894, opened 2026-09-10
Update dependencies and freeze files in preparation for release.

aschwerdfeger-galois:


RyanGlScott (approved):
`crux-mir`'s `test/symb_eval/num/checked_div.rs` test case now fails in CI because it produces slightly different output than before. Ultimately, I think these changes are benign:

* Instead of reporting `65` as the final answer, it now reports `Symbolic BV`. I think this is fine, as the answer was derived from the result of integer overflow and division-by-zero anyway, so I think Crux it free to return a symbolic value here.
* It now reports an additional error on line 22, claiming that the final addition could result in integer overflow. Similarly, I think this is fine, as the symbolic values derived from division-related overflow or division by zero could be thought of as unconstrained (therefore, adding them together could conceivably overflow under this interpretation). 

My guess is that this change was brought about by What4 becoming more precise with respect to its division-related abstract interpretation (perhaps https://github.com/GaloisInc/what4/pull/380 in particular). I don't think it's worth doing forensics to figure this out, however---I think we should just accept the new test output and move on.

PR #1897, opened 2026-09-11
`crucible-mir`: Gracefully recover from static item translation failure

### LINKED ISSUE #1534: `crux-mir` can fail in the presence of unused, unsupported `static`s

RyanGlScott:
Previously, `crucible-mir` would immediately abort if it encountered a translation error (i.e., if it called `mirFail`) when translating a static item's constant value. This was not ideal, as it meant that static items that were never accessed in practice during simulation could prevent simulation from starting in the first place.

This patch changes the behavior such that failing to translate a static item's constant value simply jumps to translating the next static item. That way, as long as the unsupported static item is never accessed during simulation, then `crucible-mir` can still make progress. See `Note [Translating unsupported static items]` in `Mir.Trans` for more details.

Fixes https://github.com/GaloisInc/crucible/issues/1534.


crucible-mir/CHANGELOG.md
```diff
@@ -4,6 +4,17 @@ This release supports [version
...
+  item's initializer. If this happens, `crucible-mir` will instead proceed to
+  simulate the `crux::test` entrypoints, and it will only fail if an
+  untranslated static item is accessed during simulation.
+
+# next
+
+This release supports [version
+13](https://github.com/GaloisInc/mir-json/blob/master/SCHEMA_CHANGELOG.md#13) of
+`mir-json`'s schema.
+
```
samcowger, line 17: This looks duplicative of the first few lines of the changelog - is this intentional?

spernsteiner (approved):

samcowger (approved):
Thanks for documenting this so thoroughly!

crucible-mir/CHANGELOG.md
RyanGlScott: This is a temporary placeholder, since I intend to rebase these changes after the next Crux release has been cut.

RyanGlScott: Since the `crucible-mir` version number wasn't bumped as part of the recent Crux release, I opted to do so here in this PR. The changelog entry added in this PR now resides in its own section, so I think this is ready to land now.

PR #1898, opened 2026-09-11
Enable path-sat option by default

podhrmic:
In most cases, `cargo crux-test -- --path-sat` is faster, so it makes sense to enable the path pruning by default. In the limited cases when this leads to slower times, it can be disabled with `--path-sat False" (at least I think that is the right syntax)


RyanGlScott:
Some observations:

* With the way that Crux's command-line parser is currently set up, I believe that `--path-sat False` is invalid syntax. Rather, path satisifiability is assumed to be disabled by default, and `--path-sat` enables it, but there is no flag counterpart for explicitly disabling it. Obviously, we'll need to do something about that, perhaps by adding a `--no-path-sat` flag.

  One complication I foresee with this is that we'll need to figure out what the interaction between `--path-sat` and `--no-path-sat` would be. Alternatively, we could remove the `--path-sat` flag, but that seems unwise given that so many Crux scripts are already relying on this flag.
* The `crucible-wasm` test suite now fails in CI because path satisfiability is causing SMT solvers to be invoked where they used not to be, and the `crucible-wasm` CI isn't yet installing SMT solvers like other jobs are. This should be relatively straightforward to address in the CI configuration.
* The test output for `strlen_test2.c` in `crux-llvm`'s test suite has changed, as path satisfiability causes Crux to exit out of a loop earlier than it normally would. Reading the comments [here](https://github.com/GaloisInc/crucible/blob/1e49f0a1c2d88ec008b848e0c3bcadea01cd3154/crux-llvm/test-data/golden/golden/strlen_test2.c#L7-L20), I don't think this is desirable, as the point of this test is to see if Crux can figure out how to exit early from a loop without the help of path satisfiability checking. I think we should add a `strlen_test.config` file that disables path satisfiability checking to avoid this.

sauclovian-g:
Also (as I've been saying on the SAW side for a while) turning off path satisfiability checking can lead to unexpected / notionally incorrect behavior and people should know what they're doing before they try it.

PR #1899, opened 2026-09-12
Cache MIR tests to speed up CI build

Ra1issa:


.github/workflows/crux-mir-build.yml
```diff
@@ -157,7 +157,8 @@ jobs:
           cargo install --locked --force
           mir-json-translate-libs
           echo "CRUX_RUST_LIBRARY_PATH=$(pwd)/rlibs" >> $GITHUB_ENV
-
+        env:
+          SCHEMA_VER: ${mir-json -V | grep "schema"}
```
aschwerdfeger-galois, line 161: It looks like entries in the `env` section only support static values, not dynamically generated ones like this. In any event, the value assigned to `SCHEMA_VER` ends up being the string literal `${mir-json -V | grep "schema"}` rather than the schema version.

As far as I can tell this value is unused and can be safely removed.

.github/workflows/crux-mir-build.yml
```diff
@@ -176,14 +177,40 @@ jobs:
...
+
       - name: Haddock
         # See Note [--disable-documentation] in `crucible-go-build.yml`.
         run: cabal haddock --disable-documentation crucible-syntax crucible-concurrency crux-mir
 
       - run: .github/ci.sh test crucible-mir-syntax
       - run: .github/ci.sh test crucible-mir-cli
+      
+      - name: Save MIR schema version
+        run: echo "SCHEMA_VER=$(mir-json -V | grep "schema" | sed -r 's/[ ()]+/-/g')" >> $GITHUB_ENV
```
aschwerdfeger-galois, line 190: This version string includes the `mir-json` version number as well as the schema version, and the variable name should reflect this (*e.g.*, `MIR_JSON_VER`).

aschwerdfeger-galois (changes requested):

RyanGlScott:
@Ra1issa and I discussed this offline. One of the more surprising outcomes of attempting to cache the residual MIR JSON files that the `crux-mir` test suite produces is that is doesn't actually save any time! That is, the time it takes to compile the `.rs` files to JSON is pretty negligible nowadays (even without caching), and the time it takes to simulate the JSON afterwards now dominates the overall time the test suite takes. I'm fairly certain that this didn't used to be the case, but I suspect that recent `mir-json`-side changes in https://github.com/GaloisInc/mir-json/pull/294 improved the compilation times substantially.

In any case, @Ra1issa and I agreed that in light of this discovery, this avenue of CI optimization isn't worth pursuing. I do think that adding SVN hashes to MIR files (as was done in https://github.com/GaloisInc/mir-json/pull/299) could still be useful in the future should we ever decide to make `mir-json`'s recompilation checking smarter.

PR #1900, opened 2026-09-15
crux-mir: add failing test for non-integer peelIndex

spernsteiner:


RyanGlScott (approved):
Thanks!

Interestingly, this program also appears to fail in Miri when using stacked borrows ([play.rust-lang.org link](https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=2350ea0b7383a9c890e917eb77e2cb45)), and you have to use tree borrows in order for it to pass.

PR #1903, opened 2026-09-22
crucible-mir: represent zero-sized enums as `MirAggregate`s

samcowger:
See https://github.com/GaloisInc/crucible/issues/1497#issuecomment-5765654645 for background, and the newly-added `Note [enum representation]` for details.


crucible-mir/src/Mir/TransTy.hs
```diff
@@ -1607,6 +1614,26 @@ buildUnion unionAdt fieldIdx (MirExp actualFieldTpr fieldExpr) = do
   fullAg <- mirAggregate_set fieldOffset fieldSize actualFieldTpr fieldExpr emptyAg
   pure (MirExp MirAggregateRepr fullAg)
 
+
+{-
+Note [enum representation]
+----------------------------------------
```
RyanGlScott: Alignment:


```suggestion
Note [enum representation]
--------------------------
```

crucible-mir/src/Mir/TransTy.hs
RyanGlScott: A minor thing, but would you be willing to combine these two `Enum` cases into a single one? Something like this:

```hs
      M.Enum discrTy
        | adt ^. M.adtSize == 0 ->
          -- See Note [enum representation]
          Right (Some MirAggregateRepr)
        | otherwise -> do
          Some discrTp <- tyToRepr col discrTy
          SomeRustEnumRepr _ ctx <- enumVariants col adt
          Right (Some (RustEnumRepr discrTp ctx))
```

This makes it easier to tell at a glance where all of the `Enum`-related logic lives. (Similar comments apply to other places in this patch that introduce `adt ^. M.adtSize == 0` special cases.)

RyanGlScott: Does the logic here need to be special-cased for zero-sized enums?

sauclovian-g: TBH, I find it easier the other way, but do it Ryan's way, he reads this code more than I do :-)

samcowger: To be honest, I'm not sure. It certainly seems like the sort of thing we should do, but I've just tried without success to write a test that would fail but for an additional special case here. I'll go ahead and implement it, though, just to be safe.

RyanGlScott: Interesting! This might be one of those cases where field references are never observable with zero-sized types due to other special cases in the code. That being said, I agree that it's better to be on the safe side.

samcowger: Done - I've also added similar special cases for `{get,set}EnumField`, by the same token.

RyanGlScott (approved):
Looks reasonable. It might be worth preparing a corresponding SAW PR before landing this, as SAW will also need several changes to accommodate the changes to enums' Crucible representation.

